PR #1, opened 2019-06-07
[WIP] Refactoring and documentation extension

mfherbst:


src/core/PlaneWaveBasis.jl
```diff
@@ -0,0 +1,216 @@
+using FFTW
+
+@doc raw"""
```
antoine-levitt: I think this is unnecessary?

src/core/PlaneWaveBasis.jl
```diff
@@ -0,0 +1,216 @@
...
+    PlaneWaveBasis(lattice::SMatrix{3, 3, T}, grid_size::SVector{3, I},
+                   Ecut::Number, kpoints, kweights) where {T <: Real, I <: Integer}
+
+Create a plane-wave basis.
+
+## Examples
+```julia-repl
+julia> b = PlaneWaveBasis(TODO)
```
antoine-levitt: Premature documentation is the root of all evil :-p

src/core/PlaneWaveBasis.jl
```diff
@@ -0,0 +1,216 @@
...
+                   on the size of the grid with respect to performing FFTs.
+- `Ecut`:          Kinetic energy cutoff in Hartree
+- `kpoints`:       List of ``k``-Points in fractional coordinats
+- `kweights`:      List of corresponding weights for the Brillouin-zone integration.
+- `kbasis`:        List of wave vectors (in integer coordinates) for each ``k``-Point,
+                   chosen such that ``|G + k|^2/2 \leq Ecut`` where ``G`` is the
+                   wave vector. The defines the wave function basis ``B_{Ψ,k}`` for
+                   each ``k``-Point.
+- `FFT` and `iFFT`:   Containers for the forward and backward FFT plans on ``B_ρ``.
+"""
```
antoine-levitt: This structure is low-level, I would not document it directly, but rather document the constructor below, and just annotate each field with a comment

antoine-levitt: The rationale is that for this kind of not particularly user-facing things you're going to look at the code anyway, so might as well just document it where it's written, ie in front of the fields

src/core/PlaneWaveBasis.jl
```diff
@@ -0,0 +1,216 @@
...
+    unit_cell_volume::T
+    recip_cell_volume::T
+
+    grid_size::SVector{3, Int}
+    Ecut::T
+    kpoints::Vector{SVector{3, T}}
+    kweights::Vector{T}
+    kbasis::Vector{Vector{SVector{3, Int}}}
+
+    FFT
```
antoine-levitt: This should probably be parametrized upon to get a struct with inferred types; this can be added later though.

src/core/PlaneWaveBasis.jl
```diff
@@ -0,0 +1,216 @@
...
+    iFFT
+end
+
+function PlaneWaveBasis(lattice::AbstractMatrix{T}, grid_size,
+                        Ecut::Number, kpoints, kweights) where {T <: Real}
+    lattice = SMatrix{3, 3, T}(lattice)
```
antoine-levitt: Careful:
```
julia> SMatrix{3,3,Float64}
SArray{Tuple{3,3},Float64,2,L} where L
```
so this is not statically typed. Applies also above. in this case though you can just omit this line and it will get converted on the call to the struct

src/core/PlaneWaveBasis.jl
```diff
@@ -0,0 +1,216 @@
...
+    # Plan a FFT, spending some time on finding an optimal algorithm
+    # for the machine on which the computation runs
+    grid_size = SVector{3, Int}(grid_size)
+    tmp = Array{Complex{T}}(undef, grid_size...)
+    fft_plan = plan_fft!(tmp, flags=FFTW.MEASURE)
+    ifft_plan = plan_ifft!(tmp, flags=FFTW.MEASURE)
+
+    dummy_kpoints = Vector{SVector{3, T}}(undef, length(kpoints))
+    dummy_kweights = Vector{T}(undef, length(kweights))
+    dummy_kbasis = Vector{Vector{SVector{3, Int}}}(undef, length(kpoints))
```
antoine-levitt: You can use `similar` for this

src/core/PlaneWaveBasis.jl
```diff
@@ -0,0 +1,216 @@
...
+                           fft_plan, ifft_plan)
+    substitute_kpoints!(pw, kpoints, kweights)
+end
+
+"""
+Replace the kpoints of an existing Plane-wave basis and make
+the internal data structures consistent.
+"""
+function substitute_kpoints!(pw::PlaneWaveBasis{T}, kpoints, kweights) where T
+    if length(kpoints) != length(kweights)
```
antoine-levitt: This kind of trivial check uses valuable screen real estate. I would personally do `length(kpoints) != length(kweights) && error()`, or, even better, `@assert length(kpoints) == length(kweights)`. You might argue that an assert is not "clean" and does not produce a legible error message, but really it's exactly the same amount of effort to understand where an `@assert` comes from than an `error`. Also, less code, less bugs also applies for error messages (you might decide someday you want points at the end of sentences, of rename the fields and forget to do it inside the error message, or misspell `length` (like I do every other sentence), or...). Again, I have a strong bias towards less LOC = better code (except for comments), which might not always be true, but is not that bad a heuristic.

src/core/PlaneWaveBasis.jl
```diff
@@ -0,0 +1,216 @@
...
+        error("Lengths of kpoints and length of kweights need to agree")
+    end
+    kweights = Vector{T}(kweights / sum(kweights))  # Normalise kweights
```
antoine-levitt: If the "user" gives an unnormalized list of weights, they might have their reasons, and it seems dangerous to silently ignore that. If you want to avoid funny business here you can have an `@assert sum(kweights) ≈ 1`

src/core/PlaneWaveBasis.jl
```diff
@@ -0,0 +1,216 @@
...
+    kpoints = [SVector{3, T}(kp) for kp in kpoints]
```
antoine-levitt: Again I would not do that, and let convert work its magic (or lack thereof, in which case the "user" converts to svector themselves))

src/core/PlaneWaveBasis.jl
antoine-levitt: Also why not just []?

src/core/PlaneWaveBasis.jl
```diff
@@ -0,0 +1,216 @@
...
+
+    resize!(pw.kpoints, length(kpoints))[:] = kpoints
```
antoine-levitt: `a[:] = b` is better written `a .= b`

src/core/PlaneWaveBasis.jl
```diff
@@ -0,0 +1,216 @@
...
+"""
+Return a generator producing the range of wave-vector coordinates contained
+in the Fourier grid ``B_ρ`` described by the plane-wave basis.
+"""
+function gcoords(pw::PlaneWaveBasis)
+    start = -ceil.(Int, (pw.grid_size .- 1) ./ 2)
+    stop  = floor.(Int, (pw.grid_size .- 1) ./ 2)
+    ci = CartesianIndices((UnitRange.(start, stop)..., ))
+    (SVector(ci[i].I) for i in 1:length(ci))
+end
```
antoine-levitt: Can you add an `@assert` that the `grid_size` is sufficient wrt Ecut here?

src/core/PlaneWaveBasis.jl
```diff
@@ -0,0 +1,216 @@
...
+                           grid_size, Ecut, dummy_kpoints, dummy_kweights, dummy_kbasis,
+                           fft_plan, ifft_plan)
+    substitute_kpoints!(pw, kpoints, kweights)
+end
+
+"""
+Replace the kpoints of an existing Plane-wave basis and make
+the internal data structures consistent.
+"""
+function substitute_kpoints!(pw::PlaneWaveBasis{T}, kpoints, kweights) where T
```
antoine-levitt: `set_kpoints!` instead?

src/core/PlaneWaveBasis.jl
```diff
@@ -0,0 +1,216 @@
...
+    dummy_kbasis = Vector{Vector{SVector{3, Int}}}(undef, length(kpoints))
+    pw = PlaneWaveBasis{T}(lattice, recip_lattice, det(lattice), det(recip_lattice),
+                           grid_size, Ecut, dummy_kpoints, dummy_kweights, dummy_kbasis,
+                           fft_plan, ifft_plan)
+    substitute_kpoints!(pw, kpoints, kweights)
+end
+
+"""
+Replace the kpoints of an existing Plane-wave basis and make
+the internal data structures consistent.
```
antoine-levitt: "and change the basis accordingly"

src/core/PlaneWaveBasis.jl
```diff
@@ -0,0 +1,216 @@
...
+
+    dummy_kpoints = Vector{SVector{3, T}}(undef, length(kpoints))
+    dummy_kweights = Vector{T}(undef, length(kweights))
+    dummy_kbasis = Vector{Vector{SVector{3, Int}}}(undef, length(kpoints))
+    pw = PlaneWaveBasis{T}(lattice, recip_lattice, det(lattice), det(recip_lattice),
+                           grid_size, Ecut, dummy_kpoints, dummy_kweights, dummy_kbasis,
+                           fft_plan, ifft_plan)
+    substitute_kpoints!(pw, kpoints, kweights)
+end
+
```
antoine-levitt: Add a convenience constructor `PlaneWaveBasis(lattice, Ecut, supersampling=2, BZ_sampling=(6,6,6))`?

src/core/PlaneWaveBasis.jl
```diff
@@ -0,0 +1,216 @@
...
+- `grid_size`:     Size of the rectangular Fourier grid used as the
+                   density basis ``B_ρ``. In each dimension `idim` the
+                   range of wave vectors (in integer coordinates) extends from
+                   `-ceil(Int, (grid_size[idim]-1) / 2)` up to
+                   `floor(Int, (grid_size[idim]-1) / 2)`. No optimisation is done
+                   on the size of the grid with respect to performing FFTs.
+- `Ecut`:          Kinetic energy cutoff in Hartree
+- `kpoints`:       List of ``k``-Points in fractional coordinats
+- `kweights`:      List of corresponding weights for the Brillouin-zone integration.
+- `kbasis`:        List of wave vectors (in integer coordinates) for each ``k``-Point,
```
antoine-levitt: G_basis maybe? Not clear from that description that it is the G vectors you're talking about

src/core/PlaneWaveBasis.jl
```diff
@@ -0,0 +1,216 @@
...
+                   chosen such that ``|G + k|^2/2 \leq Ecut`` where ``G`` is the
```
antoine-levitt: Technically nothing in the PlaneWaveBasis structure assumes this, and the rest of the code works fine without this assumption (which you could reasonably want to change to have a different basis selection). So I'd leave that out of the struct, and document this in the convenience constructors.

src/core/PlaneWaveBasis.jl
```diff
@@ -0,0 +1,216 @@
...
+    resize!(pw.kpoints, length(kpoints))[:] = kpoints
+    resize!(pw.kweights, length(kweights))[:] = kweights
+
+    # Update kbasis: For each k-Point select those Gcoords, such that
+    # the kinetic energy cutoff is satisfied
+    for (ik, k) in enumerate(kpoints)
+        energy(Gc) = sum(abs2, pw.recip_lattice * (k + Gc)) / 2
+        p = [Gc for Gc in gcoords(pw) if energy(Gc) ≤ pw.Ecut]
+        pw.kbasis[ik] = [Gc for Gc in gcoords(pw) if energy(Gc) ≤ pw.Ecut]
+    end
```
antoine-levitt: Kind of unclear that the rule is |k+Gc|^2 <= Ecut and not |Gc|^2 <= Ecut here. I'd define `energy(q)` and then check `energy(k+G)`, with the consistent rule everywhere in the code that `q=k+G` (or other names, but I really like having a consistent notation for this Fourier decomposition, which is the fundamental idea of periodic systems)

src/core/PlaneWaveBasis.jl
```diff
@@ -0,0 +1,216 @@
...
+
+    pw
+end
+
+Base.eltype(basis::PlaneWaveBasis{T}) where T = T
```
antoine-levitt: Why is this needed? If you need it somewhere you can get it from the fields.

src/core/PlaneWaveBasis.jl
mfherbst: I thought this would better go to the utilities.

src/core/PlaneWaveBasis.jl
mfherbst: I think it's needed because oft the LaTeX (See Documenter.jl examples), but I will check.

src/core/PlaneWaveBasis.jl
mfherbst: The in-line comments make the eines very long. How abou a comment block at the beginning oft a struct?

src/core/PlaneWaveBasis.jl
mfherbst: Ag good point.

src/core/PlaneWaveBasis.jl
antoine-levitt: I would weigh locality of information more than core/utilities distinction in this instance, but your call

src/core/PlaneWaveBasis.jl
antoine-levitt: I don't have your aversion to long lines :-p but comment block is good; or just a comment before each field

src/core/PlaneWaveBasis.jl
mfherbst: It is indeed required.

src/core/PlaneWaveBasis.jl
mfherbst: Of course the plan was to change that once we have settled on something.

src/core/PlaneWaveBasis.jl
mfherbst: Agree. I added a comment.

src/core/PlaneWaveBasis.jl
mfherbst: Did some restructuring.

src/core/PlaneWaveBasis.jl
mfherbst: Nah. I don't like that, since `grid_size` also defines G-vectors, but different one. How about `wfcn_basis`?

src/core/PlaneWaveBasis.jl
mfherbst: Changed.

src/core/PlaneWaveBasis.jl
mfherbst: About the automatic conversion: That's true, but then the error will occur at a later point and it is less clear what caused it.

src/core/PlaneWaveBasis.jl
mfherbst: Historic legacy reasons ... removed.

src/core/PlaneWaveBasis.jl
mfherbst: How about a compromise: `@assert` with a nice error message ;)

src/core/PlaneWaveBasis.jl
mfherbst: Hmm ... I see the point. I make an assertion.

src/core/PlaneWaveBasis.jl
mfherbst: Agree.

src/core/PlaneWaveBasis.jl
mfherbst: done.

src/core/PlaneWaveBasis.jl
mfherbst: Also: `q` is used elsewhere in the code and I fully agree about consistency on this.

src/core/PlaneWaveBasis.jl
mfherbst: Done.

src/core/PlaneWaveBasis.jl
mfherbst: Still looking at other codes, it seems to be the agreement to let the multiple dispatch automatic `convert` do as much magic as possible, let's stick to that.

src/core/PlaneWaveBasis.jl
mfherbst: I introduced it for the iterative solvers. I guess you are right, but I'll keep it for now. If I see a nice way to do the type stuff as I go along refactoring, I'll remove it.

src/core/PlaneWaveBasis.jl
mfherbst: Why here? For me this is just a fast helper function to get the wave vectors, whenever they are explicitly needed. If one wanted to check for `Ecut` and for internal consistency of `PlaneWaveBasis`, I would put it into the constructor instead. Don't you think?

src/core/PlaneWaveBasis.jl
mfherbst: I looked at it again and I am not sure it is a good idea. Right now adding the `BZ_sampling=(6,6,6)` parameter implies, that we need to do a `PyCall` into `pymatgen` to get the IR-reduced BZ mesh and thus the actual `k`-Points and their weights. This requires to build a `pymatgen.Structure`, which not only needs the `lattice` info, but also the list of atoms and their fractional coordinates. In my opinion this makes this low-level interface quite dependent on a lot of external stuff. For this reason I'd prefer the more orthogonal approach and stuff that into the utilities instead.

src/core/PlaneWaveBasis.jl
mfherbst: Now removed.

src/core/PlaneWaveBasis.jl
```diff
@@ -0,0 +1,176 @@
+using FFTW
+using StaticArrays
+
+struct PlaneWaveBasis{T <: Real}
+    # Lattice and reciprocal lattice vectors in columns
+    lattice::SMatrix{3, 3, T, 9}
```
antoine-levitt: Maybe use a `SMat3{T}` and `SVec3` typedef for this? It comes up a lot

src/core/PlaneWaveBasis.jl
```diff
@@ -0,0 +1,176 @@
...
+#
+@doc raw"""
+    G_to_R!(pw::PlaneWaveBasis, f_fourier, f_real[, gcoords])
+
+Perform an in-place FFT to translate between `f_fourier`, a fourier representation
+of a function using the wave vectors specified in `gcoords` and a representation
+on the real-space density grid ``B^∗_ρ``. The function will destroy all data
+in `f_real`. If `gcoords` is absent, the full density grid ``B_ρ`` is used.
+"""
+function G_to_R!(pw::PlaneWaveBasis, f_fourier, f_real; gcoords=gcoords(pw))
```
antoine-levitt: Add non-mutating helper function for faster prototyping? That's the downside with documentation (:-p) : if you want to be consistent you have to document the helper also, and that induces code duplication... not sure I have a good solution to that.

antoine-levitt: Also possibly G_to_r/r_to_G would be more consistent, if we say that r is in the unit cell and R is lattice vectors. Actually I really like q=k+K and so maybe it would be better to rename G to K... let's discuss next week!

src/core/PlaneWaveBasis.jl
```diff
@@ -0,0 +1,176 @@
...
+    recip_cell_volume::T
+
+    # Selected energy cutoff at construction time
+    Ecut::T
+
+    # Size of the rectangular Fourier grid used as the density basis
+    # and the k-point-specific (spherical) wave function basis
+    # The wave vectors are given in integer coordinates.
+    grid_size::SVector{3, Int}
+    idx_DC::Int  # Index of the DC component
```
antoine-levitt: In which array?

src/core/PlaneWaveBasis.jl
```diff
@@ -0,0 +1,176 @@
...
+    kpoints::Vector{SVector{3, T}}
+    wfctn_basis::Vector{Vector{SVector{3, Int}}}
```
antoine-levitt: Need to explain here when you have multi-d arrays: is it `wfctn_basis[ik][iG]` or `wfctn_basis[iG][ik]`? Maybe use `wf` for wave functions (otherwise I'll never remember what particular letters I should put here)

src/core/PlaneWaveBasis.jl
```diff
@@ -0,0 +1,176 @@
...
+
+    # Selected energy cutoff at construction time
+    Ecut::T
+
+    # Size of the rectangular Fourier grid used as the density basis
+    # and the k-point-specific (spherical) wave function basis
+    # The wave vectors are given in integer coordinates.
+    grid_size::SVector{3, Int}
+    idx_DC::Int  # Index of the DC component
+    kpoints::Vector{SVector{3, T}}
```
antoine-levitt: This is in fractional coordinates I guess?

src/core/PlaneWaveBasis.jl
mfherbst: I'll leave it unchanged until then ... it's a copy and replace anyway.

src/core/PlaneWaveBasis.jl
mfherbst: Renamed to `wf_basis` and added clearifying comments.

src/core/PlaneWaveBasis.jl
mfherbst: added a comment.

src/core/PlaneWaveBasis.jl
mfherbst: Added a comment.

src/core/PlaneWaveBasis.jl
mfherbst: Good point.

src/core/PlaneWaveBasis.jl
mfherbst: Ok now changed to `wf_basis` as discussed below.

PR #4, opened 2019-07-16
Add LDA exchange-correlation functional

mfherbst:
This PR will add support for LDA density functionals. The goal is to use libxc by default and have a fallback julia implementation for deviating precision.


src/core/xc_fallback.jl
```diff
@@ -0,0 +1,74 @@
...
+function lda_x!(ρ; α=1, E=nothing, Vρ=nothing)
+    tmp = -α * cbrt.(3/π * ρ)
+
+    if E !== nothing
+        E .= 3/4 .* tmp
+    end
+    if Vρ !== nothing
+        Vρ .= tmp
+    end
+end
```
antoine-levitt: This complexifies the code without adding too much functionality, and kwargs are tricky for inference. I think I would write this as `lda_x!(E,V,rho;alpha)`, since computing E and computing V are about the same cost; and then maybe a helper function `lda_x(E,rho;alpha) = lda_x!(E,similar(rho),alpha)` for convenience?

src/core/xc_fallback.jl
```diff
@@ -0,0 +1,74 @@
...
+    n_spin::Int
+end
+function Functional(identifier)
+    Functional(string(identifier), family_lda, 1)
+end
+# TODO end temporary
+
+
+function lda_x!(ρ; α=1, E=nothing, Vρ=nothing)
+    tmp = -α * cbrt.(3/π * ρ)
```
antoine-levitt: You want a `@.` here to avoid temporaries. See comment below to get to zero allocs.

src/core/xc_fallback.jl
antoine-levitt: Huh, so actually I'm confused; is rho meant to be scalar or vector? If scalar, it should just return the value (eg return (E,V)), not modify anything.

codegen/xc_fallback_generate/generate.jl
```diff
@@ -0,0 +1,48 @@
+# Adapted from the maple scripts in libxc
```
antoine-levitt: Could you add a README to that dir explaining the rationale here? I don't really get what this generates, and why this is needed. Also what's vwn?

mfherbst:
@antoine-levitt Thanks for your comments ... I wanted to clean up the code anyway today :smile:

codegen/xc_fallback_generate/generate.jl
mfherbst: Sure. Will do if you think it's worth it.

The point is that libxc actually uses an analytic procedure (written in Maple) to generate the C code for computing exchange correlation functionals. I.e. you only code the energy expression in a general form and the special cases and the potential / higher-order derivatives are autogenerated. So instead of recoding the LDA expressions in Julia from the papers, I wanted to see how difficult it is to use the Maple scripts and generate the expressions automatically in a useful way. It's not completely clean (in the sense that one has to manually adapt afterwards) and probably at present will not scale to the complete repository of functionals available in libxc, but if we need some selected functionals in different precision than `Float64`, this generator is probably useful.

VWN = Vosko Wilk,and Nusair. That's the term commonly used in quantum chemistry literature to refer to [this paper](https://doi.org/10.1139/p80-159), where they compare different parametrisations of LDA exchange. The "standard one" to the best of my knowledge is the one with a Pade approximation of a numerical diffusion Monte-Carlo simulation and that's the one I use as well.

src/core/xc_fallback.jl
mfherbst: What then for higher derivatives, which will need to be computed only sometimes? In the [Libxc.jl](https://github.com/unkcpz/Libxc.jl) julia package, I solved this using kwargs, which (if present) indicate a computation, otherwise not. See the code: [evaluate.jl](https://github.com/unkcpz/Libxc.jl/blob/master/src/evaluate.jl).

`ρ` is meant to be a vector.

src/core/xc_fallback.jl
mfherbst: Good point ... haven't really looked into that so far.

codegen/xc_fallback_generate/generate.jl
antoine-levitt: Just put that in the README :p

mfherbst: Done.

src/core/xc_fallback.jl
mfherbst: That's changed now.

src/core/xc_fallback.jl
mfherbst: I left that unchanged so far. Happy to discuss, however.

mfherbst:
@antoine-levitt A silicon LDA in Float64 and Float32 precision is working now. I will merge this PR once we have settled with respect to the interface question you posed.

src/core/xc_fallback.jl
antoine-levitt: I worry that this will be very slow. Also, using in-place is a bit inconvenient. But probably it doesn't really matter... One thing that might be useful is named tuples: `(a=1,b=2)`. Anyway, I don't have anything better to propose and it's pretty low-level anyway so let's go with that!

mfherbst: Named tuple is not a bad idea ... on the other hand: I wonder why you fear the kwargs version is slow. From my tests it does not appear as type unstable and unused parts of the function (where the argument is `nothing`) are completely removed in the `@code_llvm` output as far as I can tell. So that sounds exactly as one would want it.

antoine-levitt: OK! My fear appears outdated, that's great!

antoine-levitt: Yeah, that's https://github.com/JuliaLang/julia/issues/9551, gone since feb 2018

mfherbst: Great. Good to hear about that ... I guess then we are good to go for now.

PR #6, opened 2019-07-25
Use Tuple to splat

antoine-levitt:
Non-trivial gotcha that showed up under profiling (yes, I like profiling :-p)


antoine-levitt:
Possibly a better option is to not use SVector but CartesianIndex here.

mfherbst:
Wow ... that's clearly not so trivial in the end.

> Possibly a better option is to not use SVector but CartesianIndex here.

Yes, especially since the Julia standard library seems to be in favour of the `CartesianIndex` datastructure.

PR #7, opened 2019-07-25
WIP: apply_ functions add to arguments instead of setting them

antoine-levitt:
This is just a test to see what the performance looks like, it doesn't have docs or anything. Even with these changes, still an incredible amount of time (just a bit less than the FFTs) is spent zeroing stuff, so maybe it's not the right approach. Needs to be discussed more.


mfherbst:
I'm closing this for now, since the branch becomes more and more out of date. We can pick up the developments at a later point if we desire.

PR #8, opened 2019-07-27
Support smearing

antoine-levitt:
This is unit tested but I didn't try it on an actual metal. Also added an assert in the case of zero temperature that we are in an insulator (otherwise the procedure is not correct); that assert makes one test fail, do you know why that is?

### DIFF
--- a/src/core/occupation.jl
+++ b/src/core/occupation.jl
@@ -1,20 +1,74 @@
+import Roots
+using SpecialFunctions
+
+
 """
 Compute the occupation at zero temperature and without smearing
 for `n_elec` electrons and the bands `Psi` with associated `energies`.
 """
 function occupation_zero_temperature(basis, energies, Psi, n_elec)
     n_bands = size(Psi[1], 2)
-    T = real(eltype(Psi[1]))
+    T = eltype(energies[1])
 
     @assert n_elec % 2 == 0 "Only even number of electrons implemented"
     @assert n_bands ≥ n_elec / 2
 
     occupation = similar(basis.kpoints, Vector{T})
+    HOMO = -Inf
+    LUMO = Inf
     for ik in 1:length(occupation)
         occupation[ik] = zeros(T, n_bands)
-        occupation[ik][1:Int(n_elec / 2)] .= 2
+        occupation[ik][1:div(n_elec,2)] .= 2
+        HOMO = max(HOMO, energies[ik][div(n_elec,2)])
+        LUMO = min(LUMO, energies[ik][div(n_elec,2)+1])
     end
+    ## We just error in the case of metal; the proper thing to do here
+    ## is to do a bisection for the Fermi level as in the case of
+    ## finite-temperature (although that has a dreadful convergence
+    ## rate)
+    @assert HOMO < LUMO
     occupation
 end
 
-# TODO Implement smearing
+# Ref for the equations: Methfessel-Paxton paper, or
+# E. Cancès, V. Ehrlacher, D. Gontier, A. Levitt, D. Lombardi,
+# Numerical quadrature in the brillouin zone for periodic schrodinger
+# operators
+# TODO: Marzari-Vanderbilt "cold smearing"
+smearing_fermi_dirac(x) = 1/(1+exp(x))
+smearing_gaussian(x) = 1/2*(1-erf(x))
+
+H1(x) = 2x
+H2(x) = 4x^2 - 2
+H3(x) = 8x^3 - 12x
+A_coeff(n) = (-1)^n/(factorial(n)*4^n*sqrt(pi))
+const A1 = A_coeff(1); const A2 = A_coeff(2);
+smearing_methfessel_paxton_1(x) = smearing_gaussian(x) + A1*H1(x)*exp(-x^2)
+smearing_methfessel_paxton_2(x) = smearing_gaussian(x) + A1*H1(x)*exp(-x^2) + A2*H3(x)*exp(-x^2)
+
+# List of available smearing functions, useful in tests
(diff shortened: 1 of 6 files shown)


mfherbst:
Not sure. I'll look at it once I am done with the other energy computation and density computation things.

mfherbst:
I did some digging: The answer is simple. The `scf_compare` routine uses a Hamiltonian without XC potential ... and at that level of theory silicon is a metal.

antoine-levitt:
Oh interesting! For metals at 0K the routine is incorrect, I'll take care of it and remove the assert.

mfherbst:
I'm not sure that's the right way to go. I think it's better to have an `occupation_step` function (or something like that) which plainly does no smearing and one with smearing. The user should then either give a temperature and select smearing or not.

mfherbst:
@antoine-levitt Just pushed the changes as I had them in mind. Feel free to have a look.

antoine-levitt:
I agree; I just meant that there is one routine missing, which is a metal at 0K, which is different than an insulator. This is just making sure that the T=0 case of `occupation_temperature` is handled appropriately. I think the `occupation_step` function should still raise an error when HOMO > LUMO, as in that case the result is just plain wrong. If you agree I'll revert your commit, fix `occupation_temperature` to accept T=0, and add an optional flag to disable the error checking.

mfherbst:
I mainly agree ... the error flag has to be accessible from the tests, most importantly.

I do not want to find out right now, how to generate the reference data in ABINIT for silicon without XC, which is actually a metal.

antoine-levitt:
Are you sure abinit gives the result of `occupation_step` (ie with a HOMO > LUMO)? That seems weird.

mfherbst:
Yes:
- The values agree too good.
- The default of [`occopt`](https://docs.abinit.org/variables/basic/#occopt), which I did not change, puts 2 or 0 in each k-Point.

src/core/occupation.jl
```diff
@@ -1,20 +1,79 @@
...
+H2(x) = 4x^2 - 2
+H3(x) = 8x^3 - 12x
+A_coeff(n) = (-1)^n / (factorial(n) * 4^n * sqrt(π))
+const A1 = A_coeff(1)
+const A2 = A_coeff(2);
+smearing_methfessel_paxton_1(x) = smearing_gaussian(x) + A1 * H1(x) * exp(-x^2)
+function smearing_methfessel_paxton_2(x)
+    smearing_gaussian(x) + A1 * H1(x) * exp(-x^2) + A2 * H3(x) * exp(-x^2)
+end
+
```
antoine-levitt: I understand you don't like long lines, but in this case it kills off the symmetry with the previous line and makes it harder to see it's just one more term

src/utils/self_consistent_field.jl
```diff
@@ -11,15 +11,20 @@ electrons, `ρ` is the initial density, e.g. constructed via a SAD guess.
...
-    if ρ === nothing
-        ρ = guess_hcore(ham, n_bands, compute_occupation, lobpcg_prec=lobpcg_prec)
+    if smearing === nothing
+        compute_occupation(basis, energies, Psi) = occupation_step(basis, energies, Psi,
+                                                                   n_electrons)
+    else
+        function compute_occupation(basis, energies, Psi)
+            occupation_temperature(basis, energies, Psi, n_electrons, T, smearing=smearing)
+        end
     end
```
antoine-levitt: Hm, I get 
`WARNING: Method definition (::getfield(DFTK, Symbol("#compute_occupation#188")){T, smearing, n_electrons})(Any, Any, Any) in module DFTK at /home/antoine/.julia/dev/DFTK/src/utils/self_consistent_field.jl:18 overwritten at /home/antoine/.julia/dev/DFTK/src/utils/self_consistent_field.jl:22.
`
from this

antoine-levitt: WTF: https://discourse.julialang.org/t/conditional-function-definition/27348

antoine-levitt:
I left the check disabled by default for tests, because I don't have a good way of passing down the flag to the function, but this should really be enabled by default. The arguments to the scf function is starting to get huge, we should start thinking of putting it into a struct.

src/utils/self_consistent_field.jl
```diff
@@ -14,17 +14,16 @@ function self_consistent_field(ham::Hamiltonian, n_bands::Int, n_electrons::Int;
                                ρ=nothing, tol=1e-6, T=0, smearing=nothing,
                                lobpcg_prec=PreconditionerKinetic(ham, α=0.1),
                                max_iter=100, algorithm=:scf_nlsolve, kwargs...)
-    if smearing === nothing
-        compute_occupation(basis, energies, Psi) = occupation_step(basis, energies, Psi,
-                                                                   n_electrons)
+    if smearing == nothing
```
mfherbst: Tests for nothing should have `===`, see e.g. https://docs.julialang.org/en/latest/base/base/#Base.isnothing

src/utils/self_consistent_field.jl
mfherbst: Interesting ... weird one for sure.

src/core/occupation.jl
mfherbst: How about
```
smearing_methfessel_paxton_2(x) = smearing_methfessel_paxton_1(x) + A2 * H3(x) * exp(-x^2) 
```
That's shorter and makes it apparent, too?

src/utils/self_consistent_field.jl
antoine-levitt: Hm, you're right. Should not matter in principle but the compiler is sure that `x === nothing` is false as long as the type of `x` is not Nothing, whereas `==` might do stranger things. Thanks!

mfherbst:
LGTM as it is now.

PR #10, opened 2019-07-28
Batch Hamiltonian applications

antoine-levitt:


mfherbst:
> Nice commit, indeed better. This line is in a tight-ish loop and probably not that efficient though, it > allocates a slice and does not loop over the fast index first. Do an explicit loop over bands?

> What happens when the input is a single vector? Does it really still work?  

Ups ... I should not force push if you are commenting.  

Good point. I'll change the loop order ... also yes it is working, but I need to change some pieces of code here and there, but should not be too bad.

mfherbst:
> Nitpicking, but: is that kind of verbose assert with an error message really better than a plain assert? This one for instance doesn't really add any information. The other ones above do by printing the actual sizes, but this feels a bit like duplicating the work of proper tooling: I see asserts as essentially an invitation to run in a debugger, which is a better way to inspect the state of the inputs. And this kind of verboseness is killing off the nice minimalist vibe julia has going on. I suggest doing proper argument checking (throwing ArgumentError) for common user-facing errors on complex and high level functions, but just plain asserts for this kind of defensive programming that also helps understanding the invariants of the inputs. 

In my workflow (test -> edit code -> test), I cannot run a debbugger at the moment and this information really helps. I am happy to remove them once I got some alternative setup running.

antoine-levitt:
Completely fine to force push! I just randomly comment on code I see. I'll do a full pass once we get into release mode anyway.

PR #12, opened 2019-07-28
Fourier fixes

antoine-levitt:
Save one temp by using in-place FFT, and one pass over the data by using a scaled FFT plan.

I don't understand the factors here, I thought the fourier data was such that `psi = sum cG eG` with the eG orthonormal. Shouldn't there be sqrt(unit_cell_volume) on both sides?

### DIFF
--- a/src/core/PlaneWaveBasis.jl
+++ b/src/core/PlaneWaveBasis.jl
@@ -1,6 +1,6 @@
 using FFTW
 
-struct PlaneWaveBasis{T <: Real}
+struct PlaneWaveBasis{T <: Real, TFFT, TIFFT}
     # Lattice and reciprocal lattice vectors in columns
     lattice::Mat3{T}
     recip_lattice::Mat3{T}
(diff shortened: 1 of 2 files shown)


antoine-levitt:
So doing the scaling in the plan rather than on the data afterwards is about the same timing, nothing to be gained there, so you can revert that change if you prefer the old version (I think it's a bit cleaner like that but it doesn't matter much)

mfherbst:
About the volume factor: Initially I had sqrt(unit_cell_volume), but this did not reproduce the right numbers, so I removed the factor. In fact I am not too sure, why it has to be like that.

antoine-levitt:
Hm OK let's discuss it sometimes. Which convention doesn't actually matter as long as it's clearly documented somewhere. I liked the eG thing because it's clear that it's an orthogonal decomposition.

mfherbst:
I fully agree (to all three points) :smile:

PR #13, opened 2019-07-30
Brillouin zone symmetry for density computation

mfherbst:
This PR will introduce a density computation routine, which takes the symmetry of the Brillouin zone (BZ) into account in order to allow to compute the density only from an irreducible BZ wedge.


antoine-levitt:
Great, well done! Could you add a quick recap of the equations and the derivations (the ones we did on the blackboard) in the comments?

mfherbst:
Good point, will do.

PR #17, opened 2019-08-07
Doc and cleanup

antoine-levitt:


README.md
```diff
@@ -34,16 +36,26 @@ The package is not yet registered, so you need to install it from the github url
...
+from a Julia command line (version at least 1.1).
+
+Some parts of the code require a working Python installation with the libraries
+`scipy` and `spglib`. The examples require `matplotlib` and `pymatgen` as well.
+Check out which version of python is used by the
+[PyCall.jl](https://github.com/JuliaPy/PyCall.jl) package, and use the
+corresponding package manager (usually `apt`, `pip` or `conda`) to install
+these libraries, for example
+```
+pip install scipy spglib matplotlib pymatgen
```
antoine-levitt: Should that be `pip3`? I get very confused by this on a regular basis.

mfherbst: Depends on your distro ... on Debian-based systems `pip` is still python2 and `pip3` is python3, but I think that's a Debian modification. Usually it is just `pip`.

antoine-levitt: Still can't believe how annoying this whole thing is. Python3 is ten years old now... Anyway, lots of people are using Debian-based distros and have half of dozen of different python installations. I don't know if there are any general recommendations to be made here other than just "good luck".

mfherbst: Unfortunately true ... I think going with `pip install bla` is kind of standard. One could give the `apt`-line as well, but then `pymatgen` is not yet packaged, so that opens up the next issue. I think we should just have a detailed "install" section in the real documentation at some point, going in some distros in detail, but for now I think the hint we have is good enough.

PR #20, opened 2019-09-04
Add flag to only check convergence for a few bands in LOBPCG

mfherbst:


src/core/constants.jl
```diff
@@ -1,6 +1,15 @@
...
 Mat3{T} = SMatrix{3, 3, T, 9} where T
 Vec3{T} = SVector{3, T} where T
 
 """The default search location for Pseudopotential data files"""
 DFTK_DATADIR = joinpath(dirname(pathof(DFTK)), "..", "data")
+
+# Special macro to flag expensive assertions, which can be
+# disabled by setting these to nothing
+macro assert_expensive(expr) :(@assert $expr) end
+macro assert_expensive(expr, text) :(@assert($expr, $text)) end
```
antoine-levitt: Maybe have a DO_ASSERT_EXPENSIVE=true constant?

mfherbst: I thought about that, but I do not really like that. (1) You have to set that constant somehow, which is also a little sneaky and (2) it feels like we are building something, that should exist in base Julia. I haven't had a good idea about this yet.

antoine-levitt: That's #10614 in julia, it could also be toggled with a CLI arg.

mfherbst: I like [this idea](https://discourse.julialang.org/t/assert-alternatives/24775/14) more, because it can be changed from the REPL.

antoine-levitt: Oh, I was wondering if that was possible but didn't think the macro would trigger recompilation. Precompilation could also possibly mess this up but apparently not. Very cool!

mfherbst: see #21

PR #35, opened 2019-10-28
Unify local potential energy computation

antoine-levitt:
The potentials are constructed complex (eg Hamiltonian:27), but they should all be real, right?


antoine-levitt:
OK I see; they should be but it's annoying because of the way G_to_r works.

PR #40, opened 2019-10-30
Refactor Silicon example

antoine-levitt:


antoine-levitt:
My thought was that a branch is better than three copy-pasted files :-p I see examples as more of a "if you want to do something non-standard, start off from this standard thing" whereas I guess you're looking at it more from a tutorial point of view? I agree with what you're saying if the examples are short, like input files in other codes (ie if we move some helper functions to `src`), but we're not there yet. Besides, I'm quite happy with what's there right now: the examples are a good balance between low and high level. That's more in line with what we want DFTK to be (something for people interested in methods).

antoine-levitt:
I changed the pymatgen interface to avoid building a dummy model, is that OK?

mfherbst:
I came accross this annoyance today as well :)

PR #44, opened 2019-11-02
tighten hamiltonian application a bit

antoine-levitt:
So I played around with avoiding allocations a bit, it doesn't actually change much in terms of performance. Still the `f_real .= 0` and `f_fourier .= 0` show up on profiling, but I wonder where it really comes from (the OS figuring out we actually do need the memory, probably). I *think* calloc is faster than malloc+bzero, and so we might benefit from changing the `f_real = similar(); G_to_r!(f_real, ...)` pattern to something like `f_real = zeros(); G_to_r!(f_real, ...; zeroed_output = true)`. I haven't done that yet because I think we should clear up the FFT functions first, and because I can't find something generic that does what `similar` does but also allocates zero. Might just special-case to `Array` for a start if there's an actual performance difference.

### DIFF
--- a/src/core/HamiltonianBlock.jl
+++ b/src/core/HamiltonianBlock.jl
@@ -41,13 +41,16 @@ function LinearAlgebra.mul!(Y, block::HamiltonianBlock, X)
     kin = block.values_kinetic
     Vnloc = block.values_nonlocal
     Vloc = block.values_local
-    ifft(x) = G_to_r(block.basis, block.kpt, x)
-    fft!(y, x) = r_to_G!(y, block.basis, block.kpt, x)
 
-    if Vloc == nothing
-        Y .= kin * X
+    @assert kin isa Diagonal # for optimization
+
+    if Vloc === nothing
+        Y .= kin.diag .* X
     else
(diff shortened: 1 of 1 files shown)


antoine-levitt:
This PR is just saving an alloc. @mfherbst for this kind of commits, is it OK with you if I pass the tests locally and just merge directly into master, or do you prefer PRs? Happy to do either.

antoine-levitt:
Hah, so I took the trouble to avoid the fill by using `zeros`, only to figure out that it actually allocates an uninitialized array and then calls `fill!`. This is julia issue 130.

antoine-levitt:
It does seem to make a (small) difference though, so let's wait until 130 is fixed and revisit this. Apart from this there isn't much more to be gained in this kind of micro-optimization (in particular saving allocs is basically useless as far as runtime is concerned).

mfherbst:
> This PR is just saving an alloc. @mfherbst for this kind of commits, is it OK with you if I pass the
> tests locally and just merge directly into master, or do you prefer PRs? Happy to do either.
Generally I prefer PRs, because it's sometimes surprising what works on your computer, but not somewhere else. For fixing a typo in the docs or a comment or so it's of course a different thing.

PR #45, opened 2019-11-03
Fixes for 1D systems

antoine-levitt:
spglib segfaults for lattices with a very small (<1e-8) dimension in one direction. Am I doing something wrong? If not we should report it as a bug.

### DIFF
--- a/src/core/Hamiltonian.jl
+++ b/src/core/Hamiltonian.jl
@@ -2,7 +2,7 @@ using LinearAlgebra
 
 # Data structure and functionality for a one-particle Hamiltonian
 
-struct Hamiltonian
+mutable struct Hamiltonian
     basis::PlaneWaveBasis  # Discretized model
     density                # The electron density used to build this Hamiltonian
     kinetic                # Discretized kinetic operator
@@ -55,6 +55,7 @@ Update Hamiltonian in-place
 function update_hamiltonian!(ham::Hamiltonian, ρ::Density)
     basis = ham.basis
     model = basis.model
+    ham.density = ρ
     model.build_hartree(basis, nothing, ham.pot_hartree; ρ=ρ)
     model.build_xc(basis, nothing, ham.pot_xc; ρ=ρ)
     if ham.pot_local !== nothing
--- a/src/core/Hamiltonian.jl
+++ b/src/core/Hamiltonian.jl
@@ -26,7 +26,7 @@ function Hamiltonian(basis::PlaneWaveBasis{T}, ρ=nothing) where T
     # TODO This assumes CPU array
     potarray(p::Density) = similar(fourier(p))
     potarray(::Nothing) = zeros(Complex{T}, basis.fft_size)
-    ρzero = something(ρ, density_zero(basis))
+    ρzero = something(ρ, Density(basis))
 
     _, pot_external = model.build_external(basis, nothing, potarray(ρ))
     _, pot_nonlocal = model.build_nonlocal(basis, nothing, potarray(ρ))
@@ -78,10 +78,6 @@ function update_energies!(energies, ham::Hamiltonian, Psi, occupation, ρ=nothin
     ρ === nothing && (ρ = compute_density(ham.basis, Psi, occupation))
 
     energies[:Kinetic] = energy_term_operator(ham.kinetic, Psi, occupation)
-    if ham.pot_external !== nothing
-        dVol = model.unit_cell_volume / prod(basis.fft_size)
-        energies[:PotExternal] = real(sum(real(ρ) .* ham.pot_external) * dVol)
-    end
 
     function insert_energy!(key, builder; kwargs...)
         energy, _ = builder(basis, Ref{valtype(energies)}(0), nothing; kwargs...)
--- a/src/utils/bzmesh.jl
+++ b/src/utils/bzmesh.jl
@@ -38,7 +38,12 @@ function bzmesh_ir_wedge(kgrid_size, lattice, composition...; tol_symmetry=1e-5)
     # Ask spglib for symmetry operations and for irreducible mesh
     spg_symops = spglib.get_symmetry(spglib_cell(lattice, composition...),
                                      symprec=tol_symmetry)
-    spg_symops !== nothing || error("Could not get symmetry")
+
+    # If spglib does not find symmetries give an error
+    spg_symops !== nothing || error(
+        "spglib failed to get the symmetries. Check your lattice, or use a uniform BZ mesh."
+    )
+
     mapping, grid = spglib.get_stabilized_reciprocal_mesh(
         kgrid_size, spg_symops["rotations"], is_shift=[0, 0, 0], is_time_reversal=true
     )
--- a/examples/silicon.jl
+++ b/examples/silicon.jl
@@ -45,9 +45,9 @@ kcoords, ksymops = bzmesh_ir_wedge(kgrid, lattice, composition...)
 fft_size = determine_grid_size(lattice, Ecut)
 basis = PlaneWaveBasis(model, fft_size, Ecut, kcoords, ksymops)
 
-# Run SCF, note Silicon metal is an insulator, so we will assume that we do not need
-# all bands here. This will cause warnings in some models, because e.g. in the
-# :reduced_hf model silicon is a metal
+# Run SCF. Note Silicon is a semiconductor, so we use an insulator
+# occupation scheme. This will cause warnings in some models, because
+# e.g. in the :reduced_hf model silicon is a metal
 n_bands_scf = Int(model.n_electrons / 2)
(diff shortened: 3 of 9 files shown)


antoine-levitt:
The external potential energy still blows up with a diag(1,1,eps) unit cell. Need to investigate.

src/utils/bzmesh.jl
```diff
@@ -38,7 +38,10 @@ function bzmesh_ir_wedge(kgrid_size, lattice, composition...; tol_symmetry=1e-5)
     # Ask spglib for symmetry operations and for irreducible mesh
     spg_symops = spglib.get_symmetry(spglib_cell(lattice, composition...),
                                      symprec=tol_symmetry)
-    spg_symops !== nothing || error("Could not get symmetry")
+
+    # If spglib does not find symmetries, fall back to regular grid
+    spg_symops === nothing && return bzmesh_uniform(kgrid_size)
+
```
mfherbst: I'm not sure this is such a good idea, actually. Something like this should not go unnoticed. At least one should issue a warning, but I'm not even sure about that. This happens pretty early in a typical calculation, so I think it's fine to just kill everything and let the user decide what to do. When I'm working with DFTK this issue usually arises when I'm stupid with the lattice, so it's been good to force me to go back and reconsider.

src/core/Hamiltonian.jl
```diff
@@ -78,10 +79,6 @@ function update_energies!(energies, ham::Hamiltonian, Psi, occupation, ρ=nothin
     ρ === nothing && (ρ = compute_density(ham.basis, Psi, occupation))
 
     energies[:Kinetic] = energy_term_operator(ham.kinetic, Psi, occupation)
-    if ham.pot_external !== nothing
-        dVol = model.unit_cell_volume / prod(basis.fft_size)
-        energies[:PotExternal] = real(sum(real(ρ) .* ham.pot_external) * dVol)
-    end
```
mfherbst: You removed this, because it was a duplicate, right?

src/core/Hamiltonian.jl
```diff
@@ -55,6 +55,7 @@ Update Hamiltonian in-place
 function update_hamiltonian!(ham::Hamiltonian, ρ::Density)
     basis = ham.basis
     model = basis.model
+    ham.density = ρ
```
mfherbst: Don't we better want to in-place update the real and fourier part of the density. I don't fully remember, but it might be that it is assumed somewhere that the pointer addresses will not change. Did you check that?

src/core/Hamiltonian.jl
antoine-levitt: Yes, I forgot to remove it from a previous commit

src/core/Hamiltonian.jl
antoine-levitt: No, but it looked relatively harmless so I just changed the type to be mutable. We could do in-place update here. What's the way to do it properly? Should we overload `copy!`?

src/utils/bzmesh.jl
antoine-levitt: Why? It's perfectly fine to do a computation with no specific symmetries. It happens for quasi-1D.

antoine-levitt: Hm, come to think of it it should at least have time-reversal symmetry; or is it taken care of somewhere else?

src/core/Hamiltonian.jl
mfherbst: Good question ... I never looked at that. Overloading `copy!` sounds reasonable. Can one also overload the `.=` operator?

src/utils/bzmesh.jl
mfherbst: Nope, nowhere. `bzmesh_uniform` is without any symmetry taken into account.

mfherbst: > Why? It's perfectly fine to do a computation with no specific symmetries. It happens for quasi-1D.

Sure, but `spglib` can fail for different reasons, too, like when you accidentally transpose the lattice matrix. :smile: ... this happens quite quickly when intertwining Julia and python.

antoine-levitt: What do you mean fails? If you pass it the transpose then it just finds a different symmetry group, no?

antoine-levitt: OK so you're saying that even in the purely non-symmetric case it should at least have the time-reversal symmetry in? So when it returns nothing it means some kind of internal error?

src/core/Hamiltonian.jl
antoine-levitt: `methods(copy!)` has a few results so at least it's not explicitly discouraged. Not sure if it matters, but I pushed it.

mfherbst:
There's something wrong now ... the CROP and Anderson solvers converge in 1 step with phenominally low tolerances. This means:
(a) Our SCF compare test is not good enough, because it did not capture that
(b) Blocking the merge due to coverage drop is worth something :)

antoine-levitt:
Hah, point well taken :-) I'll take a look

I'm fine with having coverage info and red crosses, I just don't want coverage to be the endgoal so that we end up writing stupid tests with the only aim to increase coverage, as I've seen happen on other projects.

mfherbst:
> I'm fine with having coverage info and red crosses, I just don't want coverage to be the endgoal so that we end up writing stupid tests with the only aim to increase coverage, as I've seen happen on other projects.

I agree :smile: Only with a useful set of tests, there's meaning in coverage, otherwise not.

antoine-levitt:
Well yeah OK, it's the hamiltonian overwriting which overwrites the density all over again. I could switch to the previous version with ham.density = rho but then we'd be in trouble if, let's say, we want to do mixing on the potential. How about I just remove `update_hamiltonian!` in favor of creating new objects each time? Less potential for errors, less code duplication, cleaner concepts : hamiltonian is strongly immutable (ie is strictly read-only after creation).

antoine-levitt:
In general I think there are a couple of places we could do away with the mutation. Performance is not really critical for anything that is not done for each eigenvector, and non-mutating is simpler to understand, reason with, and AD through. Will fix the current problem a minima and open an issue.

mfherbst:
Regarding the Hamiltonian: I like the idea of having it a fixed object ... in fact that's how it all started out in my head, but then it diverged for a couple of reasons. From my point of view there are two options:

- Make it truely static as you suggest. That would come with a static density, but also with static potentials. This means that we allocate a new density grid (in real and fourier) and two new potential arrays each iteration. That's quite a bit of RAM motion each iteration. Advantage is, that this allows algorthims to do whatever with the Hamiltonian and its constituents, because the meaning of the pointers and the data they point to never changes. Perhaps this is indeed the easiest. 
- Make it fully flexible, i.e. assure that algorithms make a copy when they rely on the density or potentials not changing. That's probably better for performance and RAM motion, but of course that could introduce bloat. 

Not sure what is best here. Maybe we should keep to the motion of KISSing as you suggest and only allow new objects to be returned after an update. That's, however, a bit of a code change to get consistent (e.g. the way the potentials arrays are allocated or not if a builder is nothing) and could potentially impact the way types are derived along the way. I also suggest to open an issue and address this in a different PR.

src/utils/bzmesh.jl
mfherbst: It is not so clear ... I checked the code. Essentially it can return nothing, when two atoms are too close, when the spacegroup search failed, if it does not find symmetry operations and a few other things ... so some are internal, some are due to issues with the lattice setup. There is some sort of more detailed error message you can get, but I am not sure it is always set (it's a very c-like interface).

I think the options we have is either abort and tell the user, or warn the user and fall back to no symmetry.

antoine-levitt:
OK so I reverted the spglib and special-casing of thin cells (I'll do it properly for 0-width cells later), so there's just a couple of cosmetic fixes in here. However something is very strange: I can't see the last commit on this page, but it's there on https://github.com/mfherbst/DFTK.jl/tree/1D_fixes. Github bug, or am I missing yet another git subtlety?

antoine-levitt:
OK so I reverted the spglib and special-casing of thin cells (I'll do it properly for 0-width cells later), so there's just a couple of cosmetic fixes in here. However something is very strange: I can't see the last commit on this page, but it's there on https://github.com/mfherbst/DFTK.jl/tree/1D_fixes. Github bug, or am I missing yet another git subtlety?

mfherbst:
Me neither. Github explicitly says `Add more commits by pushing to the 1D_fixes branch on mfherbst/DFTK.jl.` for me, so I'm a bit unsure what's happening. Could indeed be a github bug.

antoine-levitt:
Seems to be fixed now. (I merged where I probably should have rebased, I guess...)

mfherbst:
No worries, I cleaned up ;)

PR #46, opened 2019-11-04
WIP Normalizations

antoine-levitt:
Builds on top of 1D_fixes, not sure what the correct git etiquette is for this. Anyway, will rebase on top of master when 1D_fixes is merged, which should fix that (?)

Preparation for normalization fix, just some cosmetic fixes and docs right now, will next commit the correct normalization conventions (which should break everything)


antoine-levitt:
OK, tests pass. @mfherbst I don't understand the code in `guess_gaussian_sad`, specifically the fact that the guess density is normalized to have total charge `charge_ionic(spec)`, so it doesn't have the correct number of electrons; shouldn't it be the number of valence electrons instead?

mfherbst:
That's exactly what `charge_ionic` is, see https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/core/Species.jl#L20. so it's the charge needed to compensate the "ionic change", which is defined in the PSP literature as the nuclear charge minus the core electrons as far as I know.

antoine-levitt:
Ah sure, there's even `n_elec_valence(spec::Species) = charge_ionic(spec)`. My bad, thanks!

examples/silicon.jl
```diff
@@ -3,8 +3,8 @@ using Printf
 using PyCall
 
 # Calculation parameters
-calculation_model = :lda
-kgrid = [4, 4, 4]        # k-Point grid
+calculation_model = :pbe
+kgrid = [1, 1, 1]        # k-Point grid
 supercell = [1, 1, 1]    # Lattice supercell
```
mfherbst: This looks like an accidential commit.

mfherbst:
Why do you want to rename it? SAD = superposition of atomic densites ... that's kind of what it is, no?

antoine-levitt:
It's a non-standard acronym (at least for me), and feels like an internal thing. Eg if we want other type of starting densities we can give them as keywords to `guess_density`.

examples/silicon.jl
antoine-levitt: Woops, thanks!

mfherbst:
I like the idea of the kwargs :smile:.

antoine-levitt:
Hah, I didn't know about that! Looks like I still have some bad jokes to discover :p

Great to have this done!

PR #51, opened 2019-11-05
Names

antoine-levitt:
Lots of rename, plus a stealth refactor sneaked in there, and a TODO to remind us we should do a higher-level interface to PlaneWaveBasis. Fixes https://github.com/mfherbst/DFTK.jl/issues/49.

I began with PWBasis, but it sounded strange (PWB is weird) so I went with PlaneWaveBasis. @mfherbst you've been pretty careful to avoid excessive abbreviation, and I think that's a good idea so let's keep that. If you agree let's merge this one to avoid too much conflict; I'll resolve the conflicts in the other PRs.


src/core/PlaneWaveBasis.jl
```diff
@@ -48,19 +46,12 @@ function build_kpoints(basis::PlaneWaveModel{T}, kcoords; Ecut=basis.Ecut) where
 
     kpoints
 end
-
+build_kpoints(basis::PlaneWaveBasis, kcoords) = build_kpoints(basis.model, basis.fft_size, kcoords, basis.Ecut)
```
mfherbst: This line looks too long.

mfherbst:
Merci :smile:

antoine-levitt:
Heh, I don't care that much but OK :p I remember seeing at some point some tooling to enforce style rules, we should probably do that at some point.

mfherbst:
Agree ... but that's by far not the first on my TODO list.

PR #52, opened 2019-11-07
Update deps, and simplify SCF code a bit

antoine-levitt:
Simplify the SCF code thanks to the NLsolve release. I also changed the order of tests a bit so it's more convenient when breaking stuff.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -23,9 +23,9 @@ version = "0.8.10"
 
 [[BinaryProvider]]
 deps = ["Libdl", "SHA"]
-git-tree-sha1 = "29995a7b317bbd06be147e1974a3541ce2502dca"
+git-tree-sha1 = "5b08ed6036d9d3f0ee6369410b830f8873d4024c"
 uuid = "b99e7846-7c00-51b0-8f62-c81ae34c0232"
-version = "0.5.7"
+version = "0.5.8"
 
 [[Blosc]]
 deps = ["BinaryProvider", "CMakeWrapper", "Compat", "Libdl"]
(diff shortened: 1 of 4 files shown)


mfherbst:
I agree with all, but the change of the test order. The problem here is that as of now Julia aborts any testing as soon as one test case fails, so if first the "building blocks" are tested and then the full thing, you know which building block failed. Otherwise you only know that something is wrong.

If you did this to get faster feedback on Travis, I suggest to split the test suite into unit tests and functionality tests and run the functionality tests first and the unit tests afterwards, but configure in a way that for sure both are run. That way you get a fast failure and feedback what went wrong.

I'm happy to do that ... it involves a little playing with the CI.

By the way: Last PR to merge before we'll move to JuliaMolSim :smile:.

antoine-levitt:
I did it to get faster feedback when ] test DFTK. Is there any way to just run all the tests even when one of them fail? At worse can just do it ourselves with a macro that does a try catch, but there should be a nicer way...

mfherbst:
I see. Not a fan of the macro, that's not such a clean solution. Instead I think what one can do is simply to use an environment variable to influence the test order. If it's set, than the big tests run first, otherwise not.

antoine-levitt:
See https://stackoverflow.com/questions/48628549/about-the-right-way-to-use-julia-base-test-run-all-tests-even-if-some-fail. So we just wrap everything with `@testset "All tests"`. Then we get a full report from CI, but it's more convenient to test interactively. Is that OK?

mfherbst:
Then I'm ok, if you move the full paragraph of "big tests" up ... I see no reason why PBE is special.

antoine-levitt:
The point was to just get one (I took PBE because it was the "most complicated") to get a fast diagnostic. I'll move them up.

PR #59, opened 2019-11-14
[WIP] Get DFTK fully generic in the floating point type

mfherbst:
This PR will get DFTK agnostic to the floating point type used.

- [x] Get DoubleFloats to work
- [x] Tests
- [x] Merge magnesium examples


src/core/energy_ewald.jl
```diff
@@ -22,7 +23,7 @@ function energy_ewald(lattice, recip_lattice, charges, positions; η=nothing)
     if η === nothing
         # Balance between reciprocal summation and real-space summation
         # with a slight bias towards reciprocal summation
-        η = sqrt(sqrt(1.69 * norm(recip_lattice ./ 2π) / norm(lattice))) / 2
+        η = T(sqrt(sqrt(1.69 * norm(recip_lattice ./ 2π) / norm(lattice))) / 2)
```
antoine-levitt: How do you more generally make sure that there are no implicit conversions to float64? Should it be T(1.69) and T(2pi) everywhere there are constant literals?

src/core/generic_fftn.jl
```diff
@@ -0,0 +1,58 @@
+import FourierTransforms
+import Primes
+
+function is_fft_size_ok_for_generic(size::Integer)
+    # TODO FourierTransforms has a bug, which is triggered
+    #      only in some factorisations
+    # Everything is fine if we have up to one prime factor,
+    # which is not two, also we want to avoid large primes
```
antoine-levitt: Add ref to bug report

src/core/lobpcg_hyper_impl.jl
```diff
@@ -8,14 +8,22 @@ vprintln(args...) = nothing
...
     F.vectors[:,1:N], F.values[1:N]
 end
 
+import LinearAlgebra: cholesky
+function cholesky(X::Union{Matrix{ComplexF16}, Hermitian{ComplexF16,Matrix{ComplexF16}}})
+    # Cholesky factorisation above may promote the type
+    # (e.g. Float16 is promoted to Float32. This undoes it)
+    U = cholesky(ComplexF32.(X)).U
+    (U=convert.(ComplexF16, U), )
+end
```
antoine-levitt: Should that be a PR to Base?

src/core/self_consistent_field.jl
```diff
@@ -75,7 +75,7 @@ function self_consistent_field!(ham::Hamiltonian, n_bands;
     energies = update_energies(ham, itres.Psi, itres.occupation, itres.ρ)
 
     # Strip off the extra (unconverged) eigenpairs
-    Psi = [p[1:end-n_ep_extra, :] for p in itres.Psi]
+    Psi = [p[:, 1:end-n_ep_extra] for p in itres.Psi]
```
antoine-levitt: Interesting... how did it work before?!

src/core/PlaneWaveBasis.jl
```diff
@@ -70,6 +76,21 @@ function determine_grid_size(model::Model, Ecut; kwargs...)
...
+    fft::opFFT
+end
+LinearAlgebra.mul!(Y, p::DummyInplace, X) = (Y .= mul!(similar(X), p.fft, X))
+LinearAlgebra.ldiv!(Y, p::DummyInplace, X) = (Y .= ldiv!(similar(X), p.fft, X))
+
+import Base: *, \, length
+*(p::DummyInplace, X) = p.fft * X
+\(p::DummyInplace, X) = p.fft \ X
+length(p::DummyInplace) = length(p.fft)
+
```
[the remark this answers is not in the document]
mfherbst: I agree. Will do.

src/core/PlaneWaveBasis.jl
```diff
@@ -105,17 +126,30 @@ Plan a FFT of type `T` and size `fft_size`, spending some time on finding an opt
...
+    # Fall back to FourierTransforms
+    # Note: FourierTransforms has no support for in-place FFTs at the moment
+    # ... also it's extension to multi-dimensional arrays is broken and
+    #     the algo only works for some cases
+    @assert all(is_fft_size_ok_for_generic.(fft_size))
+
+    # opFFT = FourierTransforms.plan_fft(tmp)  # TODO When multidim works
+    opFFT = genplan_fft(tmp)                   # Fallback for now
+    ipFFT = DummyInplace{typeof(opFFT)}(opFFT)
+
```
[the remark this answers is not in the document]
mfherbst: I'm not sure about the kwarg for the moment. FourierTransforms is still pretty unstable. Once it's properly registered and more thoroughly tested than we can think about it.

src/core/self_consistent_field.jl
mfherbst: Because we do not actually use it :)

src/core/energy_ewald.jl
mfherbst: I do still not have a better solution here ... if you find one let me know.

src/core/generic_fftn.jl
mfherbst: done

src/core/lobpcg_hyper_impl.jl
mfherbst: Well ... they've been looking for someone to do it properly. See issue 16446. At the moment I do not have time to do this, also I'm not sure it's worth the effort DFTK in Float16 sounds like a bad idea anyway...

examples/magnesium_supercell.jl
```diff
@@ -0,0 +1,64 @@
...
+    lattice[i, j] = A_to_bohr * get(get(pystruct.lattice.matrix, j-1), i-1)
+end
+composition = [Mg => [s.frac_coords for s in pystruct.sites]]
+
+# Setup PBE model with Methfessel-Paxton smearing and its discretisation
+model = model_dft(lattice, :lda_xc_teter93, composition...;
+                  temperature=Tsmear,
+                  smearing=DFTK.smearing_methfessel_paxton_1)
+fft_size = determine_grid_size(lattice, Ecut)
+basis = PlaneWaveModel(model, fft_size, Ecut, [0, 0, 0.], nothing)
```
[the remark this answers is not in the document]
mfherbst: I'll merge with the other magnesium.

mfherbst:
@antoine-levitt I'm waiting for your final verdict and then this goes in :smile:.

src/core/PlaneWaveBasis.jl
```diff
@@ -100,43 +105,30 @@ end
...
         fft_size = determine_grid_size(model, Ecut)
     end
-
     fft_size = Tuple{Int, Int, Int}(fft_size)
+
+    # TODO generic FFT is kind of broken for some fft sizes
+    if !(T in [Float32, Float64]) && !all(is_fft_size_ok_for_generic.(fft_size))
+        fft_size = next_working_fft_size_for_generic.(fft_size)
+        @info "Changing fft size to $fft_size (smallest working size for generic FFTs)"
+    end
```
antoine-levitt: Move that to `determine_grid_size`

src/core/fft.jl
```diff
@@ -0,0 +1,106 @@
+import FFTW
```
antoine-levitt: move determine_grid_size here? Also add a comment at the top saying that most of this is workarounds because FourierTransforms doesn't conform to AbstractFFTs (yet?)

src/core/PlaneWaveBasis.jl
mfherbst: I thought about that, but I disagree. If you select the wrong size here it's not a matter of making a numerical error it's a matter of actually getting a completely wrong result: It's an implementation detail specific to the generic FFT case and will (hopefully) disappear. It needs to be done at this place, however, since in the `build_fft_plans` I cannot actualy *change* `fft_size`, only read it.

src/core/occupation.jl
```diff
@@ -85,7 +91,14 @@ function find_occupation_fermi_metal(basis, energies, Psi)
     # taken with convergence criteria and the like
     εF = Roots.find_zero(εF -> compute_n_elec(εF) - n_electrons, (min_ε, max_ε),
                          Roots.Bisection())
-    @assert compute_n_elec(εF) ≈ n_electrons
+
+    n_electrons_integration = compute_n_elec(εF)
+    if abs(n_electrons_integration - n_electrons) > sqrt(eps(real(T)))
+        @warn("`find_occupation_gap_zero_temperature` has a mismatch in number " *
```
antoine-levitt: wrong copy-paste here in the error message. Shouldn't these be asserts? It really shouldn't happen...

test/silicon_runners.jl
```diff
@@ -7,7 +7,7 @@ include("testcases.jl")
 # TODO There is a lot of code duplication in this file ... once we have the ABINIT reference
 #      stuff in place, this should be refactored.
 
-function run_silicon_redHF(;Ecut=5, test_tol=1e-6, n_ignored=0, grid_size=15, scf_tol=1e-6)
+function run_silicon_redHF(T; Ecut=5, test_tol=1e-6, n_ignored=0, grid_size=15, scf_tol=1e-6)
```
antoine-levitt: `T=Float64`?

src/core/occupation.jl
mfherbst: Well but it does for some floating point types ... albeit the error is really neglible.

antoine-levitt:
Left of couple of nitpicking comments, but otherwise good to go!

test/silicon_runners.jl
mfherbst: But then also at other places like in `run_silicon_lda` and so on ...
so far we explicitly specify the type everywhere, which I think is actually what you want.

src/core/occupation.jl
mfherbst: Ah no, scratch that ... that's at the other instance, here it can be a proper assert.

src/core/PlaneWaveBasis.jl
antoine-levitt: OK that makes sense. These kinds of ugly temporary workarounds are unfortunately unavoidable, but fine as long as they are explicitly marked as such and isolated as much as possible from "real" code.

src/core/occupation.jl
antoine-levitt: really ? The other one should really be fine also... that's in float16 I guess?

src/core/fft.jl
mfherbst: Agreed.

src/core/occupation.jl
mfherbst: No, it's for more accurate types like Double32 or Double64. I think the reason is because their density of representable floats is plainly different than for IEEE floats, such that `eps(T)` might not always be the best metric, or put another way: The prefactor for an `eps(T)` error bound might just be different, not sure.

antoine-levitt: Seems a bit strange, the density thing shouldn't matter (we don't manipulate huge or tiny numbers). A more likely explanation is that something's converting to a less accurate precision under the radar

antoine-levitt: Can you test if `filled_occ = T(2)` fixes it?

mfherbst: That's also well possible. I'm not sure how `Double32` handles thinks which are not implemented generically ... could well fallback to `Float32` or so.

antoine-levitt: Hm it probably doesn't because `2` is an int literal, and so not converted to Float64 before going into a double float... that's weird.

mfherbst: Also `2` is exactly representable already for `Float16` :P

antoine-levitt: Problem is likely `kweights`.
```
        kweights = [length(symops) for symops in ksymops]
        kweights = kweights / sum(kweights)
```
computes it in Float64. Since the eps for Float64s is about sqrt of the eps for Double64, there you go. This is hugely tricky to do systematically...

mfherbst: Yes that could be it and that definitely should be changed!

antoine-levitt: Places where we have to be careful. https://github.com/stevengj/ChangePrecision.jl is probably a good source of inspiration for things that can go wrong (as well as a good diagnostic tool). So apart from the obvious ones, things that create Float64 behind our backs:
- Irrationals
- Division/sqrt/exp/... of integers

mfherbst: Hmm ... that gets me thinking: Since change precision has a way for changing default precision in `begin` ... `end` environments, we could use that for testing in the `DFTK.jl` module file sorrounding all includes. If our results are kind of agnostic to what we choose, we did a good job.

antoine-levitt: Yeah that's what I meant by "diagnostic tool". So I did a quick grep for "\", things that might get us into trouble is pretty much all of PspHgh, SphericalHarmonics, bzmesh, `A_coeff(n) = (-1)^n / (factorial(n) * 4^n * sqrt(π))`

mfherbst: Ah I see ... puh ... I can't be bothered to fix that all here. I made a separate issue for that, see #60.

mfherbst: Hmm ... but some of that might not actually matter that much, because it's later assigned to arrays of the appropriate type.

antoine-levitt: yeah but you lose precision, which matters if we want to push below Float64. Let's do that later, but keep the assert (which is pretty useful precisely because of this kind of things; in fact a better check would be to check against eps^2/3 or something)

mfherbst: Ok, I dropped the relevant commit.

antoine-levitt:
As usual I'm thoroughly confused by the package system. I get
```
(v1.3) pkg> resolve
 Resolving package versions...
ERROR: Unsatisfiable requirements detected for package FourierTransforms [65ccaadd]:
 FourierTransforms [65ccaadd] log:
 ├─FourierTransforms [65ccaadd] has no known versions!
 └─restricted to versions * by DFTK [acf6eb54] — no versions left
   └─DFTK [acf6eb54] log:
     ├─possible versions are: 0.0.1 or uninstalled
     └─DFTK [acf6eb54] is fixed to version 0.0.1
```
on this branch

mfherbst:
Ok ... I did some digging. Essentially all these problems boil down to the fact that `FourierTransforms` is unregistered. I propose the following: I'll open an issue asking about registration of the package. I would understand if that's not done to General, because it really is a little immature. In that case we have four options:
- Do not allow for the generic stuff at all
- Copy the files to DFTK
- Register a fork in MolSim under a temporary name such that it is clear its a workaround
- Register the package in MolSim as is.

I personally prefer to the third option. The fourth has issues if the package is eventually registered to General as then there might be packages with different versions but the same state around. Third option surely puts some maintenance burden on us, but I'd say that's ok.

What are your thoughts?

antoine-levitt:
As usual I'm for the simplest and dirtiest solution :-p copy the files. It's fine under MIT (with proper attribution of course), it's just a temporary workaround to try stuff and it's the most flexible solution (eg if we want to change things in the FFT code)

mfherbst:
Hmm ... ok ... after some thoughts I agree with you. I closed the issue at FourierTransforms and added the files as a copy.

Issue #60, opened 2019-11-16
Proper type stability


mfherbst:
As discussed https://github.com/JuliaMolSim/DFTK.jl/pull/59 our code is still full of expressions, where a `Float64` result implicitly generated. This should be carefully investigated and made properly type-stable.

antoine-levitt:
Problems are:
- Irrationals, eg T(pi) instead of pi
- Integer division, eg in kweights, bzmesh
- Integer special functions, eg sqrt in PspHgh, SphericalHarmonics

Grepping for `pi`, `/` and `sqrt` *should* be fine and catch them all.

mfherbst:
Once #61 is merged, there are still two aspects missing:
- LDA implementation inside `xc_fallback/`
- `occupation.jl`

antoine-levitt:
Once https://github.com/stevengj/ChangePrecision.jl/pull/11 (or similar) is merged we should be able to use that for LDA

mfherbst:
I feel this has become solved or not relevant any more. Feel free anyone to reopen if they disagree.

PR #61, opened 2019-11-16
Type instabilities

antoine-levitt:
Work towards https://github.com/JuliaMolSim/DFTK.jl/issues/60. Should take care of most of the implicit Float64 conversions.

(based on https://github.com/JuliaMolSim/DFTK.jl/pull/59)

### DIFF
--- a/src/core/PspHgh.jl
+++ b/src/core/PspHgh.jl
@@ -162,13 +162,13 @@ function eval_psp_projection_radial(psp::PspHgh, i, l, qsq::Number)
         #       The first 8 in equation (8) should not be under the sqrt-sign
         #       This is the right version (as shown in the GTH paper)
         if i == 2 return @. common *    2     / sqrt(T(15))  * (3  -   qrsq         ) end
-        if i == 3 return @. common * (4/T(3)) / sqrt(T(°)105) * (15 - 10qrsq + qrsq^2) end
+        if i == 3 return @. common * (4/T(3)) / sqrt(T(105)) * (15 - 10qrsq + qrsq^2) end
     end
 
     if l == 1  # verify expressions
         if i == 1 return @. common * 1     /    sqrt(T(3)) * q end
-        if i == 2 return @. common * 2     /  sqrt(T(105)) * q * ( 5 -   qrsq         ) end
-        if i == 3 return @. common * 4 / T(3) / sqrt(T(°)1155) * q * (35 - 14qrsq + qrsq^2) end
+        if i == 2 return @. common * 2     /    sqrt(T(105)) * q * ( 5 -   qrsq         ) end
+        if i == 3 return @. common * 4 / T(3) / sqrt(T(1155)) * q * (35 - 14qrsq + qrsq^2) end
     end
 
     error("Did not implement case of i == $i and l == $l")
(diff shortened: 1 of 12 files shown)


antoine-levitt:
It's actually quite pleasing how few of these we need. The bzmesh is in rational and the G are ints, so we have things like `recip_lattice * (G+k)` do the right thing all by itself

src/core/PspHgh.jl
```diff
@@ -161,14 +161,14 @@ function eval_psp_projection_radial(psp::PspHgh, i, l, qsq::Number)
         # Note: In the next case the HGH paper has an error.
         #       The first 8 in equation (8) should not be under the sqrt-sign
         #       This is the right version (as shown in the GTH paper)
-        if i == 2 return @. common *    2  / sqrt(15)  * (3  -   qrsq         ) end
-        if i == 3 return @. common * (4/3) / sqrt(105) * (15 - 10qrsq + qrsq^2) end
+        if i == 2 return @. common *    2     / sqrt(T(15))  * (3  -   qrsq         ) end
+        if i == 3 return @. common * (4/T(3)) / sqrt(T(°)105) * (15 - 10qrsq + qrsq^2) end
```
mfherbst: This looks like a typo.

src/core/PspHgh.jl
```diff
@@ -161,14 +161,14 @@ function eval_psp_projection_radial(psp::PspHgh, i, l, qsq::Number)
...
     end
 
     if l == 1  # verify expressions
-        if i == 1 return @. common * 1     /    sqrt(3) * q end
-        if i == 2 return @. common * 2     /  sqrt(105) * q * ( 5 -   qrsq         ) end
-        if i == 3 return @. common * 4 / 3 / sqrt(1155) * q * (35 - 14qrsq + qrsq^2) end
+        if i == 1 return @. common * 1     /    sqrt(T(3)) * q end
+        if i == 2 return @. common * 2     /  sqrt(T(105)) * q * ( 5 -   qrsq         ) end
+        if i == 3 return @. common * 4 / T(3) / sqrt(T(°)1155) * q * (35 - 14qrsq + qrsq^2) end
```
mfherbst: Here, too.

mfherbst:
I agree ... if what you committed so far is all that's needed ... that would be pretty good!

src/core/PspHgh.jl
antoine-levitt: Thanks! shift instead of ctrl...

mfherbst: :+1: Isn't everything escape-meta-alt-control-shift in your editor? :smile:

antoine-levitt: ctrl-) is slurp: if you have `105` and want to wrap it into a `T()`, you do put your cursor before and do `T ( ctrl-)` Very useful for lisp...

antoine-levitt:
So lda_c_vwn.jl would be pretty annoying to convert to generic. I think the only automatic option is to use `ChangePrecision` inside a generated function.

antoine-levitt:
I think apart from `xc_fallback/` and `occupation.jl` we're fine.

antoine-levitt:
Something like this appears to work:
```
using ChangePrecision
@generated function f(x::T) where T
    @changeprecision T begin
        :(2/3*x)
    end
end
```

Issue #25, opened 2019-10-25
File structure


antoine-levitt:
Move everything to toplevel, and then make folders for specific things, like

- eigensolvers
- scf
- utils
- pseudo
- brillouin
- terms

PR #62, opened 2019-11-19
Change file structure

antoine-levitt:
Fixes https://github.com/JuliaMolSim/DFTK.jl/issues/25.

This is the minimal fix. Everything that did not fit neatly in one folder I just left at the toplevel, which I think is fine. I thought of having a "PW" folder, but then it would have ended up containing half the code so I decided against.


mfherbst:
I generally agree, some comments:
- I would rename `ext` to `external` to make it more clear.
- The `etsf_nanoquanta.jl` is not really an external library, but a parser for a file format, so the description "Interfaces to external libraries." does not completely fit. I'd say "Interfaces to third-party libraries and resources" to make that more clear.

I'll tag 0.0.2 after this is merged to allow to explicitly depend on the new file structure.

Issue #43, opened 2019-10-30
Implement conversions between grids


antoine-levitt:
Something that would be quite useful is to be able to convert quantities (densities, wavefunctions) between grids (different lattices, different ecuts, different fft_sizes). Eg a more generalized version of `interpolate_at_kpoint`. Doesn't need to be accurate.

mfherbst:
Absolutely! I thought about having more or less a hierarchy of methods (kpoints, then full plane wave grids then lattices).

antoine-levitt:
Can we close this are there conversion methods still needed? @gkemlin

mfherbst:
I think we have a good first set. I'll close this for now.

PR #64, opened 2019-11-21
Interpolation

antoine-levitt:
Start to address https://github.com/JuliaMolSim/DFTK.jl/issues/43. It adds a dependency on Interpolations. This is pretty annoying as `using Interpolations` takes a few seconds. Don't have a better idea for now. One could use Fourier interpolation, but that's not quite as general, ie it wouldn't support arbitrary grids.


test/interpolation.jl
```diff
@@ -0,0 +1,27 @@
+using DFTK: interpolate_density
+using LinearAlgebra
+
+@testset "Interpolation" begin
+    lattice = Array{Float64}(I(3))
```
mfherbst: Does not work on Julia 1.1 ... only came later.

src/Density.jl
```diff
@@ -31,3 +32,62 @@ function check_density_real(ρreal)
         @warn "Large imag(ρ)" norm_imag=norm(imag(ρreal))
     end
 end
+
+"""
+Interpolate a function expressed in a basis `b_in` to a basis `b_out`
+This interpolation uses a very basic real-space algorithm, and makes a DWIM-y attempt to take into account the fact that b_out can be a supercell of b_in
+TODO move this outside of density
+"""
+function interpolate_density(ρ_in::Density, b_in::PlaneWaveBasis, b_out::PlaneWaveBasis)
```
mfherbst: `Density` already contains a basis object, can't you just use that?
Also: I feel `interpolate_density(b_in, b_out, ρ_in) more natural, however, `interpolate_density(ρ_in, b_out)` ...

mfherbst:
Hmm I don't have a good idea. Maybe it is possible to load a package lazily only if it is truely needed in some way ... like in `python`. That way you avoid the cost of loading it without need. I'm not sure that's anyhow supported in julia as right now imports have to be on global scope. There is https://github.com/MikeInnes/Requires.jl, which might be a step in an interesting direction, but it seems to aim at a different problem.

mfherbst:
By the way, on my machine:
```julia
julia> @time using Interpolations
  0.746871 seconds (1.55 M allocations: 96.510 MiB, 3.89% gc time)
```
That's not great, but not too bad.

mfherbst:
Actually: I just noted, that you can use `include("file.jl")` from inside other functions and then `file.jl` may contain `using` and `import` statements, which are only loaded once the function is run. Not sure about potential other performance implications, but maybe something to try.

antoine-levitt:
I think Requires is the way to do this; if we wrap this function inside a `@require Interpolations`, it will only be available once Interpolations is loaded. But let's worry about this later, I guess: load times are supposed to improve with new releases of julia.

mfherbst:
That's definitely a good thing ... still, it was fun to research this a little :).

src/Density.jl
antoine-levitt: Absolutely, I missed this, thanks!

PR #65, opened 2019-11-23
PWBasis fixes and additions

antoine-levitt:


src/PlaneWaveBasis.jl
```diff
@@ -162,10 +169,21 @@ end
...
-AbstractFFTGrid = Union{AbstractArray{T, 4}, AbstractArray{T, 3}} where T
+# There are two kinds of FFT/IFFT functions: the ones on `C_ρ` (that
+# don't take a kpoint as input) and the ones on `B_k` (that do)
+# 
+# Quantities in real-space are 3D arrays, quantities in reciprocal
+# space are 3D (`C_ρ`) or 1D (`B_k`). We take as input either a single
+# array (3D or 1D), or a bunch of them (4D or 2D)
+
+Array_3or4D = Union{AbstractArray{T, 3}, AbstractArray{T, 4}} where T
+Array_1or2D = Union{AbstractArray{T, 1}, AbstractArray{T, 2}} where T
```
mfherbst: Hmm. First of all that breaks with the camel case naming convention we use for types. Otherwise I wonder, why you do not want to use `AbstractVecOrMat`? If you do not like `AbstractFFTGrid`, how about `AbstractArray3dOr4d` or `AbstractArray3Or4`?

antoine-levitt: I don't like Grid because it's not a grid, it's an array. Abstract feels like it doesn't add too much information, so I removed it for short. I did not use VecOrMat because I wanted to show the parallelism to 3or4. I'm fine with camel casing, removing or adding the d in lower or upper case, and adding or removing underscores are you like.

mfherbst: I'd keep the abstract, because it is abstract and that's what Julia Base keeps using everywhere for such things (debatable, but usually it's best to stick to the stdlib). Let's settle for `AbstractArray1or2D` and `AbstractArray3or4D`?

Issue #31, opened 2019-10-27
Adaptive kinetic preconditioning


antoine-levitt:
The usual way of doing preconditioning with the kinetic energy is essentially `1/(<K> + K)`, where K is the kinetic energy, and <K> = <psi, K psi> where psi is the vector under consideration (right now we take <K> constant). This means that PR = P\R does not work, and one needs a more general mechanism PR = P(X,R). There are even more general kinds of preconditioning that do not even fit in P(X,R) but instead require P(X,G) where G is the unpreconditioned gradient, with which we might want to play with @benjaminstamm. We should figure out a general mechanism for doing that.

PR #66, opened 2019-11-23
Adaptive preconditioning

antoine-levitt:
Fix https://github.com/JuliaMolSim/DFTK.jl/issues/31


antoine-levitt:
So `Optim.jl` deals with it differently, by introducing a method `precondprep!` and doing `precondprep!(P, x); ldiv!(p, P, r)` with the fallback `precondprep!(P, x) = P`. Interfaces are hard.

mfherbst:
I like the `Optim.jl` interface better, since (a) it integrates better with standard Julia: `ldiv!` is really what should be used for this and makes our preconditioners compatible with e.g. `IterativeSolvers.jl`. (b) It is easier to play with preconditioners: One could just use a bare matrix and it worked.

src/eigen/diag.jl
```diff
@@ -50,7 +50,7 @@ function diagonalise_all_kblocks(eigensolver, ham::Hamiltonian, nev_per_kpoint::
         @assert size(guessk) == (length(kpoints[ik].basis), nev_per_kpoint)
 
         results[ik] = eigensolver(kblock(ham, kpt), guessk;
-                                  prec=kblock(prec, kpt), tol=tol,
+                                  prec=prec(ham, kpt), tol=tol,
```
mfherbst: Puh ... that took me a second to understand. Could you make a comment to make sure that before `prec` is a type and after an instance.

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -258,7 +258,7 @@ function LOBPCG(A, X, B=I, precon=I, tol=1e-10, maxiter=100; ortho_tol=2eps(real
             resids[i+nlocked,niter] = norm(R[:,i])
         end
         vprintln(niter, "   ", resids[:, niter])
-        R = precon\R
+        precon(R, new_X, R)
```
mfherbst: See my suggestion to use `ldiv!` as the interface.

test/lobpcg.jl
```diff
@@ -61,7 +61,7 @@ end
     basis = PlaneWaveBasis(model, Ecut, silicon.kcoords, silicon.ksymops; fft_size=fft_size)
     ham = Hamiltonian(basis)
 
-    res = diagonalise_all_kblocks(lobpcg_hyper, ham, 5, tol=1e-8, prec=PreconditionerKinetic(ham, α=0.1),
+    res = diagonalise_all_kblocks(lobpcg_hyper, ham, 5, tol=1e-8,
                                   interpolate_kpoints=false)
```
mfherbst: Drop linebeak here.

antoine-levitt:
Yeah you're right, and we'll have to do it like that for direct min anyway. Sucks a bit to have to split the computation in two but I'll have to deal with it. I'll do it when I get in front of a keyboard.

antoine-levitt:
How annoying, julia 1.1 doesn't have `ldiv!(I, x)`. Is it OK to require 1.2, or should we hack it?

mfherbst:
Agreed and done. Also I removed PreconditionerNone, because there really is no point in having it. I also made an `AbstractPreconditioner` to have a generic `ldiv!` fallback for the 2-argument version.  Actually now we could also completely drop the 2-argument version, I find it kind of unintuitive anyway.

One issue I still have is that now our LOBPCG implementation is no longer independent of the rest of our code (because it assumes `precondprep!`. If we keep going down that route it might get harder and harder to contribute it back to general package eventually. Just a remark.

antoine-levitt:
Problem with PreconditionerNone is that otherwise we need to pass (ham, kpt) - > I for the prec_type

antoine-levitt:
Not sure about the fallback. Why do we need that? It's the responsability of the individual types to support the two forms, and I even have a PR to base (I'm on my phone and can't find the number now, but it's a recent PR I made) to define the reverse fallback because Nlsolve and DiffEq did it

antoine-levitt:
Precondprep can go into a generic package, it just needs a default fallback that does nothing, and then we explicitly import and override it on our end. That's what happens with Optim

mfherbst:
> Problem with PreconditionerNone is that otherwise we need to pass (ham, kpt) - > I for the prec_type

Ok ... good point ... I use nothing now to explicitly flag that.

PR #69, opened 2019-11-28
Modifications to support IntervalArithmetic

mfherbst:
Small modifications for intervals. What's missing:

- [x] A test to ensure compatibility with IntervalArithmetic does not break randomly.


src/FourierTransforms.jl/ctfft.jl
```diff
@@ -1,4 +1,13 @@
 using Primes
+using IntervalArithmetic
+import Base: cbrt
+AbstractFloatOrInterval = Union{AbstractFloat, Interval}
+
+# Monkey-patch cbrt for Intervals ... should be done more properly by changing
+# the rounding mode ...
+# TODO Remove this once https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/310
+#      is dealt with.
+cbrt(i::Interval) = Interval(prevfloat(cbrt(i.lo)), nextfloat(cbrt(i.hi)))
```
antoine-levitt: Remove that (it can go into the test script)

src/FourierTransforms.jl/ctfft.jl
```diff
@@ -45,7 +54,7 @@ plan_inv(p::CTPlan{T}) where T =
     ScaledPlan(invCT(p), normalization(real(T), p.n, 1))
 
 # steps for pregenerated kernels:
-function ωpow(T::Type{<:Complex}, n, i)
+function ωpow(T::Type{<:Complex{<:AbstractFloatOrInterval}}, n, i)
```
antoine-levitt: Widen to `Real` instead

src/FourierTransforms.jl/ctfft.jl
```diff
@@ -374,13 +383,17 @@ const CTComplex = Union{Complex{Float32},Complex{Float64}} # types of pregenerat
...
-for forward in (true,false)
-    @eval @twiddle($(Complex{<:AbstractFloat}), $forward, 2)
-    @eval @twiddle($(Complex{<:AbstractFloat}), $forward, 4)
+for T in [AbstractFloat, Interval]
+    # This loop is kind of bad, because it creates the code twice for
+    # both abstract types respectively ... that's not what we want, but it works ...
+    for n in (1,2,4)
+        @eval @nontwiddle($(Complex{<:T}), true, $n)
+        @eval @nontwiddle($(Complex{<:T}), false, $n)
+    end
```
antoine-levitt: I don't understand why this is needed?

src/fft.jl
```diff
@@ -106,9 +107,13 @@ function determine_grid_size(lattice::AbstractMatrix{T}, Ecut; supersampling=2,
...
         Vec3([nextprod([2, 3, 5], 2gs + 1) for gs in Gmax])
     else
-        Vec3([2gs+1 for gs in Gmax])
+        Vec3([2gs + 1 for gs in Gmax])
     end
 end
+function determine_grid_size(lattice::AbstractMatrix{T}, Ecut;
+                             kwargs...) where T <: Interval
+    determine_grid_size(mid.(lattice), Ecut; kwargs...)
+end
```
antoine-levitt: Why is this needed?

src/FourierTransforms.jl/ctfft.jl
mfherbst: I did that in the first place, but I do not think that's the right thing to do here.

There is some discussion in IntervalArithmetic.jl to make (some) intervals a subtype of AbstractFloat. That seems more logical, because intervals satisfy some promises that Reals might not (e.g. they implement `eps`) and are in that sense floating point numbers.

To have a counterexample: Note, that `Int <: Real`, but the FFT code is certainly not suited for  `Int`.

src/fft.jl
mfherbst: Because `ceil(Int, ::Interval{Float64})` cannot be properly implemented (`Interval{Int}` does not exist). I'll add a comment.

src/FourierTransforms.jl/ctfft.jl
mfherbst: Not sure exactly, what you are referring to.

src/FourierTransforms.jl/ctfft.jl
mfherbst: Ok, will do.

src/FourierTransforms.jl/ctfft.jl
antoine-levitt: Sure but in the meantime it's easier as a hack to widen the FFT code more than necessary than to add a dep on intervals

src/fft.jl
antoine-levitt: I think the right thing to do here is to define ceil in IntervalArithmetic to ceil when possible and error it cannot be unambiguously defined

antoine-levitt: What I want to avoid is "polluting" our code with this. We can avoid having the main code depend on intervals by widening the FFT code and defining the methods we need in a script

mfherbst: Hmm ... that's probably fair, but I figured we'll probably depend on Intervals anyway if we start to use them more seriously. Also, if we want to test our code with Intervals, that it's at least a test dependency (though not an install one) :). I'll check how difficult it is to get this into IntervalArithmetic directly.

src/FourierTransforms.jl/ctfft.jl
mfherbst: If it's the only place we truely need the intervals in the main code, then I agree. I only changed it after I noticed the midpoint thing below.

src/fft.jl
antoine-levitt: We can isolate that in a test, while it's not in the main IntervalArithmetic. Better to pirate in a test than in the main code. Also not having to load it would help the loading times. I agree it's a test dependency though; can we express that? Ie depend on something only for tests?

mfherbst: Sure ... test-only dependencies are easy. That's done for `DoubleFloats` already.

src/fft.jl
```diff
@@ -1,5 +1,6 @@
 import FFTW
 import Primes
+using IntervalArithmetic
```
antoine-levitt: Remove this, no?

antoine-levitt: Oh yeah, CI agrees with me :D

mfherbst: That's an oops, sure will be removed.

PR #71, opened 2019-11-29
Kerker

antoine-levitt:
Test case: supercell of magnesium in one direction, with random deviations to the positions:
```
Random.seed!(0)
composition = [Mg => [x + .1*randn(3) ./ supercell for x in composition[1][2]]]
```
where Kerker clearly dominates.


mfherbst:
A few comments and questions:
- I do not see the difference between `scf_damping_solver` and `SimpleMixing()`. Does it even make sense to have both of them? Maybe we should just have an `scf_roothaan_solver`, which is completely stupid and then use `SimpleMixing()` on top to get what we now call `scf_damping_solver`.
- I'm not a big fan of the interface `mix(m::Nothing, basis, ρin, ρout) = ρout`:
    * I would not call the function `mix`, because it's a bit vague. Why not `mix_steps`?
    * The mixing is not specific to densities, if I understand the theory right, so in the `KerkerMixing` and `SimpleMixing` I would not talk about `ρin` and `ρout`, more `xold` and `xnext`. Also I would use `ρold` and `ρnext` ...
    * Why is the basis passed to `mix` and not part of e.g. `KerkerMixing`. Surely for more complex mixing schemes they might per se become dependent on the model or the basis. Perhaps it makes sense to precompute some stuff at the level of the class. It's anyway short-lived inside the SCF and so a separation like the one we did between "Model" and "Hamiltonian" and "HamiltonianBlock" is probably not neccessary.
- Some more comments I'll put into the code.

src/scf/mixing.jl
```diff
@@ -0,0 +1,27 @@
...
+struct KerkerMixing{T <: Real}
+    q0::T
+end
+KerkerMixing() = KerkerMixing(1)
+function mix(m::KerkerMixing, basis, ρin::Density, ρout::Density)
+    Gsq = [sum(abs2, basis.model.recip_lattice * G)
+           for G in basis_Cρ(basis)]
+    ρin = fourier(ρin)
+    ρout = fourier(ρout)
+    ρnext = @. ρin + (ρout - ρin) * Gsq / (m.q0 + Gsq)
```
mfherbst: Combine into one line?

src/scf/mixing.jl
```diff
@@ -0,0 +1,27 @@
...
+    density_from_fourier(basis, ρnext)
+end
+
+struct SimpleMixing{T <: Real}
+    α::T
+end
+SimpleMixing() = SimpleMixing(1)
+function mix(m::SimpleMixing, basis, ρin::Density, ρout::Density)
+    ## TODO optimize this by defining broadcasting directly on density objects?
+    density_from_real(basis, real(ρin) .+ m.α.*(real(ρout) .- real(ρin)))
```
mfherbst: Spacing :)

test/scf_compare.jl
```diff
@@ -27,4 +27,11 @@ include("testcases.jl")
         ρ_alg = fourier(scfres.ρ)
         @test maximum(abs.(ρ_alg - ρ_nl)) < 30tol
     end
+
+    # Run other mixing with nlsolve (the others are too slow...)
+    for mixing in (KerkerMixing(), SimpleMixing(), nothing)
+        scfres = self_consistent_field!(Hamiltonian(basis, ρ0), n_bands, tol=tol, solver=scf_nlsolve_solver(), mixing=mixing)
```
mfherbst: Pretty long line.

src/scf/self_consistent_field.jl
```diff
@@ -55,9 +56,12 @@ function self_consistent_field!(ham::Hamiltonian, n_bands;
     # We do density mixing in the real representation
     # TODO do the mixing in Fourier space instead?
     function fixpoint_map(x)
-        res = iterate_density!(ham, n_bands, density_from_real(basis, x);
+        ρin = density_from_real(basis, x)
```
mfherbst: `ρold`?

src/scf/self_consistent_field.jl
```diff
@@ -55,9 +56,12 @@ function self_consistent_field!(ham::Hamiltonian, n_bands;
...
+        res = iterate_density!(ham, n_bands, ρin;
                                Psi=Psi, eigensolver=eigensolver, tol=diagtol)
-        real(res.ρ)
+        ρout = res.ρ
```
mfherbst: `ρnext`

src/scf/self_consistent_field.jl
```diff
@@ -55,9 +56,12 @@ function self_consistent_field!(ham::Hamiltonian, n_bands;
...
+        ρnext = mix(mixing, basis, ρin, ρout)
+        real(ρnext)
```
mfherbst: `real(mix_steps(mixing, basis, ρold, ρnext))`?

antoine-levitt:
> I do not see the difference between scf_damping_solver and SimpleMixing(). Does it even make sense to have both of them? Maybe we should just have an scf_roothaan_solver, which is completely stupid and then use SimpleMixing() on top to get what we now call scf_damping_solver.

There's the formulation of the fixed-point, and then there's how to solve it. Both can be combined, and turn out to be equivalent in some cases. I see no disadvantages in having both.

`mix` is a bit generic yeah, and also awkward because "mixing" refers to many things in that context. But mix_steps is even worse, I think...

It's not specific to densities yeah, but that's the only thing we do now, so we'll generalize when we get to that I think. The first order of business would be to generalize `Density`, because we also store potentials in that format.

For basis, it's simple as it is but I'm fine with whatever if you don't like it.

antoine-levitt:
But I will add a comment for the mixing. (sorry these comments are a bit messy, I'm on my phone)

test/scf_compare.jl
antoine-levitt: It's a test :-p

antoine-levitt:
For basis etc, I agree with your comment, it comes up in slightly different forms in various places, and I think we should have something more systematic. Let's discuss.

test/scf_compare.jl
mfherbst: Sure, but I'll still need to read that code from time to time. :-p

mfherbst:
> There's the formulation of the fixed-point, and then there's how to solve it. Both can be combined, and turn out to be equivalent in some cases. I see no disadvantages in having both.

Ok. Just wanted to point that out.

> But mix_steps is even worse, I think...
I agree, but I couldn't really come up with something better. Some more ideas: `apply_mixing`, `mix_scf`?

> For rho in out next : no that's the point. Rho in is rho n, rho out is the scf of rho in, and rho next is rho n+1. That's more or less standard terminology and it's pretty clear like that, no?

Well, not so standard for me at least, I've never consciously encountered that notation. Some more ideas: (`ρold`, `ρnew`, `ρnext`) or (`ρprev`, `ρscf`, `ρnext`) or (`ρold`, `ρnext`, `ρmixed`)?

Btw: The part that p particularly strikes me is that in `mix(...)` the `ρout` is an *input* argument to the function.

Also: Why do we not make mixing in-place, like modify `ρout` (in your convention) directly? Then we could also have the syntax `mix!(ρnext, mixing_method, ρprev)` or the alike. For me it is then also completely clear what is the order of the arguments whereas for `mix(method, ρin, ρout)` I'm not so sure.

antoine-levitt:
I've definitely seen it in a few papers. The julia convention of making mutation explicit means that I have no trouble parsing `mix(rhoin, rhoout)` but I see your point. Will think about it ; it doesn't really matter that much.

RE mutation, I really really don't want to. I've been consistently and consciously avoiding it, and removing it when I can (see scf PR). Mutation makes it hard to see what's going on, and also makes it hard on the compiler / tooling (eg ad with zygote). It makes for nastier code, and deviates from mathematical notation. Also it's inconsistent : some functions mutate, some not, and I can never remember which do. The only reason to do it is performance, and that is not a concern here (if it becomes one later on we can always optimize; premature optimization is the root of all evil and so on).

test/scf_compare.jl
antoine-levitt: fix your editor ;-)
(I broke the line)

antoine-levitt:
I thought about names: rho_in, rho_out and rho_next are OK for me. There's no cleaner way to describe it, I find, except to give it a name (like F or something) which is nonstandard and will be confusing. It's used in https://iopscience.iop.org/article/10.1088/1361-648X/ab31c0/pdf among others. I don't like "mixing" as it can denote two distinct concepts: this PR (which is really about preconditioning, although I don't like that term either here...) and eg anderson. I don't have a better alternative though...

test/scf_compare.jl
mfherbst: I am happy with 3 letters, 5 is harder to type (and aliases are cheating!) :smile:

antoine-levitt: Ah, those weird people who type the command name of their editor. Can't remember the last time I did that: I just run emacs when my computer boots and do everything from there.

mfherbst: Sure, I mean that's why it's not an editor but an operating system :P

mfherbst:
Thanks for the discussion! I hope I'm not too much of a pain ...

Ok, I'm convinced about the `ρin`, `ρout`, `ρnext` busyness, if it's standard, than it is as it is. I see the point about mutation, too. I'll think a bit about a better name ... if nothing comes to mind, then we'll leave it as is and merge tmr.

antoine-levitt:
No, that's the point of doing these things! I'm much more of a pain for your PRs anyway...

We could call it preconditioning (because that's what it is). Then it'd be much more natural to take as input residuals (ie rho_out - rho_in).

mfherbst:
You mean the classes or the functions? I think calling the classes like `KerkerMixing` is not such a bad idea. For the interface: Did you have something in mind like we do for the eigensolvers (i.e. `precondprep!` and `ldiv!`) ? `ldiv!` to the residuals makes some sense ... since it kind of is a matrix. Also on the setup one can just pass the basis, much like in https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/eigen/preconditioners.jl.

antoine-levitt:
So, I made an honest effort, I even started renaming to preconditioners, but in the end I didn't like it. Mix is such a nice name, the only problem is that it's sometimes used for both preconditioning and acceleration. I added some comments. The interface is not that bad; this is quite internal anyway...

antoine-levitt:
The reason I didn't go with ldiv/mul is that's really supposed to be a linear operation, which mixing is not necessarily (eg there's a method that solves a nonlinear Thomas-Fermi equation to do mixing)

PR #72, opened 2019-11-29
refactor occupations, and add some comments

antoine-levitt:


src/occupation.jl
```diff
@@ -1,5 +1,17 @@
...
+function filled_occupation(basis)
+    @assert basis.model.spin_polarisation in (:none, :spinless)
+    if basis.model.spin_polarisation == :none
+        @assert basis.model.n_electrons % 2 == 0
+        filled_occ = 2
+    else
+        filled_occ = 1
+    end
+    filled_occ
+end
```
mfherbst: Put this into `PlaneWaveModel.jl` ?

src/occupation.jl
```diff
@@ -9,35 +21,31 @@ function find_occupation_gap_zero_temperature(basis, energies, Psi)
...
-        filled_occ = 2
-    else
-        n_occ = n_electrons
-        filled_occ = 1
-    end
-    @assert n_bands ≥ n_occ
     @assert basis.model.temperature == 0.0
 
+    filled_occ = filled_occupation(basis)
+    n_fill = div(n_electrons, 2)
```
mfherbst: This is not true for `spin_polarisation == :collinear` for example. I'd suggest to add a comment and an assert.

src/occupation.jl
```diff
@@ -86,27 +92,21 @@ function find_occupation_fermi_metal(basis, energies, Psi)
     εF = Roots.find_zero(εF -> compute_n_elec(εF) - n_electrons, (min_ε, max_ε),
                          Roots.Bisection())
     @assert compute_n_elec(εF) ≈ n_electrons
+    minocc = maximum(minimum.(compute_occupation(εF)))
+    if minocc > .01
+        @warn "One kpoint has minimum occupation $minocc, you should increase the number of bands"
```
mfherbst: I'd specify the issue more explicitly. `"Very low occupation of $minocc detected on one kpoint, you should ..."`

PR #73, opened 2019-11-29
DOS

antoine-levitt:
@mfherbst the band structure plotting fails with a division by zero, do you know what's up?


antoine-levitt:
Also refactored band structure a bit (I know this is a useless comment, this is to test comment posting with magit, which is pretty incredible)

examples/magnesium_pbe.jl
```diff
@@ -48,17 +48,12 @@ for key in sort([keys(energies)...]; by=S -> string(S))
...
-bsplot = plotter.BSPlotter(bs)
-plt = bsplot.get_plot()
-plt.autoscale()
-plt.savefig("magnesium_pbe.pdf")
-plt.legend()
-plt.show()
+# Plot DOS
+εs = range(minimum(minimum(scfres.orben))-1, maximum(maximum(scfres.orben))+1, length=1000)
+Ds = DOS.(εs, Ref(basis), Ref(scfres.orben), T=Tsmear*4, smearing=DFTK.smearing_methfessel_paxton_1)
+using PyPlot
```
mfherbst: Move to the top of the file.

examples/magnesium_pbe.jl
```diff
@@ -48,17 +48,12 @@ for key in sort([keys(energies)...]; by=S -> string(S))
...
-bs = pymatgen_bandstructure(basis, band_data, klabels, fermi_level=scfres.εF)
-bsplot = plotter.BSPlotter(bs)
-plt = bsplot.get_plot()
-plt.autoscale()
-plt.savefig("magnesium_pbe.pdf")
-plt.legend()
-plt.show()
+# Plot DOS
+εs = range(minimum(minimum(scfres.orben))-1, maximum(maximum(scfres.orben))+1, length=1000)
+Ds = DOS.(εs, Ref(basis), Ref(scfres.orben), T=Tsmear*4, smearing=DFTK.smearing_methfessel_paxton_1)
```
mfherbst: I'd prefer one space around the operators ...

On that note: I think we should start to settle on some syntax agreements like that at some point and write it down somewhere.

src/postprocess/band_structure.jl
```diff
@@ -30,3 +30,20 @@ function compute_bands(ham::Hamiltonian, kpoints, n_bands;
...
+    # Band structure calculation along high-symmetry path
+    kpoints, klabels, kpath = high_symmetry_kpath(basis, kline_density, composition...)
+    println("Computing bands along kpath:\n     $(join(kpath[1], " -> "))")
+    band_data = compute_bands(ham, kpoints, n_bands)
+
+    # Plot bandstructure using pymatgen
+    plotter = pyimport("pymatgen.electronic_structure.plotter")
+    bs = pymatgen_bandstructure(basis, band_data, klabels, fermi_level=εF)
+    bsplot = plotter.BSPlotter(bs)
+    plt = bsplot.get_plot()
```
mfherbst: Don't you think this should go to `src/external/pymatgen.jl` instead? I think it's wise to keep the explicit pymatgen-dependent code to one place inside DFTK.

src/postprocess/band_structure.jl
```diff
@@ -30,3 +30,20 @@ function compute_bands(ham::Hamiltonian, kpoints, n_bands;
...
+    plt.autoscale()
+    plt.legend()
+    plt.show()
```
mfherbst: Having a call to `show()` in a library function is a bad idea in my opinion, because a user might want to add more stuff to the plot, which you prevent in this way. Also actually the legend thing does not work properly anyway ... so you can remove that, too ;).

src/postprocess/DOS.jl
```diff
@@ -0,0 +1,41 @@
...
+
+"""
+Total density of states at energy ε
+"""
+function DOS(ε, basis, orben; smearing=basis.model.smearing, T=basis.model.temperature)
+    D = zero(ε)
+    T != 0 || error("DOS only supports finite temperature")
+    for ik = 1:length(orben)
+        for iband = 1:length(orben[ik])
+            D += basis.kweights[ik]*ForwardDiff.derivative(ε -> smearing((orben[ik][iband]-ε)/T), ε)
```
mfherbst: Spaces around operators missing.

src/postprocess/DOS.jl
```diff
@@ -0,0 +1,41 @@
+# Compute densities of states
+
+# IDOS (integrated)
+# N(ε) = sum_n f_n = sum_n f((εn-ε)/T)
+# DOS
+# D(ε) = N'(ε)
+# LDOS (local)
+# LD = sum_n f_n ψn
```
mfherbst: Perhaps use the non-abbreviated form of the terms at least once?

src/postprocess/DOS.jl
```diff
@@ -0,0 +1,41 @@
...
+    end
+    D
+end
+
+"""
+Local density of states, in real space
+"""
+function LDOS(ε, basis, orben, Psi; smearing=basis.model.smearing, T=basis.model.temperature)
+    D = zeros(typeof(ε), basis.fft_size)
+    T != 0 || error("DOS only supports finite temperature")
```
mfherbst: Typo: `"LDOS only supports ...`

src/postprocess/DOS.jl
```diff
@@ -0,0 +1,41 @@
...
+    for ik = 1:length(orben)
+        ψreal = G_to_r(basis, basis.kpoints[ik], Psi[ik])
+        for iband = 1:length(orben[ik])
+            D += basis.kweights[ik] *
+                ForwardDiff.derivative(ε -> smearing((orben[ik][iband]-ε)/T), ε) *
+                abs2.(ψreal[:, :, :, iband])
```
mfherbst: I like for breaks such as this to have the operator as the first thing in the next line (especially for sums this is helpful), also spacing around operators :smile:.

src/postprocess/DOS.jl
```diff
@@ -0,0 +1,41 @@
...
+        end
+    end
+    D
+end
+
+"""
+Local density of states, in real space
+"""
+function LDOS(ε, basis, orben, Psi; smearing=basis.model.smearing, T=basis.model.temperature)
+    D = zeros(typeof(ε), basis.fft_size)
```
mfherbst: I'm not sure `typeof(ε)` is the right expression to deduce the type here. Don't you think `real(eltype(Psi[1]))` or `eltype(orben[1])` is better?

src/postprocess/DOS.jl
```diff
@@ -0,0 +1,41 @@
...
+# D(ε) = N'(ε)
+# LDOS (local)
+# LD = sum_n f_n ψn
+
+using ForwardDiff
+
+"""
+Total density of states at energy ε
+"""
+function DOS(ε, basis, orben; smearing=basis.model.smearing, T=basis.model.temperature)
```
mfherbst: Maybe also have a version that takes an `scfres`, because that directly contains ε, basis and orben?

examples/magnesium_pbe.jl
antoine-levitt: Sure. Very far at the bottom of my priority list though. This is the kind of things that can be done after the fact with automatic formatting tools. There are a couple of active ones: https://github.com/julia-vscode/DocumentFormat.jl https://github.com/domluna/JuliaFormatter.jl

src/postprocess/band_structure.jl
antoine-levitt: We can move it out later if you want. That's a postprocessing function, as far as I'm concerned it can go either as API in pymatgen (in which case we need to refactor it a bit) or postprocessing

src/postprocess/band_structure.jl
antoine-levitt: I don't really see that as a library function, postprocessing is like a collection of utility functions you can call to help your analyses. If it doesn't do exactly what you want you take the code out and tweak it. I removed the legend.

examples/magnesium_pbe.jl
mfherbst: Awesome ... I did not know there was something in the pipeline :smile:

src/postprocess/DOS.jl
antoine-levitt: I'm intensely relaxed about that. If the user wants to mix precision they're responsible for ensuring everything is in the proper type. Eps is as good as anything here, but I can change it if you don't like it

src/postprocess/band_structure.jl
mfherbst: I'm really not too sure, the `high_sym_kpath` function also depends on it, so one should split the API there as well. It's just that pymatgen really is a big beast and things such as the `ZeroDivisionError` really start to annoy me a bit, so I'd like to keep the option open to easily change at some later point.

I can agree to your definition of post-processing function, but then make a quick note about that in the readme (i.e. these are just stand-alone tools, potentially depending on further third-party code and not part of DFTK, something like that)

src/postprocess/DOS.jl
antoine-levitt: I have that in my radar anyway, we should factor a number of objects into a proper `ElectronicState` struct or something, but I'm not 100% sure about the API (ie should psi be with the hamiltonian of which it is the ground state, or with the hamiltonian H[psi] it creates?). Let's discuss that next time.

src/postprocess/band_structure.jl
mfherbst: Also the show works differently in Julia and can actually be removed without any impact (I just tried it). Still, returning the `plt` object to the user is probably helpful.

src/postprocess/DOS.jl
mfherbst: Usually I used `real(eltype(Psi[1]))`, so for consistency I'd prefer that.

src/postprocess/DOS.jl
mfherbst: Ja not sure about the name, but I agree that would be a good thing to have. Let's talk next week.

src/postprocess/DOS.jl
antoine-levitt: But then it's not consistent with DOS... but sure, why not

mfherbst: hmmm ... true thing.

mfherbst:
I found the source of the pymatgen error ... the lattice vectors are transposed: We used pymatgen for the lattice setup beforehand and I did not transpose the vectors when moving to our own setup functions. With the transposed lattice everything is fine.

mfherbst:
Ups ... I made a wrong push ... can you merge on your end and push again?

antoine-levitt:
Done (I think)

mfherbst:
Yes, thanks.

antoine-levitt:
Nice! Is it feasible to have an assert somewhere to prevent this in the future?

mfherbst:
Puh ... not sure if that can be done easily, because from the point of view of the code what I gave is just very messed-up input. For me pymatgen prints a warning about a strange unit cell ... so that could be your pointer if you want.

antoine-levitt:
OK! Well I was doing my convergence tests on that system so might have that a bit... :-)

mfherbst:
This can be merged as such from my point of view.

antoine-levitt:
Thanks for the review and amends!

antoine-levitt:
Will merge when CI is green

PR #80, opened 2019-12-07
Move to new pseudo datastructures

mfherbst:
I made the filename all-lowercase to allow simple normalisation (i.e. both `LDA` and `lda` are valid and so on) and added the `hgh` extension to all files, since that's e.g. used by [abipy](https://github.com/abinit/abipy/tree/develop/abipy/data/hgh_pseudos).


antoine-levitt:
+1 to most changes, but I don't get why the special-casing in load_psp. Wouldn't it be simpler to just have the identifier be the path to a pseudo file, and not do any fancy processing? That'd make it much easier to load a pseudo not in DFTK. The only "magic" would be that paths are relative to the data paths (but you can still do absolute paths to load a pseudopotential somewhere else) Format can be read from the extension (which is why it's good that you renamed them to hgh)

mfherbst:
if you want to load a pseudo not in psp ... just give the path to the file to `load_psp`, that has always worked.

I'm not sure what you consider fancy about this implementation. It adds the extension so you do not have to type `hgh` twice. Having `hgh` as a subdirectory helps the file organisation ... so I'd keep that, too. Apart from that ... what do you mean?

data/psp/hgh/fetch.sh
```diff
@@ -1,9 +1,21 @@
...
+    H He Li Be B C N O F Ne Na Mg Al Si P S Cl Ar K Ca Sc Ti V Cr Mn Fe Co
+    Ni Cu Zn Ga Ge As Se Br Kr Rb Sr Y Zr Nb Mo Tc Ru Rh Pd Ag Cd In Sn Sb
+    Te I Xe Cs Ba La Hf Ta W Re Os Ir Pt Au Hg Tl Pb Bi Po At Rn Fr Ra Ac
+    Rf Db Sg Bh Hs Mt Ds Rg Cn Uut Fl Uup Lv Uus Uuo Ce Pr Nd Pm Sm Eu Gd
+    Tb Dy Ho Er Tm Yb Lu Th Pa U Np Pu Am Cm Bk Cf Es Fm Md No Lr
+)
+
+URL="https://www.cp2k.org/static/potentials/cp2k/"
+
+for type in pade lda; do
```
antoine-levitt: should that be `pade pbe` ?

mfherbst: Yes ;)

antoine-levitt:
Oh sorry I missed we could load the psp directly.

I meant like adding the extension, ignoring the pade symlink etc, and generally trying to provide higher-level functionality for this. It's fine right now, but if we move to other types of psp that'll become harder to manage, for a minimal added benefit (having to type the `.hgh`) and increased complexity (I find it confusing when I see that there is no extension in the call to `load_psp` but the file has an extension). Might as well state now that the `load_psp` doesn't do anything else than load the file it's given as argument, finding a suitable parser for it.

I'd also remove the pade symlink. It's the only LDA available, so if someone wants to know what exact LDA was used they can find the source.

I just think this is not worth abstracting over, and the less code there is about this the better for both us and users. But I'm not that attached to it,

mfherbst:
Fair enough ... I removed the pade bit. I like `load_psp` and `list_psp` ... I think it makes working with DFTK easier.

antoine-levitt:
OK!

mfherbst:
Nooooooo. I'm not done with this.

mfherbst:
Ah I just open another PR ;)

antoine-levitt:
Haha sorry. How should we do these things? We mark the PR as ready for review when we're done?

mfherbst:
No worries. I think actually github's draft PR thing is quite good for this, but I've not been extremely good at using it. Otherwise a "[WIP]" in the title is fairly standard and sounds reasonable.

PR #82, opened 2019-12-07
Direct minimization

antoine-levitt:
Well I thought this would take way more effort than this :-) still missing preconditioning and some interface stuff, but it works


antoine-levitt:
Preconditioning is OK, but kpoints are buggy. Array format conversions, always a pleasure.

I get convergence in 42 iterations for silicon, which should be comparable with SCF.

antoine-levitt:
Done (for now). It doesn't do temperature, but doing that properly is tricky work. It should be extremely robust, and converge to a local minimum (failure to do that is essentially a bug in optim). It should be useful for eg 2D GPE (where SCF is just not applicable).

src/scf/direct_minimization.jl
```diff
@@ -0,0 +1,132 @@
...
+end
+
+
+"""
+Computes the ground state by direct minimization. `kwargs...` are passed to `Optim.Options()`.
+"""
+direct_minimization(basis::PlaneWaveBasis; kwargs...) = direct_minimization(basis, nothing; kwargs...)
+function direct_minimization(basis::PlaneWaveBasis{T}, Psi0;
+                             prec_type=PreconditionerTPA,
+                             optim_solver=Optim.LBFGS, kwargs...) where T
```
mfherbst: General question: Does direct minimization not work with smearing, or would that just complicate things?

src/scf/direct_minimization.jl
```diff
@@ -0,0 +1,132 @@
...
+    function fg!(E, G, Psi)
+        Psi = devec(Psi)
+        ρ = compute_density(basis, Psi, occupation)
+        ham = update_hamiltonian(ham, ρ)
+        if E != nothing
+            energies = update_energies(ham, Psi, occupation, ρ)
+            E = sum(values(energies))
+        end
+
+        # TODO check for possible 1/2 factors
```
mfherbst: Check in what sense? Check the derivation / literature?

src/scf/direct_minimization.jl
```diff
@@ -0,0 +1,132 @@
...
+    for ik = 1:Nk-1
+        starts[ik+1] = starts[ik] + lengths[ik]
+    end
+    vec(Psi) = vcat(Base.vec.(Psi)...) # TODO as an optimization, do that lazily? See LazyArrays
+    devec(Psi) = [@views reshape(Psi[starts[ik]:starts[ik]+lengths[ik]-1], size(Psi0[ik])) for ik in 1:Nk]
+
+    # this will get updated
+    energies = nothing
+
+    # computes energies and gradients
```
mfherbst: Could be worth mentioning that `G` is to be updated in-place, but `E` not.

src/scf/direct_minimization.jl
```diff
@@ -0,0 +1,132 @@
...
+
+    Pks = [prec_type(ham, kpt) for kpt in basis.kpoints]
+    P = DMPreconditioner(Nk, Pks, devec)
+
+    res = Optim.optimize(Optim.only_fg!(fg!), vec(Psi0), optim_solver(P=P, manifold=manif), Optim.Options(; allow_f_increases=true, show_trace=true, kwargs...))
+    Psi = devec(res.minimizer)
+    ρ = compute_density(basis, Psi, occupation)
+    ham = update_hamiltonian(ham, ρ)
+    # These concepts do not make sense in direct minimization,
+    # although we could maybe do a final Rayleigh-Ritz
```
mfherbst: Not necessarily needed. I mean you have the `Psi`, so just forming Rayleigh quotients is enough ... and that boils down to a few matrix-vector products. I think we should do that by default, because orben is pretty important to get a quick idea of the bands and some intuition as a chemist ;). We can of course add a flag to optionally disable that.

src/scf/direct_minimization.jl
```diff
@@ -0,0 +1,132 @@
...
+    @assert model.assume_band_gap
+    filled_occ = filled_occupation(model)
+    n_bands = div(model.n_electrons, filled_occ)
+    ortho(Psik) = Matrix{Complex{T}}(Matrix(qr(Psik).Q))
+    Nk = length(basis.kpoints)
+
+    if Psi0 === nothing
+        Psi0 = [ortho(randn(Complex{T}, length(k.basis), n_bands)) for k in basis.kpoints]
+    end
+    occupation = [filled_occ*ones(T, n_bands) for ik = 1:Nk]
```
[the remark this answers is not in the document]
mfherbst: Also, why not use `find_occupation_gap_zero_temperature` here? Then you also get a (dummy) `εF` you can return at the end of the routine?

src/scf/direct_minimization.jl
```diff
@@ -0,0 +1,132 @@
...
+    P = DMPreconditioner(Nk, Pks, devec)
+
+    res = Optim.optimize(Optim.only_fg!(fg!), vec(Psi0), optim_solver(P=P, manifold=manif), Optim.Options(; allow_f_increases=true, show_trace=true, kwargs...))
+    Psi = devec(res.minimizer)
+    ρ = compute_density(basis, Psi, occupation)
+    ham = update_hamiltonian(ham, ρ)
+    # These concepts do not make sense in direct minimization,
+    # although we could maybe do a final Rayleigh-Ritz
+    orben = nothing
+    εF = nothing
```
mfherbst: See above, could just use `find_occupation_gap_zero_temperature`.

src/scf/direct_minimization.jl
```diff
@@ -0,0 +1,132 @@
+# Direct minimization of the energy
+
+using Optim
+
+# This is all a bit annoying because our Psi is represented as Psi[k][G,n], and Optim accepts only dense arrays
+# We do a bit of back and forth using custom `vec` (ours -> optim's) and `devec` (optim's -> ours) functions
```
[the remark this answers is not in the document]
antoine-levitt: Good idea, I'll use pack

src/scf/direct_minimization.jl
antoine-levitt: It complicates things quite a lot, because then you need to optimize the free energy wrt the occupations as well as the orbitals.

src/scf/direct_minimization.jl
antoine-levitt: The derivative of the energy wrt the orbitals is 2 H psi_i. And then there are the occupations to take into account. I was just too lazy to do it, I'll take a look. Related to https://github.com/JuliaMolSim/DFTK.jl/issues/92

src/scf/direct_minimization.jl
antoine-levitt: When you minimize E(X) under X'X = I, you don't necessarily get eigenvectors, but just orthogonal combinations of them. So Rayleigh quotients don't tell you what you want, and you have to do a Rayleigh-Ritz. I'll just do that.

src/scf/direct_minimization.jl
```diff
@@ -0,0 +1,132 @@
...
+direct_minimization(basis::PlaneWaveBasis; kwargs...) = direct_minimization(basis, nothing; kwargs...)
+function direct_minimization(basis::PlaneWaveBasis{T}, Psi0;
+                             prec_type=PreconditionerTPA,
+                             optim_solver=Optim.LBFGS, kwargs...) where T
+    model = basis.model
+    @assert model.spin_polarisation in (:none, :spinless)
+    @assert model.assume_band_gap
+    filled_occ = filled_occupation(model)
+    n_bands = div(model.n_electrons, filled_occ)
+    ortho(Psik) = Matrix{Complex{T}}(Matrix(qr(Psik).Q))
```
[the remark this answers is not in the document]
antoine-levitt: Cool! Can we remove it at the other place too?

src/scf/direct_minimization.jl
antoine-levitt: There's a few papers on this, but nothing really fully satisfactory - that's one of the points that would be great to investigate in the future

src/scf/direct_minimization.jl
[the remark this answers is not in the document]
mfherbst: I'd say so. That's here: https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/eigen/diag.jl#L50.

src/scf/direct_minimization.jl
mfherbst: Ok. Was just wondering why there was no comment or mention about this.

src/scf/direct_minimization.jl
mfherbst: Ah, good point. There is no constraint about the minimum being eigenvectors ... In that case it's probably even more important to do an RR, because I would intuitively expect the Psi's to be related to bands. In the end the guess is pretty good perfect, so it should just take one or two steps to converge, I guess.

antoine-levitt: Actually in some models (eg gross-pitaevskii) the solution is not aufbau (ie it's an eigenvector, but it's not associated to the lowest eigenvalue), so doing an RR would not make sense in this case. So I think let's just leave it like that, and we can RR manually if need be (maybe with `diagonalise_all_kblocks` with maxiter 1 or something). DM is not supposed to be that user-friendly yet anyway.

src/scf/direct_minimization.jl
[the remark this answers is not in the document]
antoine-levitt: Do you mean just `Matrix(qr(Psik).Q)` or do we need the explicit type?

src/scf/direct_minimization.jl
antoine-levitt: That's standard optim interface

src/scf/direct_minimization.jl
antoine-levitt: Not really for the same reason as rayleigh-ritz. Better to have explicit handling there.

src/scf/direct_minimization.jl
antoine-levitt: Hey, turns out optimization is faster when you input the correct gradients - who knew ;-)

src/scf/direct_minimization.jl
```diff
@@ -0,0 +1,132 @@
...
+    p
+end
+function LinearAlgebra.dot(x, P::DMPreconditioner, y)
+    x_devec = P.devec_fun(x)
+    y_devec = P.devec_fun(y)
+    sum(dot(x_devec[ik], P.Pks[ik], y_devec[ik])
+        for ik = 1:P.Nk)
+end
+function Optim.precondprep!(P::DMPreconditioner, x)
+    x_devec = P.devec
```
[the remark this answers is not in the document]
antoine-levitt: Good catch - the function was just not called at all.

src/scf/direct_minimization.jl
mfherbst: Surprise ;)

src/scf/direct_minimization.jl
mfherbst: Agreed, but then you should make a comment about that ("experimental code") in the code.

src/scf/direct_minimization.jl
mfherbst: Ok.

src/scf/direct_minimization.jl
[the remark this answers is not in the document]
mfherbst: I'm not fully sure about that, but it could be that the code breaks in case your Psi0 happen to have a real instead of a complex datatype. If you're willing to risk that, then a single `Matrix(` should be good. Otherwise to get type stability `Matrix{Complex{T}}(` is what you need.

src/scf/direct_minimization.jl
```diff
@@ -66,20 +67,20 @@ function direct_minimization(basis::PlaneWaveBasis{T}, Psi0;
                              optim_solver=Optim.LBFGS, kwargs...) where T
     model = basis.model
     @assert model.spin_polarisation in (:none, :spinless)
-    @assert model.assume_band_gap
+    @assert model.assume_band_gap # temperature is not yet supported
```
mfherbst: Add `@assert model.smearing === nothing` ?

src/scf/direct_minimization.jl
mfherbst: Agreed. If there is no RR, then there is no εF.

src/scf/direct_minimization.jl
antoine-levitt: But it's not really - it works fine for what it does. I just added a comment that it is not to be expected that the psis are eigenvectors

src/scf/direct_minimization.jl
antoine-levitt: I'm not too clear on how that works. Why do we have a `assume_band_gap` as kwarg to Model? Shouldn't `assume_band_gap` just be `temperature==0`?

mfherbst: The idea here is that one could have a system with a band gap and still employ temperature (!= smearing, but "real" temperature). Right now our setup breaks with this (see https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/occupation.jl#L104), but this catches that.

antoine-levitt: We should do it at the level of the model. What's a band gap with temperature? We should either do zero temperature or finite temperature, not sure what else there is?

mfherbst: Physical finite temperature versus ficticious finite temperature to improve convergence in metals?

antoine-levitt: Computationally there's no difference so why make one?

mfherbst: Because it's different physics? I don't know, maybe we shouldn't make any?

antoine-levitt: So what's going on now is a bit weird. The band gap thing switches between the two functions we have in occupation.jl, but the first one is sort of redundant : it's a particular case of the second one when there's a gap. The rational thing to do here is remove any mention of bandgap, and use the second function all the time, even at 0K. However I kinda like the first function because it's simpler and does not need a bisection. Maybe we disable the codepaths that lead to and mark it visually as inactive but leave it in the source and still test it?

antoine-levitt: Anyway, that'll go in a separate PR. I'll open an issue

mfherbst: You say weird, made perfect sense to me, when I coded it ;). I think the current structure was made, because you mentioned bisection would converge slowly if a system is gapped. Why would that be different now?

Assuming that `find_occupation_fermi_metal` is ok for gapped systems, the other functions are really redundant, especially branch `https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/occupation.jl#L105`. I still do not see why we should not have an `assume_band_gap` flag in the model, which would allow to play some tricks e.g. in the occupation for the 0K case. Probably the logic for setting it should be simplified, that I agree (like `assume_band_gap = false` by default and `true` needs to be set explicitly).

antoine-levitt: AFAICT the current structure is because we did bandgapped systems before fractional occupations, but it's kinda obsolete now. 

This is just an optimization, so it should be done at the innermost level. Bisection is pretty fast, and I don't really see a case where it should be the bottleneck (right now it should be pretty unoptimized because of type stability issues, but it never showed up on my profiling) If it turns out to be (in very specific cases, like ridiculously low Ecut with a lot of kpoints maybe? Even then I'm skeptical) we can optimize it later (it's easy enough to do, just compute the Fermi level as if it was bandgapped, and use that as starting point for bisection). I'll make a PR with that sometime after the incoming hyperglycemic coma. Merry Christmas!

mfherbst: Hohoho. Same to you ;)

PR #84, opened 2019-12-07
Simple parser for the SCF-xn tests

mfherbst:
Parser for the castep files in the scf-xn test suite. To use:

```julia
folder = ScfXnFolder("path_to_folder")
basis = load_basis(folder)
ham = Hamiltonian(basis, guess_density(basis, composition...))
scfres = self_consistent_field(ham, params.n_bands, tol=tol)
```


antoine-levitt:
Great! Does the parser expose the spin information? I guess we can just `@warn` if there's spin and then ignore it...

mfherbst:
If there is a `spin=` in the file the parser fails. Otherwise it's up to DFTK to figure that out (simply because of the uneven electron count per unit cell). That would not be too hard to `warn`, but I just haven't done it yet.

Also ... it's not yet functional as in it yields probably wrong structures, but I could not be bothered to debug further and I'll not have a lot of time in the next few days.

mfherbst:
Closed for now since `parse-castep` is not yet stable enough for our purposes. Maybe we pick this up again later.

PR #91, opened 2019-12-15
fix gross_pitaevskii, occupation, and add r_to_G for kpoints

antoine-levitt:
We should do the `r_to_G` even if it's slow: this is for convenience, not speed


antoine-levitt:
@mfherbst if there's no objection I'll merge this, since otherwise gross_pitaevskii is broken on master

mfherbst:
Sorry my internet today was, well ... sporadic

antoine-levitt:
No hurry, I just checked out master on another computer and was surprised at the bug

Issue #85, opened 2019-12-07
Magnetic fields


antoine-levitt:
For pretty picture purposes. We need to be a bit careful about how to impose the gauge condition at the discrete level (to get a hermitian hamiltonian): if N is the discrete nabla operator and B the magnetic field, then N and B do not necessarily commute at the discrete level even if divB = 0 at the continuous level, and so BN is not necessarily hermitian. Possibly a good way to do it is to allow arbitrary fields, and impose the divB=0 condition at the discrete level by adding an extra term to the local potential, ie BN-NB should be a local potential and we just calculate that? Or maybe we just do x dy - y dx for now, since the commutation is easier to see there.

Issue #86, opened 2019-12-08
Terms interface


antoine-levitt:
The current design has served us well but:
- it's specialized to "basic" DFT, doesn't really support eg Fock terms (where the hamiltonian depends on the psi and the occupations, not just the density) and magnetic fields (where the hamiltonian application needs gradients of psi)
- it hardcodes a list of predefined terms (kinetic, external, hartree, xc, nonlocal). It'd be nice to have an interface that lets the terms pick and choose what information they need
- energies sort of come as an afterthought. Eg the psp and ewald energies should be terms like the others (they just happen not to depend on the psi)
- After thinking about it a lot, I think this interface where the terms are called separately for the energy and potential, and have to ask for if things are nothing or not and fill them is annoying. A term, when called, should output its energy, and a lazy object that is able to compute gradients if they are needed. That frees the caller from having to figure out exactly what temporaries the term needs.

I think we should come up with a design for the most general situation possible, where we want to do direct minimization wrt the psi_nk and their occupation for eg Hartree-Fock. Then we try to join with the current design.

Mathematically, a term is a function E(D) of the density matrix (in our case represented as Psi and their occupations for all kpoints). From that we want to compute E(D) and H = nabla E(D). H is block diagonal in the sense that it maps Bloch waves to Bloch waves, which leads to Hk. Hk is something that takes as input a psi (represented by a list of coordinates in the G basis), and outputs Hk psi.

Performance-wise, I think we can do pretty much what we want (closures left and right, allocations of whatever), provided we don't screw up on the application of Hk to vectors.

So, just to get the ball rolling, how about Model contains a list (ie tuple) of terms. Each term is something that can be called like `term(basis, Psi, occupation)` and returns an energy, as well as an HamiltonianTerm object, that can store whatever it wants inside, and can be called like `ham_term(kpoint, Psi)`. `HamiltonianBlock` would then contain a list of these things and wrap `mul!` around that.

[...]

PR #93, opened 2019-12-15
Magnetic fields

antoine-levitt:
(this PR is on top of misc and direct_minimization).

Fix https://github.com/JuliaMolSim/DFTK.jl/issues/85

The magnetic fields work, but do too many FFTs. This will wait https://github.com/JuliaMolSim/DFTK.jl/issues/86 to be more efficient.

I also added a proof of concept for 2D gross-pitaevskii. The normalizations are probably wrong so the results are not to be trusted, but running the script gets me vortices so the basic things work.


antoine-levitt:
[pretty_picture.pdf](https://github.com/JuliaMolSim/DFTK.jl/files/3964971/pretty_picture.pdf)

(OK it's probably wrong but who cares)

antoine-levitt:
Behold! Gross-Pitaevskii fermions ! (yes, it doesn't make sense, but it works)
pretty_picture2

This 2D GPE (and its 3D counterpart) with 1 particle is a good benchmark if we want to cut down on our overhead (which doesn't matter in DFT, but is significant here).

antoine-levitt:
Closing in favor of https://github.com/JuliaMolSim/DFTK.jl/pull/130

Issue #94, opened 2019-12-24
Simplify occupation


antoine-levitt:
So what's going on now is a bit weird. The `assume_band_gap` thing switches between the two functions we have in occupation.jl, but the first one is sort of redundant : it's a particular case of the second one when there's a gap. The rational thing to do here is remove any mention of bandgap, and use the second function all the time, even at 0K. However I kinda like the first function because it's simpler and does not need a bisection. Maybe we disable the codepaths that lead to and mark it visually as inactive but leave it in the source and still test it?

PR #95, opened 2019-12-25
Rework occupations

antoine-levitt:
Fix https://github.com/JuliaMolSim/DFTK.jl/issues/94

There are a couple workarounds we could do that probably would work for doing fractional occupations at 0K (eg if there's only one energy level at the fermi level, fill it partially to fit the correct number of electrons) but that is very tricky, and figuring out the proper fractional occupations at 0K is a much more involved job than just looking at the band energies. So I prefer keeping it explicit that it doesn't work and erroring.

### DIFF
--- a/src/occupation.jl
+++ b/src/occupation.jl
@@ -6,18 +6,17 @@ import Roots
 Find the Fermi level, given a `basis`, SCF band `energies` and corresponding Bloch
 one-particle wave function `Psi`.
 """
-function find_fermi_level(basis, energies, Psi)
-    εF, _ = find_occupation(basis, energies, Psi)
+function find_fermi_level(basis, energies)
+    εF, _ = find_occupation(basis, energies)
     εF
 end
 
 """
 Find the Fermi level and occupation for the given parameters.
 """
-function find_occupation(basis, energies, Psi)
+function find_occupation(basis::PlaneWaveBasis{T}, energies) where {T}
     @assert basis.model.spin_polarisation in (:none, :spinless)
     model = basis.model
-    n_bands = size(Psi[1], 2)
     n_electrons = model.n_electrons
     temperature = model.temperature
     smearing = model.smearing
@@ -28,7 +27,15 @@ function find_occupation(basis, energies, Psi)
 
     filled_occ = filled_occupation(model)
 
-    @assert filled_occ*n_bands ≥ n_electrons
+    # assume that we can get the required number of electrons by filling every state
+    @assert filled_occ*sum(basis.kweights .* length.(energies)) ≥ n_electrons - sqrt(eps(T))
+
+    # early exit when all bands have to be filled; avoids numerical problems
+    if temperature == 0 && filled_occ*sum(basis.kweights .* length.(energies)) ≈ n_electrons
+        occ = [fill(filled_occ, length(e)) for e in energies]
+        εF = nextfloat(maximum(maximum.(energies)))
+        return εF, occ
+    end
 
     # Find εF so that
     # n_i = filled_occ * f((εi-εF)/T)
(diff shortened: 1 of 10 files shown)


src/occupation.jl
```diff
@@ -3,67 +3,39 @@
...
-              HOMO, LUMO, maxlog=5)
-    end
-
-    εF, occupation
+function find_fermi_level(basis, energies)
+    εF, _ = find_occupation(basis, energies)
+    εF
 end
 
 """
```
mfherbst: My original idea here was to have `find_occupation`, which only returns the occupation, `find_fermi_level`, which only returns the Fermi level and `find_occupation_fermi`, which does both. You might disagree, but I find it unintuitive, why `find_occupation` also returns the Fermi level and it does so as first argument.

src/occupation.jl
```diff
@@ -3,67 +3,39 @@
...
 
     filled_occ = filled_occupation(model)
 
-    @assert filled_occ*n_bands ≥ n_electrons
+    # assume that we can get the required number of electrons by filling every state
+    @assert filled_occ*sum(basis.kweights .* length.(energies)) ≥ n_electrons - sqrt(eps(T))
 
-    # Avoid numerical issues by imposing a step function for zero temp
-    temperature == 0 && (smearing(x) = x ≤ 0 ? 1 : 0)
+    # early exit when all bands have to be filled; avoids numerical problems
```
mfherbst: Just out of interest: Do you not expect problems if the lowermost levels need to be fully filled (e.g. the insulator case)?

test/silicon_runners.jl
```diff
@@ -3,6 +3,8 @@ using DFTK
 
 include("testcases.jl")
 
+# Silicon RHF is a metal, so we add a bit of temperature to it
```
mfherbst: Please do not abbreviate RHF = reduced HF, because for me RHF is restricted HF ... this messes with me each time.

src/occupation.jl
antoine-levitt: Ah, I see. The reason I removed it is because I thought the `_fermi` had something to do with Fermi broadening, rather than the Fermi level. `find_occupation_fermi_level` is a bit long... The function is literally called once in the code, so I'm fine with whatever name; in general I think it's OK to return more than asked for; we could just switch the return values: return occupation, eps_F?

src/occupation.jl
antoine-levitt: Not really, why? This is the standard way to do things for insulators.

test/silicon_runners.jl
antoine-levitt: Ah, yes. This is very very annoying. I'd rename it to Hartree, but for some people Hartree means with self-interaction removed (the Hartree potential for orbital i uses the total density minus the density of orbital i). I'll make it explicit

src/occupation.jl
mfherbst: Or use a named tuple ... there the "additional" returned object is kind of transparent. You can either use it or not.

test/silicon_runners.jl
mfherbst: Names ... always difficult :smile:

src/occupation.jl
mfherbst: What I meant is if you have an insulator and `n_bands > n_electrons // filled_occ`

src/occupation.jl
antoine-levitt: OK, let's do that. With tuple destructuring it's exactly the same for the caller.

src/occupation.jl
antoine-levitt: I simplified and commented that branch. The only trouble is when there is no wiggle room, so that bracketing doesn't work

mfherbst: Perfect. That was the part I missed so far.

antoine-levitt: Yeah, these rounds of review are pretty effective at forcing me to be less lazy :-)

Issue #36, opened 2019-10-28
Use real FFTs


antoine-levitt:
Add an option to perform real-to-complex and complex-to-real FFTs with a kwarg, and use that to do the potentials as real arrays. Could also do dispatch on the eltype of f_real, but it's not compatible with G_to_r(f_fourier).

mfherbst:
About the dispatch: I think it's fair to only do this if the in-place version of the FFTs is used or if an appropriate kwarg is used.

mfherbst:
Related to #9

antoine-levitt:
So I've been thinking about this, and the main difficulty is avoiding allocating temporary buffers.

I've been doing some more serious benchmarking on this: https://gist.github.com/antoine-levitt/3495f47c99676f26824df1fbe9462040. So if I'm interpreting this right, a reasonable mental model is that an alloc, a fill and a copy take about the same time. That unit of time is about 1/8th of an FFT. (some of that might be due to julia issue 130. So allocations do matter, unfortunately (especially when going to eg threading).

The main problem is that we use in-place ffts, which is a lot more tricky with real FFTs, essentially because the input and output don't have the same size (they also don't have the same type, but we can work around that with `reinterpret`). The real array should be of size 2*(n/2+1) (see eg http://www.fftw.org/fftw3_doc/One_002dDimensional-DFTs-of-Real-Data.html and http://www.fftw.org/fftw3_doc/Multi_002dDimensional-DFTs-of-Real-Data.html#Multi_002dDimensional-DFTs-of-Real-Data). There's a julia interface with an open PR to FFTW but really it's just as simple to do it ourselves). So it means that we can't easily use that and keep the interface we have right now for our functions and in the rest of the code.

So the simplest option is to just drop the in-place FFTs entirely. Does that hurt us very much? In my tests I see a performance difference of about 10% in favour of in-place. So acceptable for the sake of a much cleaner interface.

So, roadmap:

1 loop and thread at the level of Hamiltonian application (rather than FFTs), so FFT functions only ever have to deal with one wavefunction at a time. Ie do `for iband=1:nband` in `mul!(Y,Ham,X)`. There is no performance penalty in doing so, it simplifies the FFT code quite a bit, improves memory locality and helps us manage memory footprint.

2 do a barebones implementation with out-of-place FFTs keeping the data format for the Psi the same

[...]

PR #96, opened 2019-12-29
FFTs only work one band at a time

antoine-levitt:
First part of https://github.com/JuliaMolSim/DFTK.jl/issues/36#issuecomment-569521732

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -17,6 +17,7 @@ LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Markdown = "d6f4376e-aef5-505a-96c1-9c027394607a"
 Memoization = "6fafb56a-5788-4b4e-91ca-c0cea6611c73"
 NCDatasets = "85f8d34a-cbdd-5861-8df4-14fed0d494ab"
+NLSolversBase = "d41bc354-129a-5804-8e4c-c37616107c6c"
 NLsolve = "2774e3e8-f4cf-5e23-947b-6d7e65073b56"
 Optim = "429524aa-4258-5aef-a3af-852621145aeb"
 Primes = "27ebfcd6-29c5-5fa9-bf4b-fb8fc14df3ae"
@@ -27,6 +28,7 @@ SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 
 [compat]
+NLSolversBase = "= 7.5.0"
 julia = "1.2"
 
 [extras]
--- a/examples/graphene.jl
+++ b/examples/graphene.jl
@@ -17,6 +17,7 @@ kcoords, ksymops = bzmesh_ir_wedge(kgrid, lattice, composition...)
 basis = PlaneWaveBasis(model, Ecut, kcoords, ksymops)
 
 # Run SCF
+n_bands = 6
 ham = Hamiltonian(basis, guess_density(basis, composition...))
 scfres = self_consistent_field(ham, n_bands)
 ham = scfres.ham
--- a/examples/graphene.jl
+++ b/examples/graphene.jl
@@ -1,0 +1,32 @@
+using DFTK
+using Printf
+
+kgrid = [12, 12, 4]
+Tsmear = 0.0009500431544769484
+Ecut = 35
+
+lattice = [4.659533614391621 -2.3297668071958104 0.0;
+           0.0 4.035274479829987 0.0;
+           0.0 0.0 15.117809010356462]
+C = Species(6, load_psp("hgh/pbe/c-q4"))
+composition = [C => [[0.0, 0.0, 0.0], [0.33333333333, 0.66666666667, 0.0]]]
+
+model = model_dft(lattice, [:gga_x_pbe, :gga_c_pbe], composition...;
+                  temperature=Tsmear, smearing=DFTK.Smearing.Gaussian())
+kcoords, ksymops = bzmesh_ir_wedge(kgrid, lattice, composition...)
+basis = PlaneWaveBasis(model, Ecut, kcoords, ksymops)
+
+# Run SCF
+ham = Hamiltonian(basis, guess_density(basis, composition...))
+scfres = self_consistent_field(ham, n_bands)
+ham = scfres.ham
+
+# Print obtained energies
+energies = scfres.energies
+energies[:Ewald] = energy_nuclear_ewald(model.lattice, composition...)
+energies[:PspCorrection] = energy_nuclear_psp_correction(model.lattice, composition...)
+println("\nEnergy breakdown:")
+for key in sort([keys(energies)...]; by=S -> string(S))
+    @printf "    %-20s%-10.7f\n" string(key) energies[key]
+end
+@printf "\n    %-20s%-15.12f\n\n" "total" sum(values(energies))
--- a/src/HamiltonianBlock.jl
+++ b/src/HamiltonianBlock.jl
@@ -47,12 +47,16 @@ function LinearAlgebra.mul!(Y, block::HamiltonianBlock, X)
     if Vloc === nothing
         Y .= kin.diag .* X
     else
-        Xreal = G_to_r(block.basis, block.kpt, X)
-        Xreal .*= Vloc
-        r_to_G!(Y, block.basis, block.kpt, Xreal)
-        Y .+= kin.diag .* X
+        @views Threads.@threads for n = 1:size(X, 2)
+            Xreal = G_to_r(block.basis, block.kpt, X[:, n])
+            Xreal .*= Vloc
+            r_to_G!(Y[:, n], block.basis, block.kpt, Xreal)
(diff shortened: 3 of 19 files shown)


mfherbst:
I'd avoid using Ψ and ψ for different things, because the symbols look far too similar in most fonts. I now use Ψik instead of ψk to make the distinction clearer.

mfherbst:
Also in general I see why you did this: It simplifies code. On the other hand we now need to deal with threading at a few more places (density computation, ldos computation ...) otherwise it would only be one centralised spot. I think that's ok for now, but if we look into MPI / hybrid GPU/CPU parallelisation at some point we need to reconsider ... but probably we then have a bunch of other problems, too.

I'll add threading for LDOS and density (actually slows down the calculations I do noticably to have this single-threaded!).

antoine-levitt:
> I'll add threading for LDOS and density (actually slows down the calculations I do noticably to have this single-threaded!).

Really? Might be worth trying to optimize a bit then (eg dot fusion, making sure there's no `@code_warntype` funny business, etc. It should definitely be lower-order for large systems, but for smallish ones there might be an impact

antoine-levitt:
> Also in general I see why you did this: It simplifies code. On the other hand we now need to deal with threading at a few more places (density computation, ldos computation ...) otherwise it would only be one centralised spot. I think that's ok for now, but if we look into MPI / hybrid GPU/CPU parallelisation at some point we need to reconsider ... but probably we then have a bunch of other problems, too.

Hm, that I don't know. For optimal performance we also want to minimize the number of temp buffers of size Psi we allocate. All these will have to change things anyway, so let's see when we get to it I think.

> I'd avoid using Ψ and ψ for different things, because the symbols look far too similar in most fonts. I now use Ψik instead of ψk to make the distinction clearer.

We have way too many notations for these already. We sometimes use Psi, sometimes \Psi, sometimes \psi, and sometimes X. To complicate matters, they are really the u (periodic part of bloch waves), not the psi (bloch waves). Similar for the orben. Let's brainstorm all the notations sometime and update them all consistently?

mfherbst:
I'll profile the threading aspect on the calculations when this is in master and see what can be done about this.

> We have way too many notations for these already. We sometimes use Psi, sometimes \Psi, sometimes \psi, and sometimes X. To complicate matters, they are really the u (periodic part of bloch waves), not the psi (bloch waves). Similar for the orben. Let's brainstorm all the notations sometime and update them all consistently?

I agree. I've noticed that, too, when going through the code wrt this PR.

antoine-levitt:
> I'll profile the threading aspect on the calculations when this is in master and see what can be done about this. For sure threading should not hurt, however.

I think just adding dots and views to LDOS will already improve things massively

mfherbst:
Ok ... I got round to do some smallish testing with the density computation function and I agree, that extra threading (on top of what is done by FFTW itself) really makes little sense here and makes the code only more complicated (you need to pre-allocate thread-local buffers for the accumulation and need to take to not get into false sharing ... it's a bit messy). So I agree we should skip that.

But there clearly are some issues with threading in general. How many cores are used on the cluster seems to be a bit random. For example. I ran the graphene calculation (committed) on a cluster node with 32 virtual, 16 physical cores by just including the script. The load stayed at 100% for a pretty long time (10 minutes) with only a short spike 3200% for a couple of seconds in between. So I interrupt the execution and re-include the script in the same julia shell. Now the load is mostly 1600% with short periods (ca 5 seconds) down to 100%. The calculation terminates within about 4 minutes (but the maximal load was around 1600%). Then I re-include the script once again and get the same behaviour (1600% most of the time, short drops to 100%) ... I'm not sure but it sounds to be there is some interference between the FFTW-Threading and the Julia `@threads` macro. I'll do some more testing ...

mfherbst:
Here is a log of the processor usage from the run. One line is one second.
[dftk_FFTs_clustern10.log](https://github.com/JuliaMolSim/DFTK.jl/files/4103354/dftk_FFTs_clustern10.log)

antoine-levitt:
What did you use for `JULIA_NUM_THREADS`? I'm not sure how many threads FFTW uses by default, possibly we should do a `fftw_set_num_threads(Threads.nthreads())` by default? Can you also do a profile? It might be that a lot of time is spent on non-FFTs (eg I did a recent profile of supercell silicon, and a lot of time was being spent on the computation of the external potential (that we should cache; again dependent on the term refactor)

mfherbst:
`JULIA_NUM_THREADS = 32` obviously and FFTW enables threading automatically as well, see [here](https://github.com/JuliaMath/FFTW.jl/blob/master/src/FFTW.jl#L59). I can do some profiling, but I do not think that's the problem, because that's just one particularly odd run. Sometimes things look better (e.g. closer to 3200% usage). Also on my laptop I have not encountered such behaviour that much. I'll try to produce another case demonstrating that.

antoine-levitt:
The first run looks to be a bug, so it would be good to find where that comes from, and if that is reproducible. I have also seen weirdness in the thread usage, so probably there's something fishy going on there. One thing to try is whether doing a dummy threaded loop (`@threads for i=1:10 println(i) end` or something before including the script changes anything. Possibly FFTW thread initialization is buggy when threads are not used before or something?

mfherbst:
Now rerunning I encountered a GC bug ... fun. I should to say I've been using v1.2 on the cluster ... I'm now switching to 1.3.1 to make sure that's not the problem.

antoine-levitt:
Huh, cool! We don't do anything particularly fishy with memory, so we might be triggering a bug here... Switching to 1.3 is a good idea.

mfherbst:
Ok ... with Julia 1.3.1 things are a lot more reliable, fortunately. Still we get a drop in performance in between (from some baseline 2700% to some 400%), but things are obviously much better already. I'll do some profiling wrt. the remaining perfomance drop.

mfherbst:
Ok, I found the bottle neck: It is the single-threaded element-wise processing of the density when dealing with the symmetry operations.

I'll add a heuristic to capture the case where the symmetry operation is just the identity (the most common case, especially for supercell approaches) and add some threading to the loop to parallelise the whole processing (i.e. both computing partial densities per k-Points and the elementwise processing due to symmetry).

That should be ok for now, but perhaps we should investigate whether there is a more clever way to do symmetry than fiddeling with the Fourier coefficients elementwise ...

antoine-levitt:
It looks like `G_vectors` is pretty inefficient btw. Did you try with a `collect` in there to materialize the generator?

mfherbst:
Nope, but for this case I agree it's a good idea to collect it once.

antoine-levitt:
Also, we can 1) precompute S^-1 2) add an `@inbounds` annotation 3) see if `index_G_vectors` is the bottleneck

antoine-levitt:
this shouldn't really be a rate-limiting operation so I'd micro-optimize a bit before threading

mfherbst:
Indeed about half of the time of the compute_density function is spent in the line with the inv and the index_G_vectors at the moment.

mfherbst:
Ok so, I pushed the thread stuff, because I now did it anyway. We can see what we can find in `index_G_vectors`.

antoine-levitt:
Oof. That's some heavy thread-specific logic that seems a bit convoluted, and is pretty sure to increase the probability of hitting the infamous closure performance bug (`@threads` create a closure). Also code that is unoptimized and allocates a lot (like this one) typically does not thread well. Let's optimize the single-threaded version first and do some benchmarking. A simpler possibility that does not have load balancing issues is to thread on the G points only.

mfherbst:
I'm still working on the single-threaded version anyway ... just wanted to have it out and get checked ... we can decide later what to use.

antoine-levitt:
With the above modifications it doesn't seem to be performance-critical for me. OTOH I got a weird "corrupted size vs. prev_size” error when running with threads, did you ever get that?

antoine-levitt:
ProfileView looks like it exports and imports JLD, so that might be good to exchange profiles

antoine-levitt:
Well that's fun:

https://gist.github.com/antoine-levitt/41cc5480235d9d4c7e853354ccdb9a45
I got that by running silicon on this branch. It's in fftw_mkplan, and it has the "in expres11): Segmentation fault" characteristic of a race. This is possibly what would happen if we planned FFTs on threads (plan_fft is not thread-safe I believe), but we don't do that, do we?

mfherbst:
Yikes. Not that I'm aware of. It is done when `PlaneWaveBasis` is constructed, so here https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/PlaneWaveBasis.jl#L101.

antoine-levitt:
Got it: `inv(plan)` is not thread-safe

antoine-levitt:
Opened an issue there: https://github.com/JuliaMath/FFTW.jl/issues/134

mfherbst:
The failure in Travis is caused by https://github.com/JuliaNLSolvers/NLSolversBase.jl/issues/117. I'll pin Optim to the last working version.

mfherbst:
Puh ... glad that one's finally in :smile:.

antoine-levitt:
Heh, yeah! Hopefully the other prs are simpler

PR #98, opened 2020-01-07
Take occupation into account for DOS

mfherbst:
Small fix for DOS computations.

### DIFF
--- a/src/postprocess/DOS.jl
+++ b/src/postprocess/DOS.jl
@@ -13,11 +13,14 @@ using ForwardDiff
 Total density of states at energy ε
 """
 function DOS(ε, basis, orben; smearing=basis.model.smearing, T=basis.model.temperature)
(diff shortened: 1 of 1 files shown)


antoine-levitt:
Oh good catch! Don't know how I missed that... I'm appallingly bad at factors of 2. So that changes the hybrid mixing, not sure if in the right direction or not. It would be interesting also to compare the mixing parameters one gets with this to the ones commonly used for kerker. I did that a while back and the results were a bit inconsistent, maybe that was why...

mfherbst:
Me, too, that's why checking and comparing helps. Hybrid mixing looks pretty good for the sodium example. What do you mean by mixing parameters? The parameters of the Kerker most closely resembling the approximate Jacobian from hybrid mixing?

antoine-levitt:
Yup : simply approximate chi0 by the multiplication with the DOS (properly normalized)

PR #99, opened 2020-01-14
Amend check_real

mfherbst:
Ignore imaginary parts which are smaller than a certain threshold.

### DIFF
--- a/src/common/check_real.jl
+++ b/src/common/check_real.jl
@@ -1,7 +1,11 @@
 check_real(A::AbstractArray) = nothing
 function check_real(A::AbstractArray{Complex{T}}) where T
-    discrepancy = norm(imag(A) ./ abs.(A))
-    if discrepancy > 1000 * eps(real(eltype(A)))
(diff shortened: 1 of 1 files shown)


antoine-levitt:
Looks good. 

Use the names `atol` and `rtol` (see eg `isapprox`) in this function (not necessarily in the interface, but in the function body). 

Simplify this to a generator? `any(abs(imag(xx)) > rtol*abs(xx) && abs(imag(xx) > atol for xx in x)` or something?

Not sure what the tolerance should be. Perhaps `sqrt(eps)`?

mfherbst:
Thanks. Generator is a good idea, but I like the idea to get at least an idea of the error from the warning output. I went for a hybrid approach to solve that.

For `rtol` I use `1000eps`, because I find `sqrt(eps)` a bit large. For `atol` I went for `100eps`. I guess one could argue for different constants, but let's see how they perform.

antoine-levitt:
LGTM!

PR #100, opened 2020-01-14
Parser for pymatgen structures into DFTK structures

mfherbst:
Parser for pymatgen `lattice` and `structure` structures as they are used by `abipy` and `pymatgen`.

### DIFF
--- a/src/bzmesh.jl
+++ b/src/bzmesh.jl
@@ -90,3 +90,17 @@ function bzmesh_ir_wedge(kgrid_size, lattice, composition...; tol_symmetry=1e-5)
 
     kcoords, ksymops
 end
+
+@doc raw"""
+Selects a kgrid_size to ensure a minimal spacing (in inverse Bohrs) between kpoints.
+Default is ``2π * 0.04 \AA^{-1}``.
+"""
+function kgrid_size_from_minimal_spacing(lattice, spacing=2π * 0.04 / units.Ǎ)
+    for d in 1:3
+        @assert norm(lattice[:, d]) != 0
+        # Otherwise the formula for the reciprocal lattice
+        # computation is not correct
+    end
+    recip_lattice = 2π * inv(lattice')
+    [ceil(Int, norm(recip_lattice[:, i]) ./ spacing) for i = 1:3]
+end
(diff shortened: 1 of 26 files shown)


mfherbst:
@antoine-levitt Are you ok if I merge that?

src/bzmesh.jl
```diff
@@ -90,3 +90,17 @@ function bzmesh_ir_wedge(kgrid_size, lattice, composition...; tol_symmetry=1e-5)
...
     kcoords, ksymops
 end
+
+@doc raw"""
+Selects a kgrid_size to ensure a minimal spacing (in inverse Bohrs) between kpoints.
+Default is ``2π * 0.04 \AA^{-1}``.
+"""
+function kgrid_size_from_minimal_spacing(lattice, spacing=2π * 0.04 / units.Ǎ)
+    for d in 1:3
+        @assert norm(lattice[:, d]) != 0
```
antoine-levitt: You don't need that, the following line will fail anyway

src/common/constants.jl
```diff
@@ -1,2 +1,8 @@
 """The default search location for Pseudopotential data files"""
 DFTK_DATADIR = joinpath(dirname(pathof(DFTK)), "..", "data")
+
+# Commonly used constants. The factors convert from the respective unit to AU
+module units
+    eV = 0.03674932248  # Hartree
+    Ǎ = 1.8897261254578281  # Bohr
+end
```
antoine-levitt: Very cool idea! Maybe add a few more? Kelvins, Rydbergs and we should be OK.

src/energy_ewald.jl
```diff
@@ -90,7 +90,7 @@ function energy_ewald(lattice, recip_lattice, charges, positions; η=nothing)
         gsh += 1
     end
     # Amend sum_recip by proper scaling factors:
-    sum_recip = sum_recip * 4T(π) / det(lattice)
+    sum_recip = sum_recip * 4T(π) / abs(det(lattice))
```
antoine-levitt: Good catch on all these

src/bzmesh.jl
mfherbst: Yeah with the very helpful `SingularException` :-P . I think getting the assertion and then finding the line with the explanatory comment is more helpful. But I can remove it if you want.

src/common/constants.jl
mfherbst: Good idea.

src/bzmesh.jl
antoine-levitt: Hehe, as usual, I tend towards less code = better. But I'm fine with either

PR #101, opened 2020-01-15
Add NOS function estimating number of Kohn-Sham states for DOS / LDOS

mfherbst:
@antoine-levitt I differ a little from the definition we discussed via mail:

Instead of being fancy and and summing `-f'((εnk - εF)/T)`, I sum `T * d/dε  f((εnk - ε)/T) |_{ε=εF}`. Only a cosmetic difference, but the `T *` makes the window of width `T` aspect more apparent in my opinion.

### DIFF
--- a/src/postprocess/DOS.jl
+++ b/src/postprocess/DOS.jl
@@ -9,10 +9,17 @@
 
 using ForwardDiff
 
-"""
-The number of Kohn-Sham states in temperature window of width T contributing
-to the DOS at temperature T. This value should be larger than one for DOS and LDOS to be
-meaningful at smearing temperature T.
+@doc raw"""
+    NOS(ε, basis, orben; smearing=basis.model.smearing, T=basis.model.temperature)
+
+The number of Kohn-Sham states in a temperature window of width T around the energy ε
+contributing to the DOS at temperature T.
+
+This quantity is not a physical quantity, much rather a dimensionaless approximate measure
+for how well properties near the Fermi surface are sampled with the passed `smearing`
+and temperature `T`. It increases with both `T` and better sampling of the BZ with
+``k``-Points. A value ``\gg 1`` indicates a good sampling of properties near the
+Fermi surface.
 """
 function NOS(ε, basis, orben; smearing=basis.model.smearing, T=basis.model.temperature)
     N = zero(ε)
--- a/src/postprocess/DOS.jl
+++ b/src/postprocess/DOS.jl
@@ -10,6 +10,32 @@
 using ForwardDiff
 
 """
+The number of Kohn-Sham states in temperature window of width T contributing
+to the DOS at temperature T. This value should be larger than one for DOS and LDOS to be
+meaningful at smearing temperature T.
+"""
+function NOS(ε, basis, orben; smearing=basis.model.smearing, T=basis.model.temperature)
+    N = zero(ε)
+    T != 0 || error("NOS only supports finite temperature")
+
+    # Note the differences to the DOS and LDOS functions: We are not counting states
+    # per BZ volume (like in DOS), but absolute number of states. Therefore n_symeqk
+    # is used instead of kweigths. The number of states is counted over a temperature
+    # window of width T, where we assume that the number of states remains constant
+    # in the interval [T - T/2, T + T/2] as ( d/dε f_T(εik - ε') )|_{ε' = ε}
+    # with f_T the smearing function. This gives the factor T in front.
+    for ik = 1:length(orben)
+        n_symeqk = length(basis.ksymops[ik])  # Number of symmetry-equivalent k-Points
+        for iband = 1:length(orben[ik])
+            N += (T * n_symeqk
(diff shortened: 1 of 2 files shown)


antoine-levitt:
Looks good! NOS is not that great a name, but it's not completely bad either.

I don't get the
> where we assume that the number of states remains constant
    # in the interval [T - T/2, T + T/2]

? A better description is to say that we count the number of states in a window of size T around eps, but use instead of a rectangular window the function -f', which acts as a smooth window. I kind of liked the explicit f'((enk-e)/T), but the new version makes clearer that the NOS should be proportional to T. Maybe mention in the doc that 1) this is not a physical value 2) it increases with both T and the BZ sampling 3) it is non-dimensional, and therefore is naturally compared to 1 (the regime NOS >> 1 guaranteeing a good sampling of the properties near the Fermi surface)

mfherbst:
I thought about `NOS_for_DOS` but that's not really better ...

> I don't get the
> ```
> where we assume that the number of states remains constant
> # in the interval [T - T/2, T + T/2]
> ```

You can see it as an integration of `f_T'((enk - e)/T)` over the interval [T - T/2, T + T/2], which you approximately evaluate using the midpoint rule yielding the integral value `T * f_T'((enk - e)/T)`.

I'll extend the doc.

src/postprocess/DOS.jl
```diff
@@ -9,6 +9,39 @@
 
 using ForwardDiff
 
+@doc raw"""
+    NOS(ε, basis, orben; smearing=basis.model.smearing, T=basis.model.temperature)
+
+The number of Kohn-Sham states in a temperature window of width T around the energy ε
+contributing to the DOS at temperature T.
+
+This quantity is not a physical quantity, much rather a dimensionaless approximate measure
```
antoine-levitt: but instead of much, and dimensionless

src/postprocess/DOS.jl
```diff
@@ -9,6 +9,39 @@
...
+function NOS(ε, basis, orben; smearing=basis.model.smearing, T=basis.model.temperature)
+    N = zero(ε)
+    T != 0 || error("NOS only supports finite temperature")
+
+    # Note the differences to the DOS and LDOS functions: We are not counting states
+    # per BZ volume (like in DOS), but absolute number of states. Therefore n_symeqk
+    # is used instead of kweigths. The number of states is counted over a temperature
+    # window of width T, where we assume that the number of states remains constant
+    # in the interval [T - T/2, T + T/2] as ( d/dε f_T(εik - ε') )|_{ε' = ε}
+    # with f_T the smearing function. This gives the factor T in front.
```
antoine-levitt: I'm still confused by this description, sorry if I'm being dense

src/postprocess/DOS.jl
mfherbst: oops.

src/postprocess/DOS.jl
mfherbst: No that's fine. Maybe it's my picture of the situation, which is wrong. I tried to detail it again.

src/postprocess/DOS.jl
```diff
@@ -28,9 +28,10 @@ function NOS(ε, basis, orben; smearing=basis.model.smearing, T=basis.model.temp
     # Note the differences to the DOS and LDOS functions: We are not counting states
     # per BZ volume (like in DOS), but absolute number of states. Therefore n_symeqk
     # is used instead of kweigths. The number of states is counted over a temperature
-    # window of width T, where we assume that the number of states remains constant
-    # in the interval [T - T/2, T + T/2] as ( d/dε f_T(εik - ε') )|_{ε' = ε}
-    # with f_T the smearing function. This gives the factor T in front.
+    # window of width T yielding in analogy to the DOS an integral
+    #     ∫_(T-T/2)^(T+T/2) ( d/dε f_τ(εik - ε') )|_{ε' = ε} dτ
```
antoine-levitt: that is eps - T/2 and eps + T/2, right?

antoine-levitt: Even then I'm not completely sure of that derivation, and esp. of the analogy with the DOS (because it's quite different, and we don't normalize wrt the kpoints for instance). The one where we say that we'd like to count the number of states contributing to the integral and just use a smooth window feels the less bad. I'll try and come up with a clearer (and more mathematically informative) derivation later.

antoine-levitt: The feeling is that we should be able to derive an error bound like `exp(-N)` for the convergence of properties, I'll try and make that more precise.

mfherbst: No I did not mean eps +/- T/2 , I meant T (as in the T-dependence of the smearing function), but I'm not sure about this either. It's just what I came up with to show the analogy to the DOS in the implementation. Ok ... let's go with the window argument. In that case `-f'((εnk - ε)/T)` would just be a weight, chosen to be large and positive if a lot happens (large occupation fluctuations) and small otherwise.

Issue #57, opened 2019-11-13
Struct for `composition`?


antoine-levitt:
We pass around `composition` a lot and end up documenting it in several places. Should we just give it a struct?

mfherbst:
That's not a bad idea. I thought about it before and wanted to leave it until we see more clearly how we need it.

One advantage would be that it would be easier to hide some `pymatgen` details for things such as doubling the unit cell for supercell sampling, which is definitely something we'll be using a lot more in the future.

antoine-levitt:
- generalize Species to AbstractSpecies, with subtypes (coulomb, psp, ...)
- have a species (Vector of pairs (spec, pos)) field in Model
- make the functions that currently take `composition...` read from that field
- move the computation of number of electrons to Model (overwritable with a kwarg)

antoine-levitt:
Why `Species` and not `Atom`, btw?

antoine-levitt:
The reason I'm asking is because with `Atom` I can figure out what its plural is :D

mfherbst:
Because in coarse-graining methods the entities of the unit cell might not be atoms, but more generic species :P

mfherbst:
Plural is boring: **species** as well ... see [wiktionary](https://en.wiktionary.org/wiki/species)

antoine-levitt:
Yeah but `species::Vector{Species}` is weird.

mfherbst:
Agreed ... I don't have a better idea at the moment however. I'm also not in favour if "atoms" to be fair.

PR #104, opened 2020-01-18
Composition

antoine-levitt:
Fix https://github.com/JuliaMolSim/DFTK.jl/issues/57
Essentially renaming, plus some convenience. I didn't go as far as I wanted on redesigning the way psp interface with the terms, I'm waiting to solve https://github.com/JuliaMolSim/DFTK.jl/issues/86
The idea going forward is that all external potentials come from atoms (that may be "virtual atoms" that exist just to provide an external potential). The atoms provide their potential and nonlocal form factors centered around 0, and the translations (structure factors) are handled elsewhere in the code. This simplifies the code, and makes it more straightforward to do perturbation wrt atomic positions.


antoine-levitt:
I'm getting erfc errors because of the doublefloat workaround, but I can't find where I messed up the workaround. Now that it's fixed in DoubleFloats, can we get rid of it?

src/Model.jl
```diff
@@ -16,11 +16,13 @@ struct Model{T <: Real}
     dim::Int # Dimension of the system; 3 unless `lattice` has zero columns
 
     # Electrons, occupation and smearing function
-    n_electrons::Int
+    n_electrons::Int # not necessarily consistent with `atoms` field
     spin_polarisation::Symbol  # :none, :collinear, :full, :spinless
     temperature::T
     smearing::Union{Nothing, Function} # see smearing_functions.jl for some choices
 
+    atoms # Vector of pairs AtomType => vector of vec3 (positions, fractional coordinates)
```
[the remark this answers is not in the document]
mfherbst: Also: Personally I prefer `nothing` to disable stuff, because the compiler has an easier time optimising it away. Not that it matters in this case ...

src/Model.jl
```diff
@@ -35,11 +37,19 @@ TODO docme
...
 """
-function Model(lattice::AbstractMatrix{T}, n_electrons; external=nothing,
+function Model(lattice::AbstractMatrix{T}; n_electrons=nothing, atoms=[], external=nothing,
                nonlocal=nothing, hartree=nothing, xc=nothing, temperature=0.0,
                smearing=nothing, spin_polarisation=:none) where {T <: Real}
     lattice = SMatrix{3, 3, T, 9}(lattice)
 
+    if n_electrons === nothing
+        # get it from the atom list
+        isempty(atoms) && @error "Either n_electrons or a non-empty atoms list should be provided"
```
mfherbst: You should use `error()` here, since `@error` just prints, but does not abort execution (it does not throw).

src/postprocess/band_structure.jl
```diff
@@ -3,9 +3,9 @@ include("../external/pymatgen.jl")
 
 # Functionality for computing band structures, mostly using pymatgen
 
-function high_symmetry_kpath(basis, kline_density, composition...)
+function high_symmetry_kpath(basis, kline_density, atoms)
```
mfherbst: I guess you could add a version without the `atoms` (getting them from the `basis`) for convenience.

src/postprocess/band_structure.jl
```diff
@@ -31,10 +31,10 @@ function compute_bands(ham::Hamiltonian, kpoints, n_bands;
     select_eigenpairs_all_kblocks(band_data, 1:n_bands)
 end
 
-function plot_bands(ham, n_bands, kline_density, composition, εF)
+function plot_bands(ham, n_bands, kline_density, atoms, εF)
```
mfherbst: Here I would not have the atoms at all and use the convenience function from above.

src/terms/term_external.jl
```diff
@@ -78,6 +78,7 @@ function term_external(generators_or_composition...; compensating_background=tru
         if energy !== nothing
             dVol = model.unit_cell_volume / prod(basis.fft_size)
             energy[] = real(sum(ρ.real .* Vext)) * dVol
+            # @assert energy[] ≈ real(dot(ρ.fourier, coeffs)) # sanity check
```
mfherbst: Remove that?

test/silicon_runners.jl
```diff
@@ -134,8 +134,8 @@ function run_silicon_pbe(T ;Ecut=5, test_tol=1e-6, n_ignored=0, grid_size=15, sc
     end
 
     energies = scfres.energies
-    energies[:Ewald] = energy_nuclear_ewald(model.lattice, Si => silicon.positions)
+    energies[:Ewald] = energy_nuclear_ewald(model.lattice, [Si => silicon.positions])
     energies[:PspCorrection] = energy_nuclear_psp_correction(model.lattice,
-                                                             Si => silicon.positions)
+                                                             [Si => silicon.positions])
```
mfherbst: This should not be needed any more now that you always compute it.

test/silicon_runners.jl
```diff
@@ -86,9 +86,9 @@ function run_silicon_lda(T ;Ecut=5, test_tol=1e-6, n_ignored=0, grid_size=15, sc
...
```
mfherbst: This should not be needed any more now that you always compute it.

mfherbst:
I'm not sure which workaround you refer to, but yes the `erfc` issue should be fixed in DoubleFloats.

examples/silicon.jl
```diff
@@ -13,52 +13,45 @@ kline_density = 20       # Density of k-Points for bandstructure
...
-println("\nEnergy breakdown:")
-for key in sort([keys(energies)...]; by=S -> string(S))
-    @printf "    %-20s%-10.7f\n" string(key) energies[key]
-end
-@printf "\n    %-20s%-15.12f\n\n" "total" sum(values(energies))
+print_energies(scfres.energies)
 
 # Plot band structure
-plot_bands(ham, n_bands_plot, kline_density, composition, scfres.εF).show()
+# plot_bands(ham, n_bands_plot, kline_density, atoms, scfres.εF).show()
```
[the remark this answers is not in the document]
antoine-levitt: my bad (I comment this for my tests)

src/AtomType.jl
```diff
@@ -0,0 +1,25 @@
+struct AtomType
```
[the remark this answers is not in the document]
antoine-levitt: My reasoning was that we might want Atom (representing an atom as participating in the computation, ie with a position etc.) whereas this is the abstract struct containing the information about a type of atoms.

src/Hamiltonian.jl
```diff
@@ -85,6 +85,12 @@ function update_energies!(energies, ham::Hamiltonian, Psi, occupation, ρ=nothin
     insert_energy!(:PotHartree, model.build_hartree; ρ=ρ)
     insert_energy!(:PotXC, model.build_xc; ρ=ρ)
     insert_energy!(:PotNonLocal, model.build_nonlocal; Psi=Psi, occupation=occupation)
+    if !isempty(model.atoms)
+        energies[:Ewald] = energy_nuclear_ewald(model)
```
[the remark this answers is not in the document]
antoine-levitt: I agree: Ewald should be just a term in the energy. Pending the term redesign it seemed simpler to have it here unconditionally.

src/Hamiltonian.jl
```diff
@@ -102,3 +108,12 @@ function update_energies_hamiltonian!(energies, ham::Hamiltonian, Psi, occupatio
     update_hamiltonian!(ham, ρ)
     compute_energies!(energies, ham, Psi, occupation, ham.density)
 end
+
+# TODO move that out into a scfres structure
```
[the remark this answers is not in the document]
antoine-levitt: Agreed

src/terms/term_external.jl
antoine-levitt: I think it's nice to have these kind of invariants explicit somewhere, makes debugging and figuring out the code much simpler. Maybe with an expensive assert?

antoine-levitt:
Just the fact that the tests fail here and I'm not sure why

src/postprocess/band_structure.jl
```diff
@@ -3,7 +3,7 @@ include("../external/pymatgen.jl")
 
 # Functionality for computing band structures, mostly using pymatgen
 
-function high_symmetry_kpath(basis, kline_density, atoms)
+function high_symmetry_kpath(basis, kline_density)
     bandstructure = pyimport("pymatgen.symmetry.bandstructure")
     pystructure = pymatgen_structure(basis.model.lattice, atoms)
```
mfherbst: Isn't this `basis.model.atoms` now?

mfherbst: Ok ... just saw the second commit ;).

src/terms/term_external.jl
mfherbst: Yes. I think that's a good idea.

src/AtomType.jl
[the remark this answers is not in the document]
mfherbst: Ah I see, good point. Still I think it is a duplicate to call a type `BlaType`. To stay in the chemistry language your `AtomType` would essentially be a model for an `Element`, actually. Otherwise I guess we could just leave it as `Species` as it was. Renaming the list of species to `atoms` is a good step, I think.

src/Hamiltonian.jl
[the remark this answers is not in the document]
mfherbst: Agree.

src/Hamiltonian.jl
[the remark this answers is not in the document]
mfherbst: Tracked in #114

mfherbst:
I think we have hit another really weird one with the failing test: https://github.com/MikeInnes/Requires.jl/issues/75

src/AtomType.jl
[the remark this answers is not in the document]
mfherbst: So should we just rename it to `Element`?

PR #109, opened 2020-01-21
Add simple callback function

mfherbst:


mfherbst:
Good idea. On the other hand for the planned short-term use case I don't mind NLsolve printing stuff.

PR #116, opened 2020-01-28
Integration of expensive tests and test case for running the examples once.

### LINKED ISSUE #70: Test examples and expensive tests
It would nice to explicitly test the examples and expensive tests from time to time. Perhaps using a different CI provider.

mfherbst:
Expensive tests and example tests are now integrated into the main test suite. By default the Travis-CI runs neither expensive / example, whereas tests on a "normal" computer run expensive tests, but not the examples.

To run only the non-expensive (fast) tests, issue
```
Pkg.test("DFTK"; test_args = ["fast"])
```

To only run the example tests, issue
```
Pkg.test("DFTK"; test_args = ["examples"])
```

Closes #70.


test/runexamples.jl
```diff
@@ -0,0 +1,15 @@
+using Test
+
+if run_examples
+    @testset "Run examples" begin
+        basedir = joinpath(@__DIR__, "..", "examples")
+        for file in readdir(basedir)
+            if isfile(joinpath(basedir, file)) && endswith(file, ".jl")
+                file == "graphene.jl" && continue  # Has to be run on cluster
```
antoine-levitt: can you take that out of the loop as a list of exceptions?

antoine-levitt:
Hm, I don't like the fact that the examples are not on CI. Can't we just trim them down a bit so it's feasible to run them there? It also makes it simpler to run and see quickly what's going on. Then we comment that if the user wants more realistic results they have to increase the cutoffs.

mfherbst:
Who says they won't be on CI?

antoine-levitt:
Ah, I didn't see the draft tag!

test/runexamples.jl
mfherbst: I'll solve that by adding markers inside the file. I think that's better (info is locally inside the example file).

mfherbst:
@antoine-levitt Let me know what you think ... The examples do not yet pass, but I'm working on that.

test/runtests.jl
```diff
@@ -1,41 +1,67 @@
 using Test
 using DFTK
 
+run_expensive = true  # Run expensive tests or not?
```
antoine-levitt: the logic here is a bit hard to follow, maybe add a comment describing the conditions under which each type of tests are run?

test/silicon_lda.jl
```diff
@@ -4,6 +4,13 @@ include("silicon_runners.jl")
     run_silicon_lda(Float64, Ecut=7, test_tol=0.03, n_ignored=0, grid_size=17, scf_tol=1e-5)
 end
 
+if run_expensive
```
antoine-levitt: make that caps so it's more clear it's a global?

antoine-levitt: also it's nice when we can run a specific test by itself by including the test directly, this breaks that

test/silicon_runners.jl
```diff
@@ -13,28 +13,30 @@ function run_silicon_redHF(T; Ecut=5, test_tol=1e-6, n_ignored=0, grid_size=15,
...
     fft_size = grid_size * ones(3)
     Si = Element(silicon.atnum, psp=load_psp(silicon.psp))
-    model = model_reduced_hf(Array{T}(silicon.lattice), [Si => silicon.positions], temperature=.1)
+    model = model_reduced_hf(Array{T}(silicon.lattice), [Si => silicon.positions],
+                             temperature=0.05)
     basis = PlaneWaveBasis(model, Ecut, silicon.kcoords, silicon.ksymops; fft_size=fft_size)
     ham = Hamiltonian(basis, guess_density(basis, [Si => silicon.positions]))
 
-    scfres = self_consistent_field(ham, n_bands, tol=scf_tol)
+    scfres = self_consistent_field(ham, n_bands, tol=scf_tol, mixing=KerkerMixing())
```
antoine-levitt: i don't think we want kerker mixing for silicon?

mfherbst: We do, because Silicon reduced HF is actually a metal ;)

test/silicon_lda.jl
mfherbst: Good point, I'll add a check whether the variable is defined.

test/runtests.jl
mfherbst: You're right, the logic is a bit convoluted. I'll try to make that clearer and add some comments.

test/silicon_runners.jl
antoine-levitt: Yeah but if you don't do supercells there's no problem of coulomb divergence and you don't need preconditioning. Kerker only acts on the low wavevectors anyway.

mfherbst: Indeed: Tests work without it. I must have added it to check weather this was the cause for something else.

PR #120, opened 2020-02-07
High-level PlaneWaveBasis constructor interface

mfherbst:
Adds a high-level interfaces to PlaneWaveBasis where `kgrid` can be directly passed.

### DIFF
--- a/src/PlaneWaveBasis.jl
+++ b/src/PlaneWaveBasis.jl
@@ -84,16 +84,9 @@ fft_size is now Fourier grid size
 kcoords is vector of Vec3
 """
 function PlaneWaveBasis(model::Model{T}, Ecut::Number,
-                        kcoords::AbstractVector, ksymops=nothing, kweights=nothing;
-                        fft_size=nothing) where {T <: Real}
-    # TODO this constructor is too low-level. Write a hierarchy of
-    # constructors starting at the high level
-    # `PlaneWaveBasis(model, Ecut, kgrid)`
-
+                        kcoords::AbstractVector, ksymops, kweights=nothing;
+                        fft_size=determine_grid_size(model, Ecut)) where {T <: Real}
     @assert Ecut > 0
-    if fft_size === nothing
-        fft_size = determine_grid_size(model, Ecut)
-    end
     fft_size = Tuple{Int, Int, Int}(fft_size)
 
     # TODO generic FFT is kind of broken for some fft sizes
@@ -115,9 +108,6 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Number,
     ipIFFT = inv(ipFFT)
     opIFFT = inv(opFFT)
 
-    # Default to no symmetry
-    ksymops === nothing && (ksymops = [[(Mat3{Int}(I), Vec3(zeros(3)))]
-                                         for _ in 1:length(kcoords)])
     # Compute weights if not given
     if kweights === nothing
         kweights = [length(symops) for symops in ksymops]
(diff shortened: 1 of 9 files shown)


src/PlaneWaveBasis.jl
```diff
@@ -138,6 +128,18 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Number,
         kweights, ksymops, fft_size, grids, opFFT, ipFFT, opIFFT, ipIFFT
     )
 end
+function PlaneWaveBasis(model::Model{T}, Ecut::Number, kgrid_size::AbstractVector;
+                        bzmesh_symmetry=:full, kwargs...) where {T <: Real}
+    kcoords, ksymops = nothing, nothing
```
antoine-levitt: Don't need that

src/PlaneWaveBasis.jl
```diff
@@ -138,6 +128,18 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Number,
         kweights, ksymops, fft_size, grids, opFFT, ipFFT, opIFFT, ipIFFT
     )
 end
+function PlaneWaveBasis(model::Model{T}, Ecut::Number, kgrid_size::AbstractVector;
+                        bzmesh_symmetry=:full, kwargs...) where {T <: Real}
```
antoine-levitt: Don't need to {T}

antoine-levitt: I think just make `bzmesh_symmetry` a bool. People needing finer control just call the other constructor.

antoine-levitt:
Great, thanks for doing it! Has been in the back of my mind for a while but never got to it.

src/PlaneWaveBasis.jl
mfherbst: True, since when ... I always had in mind `if` opens a new scope ...

antoine-levitt: I always get very confused by this so for more complicated cases it's probably worth it to be a bit more explicit, but in this case it's clear

src/bzmesh.jl
```diff
@@ -33,6 +33,7 @@ representing a mapping from `Element` objects to a list of positions in fraction
 coordinates. `tol_symmetry` is the tolerance used for searching for symmetry operations.
 """
 function bzmesh_ir_wedge(kgrid_size, lattice, atoms; tol_symmetry=1e-5)
+    all(isequal.(kgrid_size, (1, 1, 1))) && return bzmesh_uniform(kgrid_size)
```
antoine-levitt: 1 instead of (1, 1, 1) should work too here.

src/PlaneWaveBasis.jl
```diff
@@ -138,6 +128,15 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Number,
         kweights, ksymops, fft_size, grids, opFFT, ipFFT, opIFFT, ipIFFT
     )
 end
+function PlaneWaveBasis(model::Model, Ecut::Number, kgrid_size::AbstractVector;
```
antoine-levitt: Do we want to support a number for the third argument to be the kpoint density? Or is that too tricky? Also for Ecut do we want to accept a 3-tuple of cutoffs for every direction, for easier support of 1d systems?

mfherbst: For me that really falls into "advanced use", actually. For 1D you can also just explicitly specify the fft_size (and then you have even more control what happens). That's not disabled by the current interface. The kpoint density should be feasible.

antoine-levitt: For 1D for sure that's advanced use. kpoint density is more user-friendly than kgrid_size, so maybe we support k density and leave the Ecut to the fft_size.

antoine-levitt: but it's a bit tricky that having `[2,2,2]` as argument is different from `2` so maybe that's too magical

mfherbst: Maybe I misunderstood your previous comment. I could leave it as is now. Is that fine?

antoine-levitt: I like kwargs but that's a bit tricky. I would naively expect no info => no kpoint. Also we have to be careful that the user doesn't pass both kspacing and kgrid. That does look too complicated, so let's just revert to what it was before (sorry!), and the user can just call `PWBasis(model, Ecut, kgrid_size_from_minimal_spacing(lattice, spacing))`

mfherbst: fair enough ... even though it would be nice to have a method, which automatically does sane defaults for kgrid and Ecut, actually ...

antoine-levitt: Yeah; that we probably want to build on top. Ideally, the interface would be to just give an accuracy epsilon and the code would figure out the rest, but we're not there yet

PR #121, opened 2020-02-07
Allow to start ABINIT calculations using DFTK datastructures

mfherbst:
Uses abipy to do the interfacing.


src/external/abinit.jl
```diff
@@ -0,0 +1,128 @@
...
+    EtsfFolder(outdir)
+end
+
+
+"""
+Run an SCF in ABINIT starting from a DFTK `Model` and some extra parameters.
+Write the result to the `output` directory in ETSF Nanoquanta format
+and return the `EtsfFolder` object.
+"""
+function run_abinit_scf(model::Model, outdir;
```
antoine-levitt: Should we maybe make it so that PWbasis contains all the info needed for that? Eg add some psp metadata. Maybe also the nbands should be part of it...

antoine-levitt: Point is it might be good to have a DFTK object that maps cleanly to "all that is needed to run a SCF in a PW code". Whether that's in PWBasis or somewhere else I'm not sure

mfherbst: With the last point I agree, with the first I'm not too sure. For me this interface to ABINIT is just convenience and I do not think this should guide our design. The psp thing could be solved for example by being able to write or read ABINIT psp files, but I just could not be bothered to implement that for now ...

test/testcases_ABINIT/graphite_E20_k8_LDA_Methfessel_Paxton/generate.jl
```diff
@@ -1,18 +1,22 @@
...
-    ecut=20,        # Hartree
-    nband=12,       # Number of bands
-    tolvrs=1e-10,   # General tolerance settings
-    ixc=1,          # Teter LDA
-    occopt=6,       # Methfessel and Paxton, Hermite polynomial degree 2
-    tsmear=0.01,    # Hartree
-)
-infile.extra = Dict("pspmap" => Dict(6 => "hgh/lda/c-q4", ), )
+# Note: This is not exactly the minimum-energy structure
+a = 1.228units.Ǎ
```
antoine-levitt: 123a.b syntax is a bit weird (I'm fine with it, just noting)

src/external/abinit.jl
antoine-levitt: Having metadata (eg what the psp file was) can't hurt, I'm keeping that in mind for the refactor. OK, let's keep it as is. PWbasis = discretization only. Solving the discretized equations is another matter.

mfherbst: > Having metadata (eg what the psp file was) can't hurt

But that is already there: Inside the atoms. It's just in this case I write stuff to disk in ABINIT format, where the DFTK pseudos are lost and then I was lazy.

antoine-levitt: Ah I see; never mind then!

PR #125, opened 2020-02-12
Fix a bug with building the total density

### LINKED ISSUE #113: No corresponding symop found
Test with silicon.jl and
```
kgrid = [1, 4, 4]        # k-Point grid
supercell = [2, 1, 1]    # Lattice supercell
```

mfherbst:
I rederived our $k$-Point symmetry stuff (will add the latex doc as well) and it all boils down to a forgotten minus [here](https://github.com/JuliaMolSim/DFTK.jl/compare/fix-kbug?expand=1#diff-a4da22d96f986b574106eb686556f859R100).

I also cleaned up the code and made it more consistent with the conventions in my document.

- [x] Find k-Point bug in density builder
- [x] Add aluminium test case
- [x] Documentation and derivation
- [x] Fix #113


antoine-levitt:
A bit scary but nice you caught that! Do you plan to commit the latex? Or as comments in the code? We didn't really settle on a doc system...

mfherbst:
For now I'm committing the latex, just that it's there. We can think about the details later.

mfherbst:
Ok now we hit #113 in the tests for this PR. On my machine it worked until I upgraded spglib from 1.12 to 1.14. I guess it's this changeset: https://github.com/atztogo/spglib/blob/master/ChangeLog#L35.

antoine-levitt:
For the latex: careful that it's not just Stilde R + tautilde = R, it also has to leave invariant the atomic environment, so that V is covariant. I'd add the proof that V(S x + tau) = V(x), since that relies on the former argument, as well as the fact that the potential (both the local and nonlocal parts) is radially symmetric

antoine-levitt:
Also for the "taking appropriate phase factors into account", we've got to be careful here, since symmetry and degeneracy sometimes have nontrivial interactions. I'd just say that if (unk) is an orthogonal eigenbasis of Hk, then (U unk) is an orthogonal eigenbasis of HSk

antoine-levitt:
Otherwise looks good! With the minor gripe that bold and underline are inconsistently applied, and that bold plus underline is way too much. I know you're not going to like that but I'd just do s/\vec// :-p

src/bzmesh.jl
```diff
@@ -45,17 +45,47 @@ function bzmesh_ir_wedge(kgrid_size, lattice, atoms; tol_symmetry=1e-5)
...
+    # Notice: In the language of the latex document in the docs
+    # spglib returns \tilde{S} and \tilde{τ} in integer real-space coordinates, such that
+    # (A Stilde A^{-1}) is the actual \tilde{S} from the document as a unitary matrix.
+    #
+    # Still we have the following properties for S and τ given is *integer* and
+    # *fractional* real-space coordinates:
+    #      - \tilde{S}^{-1} = S^T (if applied to a vector in frac. coords in reciprocal space)
+
+    # Checks: (A Stilde A^{-1}) is unitary
+    for Stilde in eachslice(Stildes; dims=1)
```
antoine-levitt: eachrow

src/bzmesh.jl
```diff
@@ -45,17 +45,47 @@ function bzmesh_ir_wedge(kgrid_size, lattice, atoms; tol_symmetry=1e-5)
...
-        Scart = lattice * S * inv(lattice)  # Form S in cartesian coords
+    Stildes = spg_symops["rotations"]
+    τtildes = spg_symops["translations"]
+    n_symops = size(Stildes, 1)
+    @assert size(τtildes, 1) == n_symops
+    # Notice: In the language of the latex document in the docs
+    # spglib returns \tilde{S} and \tilde{τ} in integer real-space coordinates, such that
+    # (A Stilde A^{-1}) is the actual \tilde{S} from the document as a unitary matrix.
+    #
+    # Still we have the following properties for S and τ given is *integer* and
```
antoine-levitt: in

src/bzmesh.jl
```diff
@@ -45,17 +45,47 @@ function bzmesh_ir_wedge(kgrid_size, lattice, atoms; tol_symmetry=1e-5)
...
+    for Stilde in eachslice(Stildes; dims=1)
+        Scart = lattice * Stilde * inv(lattice)  # Form S in cartesian coords
         if maximum(abs, Scart'Scart - I) > sqrt(eps(Float64))
             error("spglib returned non-unitary rotation matrix")
         end
     end
-    # TODO check that S * lattice + τ ∈ lattice
+
+    # Check (Stilde, τtilde) maps atoms to equivalent atoms in the lattice
+    for (Stilde, τtilde) in zip(eachslice(Stildes; dims=1), eachslice(τtildes; dims=1))
```
antoine-levitt: eachrow

src/bzmesh.jl
```diff
@@ -45,17 +45,47 @@ function bzmesh_ir_wedge(kgrid_size, lattice, atoms; tol_symmetry=1e-5)
...
+        for (elem, positions) in atoms
+            # Rationalize atomic positions
+            positions = [rationalize.(pos, tol=tol_symmetry) for pos in positions]
```
antoine-levitt: Oh that's very cool

antoine-levitt: Although that looks a bit brittle - are you sure it's not going to cause trouble, and wouldn't it be simpler to work in floats and just check the non-integer part is not too big?

src/bzmesh.jl
```diff
@@ -76,20 +106,20 @@ function bzmesh_ir_wedge(kgrid_size, lattice, atoms; tol_symmetry=1e-5)
...
 
             if isym === nothing
                 error("No corresponding symop found for $k -> $kred. This is a bug.")
             else
-                S = spg_symops["rotations"][isym, :, :]
-                τ = spg_symops["translations"][isym, :]
-                push!(ksymops[ik], (S', -inv(S) * τ))
+                S = Stildes[isym, :, :]'  # in fractional reciprocal coordinates
+                # τ in fractional real-space coordinates:
+                τ = -inv(Stildes[isym, :, :]) * τtildes[isym, :, :]
```
antoine-levitt: inv(A)*b -> A\b

src/densities.jl
```diff
@@ -97,7 +97,7 @@ function _symmetrize_ρ!(ρaccu, ρin, basis, ksymops, Gs)
         for (ig, G) in enumerate(Gs)
             igired = index_G_vectors(basis, invS * G)
             if igired !== nothing
-                @inbounds ρaccu[ig] += cis(2T(π) * dot(G, τ)) * ρin[igired]
+                @inbounds ρaccu[ig] += cis(-2T(π) * dot(G, τ)) * ρin[igired]
```
antoine-levitt: Maybe briefly rederive this if possible? Eg just write down the relationship between uSk and uk and what this implies for rho(G)

src/bzmesh.jl
mfherbst: Doesn't work for Arrays of dim 3

src/bzmesh.jl
mfherbst: see below.

src/bzmesh.jl
antoine-levitt: Ah right, sorry

mfherbst: No worries ;)

PR #127, opened 2020-02-19
Free energy for convergence and printouts with temperature

### LINKED ISSUE #105: Free energy for computations with temperature
In case of computations at finite temperature, the energy is not variational wrt the orbitals, which makes it harder to compute derivatives such as forces. We should instead compute the free energy, which has an entropic contribution.

mfherbst:
Closes #105.

### DIFF
--- a/src/scf/self_consistent_field.jl
+++ b/src/scf/self_consistent_field.jl
@@ -25,11 +25,14 @@ function next_density(ham::Hamiltonian, n_bands; Psi=nothing,
 end
 
 function scf_default_callback(info)
-    E = sum(values(info.energies))
+    E = sum(values(info.energies)) + info.entropy
     res = norm(info.ρout.fourier - info.ρin.fourier)
     neval = info.neval
+
     if neval == 1
-        println("Iter   Energy             ρout-ρin")
+        label = "Energy"
+        info.entropy < 0 && (label = "E - TS")
+        println("Iter   $label             ρout-ρin")
         println("----   ------             --------")
     end
     @printf "%3d    %-15.12f    %E\n" neval E res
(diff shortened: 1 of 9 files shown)


antoine-levitt:
Why not just add it to the energy terms? I think it makes sense as the free energy really is the natural quantity of interest. Then it can be special cased in the printing if we want to. As a test, just use the forces test at finite temp (pick it huge, because it's on silicon)

src/scf/self_consistent_field.jl
```diff
@@ -96,6 +98,7 @@ function self_consistent_field(ham::Hamiltonian, n_bands;
         Psi, orben, occupation, εF, ρout = next_density(ham, n_bands; Psi=Psi,
                                                         eigensolver=eigensolver, tol=diagtol)
         energies = update_energies(ham, Psi, occupation, ρout)
+        energies[:Entropy] = -model.temperature * compute_entropy(basis, orben, εF=εF)
```
antoine-levitt: For consistency, compute_entropy should match Entropy. So either we call Entropy -TS, or we put the -T into compute_entropy. I'm in favor of option 2 (this will become a Term anyway)

mfherbst: I see your point, but I get slight stomach ache from calling "-TS" the entropy ... I'll call it `compute_entropy_term` instead.

mfherbst: If this becomes a term, how would you handle the dependency on `ε_F`?

antoine-levitt: Then put the - T in the compute function. Good point on epsf, I'll just pass it as a kwarg

mfherbst: Sounds reasonable ...

PR #135, opened 2020-02-25
Implement fallback kpoint symmetry search

### LINKED ISSUE #128: Optimise fallback kpoint symmetry search
Implement what is mentioned in https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/bzmesh.jl#L133.

mfherbst:
A bit of refactoring in bzmesh.jl and a fallback kpoint symmetry search.

Closes #128

### DIFF
--- a/src/bzmesh.jl
+++ b/src/bzmesh.jl
@@ -21,6 +21,46 @@ function bzmesh_uniform(kgrid_size)
 end
 
 
+SymOp = Tuple{Mat3{Int}, Vec3{Float64}}
+function find_irreducible_kpoints(kpoints, Stildes, τtildes)
+    n_symops = length(Stildes)
+
+    # Flag which kpoints have already been mapped to another irred.
+    # kpoint or which have been decided to be irreducible.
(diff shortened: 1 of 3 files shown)


src/bzmesh.jl
```diff
@@ -21,6 +21,46 @@ function bzmesh_uniform(kgrid_size)
 end
 
 
+SymOp = Tuple{Mat3{Int}, Vec3{Float64}}
```
antoine-levitt: Const

src/bzmesh.jl
```diff
@@ -21,6 +21,46 @@ function bzmesh_uniform(kgrid_size)
...
+function find_irreducible_kpoints(kpoints, Stildes, τtildes)
```
antoine-levitt: Document that this is to work around spglib weirdness

src/bzmesh.jl
mfherbst: That works on types, too ... I did not know ...

PR #140, opened 2020-03-03
Implement Cohen Bergstresser

### LINKED ISSUE #136: Implement Cohen Bergstresser model
A nice test case would be to implement the [Cohen Bergstresser](https://doi.org/10.1103/PhysRev.141.789) model.

mfherbst:
closes #136 

- I first thought about having an element( ) function to multiplex between different models, but I guess we can easily add that later.
- I did not yet add the non-local Fourier stuff in the Element as we discussed. I will do that when I work on the Termsokalypse.


src/elements.jl
```diff
@@ -0,0 +1,144 @@
+import PeriodicTable
+
+# Alias to avoid similarity of elements and Element in DFTK module namespace
+periodic_table = PeriodicTable.elements
+
+# Data structure for chemical element and the potential model via which
+# they interact with electrons. A compensating charge background is
+# always assumed.
+
+abstract type AbstractElement end
```
antoine-levitt: I liked the name `Element` for this. The fact it's an abstract type doesn't have to show up in the type name, I think it's mostly used for when there's a disambiguation (eg Array vs AbstractArray).

src/elements.jl
```diff
@@ -0,0 +1,144 @@
...
+function local_potential_real(el::AbstractElement, q::AbstractVector)
+    local_potential_real(el, norm(q))
+end
+
+
+"""
+Element interacting with electrons via a bare Coulomb potential
+(for all-electron calculations)
+"""
+struct ElementAllElectron <: AbstractElement
```
antoine-levitt: `bare` instead of ae? Eg `BareAtom`?

src/elements.jl
```diff
@@ -0,0 +1,144 @@
...
+    Z::Int  # Nuclear charge
+    symbol  # Element symbol
+end
+charge_ionic(el::ElementAllElectron) = el.Z
+
+function ElementAllElectron(key)
```
antoine-levitt: What is key? String, symbol?

src/elements.jl
```diff
@@ -0,0 +1,144 @@
...
+"""Radial local potential, in Fourier space: V(q) = int_{R^3} V(x) e^{-iqx} dx."""
+function local_potential_fourier(el::ElementAllElectron, q::T) where {T <: Real}
+    q == 0 && return zero(T)  # Compensating charge background
+    # General atom => Use default Coulomb potential
+    # We use int_R^3 1/r e^{-i q⋅x} = 4π / |q|^2
+    return -4T(π) * el.Z / q^2
+end
+
+"""Radial local potential, in real space."""
+local_potential_real(el::ElementAllElectron, r::Real) = -el.Z / r
```
antoine-levitt: We should test the consistency of this and the Fourier part (not necessarily in this PR) by doing a numerical (radial) fourier transform. To avoid singularities we can test Parseval's identity numerically against a nice test function, eg a gaussian.

src/elements.jl
```diff
@@ -0,0 +1,144 @@
...
+charge_ionic(el::ElementPsp) = el.psp.Zion
+
+function local_potential_fourier(el::ElementPsp, q::T) where {T <: Real}
+    q == 0 && return zero(T)  # Compensating charge background
+    eval_psp_local_fourier(el.psp, q)
+end
+
+function local_potential_real(el::ElementPsp, r::Real)
+    # Use local part of pseudopotential defined in Element object
+    return eval_psp_local_real(el.psp, r)
```
antoine-levitt: Maybe we should make clear that this is HGH. Maybe put PspHgh in a module? But probably this is premature until we actually support other pseudos

src/elements.jl
```diff
@@ -0,0 +1,144 @@
...
+end
+charge_ionic(el::ElementCohenBergstresser) = 2
+
+function ElementCohenBergstresser(key; lattice_constant=nothing)
+    element = periodic_table[key]
+
+    # Form factors from Cohen-Bergstresser paper Table 2, converted to Bohr
+    # Lattice constants from Table 1, converted to Bohr
+    V_sym_paper = Dict{Int64,Float64}()
+    if element.symbol == "Si"
```
antoine-levitt: Put the data (vsyms and lattice constants) in a table to avoid code duplication?

src/elements.jl
mfherbst: Nah. That sounds strange to me. Maybe `ElementBare` or `ElementCoulomb`?

src/elements.jl
mfherbst: ok.

src/elements.jl
mfherbst: Actually all of them. I'll add some docs.

src/elements.jl
mfherbst: I had this idea as well. Let's put that on the list.

src/elements.jl
mfherbst: I'd wait. Probably we will understand better what we actually need.

src/elements.jl
mfherbst: Not sure what you mean exactly. Maybe what I do in the Cohen-Bergstresser example?

src/elements.jl
```diff
@@ -145,7 +144,12 @@ end
 function local_potential_fourier(el::ElementCohenBergstresser, q::T) where {T <: Real}
     q == 0 && return zero(T)  # Compensating charge background
 
+    # Number of digits to keep in rounding (depending on element type)
+    digits = 10
+    eps(T) > 1e-13 && (digits = 3)
+    eps(T) > 1e-6 && (digits = 2)
+
```
antoine-levitt: That logic looks very ad hoc. Why not just round? The argument is an integer up to machine precision, no? There could perhaps be trickiness for very large q, but we don't care about those anyway.

mfherbst: Not if you use a different lattice constant in the lattice and in the CohenBergstresser potential, which you should not, but I definitely want it to flag somehow.

src/elements.jl
antoine-levitt: pot_elements = Dict("Si" => [-.21, .04, .08]); lattice_constants = Dict("Si" => 5.66); etc

mfherbst: Nah, because it might not be just those three G^2 values, so that is actually too restrictive. And if you make Dict("Si" => Dict(4 => -.21, 5 => .04, 7 => .08]); lattice_constants = Dict("Si" => 5.66)
that becomes convoluted too.

antoine-levitt: It is restrictive, that's kind of the point : we are not going to generalize cohen-bergstresser, it's a simple demo example. I'm just saying, you get the same result with less copy/paste. And if we want to generalize later on it's simpler to have to not do changes three times.

mfherbst: Ok, I'll use a dict, but I'd like to make the units explicit. Makes it still a bit better, though.

src/elements.jl
antoine-levitt: I see. But this is conflating two different things that are unrelated: numerical precision and whether you are doing something weird with the lattice or not. I'd just use `digits=2` always, with a comment saying that this only allows for small deformations of the unit cell. A more clever option is to cook up an interpolation formula that passes through the three data points and gives V(q) for all q, but we don't particularly care anyway...

src/elements.jl
antoine-levitt: Sorry to be a pain: it's fine like this, we don't particularly care about this code anyway, I just have a visceral hate of this kind of copy-pasted code (essentially because I always screw it up)

PR #143, opened 2020-03-05
Custom plotter for plotting bandstructures

mfherbst:


src/common/units.jl
```diff
@@ -0,0 +1,18 @@
...
+    K  = 3.166810508e-6      # Kelvin -> Hartree
+    Ry = 0.5                 # Rydberg -> Hartree
+end
+
+"""
+    unit(symbol)
+Get the factor converting from the unit `symbol` to atomic units.
+E.g. `unit(:eV)` returns the conversion factor from electron volts to Hartree.
+"""
+unit_to_au(sym::Symbol) = getfield(units, sym)
```
antoine-levitt: Inconsistent with docstring (see why I hate those? :p)

examples/gross_pitaevskii.jl
```diff
@@ -5,6 +5,7 @@ using PyCall
 using DFTK
 using Printf
 using LinearAlgebra
+using PyPlot
```
antoine-levitt: Do we want to go full Plots and drop PyPlot everywhere?

src/postprocess/band_structure.jl
```diff
@@ -21,28 +22,81 @@ function high_symmetry_kpath(basis, kline_density)
...
-    bs = pymatgen_bandstructure(basis, band_data, klabels, fermi_level=εF)
-    bsplot = plotter.BSPlotter(bs)
-    plt = bsplot.get_plot()
-    plt.autoscale()
-    plt
+    data = plotter.BSPlotter(bs).bs_plot_data(zero_to_efermi=false)
+
+    # Collect some useful info in a more useful way
+    spins = [("1", :blue)]
+    bs.is_spin_polarized && (spins = [("1", :blue), ("-1", :red)])
```
antoine-levitt: Might as well do an if

antoine-levitt:
Cool! The conflicts with termpocalypse are building up and it's making me a bit uneasy, though...

mfherbst:
I know, but trust in git ;), also I really want to get this stuff working first before I loose a day for merging.

src/common/units.jl
mfherbst: :smile:

examples/gross_pitaevskii.jl
mfherbst: Yes, I'm on that. There are a few tiny things though, which came up related to labels taking real parts, etc.

src/postprocess/band_structure.jl
mfherbst: ok.

examples/gross_pitaevskii.jl
mfherbst: It's really annoying that the default GR backend does not support unicode, actually.

PR #144, opened 2020-03-06
fix LDOS

antoine-levitt:
Woopsie!

The normalizations are the same as in the old code (and are consistent with the old code whenever enable_bzmesh_symmetry). I *think* I checked them at the time...

### DIFF
--- a/src/postprocess/DOS.jl
+++ b/src/postprocess/DOS.jl
@@ -67,13 +67,14 @@ function LDOS(ε, basis, orben, Psi; smearing=basis.model.smearing, T=basis.mode
     filled_occ = filled_occupation(basis.model)
     D = zeros(real(eltype(Psi[1])), basis.fft_size)
     T != 0 || error("LDOS only supports finite temperature")
+    weights = deepcopy(orben)
     for ik = 1:length(orben)
-        @views for iband = 1:length(orben[ik])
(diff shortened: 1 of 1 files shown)


mfherbst:
That actually automatically parallelises the LDOS computation as well :)

mfherbst:
I've put in an extra comment and removed the void statement to allocate D.

mfherbst:
Let's not discuss this now. I'll revert it and we put it on the list.

I find the `=` confusing (it's not an assignment but a loop), but it seems the julia community is not clean about it.

mfherbst:
Definitely a space after an `=` though.

antoine-levitt:
Hm. C and python don't have a serious iteration syntax (python makes you write `range`, C syntax just sucks) so they don't count. Matlab and Fortran both use `=`, so I'd argue precedent in numerical computing is in the `=` camp. Numeric loops and set loops are often very different (numeric loops are usually lower-level), and it's kind of nice to distinguish them at the syntax level. That's what mathematical formulas do: you write `\sum_{i=1}^N` but `\forall x \in X`

mfherbst:
Ok I have to admit you got a point there. All right, let's go for
```
for i = 1:5
```
but 
```
for a in complex_list
```
ok?

antoine-levitt:
Yeah that's what I usually do (but might be inconsistent...)

PR #145, opened 2020-03-07
Progress bar during kpoint computation

mfherbst:
Could be a bit too much, but especially for larger grids I find this very nice.


src/postprocess/band_structure.jl
```diff
@@ -23,10 +23,11 @@ function high_symmetry_kpath(basis; kline_density=20)
 end
 
 function compute_bands(ham::Hamiltonian, kpoints, n_bands;
-                       eigensolver=lobpcg_hyper, tol=1e-5)
+                       eigensolver=lobpcg_hyper, tol=1e-5, show_progress=true)
```
antoine-levitt: kwargs and pass to diagonalise_all_kblocks ?

antoine-levitt:
It's very cool that it's so easy to add. My criticism would be that for eg supercell computations where you don't have kpoints this is useless. It would be cool if we could combine that with a progess bar for LOBPCG (would have the added benefit of showing how many LOBPCG steps are taken). Also percentage is not really relevant for a small number of steps.

mfherbst:
Yeah, that's true. ProgressMeter also has a way to display progress depending on convergence and thresholds, but it's all pretty liner and not hierarchical enough to easily combine it with inner iterations / thresholds like in an interior LOBPCG.

src/postprocess/band_structure.jl
mfherbst: Why not ...

antoine-levitt: I meant use it for tol and show_progress

mfherbst: Does not work, because it has different defaults.

antoine-levitt:
Oh I didn't see that. I think we want it to be on by default during SCF no? If it disappears after printing then it's completely fine (this should only be turned on in interactive shells anyway). I think something like

Kpt 3/15 LOBPCG iter 13 resid 1.2E-3

? At that point we might just as well roll out our own.

mfherbst:
During SCF I am not so sure. I view the SCF to be the thing you do on the server and which you expect to take long and the postprocessing which is done interactively. ProgressBar does not disappear after it is finished, but displays the total runtime. So using it during the SCF spams a bit the output in my opinion (also for interactive sessions). 

But having a small bar like you suggested kind of at the bottom of the output, which is continuously overwritten, would be awesome, but that could be a bit more tricky.

mfherbst:
@antoine-levitt Are you fine if merged as such?

antoine-levitt:
Heh, that depends on the usage. But the SCF can take several minutes on an interactive session, and it can be interesting to see what's going on, also to diagnose convergence trickiness (eg if the LOBPCG takes a lot of iterations). I don't really like total runtimes or percentage bars, as they invariably lie (I have windows PTSD). I think having a self-erasing bar should be pretty easy. Let's discuss next week, I'll merge as is.

antoine-levitt:
https://en.wikipedia.org/wiki/ANSI_escape_code#CSI_sequences
So erasing the current line is pretty simple: 
```
julia> print("a"); print("\u1b[G");print("\u1b[K"); print("b")
b
```

mfherbst:
Yes ... unless you pipe that through a grep or into a file. Then all hell breaks loose.

mfherbst:
... and that's exactly the tricky business. Essentially you need to find when one of these things happen and then not print the progress bar.

mfherbst:
Oh and if something else prints in between than you might erase that too or otherwise get inconsistent output, so one has to be careful with that, too.

antoine-levitt:
For checking for interactivity, https://rosettacode.org/wiki/Check_input_device_is_a_terminal#Julia looks fine. For in-between printing it's a fair point, I'm not sure there's a simple way to do it, so we might just assume there's no printing; if there is they'll just have to `println()` before doing their stuff

mfherbst:
That's exactly my point though ... anyway. If we add some callback mechanism to the LOBPCG, then what you propose could be easily done, I guess.

antoine-levitt:
Yup, that's what we should do. (btw, I'm leveling up my git skills by trying to merge termspocalypse against master, will let you knwo how that goes)

mfherbst:
Ah ... hmm I wanted to get started with that later today, but since this involves heavy rebasing, we definitely should not work both on it at the same time. Let me know once you can "give up the lock".

antoine-levitt:
I can never get rebases working, I'm using https://github.com/mhagger/git-imerge

mfherbst:
Interesting. So you use git-imerge rebase?

antoine-levitt:
I don't know, I'm following what magit tells me to do. Not doing so well for now :D

PR #150, opened 2020-03-14
Tests for diagonalisation routines and removal of scipy / IterativeSolver dependencies

mfherbst:
- scipy code was unused and broken (some issues with the tuple of data returned), so I removed it
- Iterative solvers is now an optional dependencies and since the routines do still not work, I added a warning.


src/DFTK.jl
```diff
@@ -149,4 +151,10 @@ export load_atoms
 include("external/etsf_nanoquanta.jl")
 include("external/abinit.jl")
 
+function __init__()
```
antoine-levitt: Could you make a comment here explaining the mechanism?

antoine-levitt:
Ok!

src/DFTK.jl
mfherbst: Yes, I probably should. Good point.

PR #151, opened 2020-03-14
Load generic FFT fallbacks in FourierTransforms only optionally

### LINKED ISSUE #132: Precompilation times
They are bearable but still annoying when developing. Empirically (meaning when I ctrl-C precompilation) it seems quite a bit of time is spent in FourierTransforms.jl; can we maybe load it conditionally, or even better upstream our changes?

mfherbst:
Builds on #150. Fixes #132 (to a large extend).


antoine-levitt:
Very nice!!

antoine-levitt:
`@time using DFTK` (including precompilation of DFTK) goes from 80s to 35s, very cool!

PR #153, opened 2020-03-14
Improve kpoint and bloch wave interpolation

mfherbst:
TODO:
- [ ] Generalise kpoint and Bloch wave interpolation?
- [ ] Tests for bloch wave interpolation


mfherbst:
- The reason for the dict is that it it's faster than using an index search into the mapping. I understand the memory concern. I guess it's a question whether we want this to be fast or don't care that much.
- The main use case is translating quantities from one basis to another. For this one could certainly go over the vectors of the k-Point directly in a double loop. I guess it depends on Ecut what is really faster. Also a question whether it matters ... I'm not so sure.

antoine-levitt:
You mean converting from one kpoint to the other, right? Otherwise we just use the mapping. For converting between kpoints, we could either do it like that, or by going to the cubic grid. In any case the convenience is probably worth it, although do add a note that it might not be the most performant way

mfherbst:
Indeed that was my idea. If we do not want to optimise this for performance, than I guess dropping the dict and just using `indexin` on the `kpoint.mapping` could be the simpler/better option. What do you thing?

antoine-levitt:
I'm OK with the current version. The dict should not consume that much more memory than the mapping anyway.

mfherbst:
I need the Bloch wave interpolation for the residual computation for the Bauer-Fike, so I added my routine here. Maybe one should spend the effort to generalise and merge with the kpoint interpolation. @antoine-levitt Do you think that would be worth it for the moment?

antoine-levitt:
Oh for sure! the kpoint interpolation stuff we have sucks, would be great to have a better one

antoine-levitt:
when I say sucks, I mean probably using random initial guess would give the same result (haven't tried it)

mfherbst:
Closed in favour of #185.

PR #158, opened 2020-03-19
Apply chi0

antoine-levitt:


antoine-levitt:
Not tested in the wild (eg with inexact eigenvectors) and not optimized, but functional.

PR #181, opened 2020-03-20
Fixing compatibilities

mfherbst:


mfherbst:
@antoine-levitt The last commit bumps us to Julia 1.3 ... let me know if you have issues with that.

antoine-levitt:
Oh no I'm completely fine with that

PR #183, opened 2020-03-21
Improve psp-related things

mfherbst:
Collection of small-ish improvements related to psps, which will help for computing errors.

- [x] Tests for qcut functions
- [x] Deal with code duplication polynomial versus eval function
- [x] Compat bound for Polynomials
- [x] Better name than `pseudize`.


mfherbst:
@antoine-levitt Two things to discuss here from my end:
- The idea of `pseudize`: I did not like that you have to pass a `pspmap` for `load_atoms` whenyou are reading from third-party datastructures, so now `load_atoms` returns Coulombic atoms if it has no clue of psps and you can later select the psps with `pseudize`. Also this allows to get rid of the annoying repetitive element definitions like:
  ```julia
  Na = ElementPsp(:Na, psp=load_psp("hgh/lda/na-q1")
  Si = ElementPsp(:Si, psp=load_psp("hgh/lda/si-q4")
  ```
  by just saying
  ```julia
  Na, Si = pseudize([:Na, :Si], functional="lda")
  ```
  What are your thoughts on that? Of course the name is horrible, but I'm not super creative right now.
- I do not like the code repetition introduced by having the polynomials. On the other hand using them for the PSP evaluation cost a factor of about 10 in runtime in those functions. Not sure we want to pay that. For big calculations computing the projectors takes quite a bit of time already (but I did not profile if the evaluation is actually time-consuming step yet).

antoine-levitt:
First point: I'm really not a fan of basically building a psp library interface. From my point of view the simplest the interface, the better; in particular I think it's good that the user specifies the path to the psp explicitly. My preferred interface would be the very basic and low-level `Na = ElementPsp(:Na, "hgh/lda/na-q1.hgh")`. That said it's maybe nice to avoid having to find a psp file... if you want on top of that we can have something like `get_pseudo_from_library(:Na, :HGH, :LDA, true)` for tests and examples and so on or something, but then 1) it should be cleanly separated from the rest of the code, and marked as a high-level convenience feature 2) it should be the only place that knows that we provide pseudos in the source distribution (ie move over the code from ElementPsp that special cases HGH)

Second point: really? Can you send a MWE?

mfherbst:
Re (1): `get_pseudo_from_library` you describe we now essentially already have ... it's called `list_psp` where you can search for almost all that stuff (modulo the semicore, but I can easily add that). There is another reason for the `pseudize`, which I did not write down. For converting from pymatgen and ase atoms datastructures the problem is that they do not always contain psp information at that level, but our "atoms" do, so one somehow has to put that back in when doing the conversion. The current solution was to pass to `load_atoms` a `pspmap` with the mappings, which is super inconvenient and actually makes the `load_atoms` code pretty cumbersome if you ask me. With `pseudize` you can "post-attach" pseudopotentials to an atoms structure, which allows to first load the atoms first from third-party not worrying about pseudos (i.e. attach them if third-party defines them, otherwise don't) and then put pseudos back in later in one line with `pseudize`. But I guess you can also argue it's the user's responsibility to do that manually with `list_psp`.

Re (2): I'll use `evalpoly`.

antoine-levitt:
I think the correct level to treat psp is at the same time as the choice of the XC functional: it's the same kind of level (between the atomic composition and the discretization parameters). So if the ase/pymatgen structures don't contain the xc, they shouldn't contain the psp either, and we should attach it in the same place. It does mean that ASE/pymatgen use a different structure than our `atoms`: they can just do `[:Si => [ones(3)]]` or something. I think it's fine for us to impose our prefered pseudos/XC functionals in the "high-level" ASE interface - we can always change the model later if needed.

mfherbst:
> I think it's fine for us to impose our prefered pseudos/XC functionals in the "high-level" ASE interface - we can always change the model later if needed.

Nah I think that will be too much hidden magic. But I agree that the model could just be changed later. I would really default to returning Coulombic atoms since it's the simplest for the parsing and then enforce people to attach pseudos (i.e. change the model).

PR #185, opened 2020-03-29
K-point index mapping

mfherbst:


src/interpolation.jl
```diff
@@ -0,0 +1,117 @@
...
+
+
+"""
+Interpolate some data from one k-Point to another. The interpolation is fast, but not
+necessarily exact or even normalised. Intended only to construct guesses for iterative
+solvers
+"""
+function interpolate_kpoint(data_in::AbstractVecOrMat, kpoint_in::Kpoint, kpoint_out::Kpoint)
+    if kpoint_in == kpoint_out
+        return data_in
```
antoine-levitt: maybe make a copy here

src/interpolation.jl
```diff
@@ -0,0 +1,117 @@
...
+    end
+    @assert length(G_vectors(kpoint_in)) == size(data_in, 1)
+    n_bands = size(data_in, 2)
+
+    data_newk = similar(data_in, length(G_vectors(kpt_new)), n_bands) .= 0
+    for (iold, inew) in enumerate(indexin(G_vectors(kpoint_in), G_vectors(kpt_new)))
+        inew !== nothing && (data_newk[inew, :] = data_in[iold, :])
```
antoine-levitt: I guess this should use the new kpoint mapping, no?

mfherbst: Absolutely!

PR #192, opened 2020-04-10
Fieldnames to propertynames

### LINKED ISSUE #191: RealFourierArray autocompletion isn't bypassed properly
It's supposed to only show the non-underscored properties (as per the fieldnames overload), but it doesn't.

mfherbst:
Fix #191

### DIFF
--- a/src/RealFourierArray.jl
+++ b/src/RealFourierArray.jl
@@ -51,8 +51,7 @@ check_real(A::RealFourierArray) = check_real(A.real)
 
 function Base.propertynames(array::RealFourierArray, private=false)
     ret = [:basis, :real, :fourier]
-    private && append!(ret, fieldnames(RealFourierArray))
-    ret
+    private ? append!(ret, fieldnames(RealFourierArray)) : ret
 end
 
 function Base.getproperty(A::RealFourierArray, x::Symbol)
--- a/src/RealFourierArray.jl
+++ b/src/RealFourierArray.jl
@@ -1,7 +1,7 @@
 # Lazy structure for an array whose real and fourier part can be accessed by A.real and A.fourier
 # For normalization conventions, see PlaneWaveBasis.jl
 
-import Base.getproperty, Base.setproperty!, Base.fieldnames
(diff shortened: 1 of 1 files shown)


src/RealFourierArray.jl
```diff
@@ -49,8 +49,11 @@ end
 check_real(A::RealFourierArray{Treal, T}) where {Treal, T <: Real} = nothing
 check_real(A::RealFourierArray) = check_real(A.real)
 
-# lie blatantly to the user, as recommended by the docs
-Base.fieldnames(RealFourierArray) = (:basis, :real, :fourier)
+function Base.propertynames(array::RealFourierArray, private=false)
+    ret = [:basis, :real, :fourier]
+    private && append!(ret, fieldnames(RealFourierArray))
```
antoine-levitt, line 54: boo mutation :-p
`private ? append(ret, fieldnames(RealFourierArray)) : ret` ?

antoine-levitt:
Huh, apparently I misread the docs. I just posted it here so I wouldn't forget but you're too damn quick :-p thanks!

src/RealFourierArray.jl
mfherbst: There is no `append` ... so if you don't want to repeat ret I don't see a way without mutation.

antoine-levitt: Really, that's weird. `[ret..., fieldnames(RealFourierArray)...]`

mfherbst: Really, what's the point? This function is not performance-critical and what you propose either allocates twice as well (if ret is an array) or is not type-stable (once returns array, once tuple if ret is a tuple). Also it's harder to read. I'm not convinced.

antoine-levitt: Ret should not be a tuple? But yeah I'm fine with what you wrote, it's just my part-time crusade against mutation.

antoine-levitt:
@gkemlin

src/RealFourierArray.jl
mfherbst: No. either tuple or vector. Both is allowed explicitly.

PR #196, opened 2020-04-14
Silicon SCF example

antoine-levitt:
This is for the paper with Gaspard and Eric. We can put it in another folder if we don't want it to be run by CI; alternatively I could put it in another repo but it's a neat self-contained example and illustrates callbacks so it's nice to have it here.

### DIFF
--- a/examples/silicon_scf_convergence.jl
+++ b/examples/silicon_scf_convergence.jl
@@ -3,7 +3,7 @@
 ## Some parameter values have been reduced for faster automatic testing
 
 using DFTK
-using PyPlot
+using Plots
 using Printf
 
 # Calculation parameters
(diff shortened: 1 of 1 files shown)


mfherbst:
If you don't want it to run by CI just add a comment line:
```
## tags: long
```
Also it uses PyPlot, which we gave up on ...

mfherbst:
But in this case there is no reason to use PyPlot, if you're saving the figures anyway... I'd stick to Plots.

mfherbst:
Oh and multiple windows in Plots you get by the kwarg `reuse=false`.

mfherbst:
Why no savefig now? And thanks for using Plots ;)

antoine-levitt:
Doesn't matter for the savefig...
Let's keep this PR open until we submit and have an arxiv number (shouldn't be long)

mfherbst:
Sure. I guess we want to have a nice page in our docs with related publications anyway ;).

mfherbst:
By the way: Do you also refer to this code inside the paper (with a footnote or so) ... could help to get people take a look at DFTK?

antoine-levitt:
Sure, it's as a ref with the URL. I did not put the zenodo there though

antoine-levitt:
The arxiv link is in, so that's good to go for me

PR #198, opened 2020-04-16
Remove some allocations in apply and LOBPCG

mfherbst:
Attempt to speed up the apply of the Hamiltonian for bigish calculations. Does two things:
- Removes the allocation of two `fft_size` grid arrays on each apply by preallocating them on Hamiltonian construction
- Cleanup of LOBPCG, removal of some `hcat`s only relevant for the general eigensolver case, usage of a a few `.=` instead of `=` to avoid allocation.

### DIFF
--- a/src/eigen/lobpcg_hyper_impl.jl
+++ b/src/eigen/lobpcg_hyper_impl.jl
@@ -120,7 +120,7 @@ end
 function ortho(X, Y, BY; tol=2eps(real(eltype(X))))
     for i=1:size(X,2)
         n = norm(@views X[:,i])
-        X[:,i] /= n
+        X[:,i] ./= n
     end
 
     niter = 1
@@ -227,16 +227,14 @@ function LOBPCG(A, X, B=I, precon=((Y, X, R)->R), tol=1e-10, maxiter=100; ortho_
 
             # Form Rayleigh-Ritz subspace
             if niter > 2
-                Y = hcat(X,R,P)
-                AY = hcat(AX,AR,AP)
-                BY = hcat(BX,BR,BP)
+                Y = hcat(X, R, P)
+                AY = hcat(AX, AR, AP)
+                BY = (B == I) ? Y : hcat(BX, BR, BP)
             else
-                Y = hcat(X,R)
-                AY = hcat(AX,AR)
-                BY = hcat(BX,BR)
+                Y = hcat(X, R)
+                AY = hcat(AX, AR)
+                BY = (B == I) ? Y : hcat(BX, BR)
             end
-
-
             cX, λs = RR(Y, AY, BY, M-nlocked)
 
             # Update X. By contrast to some other implementations, we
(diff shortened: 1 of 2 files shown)


mfherbst:
I should probably mention that this is not really doing an aweful lot. The biggest issue at the moment is that the subspace routines (mainly the orthogonalisation via Cholesky) is not properly parallelised. Even on big systems hardly more than a few CPU cores are actually employed.

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -227,64 +227,71 @@ function LOBPCG(A, X, B=I, precon=((Y, X, R)->R), tol=1e-10, maxiter=100; ortho_
 
             # Form Rayleigh-Ritz subspace
             if niter > 2
-                Y = hcat(X,R,P)
-                AY = hcat(AX,AR,AP)
-                BY = hcat(BX,BR,BP)
+                Y = hcat(X, R, P)
+                AY = hcat(AX, AR, AP)
+                if B != I
```
antoine-levitt, line 232: I would do a ternary op here and below

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -227,64 +227,71 @@ function LOBPCG(A, X, B=I, precon=((Y, X, R)->R), tol=1e-10, maxiter=100; ortho_
...
-                new_BX = BY*cX
+                new_BX = BY * cX
             else
                 new_BX = new_X
             end
         end
 
         ### Compute new residuals
-        R = new_AX - new_BX*Diagonal(λs)
+        new_R = new_AX - new_BX * Diagonal(λs)
```
antoine-levitt, line 263: why the rename?

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -355,12 +362,16 @@ function LOBPCG(A, X, B=I, precon=((Y, X, R)->R), tol=1e-10, maxiter=100; ortho_
         # Orthogonalize R wrt all X, newly active P
         if niter > 1
             Z = hcat(full_X, P)
-            BZ = hcat(full_BX, BP)
+            if B != I
```
antoine-levitt, line 365: ternary here also

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -370,11 +381,8 @@ function LOBPCG(A, X, B=I, precon=((Y, X, R)->R), tol=1e-10, maxiter=100; ortho_
             mul!(BR, B, R)
             O = Hermitian(R'*BR)
             U = cholesky(O).U
-            R = R/U
-            BR = BR/U
-            # condU = 1/LAPACK.trcon!('I', 'U', 'N', Array(U))
-        else
-            BR = R
+            R .= R / U
```
antoine-levitt, line 384: `rdiv!`

src/eigen/lobpcg_hyper_impl.jl
mfherbst: no not mix it up with the old `R` (was needed for the lazy concatenation that I tried) also follows the analogy with P and X.

src/terms/Hamiltonian.jl
```diff
@@ -11,7 +11,10 @@ struct HamiltonianBlock
     operators::Vector  # the original list of RealFourierOperator
                        # (as many as there are terms), kept for easier exploration
     optimized_operators::Vector  # the optimized list of RealFourierOperator, to be used for applying
-    HamiltonianBlock(basis, kpt, operators) = new(basis, kpt, operators, optimize_operators_(operators))
+    preallocated  # Pre-allocated scratch arrays for fast application
```
antoine-levitt, line 14: call it `scratch` instead. Possibly `_scratch` to emphasize it's very internal?

PR #199, opened 2020-04-17
Avoid using hcat in LOBPCG

mfherbst:


src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -118,22 +117,22 @@ end
...
         n = norm(@views X[:,i])
         X[:,i] ./= n
     end
 
     niter = 1
     ninners = zeros(Int,0)
     while true
-        BYX = BY'X
-        X .-= Y*BYX
+        BYX = blockdot(BY, X)
```
antoine-levitt, line 128: Here and for every instance of blockdot/blockmul, can you add the actual formula (here `BY'X`) as a comment? That makes it easier to see what mathematical operation is done

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -118,22 +117,22 @@ end
...
+        X .-= blockmul(Y, BYX)
         # If the orthogonalization has produced results below 2eps, we drop them
         # This is to be able to orthogonalize eg [1;0] against [e^iθ;0],
         # as can happen in extreme cases in the ortho(cP, cX)
         dropped = drop!(X)
-        if dropped != []
-            X[:, dropped] = X[:, dropped] - Y*BY'*X[:,dropped]
+        @views if dropped != []
+            X[:, dropped] = X[:, dropped] - blockmul(Y, blockdot(BY, X[:, dropped]))
```
antoine-levitt, line 135: dots here

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -165,6 +164,46 @@ function ortho(X, Y, BY; tol=2eps(real(eltype(X))))
 end
 
 
+# Given A as a list of blocks [A1, A2, A3] forms the matrix-matrix product
+# A' * B avoiding a concatenation of the blocks to a dense array
```
antoine-levitt, line 168: this is A * B not A' * B right? Maybe mul and adjmul then?

antoine-levitt: Move these two functions at the beginning of the file, and explain why they are needed (that we store the subspaces Y =[X, AX, P] as tuples instead of explicit structs)

antoine-levitt: Also maybe [A1, A2, ...]? To show the size is variable

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -165,6 +164,46 @@ function ortho(X, Y, BY; tol=2eps(real(eltype(X))))
...
+    res
+end
+blockmul(A, B::Tuple) = error("not implemented")
+blockmul(A, B) = A * B
+
+
+# Given A and B as two lists of blocks [A1, A2, A3], [B1, B2, B3] form the matrix
+# A' B optionally assuming that the output is a Hermitian matrix and will be used
+# with the Hermitian view, such that only the upper triangle needs to be computed.
+@views function blockdot(blocksA::Tuple, blocksB::Tuple; assume_hermitian=false)
```
antoine-levitt, line 185: Or perhaps even `overlap`?

src/eigen/lobpcg_hyper_impl.jl
mfherbst: Good point. I'd keep the block prefix, however, to distinguish them better from "mul!" but I would be ok with bmul for example.

mfherbst: yes it does A * B

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -227,32 +266,31 @@ function LOBPCG(A, X, B=I, precon=((Y, X, R)->R), tol=1e-10, maxiter=100; ortho_
 
             # Form Rayleigh-Ritz subspace
             if niter > 2
-                Y = hcat(X, R, P)
-                AY = hcat(AX, AR, AP)
-                BY = (B == I) ? Y : hcat(BX, BR, BP)
+                Y = (X, R, P)
+                AY = (AX, AR, AP)
+                BY = (B == I) ? Y : (BX, BR, BP)
```
antoine-levitt, line 271: No need to special case anymore, right?

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -349,10 +387,10 @@ function LOBPCG(A, X, B=I, precon=((Y, X, R)->R), tol=1e-10, maxiter=100; ortho_
 
         # Orthogonalize R wrt all X, newly active P
         if niter > 1
-            Z = hcat(full_X, P)
-            BZ = (B == I) ? Z : hcat(full_BX, BP)
+            Z  = (full_X, P)
+            BZ = (B == I) ? Z : (full_BX, BP)
```
antoine-levitt, line 391: same here

src/eigen/lobpcg_hyper_impl.jl
antoine-levitt: bmul and boverlap?

src/eigen/lobpcg_hyper_impl.jl
mfherbst: Sorry I don't get this comment.

src/eigen/lobpcg_hyper_impl.jl
mfherbst: sounds good, I'll go for that.

src/eigen/lobpcg_hyper_impl.jl
antoine-levitt: .= and .-

mfherbst: Ah ;) sure

PR #204, opened 2020-04-22
LOBPCG: save a matvec

antoine-levitt:


src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -210,10 +210,10 @@ function ortho(X, Y, BY; tol=2eps(real(eltype(X))))
 end
 
 
-function final_residnorms(A, X, resids, niter)
-    AX = A * X
+function final_residnorms(X, AX, resids, niter)
     λ = real(diag(X' * AX))
     residuals = AX .- X*Diagonal(λ)
+    println("Converged after $niter steps")
```
mfherbst, line 216: Left-over debug print.

antoine-levitt: Yep, sorry

Issue #33, opened 2019-10-28
Figure out an API for eigensolvers/SCF/solvers


antoine-levitt:
Ref https://github.com/mfherbst/DFTK.jl/pull/30
Something like
```
struct Eigensolver
    tol
    maxiter
    precon # Promises to implement precon(ham) -> P
    solver # Promises to implement solver(A,X0,P,tol=tol,maxiter=maxiter)
end

default_eigensolver(tol=1e-6, maxiter=100) = Eigensolver(tol, maxiter, lobpcg_hyper)
```

antoine-levitt:
Something like https://github.com/mauro3/Parameters.jl could help

antoine-levitt:
We really need to do this at one point. Something like
```
solver = Solver(basis, DensityMixing(; mixing, eigensolver); occupation_scheme, rho_init, psi_init)
scfres = solve(solver)
```
? That seems to be the pattern used in the SciML solvers.

mfherbst:
closing discussion in favour of #790.

PR #205, opened 2020-04-23
Automatic adjustment of eigensolver tolerance

mfherbst:
Eigensolver tolerance is adjusted based on the change between ρnext and ρin.

I think what we do here is ok, but one could gain flexibility by making the selection process a function `ρdiffnorm, n_iter -> diagtol`, but maybe that's overengineering for now?


antoine-levitt:
I think we should have a semi-consistent way of doing this, in the same way that we do SCF convergence. That's https://github.com/JuliaMolSim/DFTK.jl/issues/33. Can wait though.

mfherbst:
Yes I agree. This API is becoming messy ... but let's wait until we do it properly.

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -210,11 +210,13 @@ function ortho(X, Y, BY; tol=2eps(real(eltype(X))))
 end
 
 
-function final_residnorms(A, X, resids, niter)
+function final_residnorms(A, X, resids, niter, n_matvec)
```
antoine-levitt, line 213: rename to something like final_retval?

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -210,11 +210,13 @@ function ortho(X, Y, BY; tol=2eps(real(eltype(X))))
...
     AX = A * X
+    n_matvec += size(X, 2)
```
antoine-levitt, line 215: don't forget to remove this when merging against master

mfherbst: Sure ... for now I'm running some bigger tests on this branch so I'm not merging with master to stay comparable in the numbers, but I have it in mind.

src/scf/self_consistent_field.jl
```diff
@@ -115,6 +121,13 @@ function self_consistent_field(basis::PlaneWaveBasis;
             energies, ham = energy_hamiltonian(basis, ψ, occupation;
                                                ρ=ρin, eigenvalues=eigenvalues, εF=εF)
         end
+
+        # Adjust solver diagtol
+        diagtol = norm_ρnext_ρin * diagtol_ratio_ρchange
+        diagtol = min(diagtol_range[2], diagtol)
+        diagtol = max(diagtol_range[1], diagtol)
+        @assert isfinite(diagtol)
```
mfherbst, line 129: Pretty hard to digest. Perhaps better use a callback.

mfherbst: Perhaps a good name: determine_diagtol.

examples/silicon_scf_convergence.jl
```diff
@@ -29,7 +29,8 @@ function my_callback(info)
     push!(resids, err)
 end
 my_isconverged=info -> norm(info.ρout.fourier - info.ρin.fourier) < tol
-opts = (callback=my_callback, is_converged=my_isconverged, max_iter=maxiter, tol=tol, diagtol=diagtol)
+opts = (callback=my_callback, is_converged=my_isconverged, maxiter=maxiter, tol=tol,
+        determine_diagtol=DFTK.scf_determine_diagtol(0.01, diagtol, diagtol))
```
antoine-levitt, line 33: not very clear, use kwargs constructor?

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -210,10 +210,11 @@ function ortho(X, Y, BY; tol=2eps(real(eltype(X))))
 end
 
 
-function final_residnorms(X, AX, resid_history, niter)
+function final_retval(X, AX, resid_history, niter, n_matvec)
     λ = real(diag(X' * AX))
     residuals = AX .- X*Diagonal(λ)
-    λ, X, [norm(residuals[:, i]) for i in 1:size(residuals, 2)], resid_history[:, 1:niter+1]
+    (λ, X, [norm(residuals[:, i]) for i in 1:size(residuals, 2)],
+     resid_history[:, 1:niter+1], n_matvec)
```
antoine-levitt, line 217: should be named tuple?

src/scf/self_consistent_field.jl
```diff
@@ -62,6 +66,27 @@ function scf_convergence_density_difference(tolerance)
     info -> norm(info.ρout.fourier - info.ρin.fourier) < tolerance
 end
 
+"""
+Determine the tolerance used for the next diagonalisation. This function takes
+``|ρnext - ρin|`` and multiplies it with `ratio_ρdiff` to get the next `diagtol`,
+ensuring additionally that the returned value is between `diagtol_min` and `diagtol_max`.
+"""
+function scf_determine_diagtol(ratio_ρdiff=0.2, diagtol_min=nothing, diagtol_max=0.1)
+    function determine_diagtol(neval, norm_ρnext_ρin)
```
antoine-levitt, line 75: shouldn't it just take `info` like the callbacks?

src/scf/self_consistent_field.jl
```diff
@@ -62,6 +66,27 @@ function scf_convergence_density_difference(tolerance)
     info -> norm(info.ρout.fourier - info.ρin.fourier) < tolerance
 end
 
+"""
+Determine the tolerance used for the next diagonalisation. This function takes
+``|ρnext - ρin|`` and multiplies it with `ratio_ρdiff` to get the next `diagtol`,
+ensuring additionally that the returned value is between `diagtol_min` and `diagtol_max`.
+"""
+function scf_determine_diagtol(ratio_ρdiff=0.2, diagtol_min=nothing, diagtol_max=0.1)
```
antoine-levitt, line 74: call it `default_determine_diagtol`?

src/scf/self_consistent_field.jl
```diff
@@ -98,6 +123,7 @@ function self_consistent_field(basis::PlaneWaveBasis;
     neval = 0
     energies = nothing
     ham = nothing
+    norm_ρnext_ρin = Inf  # The most recent norm(ρnext - ρin)
```
antoine-levitt, line 126: can be removed if the dtermine function gets info

src/scf/self_consistent_field.jl
mfherbst: yes. Turned out to be simpler than I thought to do that.

examples/silicon_scf_convergence.jl
mfherbst: ok.

src/scf/self_consistent_field.jl
mfherbst: Nah ... I like `scf_determine_diagtol` more. But this will probably be renamed once we do the SCF params properly.

antoine-levitt: The `default` has the merit of being symmetrical with the SCF convergence criteria. What i don't like with what we have now is that it's unclear that these are functions returning function. We should perphaps capitalize them / make them structs

test/forces.jl
```diff
@@ -37,7 +39,12 @@ end
         basis = PlaneWaveBasis(model, Ecut, kgrid=[2, 1, 2])
 
         n_bands = 10
-        scfres = self_consistent_field(basis, n_bands=n_bands, tol=1e-12, diagtol=1e-12)
+        is_converged = scf_convergence_density_difference(5e-11)
+        determine_diagtol = scf_determine_diagtol(ratio_ρdiff=0.01)
```
antoine-levitt, line 43: is this really needed? I'd rather avoid tweaking the diagtol in the tests (which is a nice thing with this PR). The is_converged should be sufficient. Feel free to take bigger tolerances for the finite difference check if needed

test/hydrogen_all_electron.jl
```diff
@@ -1,19 +1,21 @@
 using Test
 using DFTK
+import DFTK: scf_convergence_density_difference
```
antoine-levitt, line 3: just nitpicking, but in general I'm not a big fan of this import business, because it tends to get lost in refactorings. I'd rather do DFTK. when needed

src/scf/self_consistent_field.jl
mfherbst: Structs I don't like for such simple stuff. Capitalisation is ok.

antoine-levitt: It's more of an outside API thing. The convention in packages seems to crystallize around options/methods/controllers being structs (eg optim, interpolations, etc.), which are capitalized

test/forces.jl
mfherbst: Playing with the FD parameter helps. I used that to get it a bit more stable.

PR #207, opened 2020-04-23
Flexible temperature in chi0 applies.

mfherbst:


src/postprocess/chi0.jl
```diff
@@ -134,6 +134,7 @@ function apply_χ0(ham, δV, ψ, occupation, εF, eigenvalues; droptol=0,
 
             # Sternheimer contributions. TODO add preconditioning here
             !(sternheimer_contribution) && continue
+            fnk = filled_occ * Smearing.occupation(model.smearing, (εmk-εF) / temperature)
```
antoine-levitt, line 137: Careful line below

antoine-levitt: Just remove occupation arg?

antoine-levitt: Here and in the other chi0 probably

mfherbst: Not sure what you mean by the "other" chi0.

antoine-levitt: I meant compute_chi0 but I see that's irrelevant (although we could have a custom temperature there as well?)

src/postprocess/chi0.jl
```diff
@@ -153,9 +153,9 @@ function apply_χ0(ham, δV, ψ, occupation, εF, eigenvalues; droptol=0,
     end
 
     # Add variation wrt εF
-    if model.temperature > 0
-        ldos = LDOS(εF, basis, eigenvalues, ψ)
-        dos = DOS(εF, basis, eigenvalues)
+    if temperature > 0
+        ldos = LDOS(εF, basis, eigenvalues, ψ, T=temperature)
+        dos = DOS(εF, basis, eigenvalues, T=temperature)
```
antoine-levitt, line 158: homogenize the `T` argument as temperature over there as well?

mfherbst: good idea. That's been bothering me for quite some time.

Issue #208, opened 2020-04-24
Non-zero forces in the silicon example


gkemlin:
When running the example file silicon.jl with `kgrid = [4,4,4]`, we get non-zero forces when computing `forces(scfres)` (the output array is `[[-0.36944959523696796, -0.23075365997138772, -2.029230333878863e-12], [0.36944959523669624, 0.23075365997089106, 1.6977489637483588e-12]]`), which does not happen with `kgrid = [1,1,1]`. 

Using `enable_bzmesh_symmetry=false` in the PlaneWaveBasis constructor enables us to avoid this issue.

antoine-levitt:
It's most likely the nonlocal part. We need to add a loop over the symmetry operations, see the latex in the docs and comments in bzmesh.jl.

antoine-levitt:
And it's not caught in the tests because we explicitly move away from equilibrium positions to avoid zero forces

antoine-levitt:
let's just declare symmetries don't exist

mfherbst:
Resolved now in #209.

PR #209, opened 2020-04-24
Fix forces

antoine-levitt:
Implementation looks OK, but forces are still nonzero so something's wrong.

### DIFF
--- a/src/PlaneWaveBasis.jl
+++ b/src/PlaneWaveBasis.jl
@@ -6,6 +6,28 @@ include("fft.jl")
 # products of orbitals. This also defines the real-space grid
 # (as the dual of the cubic basis set).
 
+# The full (reducible) Brillouin zone is implicitly represented by
+# a set of (irreducible) kpoints (see explanation in docs/). Each
+# irreducible kpoint k comes with a list of symmetry operations
+# (S,τ) (containing at least the trivial operation (I,0)), where S
+# is a rotation matrix (/!\ not unitary in reduced coordinates)
+# and τ a translation vector. The kpoint is then used to represent
+# implicitly the information at all the kpoints Sk. The
+# relationship between the Hamiltonians is
+# H_{Sk} = U H_k U*, with
+# (Uu)(x) = u(S^-1(x-τ))
+# or in Fourier space
+# (Uu)(G) = e^{-i G τ} u(S^-1 G)
+# In particular, we can choose the eigenvectors at Sk as u_{Sk} = U u_k
+
+# We represent then the BZ as a set of irreducible points `kpoints`, a
+# list of symmetry operations `ksymops` allowing the reconstruction of
+# the full (reducible) BZ, and a set of weights `kweights` (summing to
+# 1). The value of an observable (eg energy) per unit cell is given as
+# the value of that observable at each irreducible kpoint, weighted by
+# kweight
+
+
 """
 Discretization information for kpoint-dependent quantities such as orbitals.
 More generally, a kpoint is a block of the Hamiltonian;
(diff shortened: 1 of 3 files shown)


antoine-levitt:
Missing a test but functional. Fixes https://github.com/JuliaMolSim/DFTK.jl/issues/208, cc @gkemlin

src/PlaneWaveBasis.jl
```diff
@@ -43,12 +43,29 @@ Normalization conventions:
...
+    # and τ a translation vector. The kpoint is then used to represent
+    # implicitly the information at all the kpoints Sk. The
+    # relationship between the Hamiltonians is
+    # H_{Sk} = U H_k U*, with
+    # (Uu)(x) = u(S^-1(x-τ))
+    # or in Fourier space
+    # (Uu)(G) = e^{-i G τ} u(S^-1 G)
+    # In particular, we can choose the eigenvectors at Sk as u_{Sk} = U u_k
+
+    # irreducible kpoints
```
mfherbst, line 63: I think we said, that we wanted to keep struct definitions short and rather describe stuff before the struct in comments. I think we should stick to that.

src/PlaneWaveBasis.jl
```diff
@@ -43,12 +43,29 @@ Normalization conventions:
...
     kpoints::Vector{Kpoint{T}}
-    # kweights/ksymops contain the information needed to reconstruct the full (reducible) BZ
-    # Brillouin zone integration weights
+    # Brillouin zone integration weights,
+    # kweights[ik] = length(ksymops[ik]) / sum(length(ksymops[ik]) for ik=1:Nk)
+    # The value of an observable (eg energy) per unit cell is given as
+    # the value of that observable at each irreducible kpoint, weighted by kweight
```
mfherbst, line 68: Same here. A bit long for an in-struct comment.

test/forces.jl
```diff
@@ -25,7 +30,7 @@ include("testcases.jl")
 
     diff_findiff = -(E2 - E1) / ε
     diff_forces = dot(F1[1][1], disp)
-    @test abs(diff_findiff - diff_forces) < 2e-6
+    @test abs(diff_findiff - diff_forces) < 1e-4
```
mfherbst, line 33: 1e-4 is really big. Do you think this than still tests anything?

src/PlaneWaveBasis.jl
```diff
@@ -302,3 +318,52 @@ end
...
+    for i = 1:size(ψ, 2)
+        @views ψsym[:, i] .= transfer_ψ(basis, kpoint_src, kpoint_dst, ψ[:, i], S, τ)
+    end
+    ψsym
+end
+
+# Convert a basis into one that uses or doesn't use BZ symmetrization
+# Mainly useful for debug purposes and in cases we don't want to
+# bother with symmetry
+function PlaneWaveBasis(basis::PlaneWaveBasis; enable_bzmesh_symmetry)
```
mfherbst, line 355: I'm not sure I would make this a constructor. For me it feels more natural to just define a function `disable_kpoint_symmetry` or something and do it in there. Also gets rid of this strange kwarg.

test/forces.jl
antoine-levitt: really? 1e-4 is not that big... Usually I just take twice what I compute or something, but then very quickly we have to adjust when we change things (like here when I added a kpoint). Having larger tolerances means we don't have to worry about it so much

src/PlaneWaveBasis.jl
```diff
@@ -302,3 +318,52 @@ end
...
+
+#
+# Functions to handle BZ symmetry
+#
+
+"""
+Converts an eigenfunction at point `kpoint_src` to one at
+`kpoint_dst`, the two being related by symmetry operation S.
+"""
+function transfer_ψ(basis::PlaneWaveBasis{T}, kpoint_src::Kpoint, kpoint_dst::Kpoint, ψ::AbstractVector, S, τ) where {T}
```
mfherbst, line 330: I'm not a fan of the name. It's a bit too generic. From the name it could be related to the things in `interpolation.jl` (which it is in some sense, but not in the sense that it is an interpolation, but exact). Also I think the order of the arguments should be more similar to the way we do in in `interpolation.jl`. I would suggest `transfer_kpoint` or `transfer_to_kpoint` with the argument order `(basis, ψ,  kpoint_in, kpoint_out, S, τ)`

src/PlaneWaveBasis.jl
antoine-levitt: I actually quite like it like a constructor, it's what it is, no? And the kwarg is the same as the other constructors. The less extra functions we have, the less namespace pollution, and the less I have to remember stuff.

src/PlaneWaveBasis.jl
antoine-levitt: completely agree, I didn't like the name either. Will go with transfer_to_kpoint

src/PlaneWaveBasis.jl
mfherbst: Whether you not remember a name or not remember which of the 15 interfaces it was, what's the difference? It's pretty specific and low-level. It should not be honored to live on the same level as the high-level constructor with the MP-grid in my opinion.

test/forces.jl
mfherbst: I don't like this strategy. So far it surely was annoying to tweak tolerances every now and then, but each time I learned a lot about details of our code or of a particular feature (like forces computation). Also you don't miss unforseen influences of e.g. SCF convergence parameters so easily. I really would for keeping tolerances as tight as possible at the expense to tweak them when needed.

src/PlaneWaveBasis.jl
antoine-levitt: we can move it to interpolations also

test/forces.jl
antoine-levitt: I think it really depends. We should definitely have a few core tests with tight tolerance criteria to catch these sort of things. But here we just want to see if the formulas for implementing the forces are correct, so we shouldn't really focus on a very tight convergence criterion

src/PlaneWaveBasis.jl
antoine-levitt: From my point of view the more constructors the better. There are some high level, some not. What distinguishes them is whether we use them in examples and whether we add a docstring to them (ie I wouldn't for this one). Having this kind of constructors is very convenient. Eg what if you want to make a pwbasis that has the same parameters as an existing one but with a larger fft_size? What if you want to do it as the same time as removing the bz symmetry? With this kind of constructors you can do it

mfherbst: Ok, I see your point, but then we might need to clean it up a little from time to time. I have the feeling for PlaneWaveBasis it's getting out of hand.

antoine-levitt: I agree pwbasis is a bit of a mess yeah...

PR #214, opened 2020-04-29
GaAs example

mfherbst:
Adds as example constructing a GaAs surface using ase.

### DIFF
--- a/examples/GaAs.jl
+++ b/examples/GaAs.jl
@@ -1,0 +1,42 @@
+using DFTK
+using PyCall
+
+#
+# Example using ASE to build a (1, 1, 0) GaAs surface separated by vacuum.
+# See https://wiki.fysik.dtu.dk/ase/install.html for details
+## tags: long
+#
+
+miller = (1, 1, 0)              # Surface Miller indices
+n_GaAs = 20                     # Number of GaAs layers
+n_vacuum = 20                   # Number of vacuum layers
+Ecut = 15                       # Hartree
+kgrid = (4, 4, 1)               # Monkhorst-Pack mesh
+# kgrid = (13, 13, 1)           # Better Monkhorst-Pack mesh
+
+# Use ASE to build the structure
+ase_build = pyimport("ase.build")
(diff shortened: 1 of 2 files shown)


examples/GaAs.jl
```diff
@@ -0,0 +1,42 @@
...
+atoms = load_atoms(surface)
+atoms = [ElementPsp(el.symbol, psp=load_psp(el.symbol, functional="pbe")) => position
+         for (el, position) in atoms]
+
+lattice = load_lattice(surface)
+model = model_DFT(lattice, atoms, [:gga_x_pbe, :gga_c_pbe])
+basis = PlaneWaveBasis(model, Ecut; kgrid=kgrid)
+
+scfres = self_consistent_field(basis, tol=1e-8,
+                               n_bands=ceil(Int, model.n_electrons / 2),
```
antoine-levitt, line 37: I usually do `div(N,2)`

antoine-levitt: also maybe not necessary?

antoine-levitt:
Great! Rename to GaAs_surface maybe?

antoine-levitt:
Also mention the paper on TF mixing for the source

examples/GaAs.jl
mfherbst: no this is leftover from my playing. I'll remove it.

antoine-levitt:
Can you also add that this needs `ase`? Or maybe we put it somewhere in the README that ASE is useful?

antoine-levitt:
Reading the code also it's not immediately clear that `load_atoms` refers to ASE. Since this is python it's not clear we emulate dispatch here, so probably it's better to have explicit `load_atoms`? This is just nitpicking though.

mfherbst:
No I disagree. This actually saves code because abipy structures are also pymatgen structures so dispatch works for them, too. The same is true for various kinds of Atoms you can make out of ASE atoms etc. Also `load_atoms` is just "the" function to convert to DFTK atoms no matter where you are coming from, so it is not an unexpected interface at all.

antoine-levitt:
Alright

PR #216, opened 2020-04-30
Use default mixing parameters from KresseFurtmüller

mfherbst:
Any opinion on that @antoine-levitt. Seems to be suggested my many sources as the default.


antoine-levitt:
Looks very reasonable to me... 0.8 and 1 aren't that different.

PR #218, opened 2020-04-30
Misc fixes and adds

antoine-levitt:
The matrices are to simplify the manipulation of operators (eg pass chi0 back to fourier space).


antoine-levitt:
The matrices version of the kpt-specific transforms seem less useful so I didn't bother

PR #221, opened 2020-05-05
Iterative computation of the dielectric matrix

antoine-levitt:


antoine-levitt:
TODO:
[ ] Figure out the symmetrization business
[ ] Tweak tolerances in chi0
[ ] Implement analytical xc derivatives
[ ] Have a nice API for chi0 and hxc kernels. Something like a linear-operator type struct?

test/chi0.jl
```diff
@@ -7,46 +7,71 @@ include("testcases.jl")
...
+                dV_sym = DFTK.symmetrize(from_real(basis, dV)).real
+                if symmetry
+                    dV = dV_sym
+                else
+                    @test dV_sym ≈ dV
+                end
+
+                term_builder = basis -> DFTK.TermExternal(basis, ε .* dV)
+                model = model_LDA(testcase.lattice, [spec => testcase.positions], temperature=temperature, extra_terms=[term_builder], symmetry=(symmetry ? :force : :off))
+                basis = PlaneWaveBasis(model, Ecut; kgrid=kgrid, fft_size=fft_size, use_symmetry=use_symmetry)
```
mfherbst, line 39: And again :smile:

test/chi0.jl
```diff
@@ -7,46 +7,71 @@ include("testcases.jl")
...
+                occ, εF = find_occupation(basis, Es)
+
+                # Test apply_χ0
+                diff_applied_χ0 = apply_χ0(ham0, dV, Vs, εF, Es; droptol=0)
+                @test norm(diff_findiff - diff_applied_χ0) < testtol
+
+                # Test compute_χ0 against finite differences
+                if !symmetry
+                    χ0 = compute_χ0(ham0)
+                    diff_computed_χ0 = real(reshape(χ0*vec(dV), basis.fft_size))
```
mfherbst, line 57: spaces around `*`.

examples/dielectric.jl
```diff
@@ -0,0 +1,54 @@
+# Compute a few eigenvalues of the dielectric matrix (q=0,ω=0) iteratively
```
mfherbst, line 1: Do you want to convert this into a "proper" example for the website?

examples/dielectric.jl
```diff
@@ -0,0 +1,54 @@
...
+Si = ElementPsp(:Si, psp=load_psp("hgh/lda/Si-q4"))
+atoms = [Si => [ones(3)/8, -ones(3)/8]]
+
+# change the symmetry to compute the dielectric operator with and without symmetries
+model = model_LDA(lattice, atoms, symmetry=:off)
+basis = PlaneWaveBasis(model, Ecut; kgrid=kgrid)
+scfres = self_consistent_field(basis, tol=1e-14)
+
+# Apply ε = 1 - K χ0
+function epsfun(dv)
```
mfherbst, line 23: Damn you have fun with eps.

examples/dielectric.jl
```diff
@@ -0,0 +1,54 @@
...
+
+# A straightfoward Arnoldi eigensolver that diagonalizes the matrix at each step
+# This is more efficient than Arpack when `f` is very expensive
+function arnoldi(f, x0, howmany; tol=1e-4, maxiter=30)
+    for (V, B, r, nr, b) in ArnoldiIterator(f, x0)
+        # A * V = V * B + r * b'.
+        V = hcat(V...)
+        AV = V*B + r*b'
+
+        ew, ev = eigen(B, sortby=real)
```
mfherbst, line 38: Can't you additionally exploit that `B` is upper/lower Hessenberg `LinearAlgebra.Hessenberg`?

examples/dielectric.jl
```diff
@@ -0,0 +1,54 @@
...
+        Vr = V*ev
+        AVr = AV*ev
+        R = AVr - Vr * Diagonal(ew)
```
mfherbst, line 41: Can you indicate by a comment that this is the residual (or call it residual) ... just to make it clear.

examples/dielectric.jl
```diff
@@ -0,0 +1,54 @@
...
+
+        N = size(V, 2)
+        inds = 1:min(N, howmany)
+        inds = [inds..., (inds .+ N .- min(N, howmany))...]
```
mfherbst, line 45: Puh ... that's a mouthful. Can you add a comment?

src/postprocess/chi0.jl
```diff
@@ -3,6 +3,12 @@ using IterativeSolvers
 using ForwardDiff
 using ProgressMeter
 
+struct FunctionPreconditioner
+    f_ldiv!
+end
+LinearAlgebra.ldiv!(y, P::FunctionPreconditioner, x) = P.f_ldiv!(y, x)
+LinearAlgebra.ldiv!(P::FunctionPreconditioner, x) = P.f_ldiv!(copy(x), x)
```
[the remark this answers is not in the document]
antoine-levitt: Good catch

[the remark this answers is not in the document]
antoine-levitt: Yeah but which? This kind of wrapper things is better put close to the point of use

src/postprocess/chi0.jl
```diff
@@ -129,10 +153,10 @@ function apply_χ0(ham, δV, ψ, εF, eigenvalues; droptol=0,
                 # ∑_{n,m != n} (fn-fm)/(εn-εm) ρnm <ρmn|δV>
                 ρnm = conj(ψnk_real) .* ψmk_real
                 weight = dVol*dot(ρnm, δV)
-                δρ .+= real(basis.kweights[ik] .* ratio .* weight .* ρnm)
+                δρk .+= real(ratio .* weight .* ρnm)
             end
 
-            # Sternheimer contributions. TODO add preconditioning here
+            # Sternheimer contributions.
             !(sternheimer_contribution) && continue
```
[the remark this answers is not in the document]
antoine-levitt: yeah very good call

src/scf/direct_minimization.jl
```diff
@@ -141,5 +141,5 @@ function direct_minimization(basis::PlaneWaveBasis{T}, ψ0;
     # We rely on the fact that the last point where fg! was called is the minimizer to
     # avoid recomputing at ψ
     (H=H, energies=energies, converged=true,
-     ρ=ρ, ψ=ψ, eigenvalues=eigenvalues, occupation=occupation, εF=εF, optim_res=res)
+     ρ=ρ, ψ=ψ, eigenvalues=eigenvalues, occupation=occupation, εF=εF, optim_res=res, ham=H)
```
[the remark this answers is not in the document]
antoine-levitt: Yeah somehow I missed that

test/chi0.jl
antoine-levitt: yeah for tests I tend to care even less than usual :-p

examples/dielectric.jl
antoine-levitt: I don't think so; not unless we gain a bit more experience with it at least, and maybe generalize it to q, omega

examples/dielectric.jl
antoine-levitt: probably krylovkit returns it by itself and julia then dispatches appropriately. I don't particularly care since this is going to be 30x30 at the most

examples/dielectric.jl
antoine-levitt: if I do that I should probably comment every line... this is kind of a hack right now, it should be turned into a proper eigensolver eventually, but I'm too lazy to do that, so this is an exploratory tool for now

mfherbst: ok.

src/eigen/preconditioners.jl
```diff
@@ -67,6 +68,7 @@ ldiv!(P::PreconditionerTPA, R) = ldiv!(R, P, R)
     Y
 end
 mul!(P::PreconditionerTPA, R) = mul!(R, P, R)
```
mfherbst, line 70: There is no single-argument mul! in LinearAlgebra. I don't think we should define that ... (note: It's also not used in IterativeSolvers.jl)

src/postprocess/chi0.jl
```diff
@@ -3,12 +3,12 @@ using IterativeSolvers
 using ForwardDiff
 using ProgressMeter
 
+# make ldiv! act as a given function
 struct FunctionPreconditioner
-    f_ldiv!
+    f! # f!(y, x) applies f to x and puts it into y
```
mfherbst, line 8: Space before `#`. Also I don't like `f!`. It's too easy to overlook alltogether. how about `precondition!` or `apply!` or something like that?

src/postprocess/chi0.jl
```diff
@@ -3,12 +3,12 @@ using IterativeSolvers
...
 end
-LinearAlgebra.ldiv!(y, P::FunctionPreconditioner, x) = P.f_ldiv!(y, x)
-LinearAlgebra.ldiv!(P::FunctionPreconditioner, x) = P.f_ldiv!(copy(x), x)
-# (Base.:\)(P::FunctionPreconditioner, x) = P.f_ldiv!(copy(x), x)
+LinearAlgebra.ldiv!(y, P::FunctionPreconditioner, x) = P.f!(y, x)
+LinearAlgebra.ldiv!(P::FunctionPreconditioner, x) = (y=similar(x); P.f!(y, x); x .= y)
```
mfherbst, line 11: `= (x .= P.f!(similar(x), x)` ? Also I would make `f!(x)` the default, because it is the least-memory paradigm.

src/postprocess/chi0.jl
```diff
@@ -88,6 +86,32 @@ function compute_χ0(ham; droptol=0, temperature=ham.basis.model.temperature)
...
+    Q(ϕ) = ϕ - ψk * (ψk' * ϕ)
+    function QHQ(ϕ)
+        Qϕ = Q(ϕ)
+        Q(Hk * Qϕ - εnk * Qϕ)
+    end
+    precon = PreconditionerTPA(basis, kpoint)
+    precondprep!(precon, ψnk)
+    function f_ldiv!(x, y)
+        ldiv!(x, precon, Q(y))
+        x .= Q(x)
```
mfherbst, line 103: I'm kind of tempted to write this in one line: `f_ldix!(x, y) = x .= Q(ldiv!(x, precon, Q(y)))`

src/postprocess/chi0.jl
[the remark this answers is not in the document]
mfherbst: preconditioners.jl, which we move to the top level from `eigen` ?

src/postprocess/chi0.jl
antoine-levitt: how do you mean it's easy to overlook? It's literally the only thing in this struct. I can make it `fun` if you want

src/postprocess/chi0.jl
antoine-levitt: it really depends on the underlying implementation. Anyway this doesn't matter at all, it's way lower order, so let's revisit this if we need to optimize it

src/postprocess/chi0.jl
mfherbst: Yes. I meant if it is used or set in the code. I'm ok with `fun!`.

src/postprocess/chi0.jl
mfherbst: I don't mind strongly too be fair. We can leave it as is.

src/postprocess/chi0.jl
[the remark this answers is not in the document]
antoine-levitt: nah I don't think we should have this as a proper thing. It's just a kludge needed to communicate with the iterative solver because the whole ldiv/lmul is a bit of a mess. Might as well make it clear it's a workaround and have it just when we use it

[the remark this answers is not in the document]
mfherbst: well my point kind of is: What if we also need it somewhere else? Defining it again with the same name is an issue and using yet another struct is a bit stupid, too. But maybe we should burn that bridge when we get there.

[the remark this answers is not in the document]
antoine-levitt: yeah we'll see when we see the requirement of that hypothetical other usage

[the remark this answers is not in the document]
antoine-levitt: I'm very bad at anticipating what future code will need so let's just not do it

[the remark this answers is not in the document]
mfherbst: ok.

examples/dielectric.jl
mfherbst: Why not just `[inds..., (min(N, howmany)+1:N)...]`?

antoine-levitt: it's a hack anyway just to get some nice output for a quick look at the spectrum, so as long as it works I'm happy...

mfherbst: I understand, but stuff like this really bothers me. It's tons more complicated than it should be. Also it repeats one index (`min(N, howmany)`). Not sure that is what you wanted ...

antoine-levitt: ok

Issue #203, opened 2020-04-20
Debug spglib symmetry handling


antoine-levitt:
And see if we can remove the warning for recent versions

antoine-levitt:
Repro

```
using DFTK
using Plots

# Calculation parameters
kgrid = [3, 3, 3]       # k-Point grid
supercell = [1, 1, 1]   # Lattice supercell
Ecut = 15               # kinetic energy cutoff in Hartree
n_bands = 8             # number of bands to plot in the bandstructure

# Setup silicon lattice
a = 10.263141334305942  # Silicon lattice constant in Bohr
lattice = a / 2 .* [[0 1 1.]; [1 0 1.]; [1 1 0.]]
lattice = a * randn(3, 3)
Si = ElementPsp(:Si, psp=load_psp("hgh/lda/Si-q4"))
atoms = [Si => [rand(3), rand(3), rand(3)]]

model = model_LDA(lattice, atoms)
basis = PlaneWaveBasis(model, Ecut; kgrid=kgrid)
```
One possible thing is that we're missing time-reversal symmetry? Should investigate.

antoine-levitt:
Yup, pretty sure that's what going on. The unmapped kpoints all can be mapped with the symmetry k -> -k, which we just happen to not implement. This symmetry is special, because there's a conjugation of the wavefunctions, which however should not matter for most operations. Also this symmetry is never broken unless there's a magnetic field, so it should be "safe" for eg forces etc.

antoine-levitt:
And the reason why it's pretty rare is that 1) apparently it only applies to odd grids (not sure why, should investigate) 2) it doesn't happen for crystals with inversion symmetry

antoine-levitt:
To be clear: this is only about the `[ Info: 62 reducible kpoints could not be generated from the irreducible kpoints returned by spglib. 62 of these are added as extra irreducible kpoints.` message. It's separate from the bug related to recent versions of spglib.

antoine-levitt:
As a workaround if we can't be bothered to implement time-reversal symmetry we can just set the `is_time_reversal` argument to `false`

mfherbst:
Cool! Thanks for investigating this. Yes time-reversal is something I never checked or took into consideration. It makes a lot of sense from the experience I've had during the calculations. I think for now setting `is_time_reversal` to `false` sounds like a good way to go to me.

antoine-levitt:
[...]

PR #222, opened 2020-05-07
Disable time-reversal symmetry

antoine-levitt:
Still failing on the test reduced in https://github.com/JuliaMolSim/DFTK.jl/issues/203

### DIFF
--- a/src/DFTK.jl
+++ b/src/DFTK.jl
@@ -118,6 +118,7 @@ include("scf/direct_minimization.jl")
 #
 # Utilities
 #
+export standardize_atoms
 export bzmesh_uniform
 export bzmesh_ir_wedge
 export kgrid_size_from_minimal_spacing
--- a/src/bzmesh.jl
+++ b/src/bzmesh.jl
@@ -95,9 +95,10 @@ function bzmesh_ir_wedge(kgrid_size, lattice, atoms; tol_symmetry=1e-5)
     n_symops = length(Stildes)
 
     # Use determined symmetries to deduce irreducible k-Point mesh
+    # TODO implement time-reversal symmetry and turn the flag to true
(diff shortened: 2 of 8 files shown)


src/bzmesh.jl
```diff
@@ -139,14 +140,12 @@ function bzmesh_ir_wedge(kgrid_size, lattice, atoms; tol_symmetry=1e-5)
...
-              "these are added as extra irreducible kpoints.")
-
-        append!(kirreds, eirreds)
-        append!(ksymops, esymops)
+        error("$(length(kreds_notmapped)) reducible kpoints could not be generated from " *
+              "the irreducible kpoints returned by spglib.")
+        # # add them as reducible anyway
+        # eirreds, esymops = find_irreducible_kpoints(kreds_notmapped, Stildes, τtildes)
+        # append!(kirreds, eirreds)
+        # append!(ksymops, esymops)
```
mfherbst, line 148: I would vote to still keep this, even if it should not happen. If you are in a production calculation and the code crashes because of this you are not happy, believe me.

antoine-levitt: Well we should fix it first :p but if something that should not happen happens, I have zero confidence in the result and would rather like it crashes than produce potentially bad results. We are not particularly production oriented anyway...

src/DFTK.jl
```diff
@@ -118,6 +118,7 @@ include("scf/direct_minimization.jl")
 #
 # Utilities
 #
+export standardise_atoms
```
antoine-levitt, line 121: ize, I'm afraid :-p

mfherbst: true

mfherbst: Just doesn't compile in my brain

mfherbst:
Any good idea why the test on the forces fails if I use a shifted mesh?

PR #226, opened 2020-05-16
Add funding sources

mfherbst:


antoine-levitt:
Also Inria, and possibly enpc?

mfherbst:
Good point. Feel free to amend by the way.

PR #228, opened 2020-05-19
Add scaling to Hartree/XC

antoine-levitt:
Maybe we should have a more systematic way of doing this, but these two are the ones for which this is likely to be helpful.

### DIFF
--- a/src/terms/hartree.jl
+++ b/src/terms/hartree.jl
@@ -15,7 +15,7 @@ struct Hartree
     scaling::Real  # to scale by an arbitrary factor (useful for exploration)
 end
 Hartree() = Hartree(1)
-(H::Hartree)(basis) = TermHartree(basis, H.scaling)
+(hartree::Hartree)(basis) = TermHartree(basis, hartree.scaling)
 
 struct TermHartree <: Term
     basis::PlaneWaveBasis
--- a/src/terms/hartree.jl
+++ b/src/terms/hartree.jl
@@ -12,14 +12,14 @@ where G is the Green's function of the periodic Laplacian with zero
 mean (-Δ G = sum_{R} 4π δ_R, integral of G zero on a unit cell).
 """
 struct Hartree
-    scaling::Real
+    scaling::Real  # to scale by an arbitrary factor (useful for exploration)
 end
 Hartree() = Hartree(1)
 (H::Hartree)(basis) = TermHartree(basis, H.scaling)
 
 struct TermHartree <: Term
     basis::PlaneWaveBasis
-    scaling::Real  # to scale by an arbitrary factor (useful for exploration)
+    scaling::Real
     # Fourier coefficients of the Green's function of the periodic Poisson equation
     poisson_green_coeffs
 end
@@ -30,7 +30,7 @@ function TermHartree(basis::PlaneWaveBasis{T}, scaling) where T
                                                 for G in G_vectors(basis)]
     # Zero the DC component (i.e. assume a compensating charge background)
     poisson_green_coeffs[1] = 0
-    TermHartree(basis, poisson_green_coeffs)
+    TermHartree(basis, scaling, poisson_green_coeffs)
 end
 
 function ene_ops(term::TermHartree, ψ, occ; ρ, kwargs...)
--- a/src/terms/hartree.jl
+++ b/src/terms/hartree.jl
@@ -14,7 +14,7 @@ mean (-Δ G = sum_{R} 4π δ_R, integral of G zero on a unit cell).
 struct Hartree
     scaling::Real  # to scale by an arbitrary factor (useful for exploration)
 end
-Hartree() = Hartree(1)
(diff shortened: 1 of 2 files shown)


src/terms/hartree.jl
```diff
@@ -11,22 +11,26 @@ For the Coulomb potential with periodic boundary conditions, this is rather
 where G is the Green's function of the periodic Laplacian with zero
 mean (-Δ G = sum_{R} 4π δ_R, integral of G zero on a unit cell).
 """
-struct Hartree end
-(::Hartree)(basis) = TermHartree(basis)
+struct Hartree
+    scaling::Real  # to scale by an arbitrary factor (useful for exploration)
+end
+Hartree() = Hartree(1)
+(H::Hartree)(basis) = TermHartree(basis, H.scaling)
```
mfherbst, line 18: More than one letter, please.

src/terms/hartree.jl
```diff
@@ -11,22 +11,26 @@ For the Coulomb potential with periodic boundary conditions, this is rather
...
 
 struct TermHartree <: Term
     basis::PlaneWaveBasis
+    scaling::Real
```
mfherbst, line 22: Scaling might be a slight misnomer (sounds like it might not be a factor). How about `prefactor` ?

src/terms/xc.jl
```diff
@@ -6,18 +6,21 @@ Exchange-correlation term, defined by a list of functionals and usually evaluate
 """
 struct Xc
     functionals::Vector{Libxc.Functional}
+    scaling::Real  # to scale by an arbitrary factor (useful for exploration)
```
mfherbst, line 9: See above comment on the name.

src/terms/xc.jl
```diff
@@ -40,6 +43,10 @@ function ene_ops(term::XcTerm, ψ, occ; ρ, kwargs...)
         dVol = basis.model.unit_cell_volume / prod(basis.fft_size)
         E += sum(Epp .* ρ.real) * dVol
     end
+    if term.scaling != 1
```
mfherbst, line 46: Is this needed? Doesn't Julia optimise this away by itself?

src/terms/hartree.jl
antoine-levitt: But the planet!

antoine-levitt: In fact I'm not sure what would be a good name except for H here.

src/terms/xc.jl
antoine-levitt: It doesn't, because it's not clever enough. In fact even `x .= x` is O(n). Also there is trickery afoot with floating point numbers which makes this kind of optimization difficult (eg `0.0 + x` is not always `x` when x is `-0.0`)

antoine-levitt: Sorry `x .= x` is fast.

src/terms/hartree.jl
mfherbst: term? operator? hartree?

antoine-levitt:
Between scaling and prefactor I think I prefer scaling. Or scale_factor?

antoine-levitt:
scaling_factor rather

src/terms/xc.jl
```diff
@@ -8,7 +8,7 @@ struct Xc
     functionals::Vector{Libxc.Functional}
     scaling::Real  # to scale by an arbitrary factor (useful for exploration)
 end
-Xc(functionals::Vector{Libxc.Functional}; scaling=1) = Xc(functionals, scaling)  # default to no scaling
```
mfherbst: Any reason why you removed this here. I would rather remove the type specialisation in `Vector{Symbol}` and keep it specific here as a guard to the type in the functionals field.

src/terms/hartree.jl
mfherbst: So we settle on scaling because we find no better name?

src/terms/xc.jl
antoine-levitt: Tests fail without it because it's called with Vector{Any}

src/terms/hartree.jl
antoine-levitt: As I said above, I don't like prefactor but scaling_factor is good if you want.

mfherbst: Then I think I'd prefer that.

mfherbst: Sorry I missed the comment below.

src/terms/xc.jl
mfherbst: That smells like a bug somewhere else. Why was this not a problem before?

antoine-levitt: Because Vector{Libxc.Functional}(arr) where arr is a Vector{Any} works if all terms in the vector are functionals, I would guess. As long as it works I'm happy, it's not as if it's going to be a subtle bug: either it works or it crashes

PR #229, opened 2020-05-21
Move docs to MD

antoine-levitt:
Moved over things from latex to markdown.


antoine-levitt:
Ready for review. I think the next step is to move the examples (or at least a subset) to this format (but maybe tricky due to run time limitations ?)

antoine-levitt:
Hm there's a path issue, but I'm not sure how to solve it properly...

mfherbst:
Very good go at it! I really like it. I have a few amendments in mind, which I'll do, but I agree the combination of Literate and Documenter seems like a good path to go down.

antoine-levitt:
For sure. The thing to decide is how much we want to move there. We could make some of the simpler (and CPU time-friendly) examples in a literate-friendly format and include them in the docs, while keeping them in the examples/ folder. We should also have a better structure (this is a bit awkward to do with documenter so I didn't do it yet)

mfherbst:
Ok. I think we have a nice first draft. I'm also not completely happy with the structure and everything. Especially I am not too sure using Literate in the way we do makes sense. I mean after all we mostly have text and only a little bit code, but Literate feels more for cases where this is the other way round (like tutorials maybe), but not so much for giving people a general overview over DFTK on an Index page or giving them the "tour". There writing it directly in documenter (maybe using `@example` and `@repl` blocks) makes more sense to me. I'll play with that tomorrow and see what can be done. But from my point of view this is no blocker to not merge this PR as it is and make the amendments later. The current version is already 1Mio times better compared to what is live at the moment :smile:.

antoine-levitt:
You're saying that because you have to put a # before every line? It's a bit annoying yeah but I can live with it. Also we can very easily have different format in different files: just include everything as md and have literate convert every jl to a md.

mfherbst:
Oh yes sure. I completely agree: I think mixing both depending on the content of the file is the way to go. I just think "everything is literate" is not ideal.

PR #230, opened 2020-05-26
Cosmetic improvements to band structure plotting

antoine-levitt:


antoine-levitt:
The detexify we can just remove, it's a bunch of not super fascinating extra code

antoine-levitt:
interpolate_kpoints makes a ~3x difference, so we should definitely turn that on

antoine-levitt:
Rough band structure (kline_density=5) is now 2x the time of an SCF for silicon (4x4x4 kpoints). Total 12s on my laptop, which is reasonable enough to do during docs build.

src/postprocess/band_structure.jl
```diff
@@ -21,14 +21,14 @@ function high_symmetry_kpath(model; kline_density=20)
...
-                       eigensolver=lobpcg_hyper, tol=1e-5, show_progress=true, kwargs...)
+                       eigensolver=lobpcg_hyper, tol=1e-3, show_progress=true, kwargs...)
     # Create basis with new kpoints, where we cheat by using any symmetry operations.
     ksymops = [[(Mat3{Int}(I), Vec3(zeros(3)))] for _ in 1:length(kcoords)]
     bs_basis = PlaneWaveBasis(basis, kcoords, ksymops)
     ham = Hamiltonian(bs_basis; ρ=ρ)
 
     band_data = diagonalize_all_kblocks(eigensolver, ham, n_bands + 3;
-                                        n_conv_check=n_bands, interpolate_kpoints=false,
+                                        n_conv_check=n_bands,
```
mfherbst, line 31: I had trouble when this was on that band structures could give rise to very strange kinks (because it is not guaranteed that when going from one branch to another interpolation makes sense). Notice that this is *not* the case for silicon. There it is perfectly reasonable to switch it on.

src/postprocess/band_structure.jl
```diff
@@ -113,7 +113,7 @@ function plot_band_data(band_data; εF=nothing,
             energies = (data.λ[ibranch][spin][iband, :] .- eshift) ./ unit_to_au(unit)
 
             color = (spin == :up) ? :blue : :red
-            Plots.plot!(p, kdistances, energies, color=color, label="", yerror=yerror)
+            Plots.plot!(p, kdistances, energies, color=color, label="", yerror=yerror, marker=4)
```
mfherbst, line 116: I prefer to use symbols here.

src/postprocess/band_structure.jl
```diff
@@ -135,14 +135,24 @@ function plot_band_data(band_data; εF=nothing,
...
+function detexify_kpoint(string)
+    # For some reason Julia doesn't support this naively: https://github.com/JuliaLang/julia/issues/29849
+    replacements = ("\\Gamma" => "Γ",
+                    "\\Delta" => "Δ",
+                    "\\Sigma" => "Σ")
+    for r in replacements
+        string = replace(string, r)
+    end
+    string
+end
```
mfherbst, line 147: Is there really no shorter way to do this ... that's a lot of code for just 3 replacements.

src/postprocess/band_structure.jl
antoine-levitt: how do you do that?

src/postprocess/band_structure.jl
antoine-levitt: Really? Remember on what system? Tol is pretty tight and should directly translate to a bound on the eigenvalues, so there might be another bug there.

src/postprocess/band_structure.jl
antoine-levitt: I agree! But I don't see a faster way (except for some fold trickery). We can just remove this...

src/postprocess/band_structure.jl
mfherbst: It's not a bug. It found true eigenvalues, but it missed some in the middle. The problem is that using the interpolation across kpath branches is just not a very good guess. Could have been Aluminium or something like that, but I don't remember.

src/postprocess/band_structure.jl
antoine-levitt: 4 is just the size, and it puts a nice round so I think that's OK?

mfherbst: ah so it's markersize ... yes plots maps these keys around. But for library code the long form is favoured.

src/postprocess/band_structure.jl
antoine-levitt: Then yeah it's a bug, the eigensolver is not supposed to miss eigenvalues (the same could happen in a SCF for instance). I'd be very interested in a repro.

mfherbst: Uff ... I don't remember ... was a long time ago. But ok switch it off and then we see.

src/postprocess/band_structure.jl
antoine-levitt: So I'm confused, you want `marker=:o`?

mfherbst: markersize=4 would be what I want ;)

antoine-levitt: Is what is now OK?

mfherbst: Actually for the default settings using this flag looks pretty crap. I'd suggest to use `markershape=:+` and only if the density of kpoints is not too large.

antoine-levitt: I don't like the + because the directionality just looks weird. I can decrease markersize if you want? I don't particularly care just edit the line with what you want

PR #232, opened 2020-05-27
change the real space grid to [-1/2, 1/2)

antoine-levitt:
Weirdness in the GPE_2D example


antoine-levitt:
Maybe very wrong, need to think about it: the FFT forces a particular choice of grid

antoine-levitt:
Yeah that was stupid. The way we do the FFTs strongly suggests that the natural unit cell is [0, 1)^3, so rather than change the way FFTs are computed we just take [0, 1) as the unit cell. Slight inconsistency with the BZ but it's OK. I just left a clarification commit but otherwise reverted the rest, let's merge this one pending CI

PR #234, opened 2020-05-28
Bzmesh stuff

antoine-levitt:


antoine-levitt:
Hopefully our last change to this part of the code for a while...

src/PlaneWaveBasis.jl
```diff
@@ -180,12 +180,14 @@ end
...
-Creates a `PlaneWaveBasis` using the kinetic energy cutoff `Ecut` a Monkhorst-Pack
-kpoint grid `kgrid` shifted by `kshift` (which gives the fraction of adjacent mesh points
-to shift the mesh in each direction). If `enable_bzmesh_symmetry` is `true` (default)
-lattice symmetries are used to reduce the number of ``k``-Points which are treated
-explicitly. In this case all guess densities, (analytic) potential functions
-and guess wavefunctions used with DFTK must agree with the employed symmetry operations.
+Creates a `PlaneWaveBasis` using the kinetic energy cutoff `Ecut` and a Monkhorst-Pack
+kpoint grid `kgrid` shifted by `kshift` (0 or 1/2 in each direction
+
+If). `enable_bzmesh_symmetry` is `true` (default) the symmetries of the
```
mfherbst, line 186: Typo at the beginning.

src/PlaneWaveBasis.jl
```diff
@@ -403,19 +412,18 @@ function apply_ksymop(ksymop, basis, kpoint, ψk::AbstractVecOrMat)
...
 
     Skpoint, ψSk
 end
 
 """"
-Convert a basis into one that uses or doesn't use BZ symmetrization
-Mainly useful for debug purposes and in cases we don't want to
-bother with symmetry
+Convert a `basis` into one that uses or doesn't use BZ symmetrization
+Mainly useful for debug purposes (eg in cases we don't want to
```
mfherbst, line 425: e.g.

src/bzmesh.jl
```diff
@@ -1,14 +1,13 @@
 include("external/spglib.jl")
 
 
-"""Bring kpoint coordinates into the range (-0.5, 0.5]"""
-function normalise_kpoint_coordinate(x::Real)
```
mfherbst, line 5: Believe me I don't do the British thing on purpose ...

src/densities.jl
```diff
@@ -112,11 +112,11 @@ end
...
 """
-Symmetrise a `RealFourierArray` by applying all symmetry operations of
-the bases (or all symmetries passed as the second argument) and forming
-the average. Notice that an identity needs to be passed explicitly.
+Symmetrize a `RealFourierArray` by applying all symmetry operations of
+the basis (or all symmetries passed as the second argument) and forming
+the average.
 """
-function symmetrize(ρin::RealFourierArray, ksymops=ksymops(ρin.basis))
+function symmetrize_density(ρin::RealFourierArray, ksymops=ksymops(ρin.basis))
```
mfherbst, line 119: I called this symmetrize only on purpose because than we can also have a symmetrise function for wave functions at some point in the future.

src/PlaneWaveBasis.jl
```diff
@@ -33,15 +33,15 @@ Discretization information for kpoint-dependent quantities such as orbitals.
...
-                                  # ({G, 1/2 |k+G|^2 ≤ Ecut})
+struct Kpoint
+    spin::Symbol                       # :up, :down, :both or :spinless
+    coordinate::Vec3{Rational{Int}}    # Fractional coordinate of k-Point
+    mapping::Vector{Int}               # Index of G_vectors[i] on the FFT grid:
+                                       # G_vectors(basis)[kpt.mapping[i]] == G_vectors(kpt)[i]
+    mapping_inv::Dict{Int, Int}        # Inverse of `mapping`:
+                                       # G_vectors(basis)[i] = G_vectors(kpt)[kpt.mapping_inv[i]]
+    G_vectors::Vector{Vec3{Int}}       # Wave vectors in integer coordinates:
+                                       # ({G, 1/2 |k+G|^2 ≤ Ecut})
```
mfherbst, line 44: I think I only would have two spaces of space between the `Vec3{Rational{Int}}` and the first comment to avoid the lines getting too long :smile:

src/densities.jl
antoine-levitt: I changed it because symmetrize was a bit ambiguous to me, but now I think about it it's the only meaning of symmetrize I can think about. Will revert.

PR #236, opened 2020-05-29
TimerOutputs()

### LINKED ISSUE #213: Detailed timings
using [TimerOutputs](https://github.com/KristofferC/TimerOutputs.jl) for better comparison against other packages.

gkemlin:
Adding `timer`, a `TimerOutputs()` object in `src/DFTK.jl` to measure easily the execution time of some parts of the code or new features. An example of basic usage is available in `example/silicon_timer.jl`.

Closes #213.


examples/silicon_timer.jl
```diff
@@ -0,0 +1,39 @@
+using DFTK
```
antoine-levitt, line 1: I wouldn't have this as a separate example. Either remove it or just add the timers to another example

src/DFTK.jl
```diff
@@ -10,6 +10,9 @@ using LinearAlgebra
 using Interpolations
 using Requires
 
+using TimerOutputs
+const timer = TimerOutput()
```
antoine-levitt, line 14: Should we have a macro that uses this? Eg `@timer begin ... end` which would expand to `@timeit timer begin ... end`?

src/bzmesh.jl
```diff
@@ -118,7 +118,7 @@ representing a mapping from `Element` objects to a list of positions in fraction
 coordinates. `tol_symmetry` is the tolerance used for searching for symmetry operations.
 """
 function bzmesh_ir_wedge(kgrid_size, lattice, atoms;
-                         tol_symmetry=1e-5, kshift=[0, 0, 0])
+                                    tol_symmetry=1e-5, kshift=[0, 0, 0])
```
antoine-levitt, line 121: leftover

src/eigen/diag.jl
```diff
@@ -9,10 +9,10 @@ that really does the work, operating on a single ``k``-Block.
 `prec_type` should be a function that returns a preconditioner when called as `prec(ham, kpt)`
 """
 function diagonalize_all_kblocks(eigensolver, ham::Hamiltonian, nev_per_kpoint::Int;
```
antoine-levitt, line 11: leftover

src/eigen/diag.jl
```diff
@@ -31,13 +31,17 @@ function diagonalize_all_kblocks(eigensolver, ham::Hamiltonian, nev_per_kpoint::
             # guess provided
             guessk = guess[ik]
         elseif interpolate_kpoints && ik > 1
-            # use information from previous kpoint
-            X0 = interpolate_kpoint(results[ik - 1].X, kpoints[ik - 1], kpoints[ik])
-            guessk = Matrix{T}(qr(X0).Q)  # Re-orthogonalize and renormalize
+            @timeit timer "QR orthonormalization" begin
```
antoine-levitt, line 34: factorize the QR here so you just have one timer

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -17,13 +17,15 @@ using GenericLinearAlgebra
 # Given A as a list of blocks [A1, A2, A3] this forms the matrix-matrix product
 # A * B avoiding a concatenation of the blocks to a dense array
 @views function bmul(blocks::Tuple, B)
-    res = blocks[1] * B[1:size(blocks[1], 2), :]  # First multiplication
-    offset = size(blocks[1], 2)
-    for block in blocks[2:end]
-        mul!(res, block, B[offset .+ (1:size(block, 2)), :], 1, 1)
-        offset += size(block, 2)
+    @timeit timer "bmul" begin
```
antoine-levitt, line 20: timeit in front of function?

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -56,12 +58,14 @@ boverlap(A, B) = A' * B
 
 # Perform a Rayleigh-Ritz for the N first eigenvectors.
 function RR(X, AX, BX, N)
```
antoine-levitt, line 60: might as well rename it to rayleigh_ritz so you don' thave to name the timer

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -227,8 +231,8 @@ end
 ### We reuse applications of A/B when it is safe to do so, ie only with orthogonal transformations
 
 function LOBPCG(A, X, B=I, precon=((Y, X, R)->R), tol=1e-10, maxiter=100;
-                miniter=1, ortho_tol=2eps(real(eltype(X))),
-                n_conv_check=nothing, display_progress=false)
+                           miniter=1, ortho_tol=2eps(real(eltype(X))),
```
antoine-levitt, line 234: leftover

src/terms/Hamiltonian.jl
```diff
@@ -31,43 +31,45 @@ end
 # As an optimization we special-case nonlocal operators to apply them
 # instead on the full block and benefit from BLAS3
 @views function LinearAlgebra.mul!(Hψ::AbstractArray, H::HamiltonianBlock, ψ::AbstractArray)
```
antoine-levitt, line 33: I'd rather avoid the extra level of indentation here...

src/terms/ewald.jl
```diff
@@ -25,23 +25,25 @@ function ene_ops(term::TermEwald, ψ, occ; kwargs...)
...
-    forces_ewald = zeros(Vec3{T}, sum(length(positions) for (elem, positions) in atoms))
-    energy_ewald(term.basis.model; forces=forces_ewald)
-    # translate to the "folded" representation
-    f = [zeros(Vec3{T}, length(positions)) for (type, positions) in atoms]
-    count = 0
-    for i = 1:length(atoms)
-        for j = 1:length(atoms[i][2])
-            count += 1
-            f[i][j] += forces_ewald[count]
+    @timeit timer "ewald" begin
```
antoine-levitt, line 28: ewald forces, or forces/ewald?

antoine-levitt: We should really find a way of annotating the function directly and avoid these huge blocks

src/terms/local.jl
```diff
@@ -89,26 +89,28 @@ function (E::AtomicLocal)(basis::PlaneWaveBasis{T}) where {T}
...
-                        for G in G_vectors(term.basis)]
-
-        for (ir, r) in enumerate(positions)
-            forces[iel][ir] = -real(sum(conj(ρ.fourier[iG])
-                                    .* form_factors[iG]
-                                    .* cis(-2T(π) * dot(G, r))
-                                    .* (-2T(π)) .* G .* im
-                                    ./ sqrt(unit_cell_volume)
-                                   for (iG, G) in enumerate(G_vectors(term.basis))))
+    @timeit timer "local" begin
```
antoine-levitt, line 92: forces/local

src/terms/nonlocal.jl
```diff
@@ -41,58 +41,60 @@ function ene_ops(term::TermAtomicNonlocal, ψ, occ; kwargs...)
...
-                        P = structure_factors .* form_factors ./ sqrt(unit_cell_volume)
-                        dPdR = [-2T(π)*im*(Skcoord + G)[α] for G in G_vectors(Skpoint)] .* P
-
-                        # TODO BLASify this further
-                        for iband = 1:size(ψ[ik], 2)
-                            ψnSk = @view ψSk[:, iband]
-                            fr[α] -= (occ[ik][iband] / tot_red_kpt_number
-                                      * real(  dot(ψnSk, P * C * dPdR' * ψnSk)
-                                             + dot(ψnSk, dPdR * C * P' * ψnSk)))
+    @timeit timer "nonlocal" begin
```
antoine-levitt, line 44: forces/nonlocal

src/terms/operators.jl
```diff
@@ -99,7 +107,11 @@ struct NonlocalOperator <: RealFourierOperator
     P
     D
 end
-apply!(Hψ, op::NonlocalOperator, ψ) = Hψ.fourier .+= op.P * (op.D * (op.P' * ψ.fourier))
+function apply!(Hψ, op::NonlocalOperator, ψ)
+    @timeit timer "Apply nonlocal operator" begin
```
antoine-levitt, line 111: Remove the apply for consistency with the others

src/terms/operators.jl
```diff
@@ -110,7 +122,7 @@ struct MagneticFieldOperator <: RealFourierOperator
     kpoint
     Apot  # Apot[α][i,j,k] is the A field in direction α
 end
-function apply!(Hψ, op::MagneticFieldOperator, ψ)
+@timeit timer function apply!(Hψ, op::MagneticFieldOperator, ψ)
```
antoine-levitt, line 125: Give it a name

antoine-levitt:
Some nitpicking, but my biggest gripe is that we should have a way to do `@timeit timer "something" function ...`.

src/DFTK.jl
mfherbst: Pros: Less typing, Cons: What if we want a second kind of timer?
I think since this rarely happens an `@timer` macro sounds like a good idea.

src/terms/Hamiltonian.jl
mfherbst: We could make a function of the same signature with a sensible name and then just alias `mul!` to that.

src/terms/ewald.jl
mfherbst: Same idea as above could apply here as well, just have a `forces_ewald` function and then ` forces(bla) = forces_ewald(bla)` in one line

mfherbst:
I agree: https://github.com/KristofferC/TimerOutputs.jl/issues/86

Project.toml
```diff
@@ -33,6 +33,7 @@ Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 Roots = "f2b01f46-fcfa-551c-844a-d8ac1e96c665"
 SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
+TimerOutputs = "a759f4b9-e2f1-59dc-863e-4aeb61b1ea8f"
 
 [compat]
 AbstractFFTs = "0.5"
```
mfherbst, line 39: Pin the TimerOutput version (to "0.5")

src/Model.jl
```diff
@@ -1,7 +1,7 @@
 # Contains the physical specification of the model
 
 """
-A physical specification of a model. 
+A physical specification of a model.
```
mfherbst, line 4: Whitespace commit.

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -56,12 +58,14 @@ boverlap(A, B) = A' * B
...
-    F.vectors[:,1:N], F.values[1:N]
+    @timeit timer "Rayleigh-Ritz" begin
+        F = eigen(Hermitian(boverlap(X, AX, assume_hermitian=true)))  # == Hermitian(X' AX)
+        F.vectors[:,1:N], F.values[1:N]
+    end
 end
 
 import LinearAlgebra: cholesky
-function cholesky(X::Union{Matrix{ComplexF16}, Hermitian{ComplexF16,Matrix{ComplexF16}}})
+@timeit timer function cholesky(X::Union{Matrix{ComplexF16}, Hermitian{ComplexF16,Matrix{ComplexF16}}})
```
mfherbst, line 68: Remove this. This is a super special function and we never time the other versions in LinearAlgebra.

src/terms/ewald.jl
```diff
@@ -1,7 +1,7 @@
 using SpecialFunctions: erfc
 using ForwardDiff
 
-""" 
+"""
```
mfherbst, line 4: whitespace commit.

src/terms/Hamiltonian.jl
mfherbst: I would suggest calling the timed function `hamiltonian_mul` and then just alias the `LinearAlgebra.mul!` to that function in one line.

src/terms/local.jl
mfherbst: aliasing trick again

src/terms/operators.jl
```diff
@@ -52,7 +52,11 @@ struct RealSpaceMultiplication <: RealFourierOperator
     kpoint
     potential::AbstractArray
 end
-apply!(Hψ, op::RealSpaceMultiplication, ψ) = Hψ.real .+= op.potential .* ψ.real
+function apply!(Hψ, op::RealSpaceMultiplication, ψ)
```
mfherbst, line 55: It's really a shame this distroys the nice one line. Aliasing to apply_real_space_multiplication?

mfherbst:
https://github.com/KristofferC/TimerOutputs.jl/pull/87 is now merged, so we could incorporate that here directly.

examples/silicon_timer.jl
gkemlin: What do you mean ? To remove the `silicon_timer.jl` example and use timers in another example like `silicon.jl` ?

src/DFTK.jl
gkemlin: Yes, I think it would be better too.

examples/silicon_timer.jl
antoine-levitt: Yeah. Or preferably even a more developer-oriented one like 1D GPE or https://github.com/JuliaMolSim/DFTK.jl/pull/246

gkemlin:
Aliasing functions is not necessary anymore : thanks to https://github.com/KristofferC/TimerOutputs.jl/pull/87 we can do `@timer "bla" function foo() ... end` to alias them with the new macro.

src/eigen/diag_lobpcg_hyper.jl
```diff
@@ -1,7 +1,8 @@
 include("lobpcg_hyper_impl.jl")
 
-@timeit timer function lobpcg_hyper(A, X0; maxiter=100, prec=nothing, tol=20size(A, 2)*eps(real(eltype(A))),
-                                    largest=false, n_conv_check=nothing, kwargs...)
+@timer "LOBPCG" function lobpcg_hyper(A, X0; maxiter=100, prec=nothing,
+                                      tol=20size(A, 2)*eps(real(eltype(A))),
+                                      largest=false, n_conv_check=nothing, kwargs...)
```
mfherbst, line 5: If you move this to the LOBPCG function itself you can again save some boilerplate.

src/common/timer.jl
```diff
@@ -0,0 +1,6 @@
+const to = TimerOutput()
+# creating a new macro to shorten the call :
+# replaces `@timeit to [label] [block]` by `@timer [label] [block]`
+macro timer(args...)
+    TimerOutputs.timer_expr(__module__, false, to, args...)
+end
```
mfherbst, line 6: I'm starting to think whether we should call this `@timed` instead. `@timed function` reads nicer than `@timer function`.

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -77,7 +73,7 @@ end
 # Returns the new X, the number of Cholesky factorizations algorithm, and the
 # growth factor by which small perturbations of X can have been
 # magnified
-@timeit timer function ortho(X; tol=2eps(real(eltype(X))))
+@timer "orthogonalization" function ortho(X; tol=2eps(real(eltype(X))))
```
mfherbst, line 76: Might as well rename the function again.

src/scf/self_consistent_field.jl
```diff
@@ -93,21 +93,21 @@ end
...
-                                             solver=scf_nlsolve_solver(),
-                                             eigensolver=lobpcg_hyper,
-                                             n_ep_extra=3,
-                                             determine_diagtol=ScfDiagtol(),
-                                             mixing=SimpleMixing(),
-                                             callback=scf_default_callback,
-                                             is_converged=ScfConvergenceEnergy(tol),
-                                             compute_consistent_energies=true,
-                                            )
+@timer "SCF" function self_consistent_field(basis::PlaneWaveBasis;
```
mfherbst, line 96: You think `self_consistent_field` is too long? I think it's fine.

src/terms/Hamiltonian.jl
```diff
@@ -30,46 +30,46 @@ end
 # Do this band by band to conserve memory
 # As an optimization we special-case nonlocal operators to apply them
 # instead on the full block and benefit from BLAS3
-@views function LinearAlgebra.mul!(Hψ::AbstractArray, H::HamiltonianBlock, ψ::AbstractArray)
-    @timeit timer "mul!" begin
-        basis = H.basis
-        kpt = H.kpoint
-        nband = size(ψ, 2)
+@views @timer "hamiltonian multiplication" function LinearAlgebra.mul!(Hψ::AbstractArray,
```
mfherbst, line 33: `Hamiltonian`

src/terms/operators.jl
```diff
@@ -122,7 +116,7 @@ struct MagneticFieldOperator <: RealFourierOperator
     kpoint
     Apot  # Apot[α][i,j,k] is the A field in direction α
 end
-@timeit timer function apply!(Hψ, op::MagneticFieldOperator, ψ)
+@timer "apply magnetic field operator" function apply!(Hψ, op::MagneticFieldOperator, ψ)
```
mfherbst, line 119: There are two patters here: Either prefix them all by `appy` or none.

src/common/timer.jl
gkemlin: I don't know, on the one hand it's true that it reads nicer but on the other hand, it's an object that measures time, so calling it `timer` makes sense :thinking: .

mfherbst: Actually the thing which measures time is `to`. So from that point of view I suggest to rename `to -> timer`, `timer -> timed`.

gkemlin: Sounds good to me !

gkemlin: But then we would overwrite the original `Main.@timed`macro also.

mfherbst: Ah I forgot about that ... probably that we should not do. `@timing` maybe?

gkemlin: Yeah, good for me.

PR #237, opened 2020-05-30
Bzmesh stuff refactor

antoine-levitt:
Disabling the symmetrization for now. This should all be non-functional changes.


antoine-levitt:
We can merge this as is if we don't want to wait for release. Symmetrization is still wrong though; fixing it will require changing the pwbasis structure, and probably disallowing the (kcoords, ksymops) constructor

src/symmetry.jl
```diff
@@ -0,0 +1,220 @@
...
+        @assert all(0 .≤ τ .< 1)
+        push!(symops, (S, τ))
+    end
+
+    symops = unique(symops)
+
+    if kcoords !== nothing
+        # filter only the operations that respect the symmetries of the discrete BZ grid
+        function preserves_grid(S)
+            all(normalize_kpoint_coordinate(S * k) in kcoords
```
mfherbst, line 56: Probably normalize_kpoint_coordinate should live in this file as well?

src/symmetry.jl
```diff
@@ -0,0 +1,220 @@
...
+
+    symops
+end
+symmetry_operations(model::Model; kwargs...) = symmetry_operations(model.lattice, model.atoms; kwargs...)
+
+function symmetry_operations(basis::PlaneWaveBasis)
+    res = Set()
+    for ik = 1:length(basis.ksymops)
+        for isym = 1:length(basis.ksymops[ik])
+            push!(res, basis.ksymops[ik][isym])
```
mfherbst, line 70: I wonder whether we should add a flag to explicitly make the returned operations a group by closing over the determined set.

src/symmetry.jl
```diff
@@ -0,0 +1,220 @@
...
+"""
+function apply_ksymop(symop, ρin::RealFourierArray)
+    symop[1] == I && iszero(symop[2]) && return ρin
+    from_fourier(ρin.basis, symmetrize(ρin, [symop]))
+end
+
+
+# Accumulates the symmetrized versions of the density ρin into ρout (in Fourier space).
+# No normalization is performed
+function _symmetrize!(ρaccu, ρin, basis, symops, Gs)
```
mfherbst, line 183: Change name, see above.

src/bzmesh.jl
```diff
@@ -182,27 +98,26 @@ function bzmesh_ir_wedge(kgrid_size, lattice, atoms;
...
-                @assert all(0 .≤ τ .< 1)
-                push!(ksymops[ik], (S, τ))
+                push!(ksymops[ik], symops[isym])
             end
         end
     end
 
     if !isempty(kreds_notmapped)
         # add them as reducible anyway
+        Stildes = [s[1]' for s in symops]
```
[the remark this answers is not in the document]
antoine-levitt: ? that won't work. (Stilde, _) in symops you mean?

src/bzmesh.jl
```diff
@@ -182,27 +98,26 @@ function bzmesh_ir_wedge(kgrid_size, lattice, atoms;
...
+        τtildes = [-s[1]' * s[2] for s in symops]
```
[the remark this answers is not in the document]
antoine-levitt: yaeh I agree, much better

src/bzmesh.jl
```diff
@@ -182,27 +98,26 @@ function bzmesh_ir_wedge(kgrid_size, lattice, atoms;
         for ired in k_all_reducible[ik]
             kred = (kshift .+ Vec3(grid[ired, :])) .// kgrid_size
 
-            isym = findfirst(1:length(Stildes)) do isym
+            # Note that this relies on the identity coming up first in symops
+            isym = findfirst(symops) do symop
```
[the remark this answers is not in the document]
antoine-levitt: wait, how did that work ?!?!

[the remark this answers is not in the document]
antoine-levitt: no, it is an index, so what I wrote is OK?

src/bzmesh.jl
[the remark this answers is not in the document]
mfherbst: yes, sorry

src/bzmesh.jl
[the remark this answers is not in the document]
mfherbst: ah no, sorry `findfirst` was the one that returns indices.

src/densities.jl
```diff
@@ -72,61 +72,11 @@ function compute_density(basis::PlaneWaveBasis, ψ::AbstractVector,
         ρaccu .= 0
         for ik in ikpts
             ρ_k = compute_partial_density(basis, basis.kpoints[ik], ψ[ik], occupation[ik])
-            _symmetrize_ρ!(ρaccu, ρ_k, basis, basis.ksymops[ik], Gs)
+            # accumulates all the symops of ρ_k into ρaccu
+            _symmetrize!(ρaccu, ρ_k, basis, basis.ksymops[ik], Gs)
```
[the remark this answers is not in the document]
antoine-levitt: Yeah I agree. The thing is that this function is pretty internal because it accumulates rather than erases, as would be more conventional with a `symmetrize!`, and doesn't normalize, as you would expect from a symmetrize function. Let's just rename it to `add_all_symops` or something?

src/symmetry.jl
antoine-levitt: Not really, normalize has nothing to do with symmetries. It's fine that symmetry.jl and bzmesh.jl depend from each other I think

src/symmetry.jl
antoine-levitt: This function will go anyway because it's wrong. We should just store the "full" symops in the pwbasis. To do that I need to restructure pwbasis a bit which I've put off for now.

src/densities.jl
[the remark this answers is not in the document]
mfherbst: accumulate_over_symops! is better

src/symmetry.jl
mfherbst: hmm ... I prefer hierarchical dependencies. Maybe normalize_kpoint_coordinate should go to a third file?

src/symmetry.jl
mfherbst: ok

src/densities.jl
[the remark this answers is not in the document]
antoine-levitt: Agreed

src/symmetry.jl
antoine-levitt: I thought the exact same thing, but after thinking about it I'm pretty sure it's leftover from static languages. There's very little dependency problems in julia, you just need to define datastructures before functions (eg the symops declaration will have to go to types.jl if we want to use it in Model) but for functions there's no such constraints. The real question is: will somebody reading this be able to figure out where this function is defined? I think bzmesh.jl is a pretty natural place to look for.

mfherbst: You might have a point there. In either way our code is still small enough that it's not too hard to grep for a function. Let's leave it like that for now.

antoine-levitt: Btw it's a point against the index: it screws up my `grep -ir` at DFTK toplevel :-p

PR #238, opened 2020-05-31
Split tutorial & misc changes

antoine-levitt:
I was only able to get it to build with strict=false, but I'm not quite sure what's going on, it says something about multiply defined labels but I don't see them...

After this let's get feedback from the others and release!


mfherbst:
There are still a few issues with deploying the docs I have to fix. I'll do that first and then get to this PR afterwards.

PR #239, opened 2020-05-31
make.jl cleanup

mfherbst:


antoine-levitt:
Don't const in scripts, it just makes life more annoying

PR #240, opened 2020-06-01
Doc typos

gkemlin:
2 typos plus 2 questions where I might be mistaking but we never know : 
- in https://juliamolsim.github.io/DFTK.jl/dev/advanced/conventions/#Units-1 do you use bohr or Bohr ?
- I think the link at the bottom of https://juliamolsim.github.io/DFTK.jl/dev/advanced/conventions/#Normalization-conventions-1 sends to the tutorial page whereas the normalization of the wave functions is in https://juliamolsim.github.io/DFTK.jl/dev/advanced/data_structures/#PlaneWaveBasis-and-plane-wave-discretisations-1

### DIFF
--- a/docs/src/guide/tutorial.jl
+++ b/docs/src/guide/tutorial.jl
@@ -56,7 +56,7 @@ scfres.energies
 # number of eigenvalues that are computed and 8 the number of kpoints.
 hcat(scfres.eigenvalues...)
 # Use `scfres.occupation` to see the number of electrons on each energy level.
-scfres.occupation
+hcat(scfres.occupation...)
 # And density:
 rvecs = collect(r_vectors(basis))[:, 1, 1]  # slice along the x axis
 x = [r[1] for r in rvecs]                   # only keep the x coordinate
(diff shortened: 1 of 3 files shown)


mfherbst:
Both bohr and Bohr are used, but we should stick to one, let's go for "Bohr" (as you already did). Yes the link is indeed wrong. Feel free to fix (should be `[Data structures](@ref)`.

docs/src/guide/tutorial.jl
```diff
@@ -52,8 +52,11 @@ scfres = self_consistent_field(basis, tol=1e-8);
 # That's it! Now you can get various quantities from the result of the SCF.
 # For instance, the energies:
 scfres.energies
-# Eigenvalues:
+# Eigenvalues: here `scfres.eigenvalues` returns a 7x8 Array where 7 is the
+# number of eigenvalues that are computed and 8 the number of kpoints.
 hcat(scfres.eigenvalues...)
+# Use `scfres.occupation` to see the number of electrons on each energy level.
+scfres.occupation
```
mfherbst, line 59: I think you should add an hcat here as well for consistency.

gkemlin: True

mfherbst:
Awesome, thanks!

antoine-levitt:
Great, thanks!

PR #243, opened 2020-06-02
Convenience and documentation for docs generation

mfherbst:
@gkemlin Just tried in a fresh folder. Is that sufficient from your point of view? Clearly says that you need to install the python packages.


gkemlin:
@mfherbst Yes, sounds good to me !

PR #248, opened 2020-06-04
Update Libxc and adapt to its new interface

mfherbst:
- [x] Tag and update Libxc.
- [x] Apply and compute for Hartree kernel
- [ ] Test FD Hartree kernel
- [x] Apply for LDA XC kernel
- [x] Test FD LDA kernel
- [ ] GGA XC kernel


src/terms/xc.jl
```diff
@@ -53,7 +55,31 @@ function ene_ops(term::TermXc, ψ, occ; ρ, kwargs...)
 end
 
 
-# Functionality for building the XC potential term and constructing the builder itself.
+function compute_kernel(term::TermXc; kwargs...)
+    error("Not implemented")
```
antoine-levitt, line 59: just `Diagonal(kernel)` no? (for LDA at least)

src/terms/xc.jl
```diff
@@ -116,6 +142,11 @@ sum(V .* dρ) = sum(V .* IF(Xϕ) .* IF(X dϕ))
 and the result follows.
 =#
 
+function max_required_derivative(functional)
+    functional.family == :lda && return 0
+    functional.family == :gga && return 1
+    return -1
```
antoine-levitt, line 148: error?

src/terms/xc.jl
mfherbst: Yes for LDA it's just what I compute as "kernel" in the apply function. Not yet sure about GGA. I'll look at that tmr. I just wanted to get finite-differences to agree for the first checkpoint ;).

antoine-levitt: GGA once you know what to do for applying the kernel you just have to convert G_to_r calls to G_to_r_matrix and so on. That would also be a nice testcase for forwarddiff. But again if it's too much trouble just forget about it...

mfherbst:
@antoine-levitt Ok I give up on the GGA kernel for now. I included a reference and what I tried to derive it, but the results are not consistent. The corresponding test is marked as broken and the routine returns an error.

antoine-levitt:
Heh OK. Let's merge this and I'll take a look next week. The trick is always to forget about the continuous stuff and simply differentiate at the discrete level.

PR #250, opened 2020-06-05
Remove part about MolSim from installation docs

mfherbst:


antoine-levitt:
Any other *registered* package. Otherwise good to go!

PR #251, opened 2020-06-05
adding ElementCustomIonPotential + new GP example using it (with doc)…

gkemlin:
- new element for to generate AtomicLocal term with custom potential
- new example of 1D GP with SCF and forces using ElementCustomIonPotential with docs
- small typo in 2D GP example
- modified Model.jl to detect 1D/2D models and automatically disable symmetry


docs/src/index.md
```diff
@@ -59,6 +59,7 @@ Pages = [
     "examples/pymatgen.md",
     "examples/ase.md",
     "examples/gross_pitaevskii.md",
+    "examples/gross_pitaevskii_custom_potential.md",
```
antoine-levitt, line 62: put it towards the end as it's more technical

examples/gross_pitaevskii_custom_potential.jl
```diff
@@ -0,0 +1,114 @@
+# # Gross-Pitaevskii equation in one dimension with forces and custom potential
+
+# We solve the 1D Gross-Pitaevskii equation with a custom potential.
+# This is similar to the
+# previous example ([Gross-Pitaevskii equation in one dimension](@ref)) where
+# computing forces was not possible as we used an `ExternalFromReal` term
+# instead of an `AtomicLocal` term (see [`Model` datastructure](@ref)).
+# Here, we use `ElementCustomIonPotential`
+# with a custom potential to generate an `AtomicLocal` term for which we can
+# compute forces.
```
antoine-levitt, line 10: Instead of turning that negative I would simply say that in this example you're going to show how to define local potentials attached to atoms, which allows reusing the machinery for atoms (forces, etc)

examples/gross_pitaevskii_custom_potential.jl
```diff
@@ -0,0 +1,114 @@
...
+# We define the width of the Gaussian potential generated by one ion
+L = 0.5;
+# We set the potential in its real and Fourier forms
+pot_real(x) = exp(-(x/L)^2)
+pot_fourier(q::T) where {T <: Real} = exp(- (q*L)^2 / 4);
+# And finally we build the elements and set their positions in the `atoms` array
+ion = ElementCustomIonPotential(1, pot_real, pot_fourier) # 1 is the charge of each ion
+atoms = [ion => [x1*[1,0,0], x2*[1,0,0]]];
+
+# Parameters for the nonlinear potential ``α C ρ^{α-1}``
```
antoine-levitt, line 34: these comments were already done in the GP example, you can remove them so that the reader knows to focus on the new stuff

examples/gross_pitaevskii_custom_potential.jl
```diff
@@ -0,0 +1,114 @@
...
+# We discretize using a moderate Ecut
+# and run a SCF algorithm to compute forces afterwards:
+Ecut = 500
+basis = PlaneWaveBasis(model, Ecut)
+scfres = self_consistent_field(basis, tol=1e-8)
+scfres.energies
+# Computing forces is then automatic:
+hcat(forces(scfres)...)
+
+# ## Internals
```
antoine-levitt, line 56: Remove that whole section (that example should only be about atomic potentials)

src/Model.jl
```diff
@@ -118,8 +118,13 @@ function Model(lattice::AbstractMatrix{T};
 
     @assert symmetry in (:auto, :force, :off)
     if symmetry == :auto
-        # ask the terms if they break symmetry
-        compute_symmetry = !(any(breaks_symmetries, terms))
+        # check dimension, if 1D or 2D : no symmetry
```
antoine-levitt, line 121: That comment is not very informatice. Instead say that in theory we could have symmetries in 1D and 2D, but spglib sucks and doesn't symmetry detection in 1D and 2D that so we don't either.

src/elements.jl
```diff
@@ -145,3 +145,20 @@ function local_potential_fourier(el::ElementCohenBergstresser, q::T) where {T <:
...
+Element for an ion with charge Z that generates a custom local potential.
+There is no check of the relation between pot_real and pot_fourier, the user
+has to ensure that point.
+"""
+struct ElementCustomIonPotential <: Element
+    Z::Int                  # Nuclear charge
+    pot_real::Function      # Real potential
+    pot_fourier::Function   # Fourier potential
+end
+charge_ionic(el::ElementCustomIonPotential) = el.Z
```
antoine-levitt, line 159: For this model, the nuclear charge (which is really the ionic charge, corresponding to the number of valence electrons only) does not really make sense, so I would rather leave it unimplemented and not have a Z field

src/elements.jl
```diff
@@ -145,3 +145,20 @@ function local_potential_fourier(el::ElementCohenBergstresser, q::T) where {T <:
...
+
+"""1D custom Fourier potential: V(q) = int_{R^3} V(x) e^{-iqx} dx."""
```
antoine-levitt, line 161: Not 1D

src/elements.jl
```diff
@@ -145,3 +145,20 @@ function local_potential_fourier(el::ElementCohenBergstresser, q::T) where {T <:
     qsq_pi = Int(round(q^2 / (2π / el.lattice_constant)^2, digits=2))
     T(get(el.V_sym, qsq_pi, 0.0))
 end
+
+"""
+Element for an ion with charge Z that generates a custom local potential.
```
antoine-levitt, line 150: add "radial"

src/elements.jl
```diff
@@ -145,3 +145,20 @@ function local_potential_fourier(el::ElementCohenBergstresser, q::T) where {T <:
...
+    Z::Int                  # Nuclear charge
+    pot_real::Function      # Real potential
+    pot_fourier::Function   # Fourier potential
+end
+charge_ionic(el::ElementCustomIonPotential) = el.Z
+
+"""1D custom Fourier potential: V(q) = int_{R^3} V(x) e^{-iqx} dx."""
+function local_potential_fourier(el::ElementCustomIonPotential, q::T) where {T <: Real}
+    return el.pot_fourier(q)
+end
```
antoine-levitt, line 164: Since you're there, can you clean up the docstrings a bit? The docstring (eg for local_potential_real) should be docuumented once before the specific implementation for each type. You can do this with eg
```
"""
doc
"""
function X end
``` which defines a function but no methods.

Might as well mention that local_potential_radial is not currently used for anything other than debug/exploration

**Comment by antoine-levitt** [comment]
Also should that custom type be in the example file rather than in elements.jl? I think it makes sense since it's so simple. Also if in the DFTK source it poses the question of should Coulomb be a particular case of this, which I'm not sure we want to bother with.

File: src/elements.jl (hunk 6 above)
Comment (gkemlin) [reply src/elements.jl:159]: I agree but I think we still need to have a method `charge_ionic(el::ElementCustomIonPotential)` to avoid errors when using this element ?

Comment (antoine-levitt) [reply src/elements.jl:159]: It should only get called if n_electrons is not specified.

**Comment by gkemlin** [comment]
I think that it's only 18 lines of code at the end of the `element.jl` file so having it implemented in the code or in an example doesn't change that much. In the former case, users only have to call `ElementCustomIonPotential(Z, pot_real, pot_fourier)` to implement their potential without thinking about this `Element` structure and it might be a good thing. But if you prefer the other way round, it's good for me. 
Keeping Coulomb as a particular `Element` might be better ? This custom type is more for "exploration" whereas Coulomb has real interest in practice.

File: src/elements.jl (hunk 6 above)
Comment (gkemlin) [reply src/elements.jl:159]: Even with `n_electrons = 1` in the example, I get the error `Implement charge_ionic` if I don't keep that line.

Comment (antoine-levitt) [reply src/elements.jl:159]: Can you post a backtrace?

Comment (gkemlin) [reply src/elements.jl:159]:
```
ERROR: LoadError: Implement charge_ionic
Stacktrace:
 [1] charge_ionic(::ElementCustomIonPotential) at /home/kemling/Documents/Nextcloud/CERMICS/codes/DFTK/DFTK.jl/src/elements.jl:16
 [2] n_elec_valence(::ElementCustomIonPotential) at /home/kemling/Documents/Nextcloud/CERMICS/codes/DFTK/DFTK.jl/src/elements.jl:19
 [3] macro expansion at /home/kemling/Documents/Nextcloud/CERMICS/codes/DFTK/DFTK.jl/src/guess_density.jl:20 [inlined]
 [4] (::DFTK.var"#476#477"{Float64,PlaneWaveBasis{Float64},Array{Pair,1}})() at /home/kemling/.julia/packages/TimerOutputs/dVnaw/src/TimerOutput.jl:237
 [5] timeit(::DFTK.var"#476#477"{Float64,PlaneWaveBasis{Float64},Array{Pair,1}}, ::TimerOutputs.TimerOutput, ::String) at /home/kemling/.julia/packages/TimerOutputs/dVnaw/src/TimerOutput.jl:285
 [6] guess_density at /home/kemling/.julia/packages/TimerOutputs/dVnaw/src/TimerOutput.jl:236 [inlined]
 [7] guess_density at /home/kemling/Documents/Nextcloud/CERMICS/codes/DFTK/DFTK.jl/src/guess_density.jl:12 [inlined]
 [8] top-level scope at /home/kemling/Documents/Nextcloud/CERMICS/codes/DFTK/DFTK.jl/examples/gross_pitaevskii_custom_potential.jl:45
 [9] top-level scope at REPL[4]:1
 [10] run_backend(::REPL.REPLBackend) at /home/kemling/.julia/packages/Revise/kqqw8/src/Revise.jl:1163
 [11] top-level scope at none:0
in expression starting at /home/kemling/Documents/Nextcloud/CERMICS/codes/DFTK/DFTK.jl/examples/gross_pitaevskii_custom_potential.jl:45
```
(Indeed, the timer pollutes the stacktrace :stuck_out_tongue: )

**Comment by antoine-levitt** [comment]
I think it's better outside: better to have a small code base that makes simple extensions easy than have all the extensions in the code.

File: src/elements.jl (hunk 6 above)
Comment (antoine-levitt) [reply src/elements.jl:159]: Ah yes, the guess density. We should probably switch to direction minimization here anyway... You can bypass this by giving an explicit density.

File: src/Model.jl [hunk 10]
```diff
@@ -118,8 +118,14 @@ function Model(lattice::AbstractMatrix{T};
...
-        # ask the terms if they break symmetry
-        compute_symmetry = !(any(breaks_symmetries, terms))
+        # check dimension : in 1D or 2D we might have symmetries but it's not
+        # detected by spglib so we ignore them
+        if d < 3
+            compute_symmetry = false
+        else
+            # ask the terms if they break symmetry
+            compute_symmetry = !(any(breaks_symmetries, terms))
+        end
```
mfherbst, line 128: I'd just do `!(any(breaks_symmetries, terms)) && d == 3`

docs/src/index.md
```diff
@@ -63,6 +63,7 @@ Pages = [
     "examples/cohen_bergstresser.md",
     "examples/arbitrary_floattype.md",
     "examples/custom_solvers.md",
+    "examples/gross_pitaevskii_custom_potential.md",
```
mfherbst, line 66: Since the file names are also the names of the websites, I'd suggest to keep the name in sync with the point of the example. So here the point is not GPE (that was before) it is the custom potential, so I'd just use `custom_potential.md`.

examples/gross_pitaevskii_custom_potential.jl
```diff
@@ -0,0 +1,84 @@
+# # Gross-Pitaevskii equation in one dimension with forces and custom potential
+
+# We solve the 1D Gross-Pitaevskii equation with a custom potential.
+# This is similar to the
+# previous example ([Gross-Pitaevskii equation in one dimension](@ref)) and we
```
mfherbst, line 5: it's not the previous one, so just say `similar to the example ...`

examples/gross_pitaevskii_custom_potential.jl
```diff
@@ -0,0 +1,84 @@
...
+# show how to define local potentials attached to atoms, which allows for
+# instance to compute forces.
+using DFTK
+using LinearAlgebra
+
+# First, we define a new element which represents an ion generating a custom
+# potential
```
mfherbst, line 12: ion sounds strange. I'd go for `nucleus`

examples/gross_pitaevskii_custom_potential.jl
```diff
@@ -0,0 +1,84 @@
...
+struct ElementCustomIonPotential <: DFTK.Element
+    pot_real::Function      # Real potential
+    pot_fourier::Function   # Fourier potential
+end
+
+# We need to extend two methods to access the real and Fourier forms of
+# the potential during the computations performed by DFTK
+function DFTK.local_potential_fourier(el::ElementCustomIonPotential, q::T) where {T <: Real}
```
mfherbst, line 20: This should also be `q::Real`. No reason to drag around the `T`.

examples/gross_pitaevskii_custom_potential.jl
```diff
@@ -0,0 +1,84 @@
...
+# This is similar to the
+# previous example ([Gross-Pitaevskii equation in one dimension](@ref)) and we
+# show how to define local potentials attached to atoms, which allows for
+# instance to compute forces.
+using DFTK
+using LinearAlgebra
+
+# First, we define a new element which represents an ion generating a custom
+# potential
+struct ElementCustomIonPotential <: DFTK.Element
```
mfherbst, line 13: I'd cut the `Ion`, just ElementCustomPotential?

examples/gross_pitaevskii_custom_potential.jl
```diff
@@ -0,0 +1,84 @@
...
+# break symmetry and get nonzero forces.
+x1 = 0.2
+x2 = 0.8;
+# We define the width of the Gaussian potential generated by one ion
+L = 0.5;
+# We set the potential in its real and Fourier forms
+pot_real(x) = exp(-(x/L)^2)
+pot_fourier(q::T) where {T <: Real} = exp(- (q*L)^2 / 4);
+# And finally we build the elements and set their positions in the `atoms`
+# array. Not that in this exemple, `pot_real` is not necessary as all application
```
mfherbst, line 43: typo example

examples/gross_pitaevskii_custom_potential.jl
```diff
@@ -0,0 +1,84 @@
...
+# of local potentials are done in the Fourier space.
+ion = ElementCustomIonPotential(pot_real, pot_fourier)
+atoms = [ion => [x1*[1,0,0], x2*[1,0,0]]];
+
+# Setup the energy terms and the model
```
mfherbst, line 48: Setup the Gross-Pitaevskii model

examples/gross_pitaevskii_custom_potential.jl
```diff
@@ -0,0 +1,84 @@
...
+terms = [Kinetic(),
+         AtomicLocal(),
+         PowerNonlinearity(C, α),
+]
+model = Model(lattice; atoms=atoms, n_electrons=n_electrons, terms=terms,
+              spin_polarization=:spinless);  # use "spinless electrons"
+
+# We discretize using a moderate Ecut and run a SCF algorithm to compute forces
+# afterwards. As there is no ionic charge associated to `ion` we have to specify
+# a starting density and we choose to start from the null density.
```
mfherbst, line 61: the null -> a zero

examples/gross_pitaevskii_custom_potential.jl
```diff
@@ -0,0 +1,84 @@
...
+Ecut = 500
+basis = PlaneWaveBasis(model, Ecut)
+ρ = zeros(complex(eltype(basis)), basis.fft_size)
+scfres = self_consistent_field(basis, tol=1e-8, ρ=from_fourier(basis, ρ))
+scfres.energies
+# Computing forces is then automatic
```
mfherbst, line 67: Computing the forces can then be done as usual:

examples/gross_pitaevskii_custom_potential.jl
```diff
@@ -0,0 +1,84 @@
...
+hcat(forces(scfres)...)
+
+# Extract the converged total local potential
+tot_local_pot = DFTK.total_local_potential(scfres.ham)[:,:,1][:];
```
mfherbst, line 71: spaces after the commas

mfherbst: also explain what you are doing here. It's a bit magic with all the `:`

antoine-levitt: This is weird btw. Should be [:, 1, 1] or [:]

gkemlin: Yep, true !

examples/custom_potential.jl
```diff
@@ -1,51 +1,50 @@
...
 x2 = 0.8;
-# We define the width of the Gaussian potential generated by one ion
+# We define the width of the Gaussian potential generated by one nucleus
 L = 0.5;
 # We set the potential in its real and Fourier forms
 pot_real(x) = exp(-(x/L)^2)
 pot_fourier(q::T) where {T <: Real} = exp(- (q*L)^2 / 4);
 # And finally we build the elements and set their positions in the `atoms`
-# array. Not that in this exemple, `pot_real` is not necessary as all application
+# array. Note that, in this example, `pot_real` is not required as all applications
```
mfherbst, line 42: No comma after that.

PR #252, opened 2020-06-06
Allow to disable taking timings

mfherbst:
This allows to disable taking timings in DFTK by defining the environment variable `DFTK_TIMING=0`. Also it disables timings is the parallelised regions, since otherwise undefined references and other funs due to race conditions in TimerOutputs are the result.

### DIFF
--- a/src/PlaneWaveBasis.jl
+++ b/src/PlaneWaveBasis.jl
@@ -262,12 +262,12 @@ end
 """
 In-place version of `G_to_r`.
 """
-@timing function G_to_r!(f_real::AbstractArray3, basis::PlaneWaveBasis,
-                         f_fourier::AbstractArray3) where {Tr, Tf}
+@timing_seq function G_to_r!(f_real::AbstractArray3, basis::PlaneWaveBasis,
+                             f_fourier::AbstractArray3) where {Tr, Tf}
     mul!(f_real, basis.opIFFT, f_fourier)
 end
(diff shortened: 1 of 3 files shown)


antoine-levitt:
Should we just disable it by default?

mfherbst:
Not sure. We could. I have no strong preference. I think in the current version it does not hurt and still gives a nice overview of the number of calls to the LOBPCG and the Hamiltonian. I think I would leave it on.

antoine-levitt:
Yeah but it doesn't do anything unless you print the timer. If it gives nicer stacktraces I'd be for deactivating it by default. The usecase for enabling it by default would be that you run a computation that you don't know you want to time, then it finishes and you want to time it. If it's a short computation, just enable the timer and run it again. If it's a long computation, you probably have a script to run it, in which case just enable the timer at the beginning of your script. I don't know, I don't feel hugely about it either, so could go both ways.
(CC @gkemlin btw)

gkemlin:
I have no strong preference either. It's good to have it enabled by default in case you're wondering about the timing after running the code, but if disabling it cleans the stacktrace (which I'm not sure, in case it is disabled, isn't it just an empty macro which will still pollutes the stacktrace ?), it might be better to deactivate it by default, it will make debugging easier.

mfherbst:
No the macro will not pollute the stacktrace since it does not transform the expression then. Ok lets go for enabled by default. The argument with the stacktraces tips the point for me.

PR #256, opened 2020-06-10
Two changes in examples

ssirajdine:
Found two errors when trying to run examples on a clean install: in silicon_scf_convergence.jl, norm was not in the scope, and in graphene.jl, the syntax for ElementPsp was missing "psp=". Somehow Git also includes all the previous changes in formatting from my previous PR.


examples/graphene.jl
```diff
@@ -13,7 +13,7 @@ Ecut = 15
 lattice = [4.659533614391621 -2.3297668071958104 0.0;
            0.0 4.035274479829987 0.0;
            0.0 0.0 15.117809010356462]
-C = ElementPsp(:C, load_psp("hgh/pbe/c-q4"))
+C = ElementPsp(:C, psp=load_psp("hgh/pbe/c-q4"))
```
antoine-levitt, line 16: use `;` instead of `,` here, it's more consistent for keyword arguments (it's not applied consistently in the code base but I try to apply it when I can)

antoine-levitt:
Thanks! For the git, you should have started from a fresh branch based on an up-to-date master. I've merged master into your branch, which fixes the problem (although Michael is not going to like it :-p)

ssirajdine:
Right! I will next time. Sorry for the mix-up.

antoine-levitt:
No problem, thanks for the fix!

mfherbst:
Indeed. I don't like it ;). Thanks for the fix ... again the stuff not covered by CI ... too bad.

antoine-levitt:
well I did a squash so it's fine in the end :-p

PR #260, opened 2020-06-11
Small tweak with miniter

mfherbst:
I've noticed in the calculations I currently do that often in the second SCF step only 1 LOBPCG per k-Point is performed, but in the third step many more (about 10 to 20) due to our adaptive tolerance selection kicking in. I think in order to easy making progress it makes sense to enforce a few more iterations for the first two steps. I've selected `2` because that's the typical average which is done in the first step anyway so this will not increase the number of applies a lot.

### DIFF
--- a/src/postprocess/chi0.jl
+++ b/src/postprocess/chi0.jl
@@ -108,7 +108,7 @@ function sternheimer_solver(basis, kpoint, Hk, ψk, ψnk, εnk, rhs, cgtol)
     # don't commute enough, and an oversolving of the linear
     # system can lead to spurious solutions
     rhs = Q(rhs)
-    δψnk = cg(J, rhs, Pl=FunctionPreconditioner(f_ldiv!), tol=cgtol / norm(rhs), verbose=true)
+    δψnk = cg(J, rhs, Pl=FunctionPreconditioner(f_ldiv!), tol=cgtol / norm(rhs), verbose=false)
     δψnk
 end
 
--- a/examples/dielectric.jl
+++ b/examples/dielectric.jl
@@ -29,6 +29,7 @@ end
 
 # A straightfoward Arnoldi eigensolver that diagonalizes the matrix at each step
 # This is more efficient than Arpack when `f` is very expensive
+println("Starting Arnoldi ...")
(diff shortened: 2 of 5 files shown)


antoine-levitt:
Really? We only do one or two iterations for the first SCF? I would have expected a lot more

mfherbst:
The average is around 4, but many k-points only need 2.

antoine-levitt:
That means that our heuristic for the determine_diagtol is pretty bad. Diagtol_max is pretty large at 0.1, it should be maybe 0.01?

antoine-levitt:
The 20 iterations is pretty ridiculous, not sure what was going on there...

mfherbst:
Well it's the first time the density turns towards convergence and thus diagtol is tightened. I agree let's go with a tighter maximal diagtol.

antoine-levitt:
Still 20 iterations is a bit overkill. That might just be the last bands having trouble converging, maybe try more bands

src/postprocess/chi0.jl
```diff
@@ -108,7 +108,7 @@ function sternheimer_solver(basis, kpoint, Hk, ψk, ψnk, εnk, rhs, cgtol)
     # don't commute enough, and an oversolving of the linear
     # system can lead to spurious solutions
     rhs = Q(rhs)
-    δψnk = cg(J, rhs, Pl=FunctionPreconditioner(f_ldiv!), tol=cgtol / norm(rhs), verbose=true)
+    δψnk = cg(J, rhs, Pl=FunctionPreconditioner(f_ldiv!), tol=cgtol / norm(rhs), verbose=false)
```
antoine-levitt, line 111: sorry about that, leftover debug

mfherbst: Thought so ;)

mfherbst: Converges pretty quick though 4 to 10 iterations. Not too bad.

antoine-levitt: Yeah, it's supposed to be pretty quick, especially if you've got a few empty bands computed

PR #261, opened 2020-06-11
Use floating-point numbers for k-Point coordinates

antoine-levitt:


src/PlaneWaveBasis.jl
```diff
@@ -94,13 +94,22 @@ Base.eltype(basis::PlaneWaveBasis{T}) where {T} = T
 
     kpoints = Vector{Kpoint}()
     for k in kcoords
-        energy(q) = sum(abs2, model.recip_lattice * q) / 2
-        pairs = [(i, G) for (i, G) in enumerate(G_vectors(fft_size)) if energy(k + G) ≤ Ecut]
-
-        mapping = first.(pairs)
+        k = Vec3{T}(k) # rationals are sloooow
+        mapping = Vector{Int}()
```
mfherbst, line 98: `mapping = Int[]` ?

src/PlaneWaveBasis.jl
```diff
@@ -94,13 +94,22 @@ Base.eltype(basis::PlaneWaveBasis{T}) where {T} = T
...
+        Gvecs_k = Vector{Vec3{Int}}()
```
mfherbst, line 99: `Gvecs_k = Vec3{Int}[]`

src/terms/nonlocal.jl
```diff
@@ -182,13 +180,15 @@ function build_projection_vectors_(basis::PlaneWaveBasis, atoms, kpt::Kpoint)
             # k+G in this formula can also be G, this only changes an unimportant phase factor
             structure_factors = [cis(-2T(π)*dot(kpt.coordinate + G, r)) for G in G_vectors(kpt)]
             for iproj = 1:count_n_proj_(psp)
-                proj_vectors[:, offset+iproj] = structure_factors .* form_factors[:, iproj]
+                @views proj_vectors[:, offset+iproj] .= structure_factors .*
+                                                        form_factors[:, iproj] ./ 
+                                                        sqrt(model.unit_cell_volume)
```
mfherbst, line 185:
```
                @views proj_vectors[:, offset+iproj] .= (structure_factors
                                                         .* form_factors[:, iproj] 
                                                         ./ sqrt(model.unit_cell_volume))
```
Also why not put the `@views` in front of the `for`? The line above also does a copy.

src/PlaneWaveBasis.jl
mfherbst: Also space before `#`

antoine-levitt: Oh nice

src/terms/nonlocal.jl
antoine-levitt: I don't think so?

mfherbst: ah sorry ... stupid me ... that's the line you removed. All is well ;). I think for such cases I would actually write `.* @view form_factors[:, iproj]` to avoid the deep indention, but I'm happy to leave it as is.

PR #263, opened 2020-06-16
Unify kernel interfaces to RealFourierArrays

mfherbst:


src/RealFourierArray.jl
```diff
@@ -27,7 +27,9 @@ mutable struct RealFourierArray{Treal <: Real, T <: Union{Treal, Complex{Treal}}
     _state::RFA_state
 end
 # Type of the real part
-Base.eltype(A::RealFourierArray{Treal, T}) where {Treal, T} = T
+Base.eltype(::RealFourierArray{Treal, T}) where {Treal, T} = T
+Base.size(array::RealFourierArray) = size(array._real)
+Base.size(array::RealFourierArray, i) = size(array._real, i)
```
antoine-levitt, line 32: I'm not sure we should do this (the size), since it's ambiguous whether that refers to the real or fourier part

src/postprocess/chi0.jl
```diff
@@ -204,5 +203,5 @@ function apply_χ0(ham, δV, ψ, εF, eigenvalues;
         δρ .+= ldos .* dot(ldos, δV) .* dVol ./ dos
     end
 
-    δρ .* normδV
+    from_real(basis, δρ .* normδV, assume_real=true)
```
antoine-levitt, line 206: `;` for the kwarg. I'm not sure our interface is great here... Maybe we should make assume_real be true by default? Anything that uses RFA is real, right? Only the orbitals are complex, and they are not RFA. Of course we should also optimize by cutting the number of G vectors in half, but I'm too lazy to do that

src/RealFourierArray.jl
mfherbst: Doesn't matter since it's the same.

src/postprocess/chi0.jl
mfherbst: I'm ok with going for real by default if we keep the flag to switch off if needed.

src/RealFourierArray.jl
antoine-levitt: Ah right.

src/postprocess/chi0.jl
antoine-levitt: Why?

antoine-levitt: Making it real-only would emphasize the difference between this (for densities/potentials) and wavefunctions

antoine-levitt:
@gkemlin, @tzsuzsi : you might need some adjustment in your code after this: basically, don't use `assume_real=true` if you were doing that before.

src/RealFourierArray.jl
```diff
@@ -12,46 +13,41 @@ import Base.eltype, Base.size, Base.length
...
+Base.eltype(::RealFourierArray{T}) where {T} = T
 Base.size(array::RealFourierArray) = size(array._real)
 Base.size(array::RealFourierArray, i) = size(array._real, i)
 Base.length(array::RealFourierArray) = length(array._real)
 
 # Constructors
-RealFourierArray(basis, real, fourier) = RealFourierArray(basis, real, fourier, 2)
-function RealFourierArray(basis::PlaneWaveBasis{T}; iscomplex=false) where {T}
-    from_real(basis, zeros(iscomplex ? complex(T) : T, basis.fft_size...))
+RealFourierArray(basis, real, fourier) = RealFourierArray(basis, real, fourier, RFA_both)
```
mfherbst, line 34: I think you need a `{T}` here.

antoine-levitt: ? why?

test/chi0.jl
```diff
@@ -44,7 +44,7 @@ include("testcases.jl")
                                        use_symmetry=use_symmetry)
                 energies, ham = energy_hamiltonian(basis, nothing, nothing; ρ=ρ0)
                 ρ2 = DFTK.next_density(ham, tol=tol, eigensolver=diag_full, n_bands=n_bands).ρ
-                diff_findiff = (ρ2 - ρ1) / ε
+                diff_findiff = from_real(basis, ρ2.real - ρ1.real) / ε
```
mfherbst, line 47: Why? This should work otherwise as well if we implemented things properly.

antoine-levitt: It's failing the assert that both are on the same basis. I guess we could remove the assert?

antoine-levitt: yeah let's trust the user on this one

mfherbst: Ah ok. That I did not have in mind. No I'd rather keep the assert. Leave it as is.

src/RealFourierArray.jl
mfherbst: How else would it resolve the type of the RealFourierArray. Maybe I'm wrong, but I have in mind in such cases you have to help Julia a little.

test/chi0.jl
antoine-levitt: Too late I removed it :-p I think it's perfectly legitimate to want to eg compare densities between two systems with different atomic positions

mfherbst: all right, whatever, not sure I care that much

src/RealFourierArray.jl
antoine-levitt: well from the types of its arguments. There's a single match so it'll just take that. Anyway it didn't seem to complain...

mfherbst: true. So it seems it works.

mfherbst:
@antoine-levitt I don't like the hackish solution of my previous commit. Maybe we should properly implement broadcasting on the real part of the RFA, but I had a brief look just now and it takes a bit to understand which triggers one has to push, so I did not bother right now.

antoine-levitt:
Indeed. In that case just have apply_kernel return `nothing` and check for it, the same way we do with forces

mfherbst:
Good idea. I'll do that instead.

PR #264, opened 2020-06-19
Printing of SCF

antoine-levitt:


src/scf/self_consistent_field.jl
```diff
@@ -125,6 +127,8 @@ Solve the Kohn-Sham equations with a SCF algorithm, starting at ρ.
     n_iter = 0
     energies = nothing
     ham = nothing
+    nonconsistent_energies = nothing
+    previous_energies = nothing
```
mfherbst, line 131: I think this should not live here, much rather the convergence / printing functors should cache that by themselves.

src/scf/self_consistent_field.jl
```diff
@@ -170,10 +177,10 @@ Solve the Kohn-Sham equations with a SCF algorithm, starting at ρ.
         ρnext.real
     end
 
-    fpres = solver(fixpoint_map, ρout.real, maxiter; tol=min(10eps(T), tol / 10))
-    # Tolerance is only dummy here: Convergence is flagged by is_converged
+    # Tolerance and maxiter are only dummy here: Convergence is flagged by is_converged
     # inside the fixpoint_map. Also we do not use the return value of fpres but rather the
     # one that got updated by fixpoint_map
+    fpres = solver(fixpoint_map, ρout.real, typemax(Int); tol=0)
```
mfherbst, line 183: Sure tol=0 is a good idea. I tried that at some point and this caused numerical problems in *some* fixed point solvers, so I changed it to the value it has now. Have you tested with the crop and the custom anderson and the custom simple we have?

src/scf/self_consistent_field.jl
```diff
@@ -29,14 +29,17 @@ function next_density(ham::Hamiltonian;
...
 
 function scf_default_callback(info)
-    E = info.energies === nothing ? Inf : sum(values(info.energies))
-    res = norm(info.ρout.fourier - info.ρin.fourier)
     if info.n_iter == 1
-        label = haskey(info.energies, "Entropy") ? "Free energy" : "Energy"
-        @printf "Iter   %-15s    ρout-ρin\n" label
-        @printf "----   %-15s    --------\n" "-"^length(label)
+        E_label = haskey(info.energies, "Entropy") ? "Free energy" : "Energy"
+        @printf "n     %-12s      ρout-ρin   Eₙ₋₁-Eₙ     diag\n" E_label
```
mfherbst, line 34: `diag` or `diag #it`? No strong opinion here.

mfherbst: Also diag is off by one space and for some reason I would indent the "Energy" label by one space (to not align with the - but the number) ... of course debatable. Also the capitalisation of the headers should be consistent (I'd just lower-case energy and entropy)

docs/src/guide/tutorial.jl
```diff
@@ -45,10 +45,6 @@ basis = PlaneWaveBasis(model, Ecut; kgrid=kgrid)
 ## 3. Run the SCF procedure to obtain the ground state
 scfres = self_consistent_field(basis, tol=1e-8);
 
-# Note that DFTK by default applies the convergence tolerance `tol`
-# to the energy difference, so that the norm in the density difference
-# is not yet converged to 8 digits.
-#
```
mfherbst, line 51: Why do you want to remove that?

antoine-levitt: Because it's clearer now that there is the difference in energy why this stops

mfherbst: true. :+1:

src/scf/self_consistent_field.jl
antoine-levitt: agree

src/scf/self_consistent_field.jl
antoine-levitt: Well I'll see with the CI. We should just fix those solvers in that case

mfherbst: Seems good. I'll merge it.

PR #265, opened 2020-06-20
Move to spglib_jll

louisponet:
All tests work. Not sure if there are the same amount of warnings as on the master branch tests.

### DIFF
--- a/src/bzmesh.jl
+++ b/src/bzmesh.jl
@@ -65,10 +65,9 @@ function bzmesh_ir_wedge(kgrid_size, symops; kshift=[0, 0, 0])
     # Give the remaining symmetries to spglib to compute an irreducible k-Point mesh
     # TODO implement time-reversal symmetry and turn the flag to true
     is_shift = Int.(2 * kshift)
-    Stildes = [S for (S, τ) in symops]
-    spg_rotations = cat(Stildes..., dims=3)
+    Stildes = [S' for (S, τ) in symops]
     _, mapping, grid = spglib_get_stabilized_reciprocal_mesh(
-        kgrid_size, spg_rotations, is_shift=is_shift, is_time_reversal=false
+        kgrid_size, Stildes, is_shift=is_shift, is_time_reversal=false
     )
     # Convert irreducible k-Points to DFTK conventions
     kgrid_size = Vec3{Int}(kgrid_size)
--- a/src/bzmesh.jl
+++ b/src/bzmesh.jl
@@ -65,14 +65,14 @@ function bzmesh_ir_wedge(kgrid_size, symops; kshift=[0, 0, 0])
     # Give the remaining symmetries to spglib to compute an irreducible k-Point mesh
     # TODO implement time-reversal symmetry and turn the flag to true
     is_shift = Int.(2 * kshift)
-    Stildes = [Cint.(S) for (S, τ) in symops]
+    Stildes = [S for (S, τ) in symops]
     spg_rotations = cat(Stildes..., dims=3)
-    num_kpts, mapping, grid = spglib_get_stabilized_reciprocal_mesh(
+    _, mapping, grid = spglib_get_stabilized_reciprocal_mesh(
         kgrid_size, spg_rotations, is_shift=is_shift, is_time_reversal=false
     )
     # Convert irreducible k-Points to DFTK conventions
     kgrid_size = Vec3{Int}(kgrid_size)
-    kirreds = [(kshift .+ Vec3{Int}(grid[:, ik + 1])) .// kgrid_size
+    kirreds = [(kshift .+ grid[ik + 1]) .// kgrid_size
                for ik in unique(mapping)]
     kirreds = normalize_kpoint_coordinate.(kirreds)
 
@@ -94,7 +94,7 @@ function bzmesh_ir_wedge(kgrid_size, symops; kshift=[0, 0, 0])
     for (ik, k) in enumerate(kirreds)
         ksymops[ik] = Vector{SymOp}()
         for ired in k_all_reducible[ik]
-            kred = (kshift .+ Vec3(grid[:, ired])) .// kgrid_size
+            kred = (kshift .+ grid[ired]) .// kgrid_size
 
             # Note that this relies on the identity coming up first in symops
             isym = findfirst(symops) do symop
--- a/src/bzmesh.jl
+++ b/src/bzmesh.jl
@@ -65,15 +65,14 @@ function bzmesh_ir_wedge(kgrid_size, symops; kshift=[0, 0, 0])
     # Give the remaining symmetries to spglib to compute an irreducible k-Point mesh
     # TODO implement time-reversal symmetry and turn the flag to true
     is_shift = Int.(2 * kshift)
-    Stildes = [S' for (S, τ) in symops]
-    spg_rotations = permutedims(cat(Stildes..., dims=3), (3, 1, 2))
-    mapping, grid = pyimport("spglib").get_stabilized_reciprocal_mesh(
+    Stildes = [Cint.(S) for (S, τ) in symops]
+    spg_rotations = cat(Stildes..., dims=3)
+    num_kpts, mapping, grid = spglib_get_stabilized_reciprocal_mesh(
         kgrid_size, spg_rotations, is_shift=is_shift, is_time_reversal=false
     )
-
     # Convert irreducible k-Points to DFTK conventions
     kgrid_size = Vec3{Int}(kgrid_size)
-    kirreds = [(kshift .+ Vec3{Int}(grid[ik + 1, :])) .// kgrid_size
+    kirreds = [(kshift .+ Vec3{Int}(grid[:, ik + 1])) .// kgrid_size
                for ik in unique(mapping)]
     kirreds = normalize_kpoint_coordinate.(kirreds)
 
@@ -95,7 +94,7 @@ function bzmesh_ir_wedge(kgrid_size, symops; kshift=[0, 0, 0])
     for (ik, k) in enumerate(kirreds)
         ksymops[ik] = Vector{SymOp}()
         for ired in k_all_reducible[ik]
-            kred = (kshift .+ Vec3(grid[ired, :])) .// kgrid_size
+            kred = (kshift .+ Vec3(grid[:, ired])) .// kgrid_size
 
             # Note that this relies on the identity coming up first in symops
             isym = findfirst(symops) do symop
--- a/src/external/spglib.jl
+++ b/src/external/spglib.jl
@@ -70,6 +70,7 @@ spglib_atoms(atoms) = first(spglib_atommapping(atoms))
               "uniform BZ mesh or disable symmetries. Spglib reported : " * err_message)
     end
 
+    # Note: Transposes are performed to convert between spglib row-major to julia column-major
     Stildes = [Mat3{Int}(spg_rotations[:, :, i])' for i in 1:spg_n_ops]
     τtildes = [rationalize.(Vec3{Float64}(spg_translations[:, i]), tol=tol_symmetry) for i in 1:spg_n_ops]
     
@@ -127,6 +128,7 @@ function spglib_get_stabilized_reciprocal_mesh(kgrid_size, rotations::Vector;
                                                is_time_reversal=false,
                                                qpoints::Vector{Vec3{Float64}}=[Vec3(0.0, 0.0, 0.0)],
                                                isdense=false)
+    # Note: Transposes are performed to convert from julia column-major to spglib row-major
     spg_rotations = cat([Cint.(S)' for S in rotations]..., dims=3)
     nkpt = prod(kgrid_size)
     mapping = Vector{Cint}(undef, nkpt)
--- a/src/external/spglib.jl
+++ b/src/external/spglib.jl
@@ -34,7 +34,7 @@ end
 spglib_atoms(atoms) = first(spglib_atommapping(atoms))
 
 
-@timing function spglib_get_symmetry(lattice, atoms; tol_symmetry=1e-5, max_symops=384)
+@timing function spglib_get_symmetry(lattice, atoms; tol_symmetry=1e-5)
     lattice = Matrix{Float64}(lattice)  # spglib operates in double precision
 
     if isempty(atoms)
@@ -44,26 +44,29 @@ spglib_atoms(atoms) = first(spglib_atommapping(atoms))
     end
 
     # Ask spglib for symmetry operations and for irreducible mesh
-    spg_rotations    = Array{Cint}(undef, 3, 3, max_symops)
-    spg_translations = Array{Cdouble}(undef, 3, max_symops)
     spg_positions, spg_numbers = spglib_atoms(atoms)
-    
-    spg_num_ops = ccall((:spg_get_symmetry, SPGLIB), Cint,
-       (Ptr{Cint}, Ptr{Cdouble}, Cint, Ptr{Cdouble}, Ptr{Cdouble}, Ptr{Cint}, Cint, Cdouble),
-       spg_rotations, spg_translations, Cint(max_symops), lattice, spg_positions, spg_numbers,
-       Cint(length(spg_numbers)), tol_symmetry)
-       
-    # Note: spglib_jll returns transposed rotation matrix vs the python spglib interface
 
+    spg_n_ops = ccall((:spg_get_multiplicity, SPGLIB), Cint,
+        (Ptr{Cdouble}, Ptr{Cdouble}, Ptr{Cint}, Cint, Cdouble),
+        lattice, spg_positions, spg_numbers, Cint(length(spg_numbers)), tol_symmetry)
+    
+    spg_rotations    = Array{Cint}(undef, 3, 3, spg_n_ops)
+    spg_translations = Array{Cdouble}(undef, 3, spg_n_ops)
+    
+    ccall((:spg_get_symmetry, SPGLIB), Cint,
+         (Ptr{Cint}, Ptr{Cdouble}, Cint, Ptr{Cdouble}, Ptr{Cdouble}, Ptr{Cint}, Cint, Cdouble),
+         spg_rotations, spg_translations, spg_n_ops, lattice, spg_positions, spg_numbers,
+         Cint(length(spg_numbers)), tol_symmetry)
+  
     # If spglib does not find symmetries give an error
-    if spg_num_ops == 0
+    if spg_n_ops == 0
         err_message = spglib_get_error_message()
         error("spglib failed to get the symmetries. Check your lattice, use a " *
               "uniform BZ mesh or disable symmetries. Spglib reported : " * err_message)
     end
 
-    Stildes = [Mat3{Int}(spg_rotations[:, :, i])' for i in 1:spg_num_ops]
-    τtildes = [rationalize.(Vec3{Float64}(spg_translations[:, i]), tol=tol_symmetry) for i in 1:spg_num_ops]
+    Stildes = [Mat3{Int}(spg_rotations[:, :, i])' for i in 1:spg_n_ops]
+    τtildes = [rationalize.(Vec3{Float64}(spg_translations[:, i]), tol=tol_symmetry) for i in 1:spg_n_ops]
     
     # Checks: (A Stilde A^{-1}) is unitary
     for Stilde in Stildes
@@ -96,7 +99,6 @@ end
 
 function spglib_standardize_cell(lattice::AbstractArray{T}, atoms; correct_symmetry=true,
                                  primitive=false, tol_symmetry=1e-5) where {T}
-                                 
     # Convert lattice and atoms to spglib and keep the mapping between our atoms
     spg_lattice = Matrix{Float64}(lattice) 
     # and spglibs atoms
@@ -109,9 +111,6 @@ function spglib_standardize_cell(lattice::AbstractArray{T}, atoms; correct_symme
       spg_lattice, spg_positions, spg_numbers, length(spg_numbers), Cint(primitive),
       Cint(!correct_symmetry), tol_symmetry)
 
-    # Note: In the python interface of spglib the lattice vectors
-    #       are given in rows, but DFTK uses columns
-    #       For future reference: The C interface spglib also uses columns.
     newatoms = [(atommapping[iatom]
                  => T.(spg_positions[findall(isequal(iatom), spg_numbers), :]))
                 for iatom in unique(spg_numbers)]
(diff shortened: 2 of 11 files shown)


src/external/spglib.jl
```diff
@@ -1,15 +1,31 @@
...
+function spglib_get_symmetry_(unit_cell; maxsize = 52, symprec = 1e-5)
+    rotations    = Array{Cint}(undef, 3, 3, maxsize)
+    translations = Array{Cdouble}(undef, 3, maxsize)
+    cell, positions, indices = unit_cell
+    
+    num_ops = ccall((:spg_get_symmetry, SPGLIB), Cint,
+      (Ptr{Cint}, Ptr{Cdouble}, Cint, Ptr{Cdouble}, Ptr{Cdouble}, Ptr{Cint}, Cint, Cdouble),
+       rotations, translations, Cint(maxsize), cell, positions, indices, Cint(length(indices)), symprec)
+    #Note: spglib_jll returns transposed rotation matrix vs the python spglib interface
+    return num_ops, [Vec3(translations[:, i]) for i in 1:num_ops], [Mat3(rotations[:, :, i])' for i in 1:num_ops]
```
mfherbst, line 18: Use Vec3{Float64} and Mat3{Int} to ensure the c-types are converted to Julia types. Also this lines is too long.

src/external/spglib.jl
```diff
@@ -1,15 +1,31 @@
...
+end
+
+function spglib_get_symmetry_(unit_cell; maxsize = 52, symprec = 1e-5)
+    rotations    = Array{Cint}(undef, 3, 3, maxsize)
+    translations = Array{Cdouble}(undef, 3, maxsize)
+    cell, positions, indices = unit_cell
+    
+    num_ops = ccall((:spg_get_symmetry, SPGLIB), Cint,
+      (Ptr{Cint}, Ptr{Cdouble}, Cint, Ptr{Cdouble}, Ptr{Cdouble}, Ptr{Cint}, Cint, Cdouble),
+       rotations, translations, Cint(maxsize), cell, positions, indices, Cint(length(indices)), symprec)
```
mfherbst, line 16: Add more line breaks please.

src/external/spglib.jl
```diff
@@ -88,26 +99,49 @@ spglib_cell(lattice, atoms) = first(spglib_cell_atommapping_(lattice, atoms))
     Stildes, τtildes
 end
 
+function spglib_standardize_cell_(unit_cell, to_primitive, no_idealize, symprec)
+    cell, positions, indices = copy.(unit_cell)
+    num_atoms = ccall((:spg_standardize_cell, SPGLIB), Cint,
+      (Ptr{Cdouble}, Ptr{Cdouble}, Ptr{Cint}, Cint, Cint, Cint, Cdouble),
+       cell, positions, indices, length(indices), Cint(to_primitive), Cint(no_idealize), symprec)
```
mfherbst, line 106: More line breaks please.

src/external/spglib.jl
```diff
@@ -88,26 +99,49 @@ spglib_cell(lattice, atoms) = first(spglib_cell_atommapping_(lattice, atoms))
...
+function spglib_get_stabilized_reciprocal_mesh(kgrid_size, rotations::AbstractArray{Int32};
+                                               is_shift = Vec3(0, 0, 0),
+                                               is_time_reversal = false,
+                                               qpoints = [Vec3(0.0, 0.0, 0.0)],
+                                               isdense = false)
+    nkpt = prod(kgrid_size)
+    mapping = Vector{Cint}(undef, nkpt)
+    grid_address = Matrix{Cint}(undef, 3, nkpt)
+    qpoints = eltype(qpoints) == Real ? [qpoints] : qpoints #Note: Done similarly to the python spglib
+    # numrot = rotations isa Vector ? length(rotations) : size(rotations, 3)
```
mfherbst, line 141: Debug comment.

src/external/spglib.jl
```diff
@@ -88,26 +99,49 @@ spglib_cell(lattice, atoms) = first(spglib_cell_atommapping_(lattice, atoms))
...
+    numrot = size(rotations, 3)
+    num_kpts = ccall((:spg_get_stabilized_reciprocal_mesh, SPGLIB), Cint,
+      (Ptr{Cint}, Ptr{Cint}, Ptr{Cint}, Ptr{Cint}, Cint, Cint, Ptr{Cint}, Cint, Ptr{Cdouble}),
+       grid_address, mapping, [Cint.(kgrid_size)...], [Cint.(is_shift)...], Cint(is_time_reversal), Cint(numrot), rotations, Cint(length(qpoints)), qpoints)
```
mfherbst, line 145: Too long line.

src/external/spglib.jl
```diff
@@ -88,26 +99,49 @@ spglib_cell(lattice, atoms) = first(spglib_cell_atommapping_(lattice, atoms))
     Stildes, τtildes
 end
 
+function spglib_standardize_cell_(unit_cell, to_primitive, no_idealize, symprec)
```
mfherbst, line 102: I don't like that there is a function with and without underscore. Can't you just merge them?

test/bzmesh.jl
```diff
@@ -7,16 +7,16 @@ include("testcases.jl")
...
 @testset "bzmesh_uniform agrees with spglib" begin
-    spglib = pyimport_conda("spglib", "spglib")
 
     function test_against_spglib(kgrid_size; kshift=[0, 0, 0])
         kgrid_size = Vec3(kgrid_size)
-        identity = [reshape(Mat3{Int}(I), 1, 3, 3)]
+        identity = collect(reshape(Matrix{Cint}(I, 3, 3), 1, 3, 3))
         is_shift = ifelse.(kshift .== 0, 0, 1)
-        _, grid = spglib.get_stabilized_reciprocal_mesh(kgrid_size, identity,
+        num_kpts, _, grid = DFTK.spglib_get_stabilized_reciprocal_mesh(kgrid_size, identity,
```
mfherbst, line 15: If you don't use `num_kpts` please also use an underscore for that arg.

test/bzmesh.jl
```diff
@@ -7,16 +7,16 @@ include("testcases.jl")
 
 
 @testset "bzmesh_uniform agrees with spglib" begin
-    spglib = pyimport_conda("spglib", "spglib")
 
     function test_against_spglib(kgrid_size; kshift=[0, 0, 0])
         kgrid_size = Vec3(kgrid_size)
-        identity = [reshape(Mat3{Int}(I), 1, 3, 3)]
+        identity = collect(reshape(Matrix{Cint}(I, 3, 3), 1, 3, 3))
```
mfherbst, line 13: I feel a bit uneasy that c types appear here. Maybe you should add a higher-level way to call this function without C types?

test/bzmesh.jl
```diff
@@ -7,16 +7,16 @@ include("testcases.jl")
 
 
 @testset "bzmesh_uniform agrees with spglib" begin
-    spglib = pyimport_conda("spglib", "spglib")
 
     function test_against_spglib(kgrid_size; kshift=[0, 0, 0])
         kgrid_size = Vec3(kgrid_size)
-        identity = [reshape(Mat3{Int}(I), 1, 3, 3)]
+        identity = collect(reshape(Matrix{Int64}(I, 3, 3), 3, 3, 1))
```
mfherbst, line 13: Why do you need the `collect` here? Reshape should return an array, I think.

test/bzmesh.jl
mfherbst: Actually I just had an idea about this: Basically what this tries to test is that we agree with spglib, so from that point of view the best way to test would be (contrary to what I said before) to directly do a ccall into spglib_jll from the test and bypass as much of the extra icing DFTK spglib.jl adds on top. You're call if you think that's too much effort to do or not ...

mfherbst: Also the line starts to become a bit long ;)

src/external/spglib.jl
```diff
@@ -1,40 +1,40 @@
...
+    spg_numbers = Vector{Cint}(undef, n_attypes)
+    spg_positions = Matrix{Cdouble}(undef, 3, n_attypes)
 
     offset = 0
     nextnumber = 1
     atommapping = Dict{Int, Any}()
-    for (iatom, (type, positions)) in enumerate(atoms)
-        atommapping[nextnumber] = type
+    for (iatom, (typ, positions)) in enumerate(atoms)
+        atommapping[nextnumber] = typ
```
mfherbst, line 22: Just out of curiousity: Any reason you removed the `e`?

src/external/spglib.jl
```diff
@@ -1,40 +1,40 @@
...
-    (lattice', spg_positions, spg_numbers), atommapping
+    (spg_positions, spg_numbers), atommapping
 end
-spglib_cell(lattice, atoms) = first(spglib_cell_atommapping_(lattice, atoms))
+spglib_atoms(atoms) = first(spglib_atommapping(atoms))
 
 
-@timing function spglib_get_symmetry(lattice, atoms; tol_symmetry=1e-5)
-    spglib = pyimport("spglib")
+@timing function spglib_get_symmetry(lattice, atoms; tol_symmetry=1e-5, max_symops=384)
```
mfherbst, line 37: This `max_symops` is pretty much a magic number, so it should be documented where it comes from. If I recall correctly this is used for some pointer magic inside spglib, so if you get this wrong DFTK could segfault, which we definitely don't want. Is there no way to ask the library for this? I mean it might change in future versions or not?

src/external/spglib.jl
```diff
@@ -44,21 +44,27 @@ spglib_cell(lattice, atoms) = first(spglib_cell_atommapping_(lattice, atoms))
     end
 
     # Ask spglib for symmetry operations and for irreducible mesh
-    spg_symops = spglib.get_symmetry(spglib_cell(lattice, atoms),
-                                     symprec=tol_symmetry)
+    spg_rotations    = Array{Cint}(undef, 3, 3, max_symops)
+    spg_translations = Array{Cdouble}(undef, 3, max_symops)
+    spg_positions, spg_numbers = spglib_atoms(atoms)
```
mfherbst, line 49: Awesome! That's exactly how I imagined it! :+1:

src/external/spglib.jl
```diff
@@ -44,21 +44,27 @@ spglib_cell(lattice, atoms) = first(spglib_cell_atommapping_(lattice, atoms))
...
               "uniform BZ mesh or disable symmetries. Spglib reported : " * err_message)
     end
 
-    Stildes = [St for St in eachslice(spg_symops["rotations"]; dims=1)]
-    τtildes = [rationalize.(τt, tol=tol_symmetry)
-               for τt in eachslice(spg_symops["translations"]; dims=1)]
-    @assert length(τtildes) == length(Stildes)
-
+    Stildes = [Mat3{Int}(spg_rotations[:, :, i])' for i in 1:spg_num_ops]
+    τtildes = [rationalize.(Vec3{Float64}(spg_translations[:, i]), tol=tol_symmetry) for i in 1:spg_num_ops]
```
mfherbst, line 66: Line getting a bit long.

src/external/spglib.jl
```diff
@@ -44,21 +44,27 @@ spglib_cell(lattice, atoms) = first(spglib_cell_atommapping_(lattice, atoms))
...
+       (Ptr{Cint}, Ptr{Cdouble}, Cint, Ptr{Cdouble}, Ptr{Cdouble}, Ptr{Cint}, Cint, Cdouble),
+       spg_rotations, spg_translations, Cint(max_symops), lattice, spg_positions, spg_numbers,
+       Cint(length(spg_numbers)), tol_symmetry)
+       
+    # Note: spglib_jll returns transposed rotation matrix vs the python spglib interface
 
     # If spglib does not find symmetries give an error
-    if spg_symops === nothing
-        err_message=spglib.get_error_message()
+    if spg_num_ops == 0
```
mfherbst, line 59: We usually use `n_` to abbreviate "Number of"

src/external/spglib.jl
```diff
@@ -88,26 +94,47 @@ spglib_cell(lattice, atoms) = first(spglib_cell_atommapping_(lattice, atoms))
...
-    newlattice, newatoms
+    spg_lattice, newatoms
+end
+
+function spglib_get_stabilized_reciprocal_mesh(kgrid_size, rotations;
+                                               is_shift=Vec3(0, 0, 0),
+                                               is_time_reversal=false,
+                                               qpoints=[Vec3(0.0, 0.0, 0.0)],
+                                               isdense=false)
+
```
mfherbst, line 126: No empty line after function definition please.

src/external/spglib.jl
```diff
@@ -88,26 +94,47 @@ spglib_cell(lattice, atoms) = first(spglib_cell_atommapping_(lattice, atoms))
...
+    rotations = eltype(rotations) == Int32 ? rotations : Cint.(rotations)
+    nkpt = prod(kgrid_size)
+    mapping = Vector{Cint}(undef, nkpt)
+    grid_address = Matrix{Cint}(undef, 3, nkpt)
+    qpoints = eltype(qpoints) == Real ? [qpoints] : qpoints # Note: Done similarly to the python spglib
```
mfherbst, line 131: I'd put the comment on the preceeding line. If you reference to the python implementation, could you add a link to the file on github (use a commit hash which is static and not a branch name when you deep-link)?

mfherbst:
Ah also could you rebase against master at some point to prepare for the merge @louisponet, thanks!

src/bzmesh.jl
```diff
@@ -65,15 +65,14 @@ function bzmesh_ir_wedge(kgrid_size, symops; kshift=[0, 0, 0])
     # Give the remaining symmetries to spglib to compute an irreducible k-Point mesh
     # TODO implement time-reversal symmetry and turn the flag to true
     is_shift = Int.(2 * kshift)
-    Stildes = [S' for (S, τ) in symops]
-    spg_rotations = permutedims(cat(Stildes..., dims=3), (3, 1, 2))
-    mapping, grid = pyimport("spglib").get_stabilized_reciprocal_mesh(
+    Stildes = [S for (S, τ) in symops]
```
antoine-levitt, line 68: I'm confused by this: symops has rotations as S, and Stildes are transposes of S, so this line runs counter to the rest of the code. Is this done because C spglib is row major? In this case I think it'd be better if `spglib_get_stabilized_reciprocal_mesh` took column-major arrays and did the conversion internally (ie transpose twice)

src/external/spglib.jl
```diff
@@ -1,40 +1,40 @@
-using PyCall
 # Routines for interaction with spglib
+const SPGLIB = spglib_jll.libsymspg
```
antoine-levitt, line 2: Can you add a bit of context here, eg if spglib is row-major or column-major?

src/external/spglib.jl
antoine-levitt: If it's internal it shouldn't be a kwarg and just set inside this file. Eg `const SPGLIB_MAX_SYMOPS` at the beginning of the file

src/external/spglib.jl
```diff
@@ -88,26 +94,47 @@ spglib_cell(lattice, atoms) = first(spglib_cell_atommapping_(lattice, atoms))
...
-                                  no_idealize=!correct_symmetry, symprec=tol_symmetry)
-    spg_lattice, spg_scaled_positions, spg_numbers = res
+    num_atoms = ccall((:spg_standardize_cell, SPGLIB), Cint,
+      (Ptr{Cdouble}, Ptr{Cdouble}, Ptr{Cint}, Cint, Cint, Cint, Cdouble),
+      spg_lattice, spg_positions, spg_numbers, length(spg_numbers), Cint(primitive),
+      Cint(!correct_symmetry), tol_symmetry)
 
     # Note: In the python interface of spglib the lattice vectors
     #       are given in rows, but DFTK uses columns
     #       For future reference: The C interface spglib also uses columns.
```
antoine-levitt, line 114: Can remove this?

src/external/spglib.jl
```diff
@@ -44,21 +44,27 @@ spglib_cell(lattice, atoms) = first(spglib_cell_atommapping_(lattice, atoms))
...
+    spg_rotations    = Array{Cint}(undef, 3, 3, max_symops)
+    spg_translations = Array{Cdouble}(undef, 3, max_symops)
+    spg_positions, spg_numbers = spglib_atoms(atoms)
+    
+    spg_num_ops = ccall((:spg_get_symmetry, SPGLIB), Cint,
+       (Ptr{Cint}, Ptr{Cdouble}, Cint, Ptr{Cdouble}, Ptr{Cdouble}, Ptr{Cint}, Cint, Cdouble),
+       spg_rotations, spg_translations, Cint(max_symops), lattice, spg_positions, spg_numbers,
+       Cint(length(spg_numbers)), tol_symmetry)
+       
+    # Note: spglib_jll returns transposed rotation matrix vs the python spglib interface
```
antoine-levitt, line 56: We don't reference the python interface anymore so this isn't helpful; rather at the top of the file I'd write down what spglib's convention is

src/bzmesh.jl
antoine-levitt: Or is it that spglib returns the S and not the Stilde?

antoine-levitt:
In retrospect I'm not sure our tilde and not tilde notation was very good. We should maybe have chosen Shat for what we call S and S for what we call Stilde, to emphasize Shat works in reciprocal space and S in real space? Or is there any standard conventions here?

src/bzmesh.jl
mfherbst: No spglib should return Stilde in our convention. It's the row-major versus column-major thing I think.

mfherbst:
Well the other point is that there is `\tau` and `\tilde{\tau}`. At that end `\tau^\ast` makes no sense either ...

antoine-levitt:
The parallelism between S and Stilde and tau and tautilde is not very good, actually. Stilde is a rotation in real space, S a rotation in recp space. tautilde is a translation in real space. tau is also a translation in real space

antoine-levitt:
I'll open another issue for that

src/external/spglib.jl
louisponet: I always feel uneasy using built in things for variable names, and so does my syntax highligher :p. If you want I can put it back to type though

mfherbst: I see, but where does `type` live in Julia? I did not find it in a fresh REPL at least.

antoine-levitt: `type` isn't a keyword in julia 1.0 actually, only `abstract type` is

src/external/spglib.jl
louisponet: It is because one needs to pass in preallocated arrays to store the translations and rotations, it's indeed to force spglib not to try and fit in more symmetry ops than allowed. There's `get_multiplicity` actually to figure that out. I'll use that!

src/external/spglib.jl
louisponet: No clue where to break this line without it being ugly, any suggestions?

src/external/spglib.jl
louisponet: I actually opted out of doing this and just forcing a Vector{Vec3{Float64}} as the type for qpoints, there are too many possible things one could throw in and catching all of them may not be easy/error prone (I'm not sure this line was actually correct).

src/bzmesh.jl
louisponet: I'm still confused on this actually, even though like this it seems everything works. According to the spglib documentation, transformations P, and coordinates x are defined such that 
P*x 
gives you the transformed coordinate. 

This together with my idea that C was column major like Julia, led me to believe that the rotations we got back from spglib should be in the order of DFTK, but somehow that turns out not to be true, hence the transpose in the get_symmetry part. 

Then originally I also had the transpose in the above line, but that then didn't work and produced garbage kpoints. This seems like an obvious inconsistency but I'm not clear on why that happens.

src/external/spglib.jl
```diff
@@ -1,40 +1,45 @@
-using PyCall
 # Routines for interaction with spglib
+# Note: spglib/C uses the row-major convention, thus we need to perform transposes
+#       between julia and spglib.
+#       However, spglib expects the lattice to be passed in as a C Array of the cell vectors,
```
antoine-levitt, line 4: I like the term "array of arrays" for this. We don't usually indent like this in comments (but I'm fine with it)

src/external/spglib.jl
```diff
@@ -44,21 +49,30 @@ spglib_cell(lattice, atoms) = first(spglib_cell_atommapping_(lattice, atoms))
...
         error("spglib failed to get the symmetries. Check your lattice, use a " *
               "uniform BZ mesh or disable symmetries. Spglib reported : " * err_message)
     end
 
-    Stildes = [St for St in eachslice(spg_symops["rotations"]; dims=1)]
-    τtildes = [rationalize.(τt, tol=tol_symmetry)
-               for τt in eachslice(spg_symops["translations"]; dims=1)]
-    @assert length(τtildes) == length(Stildes)
-
+    Stildes = [Mat3{Int}(spg_rotations[:, :, i])' for i in 1:spg_n_ops]
```
antoine-levitt, line 73: Add a comment like "transpose for row-major -> column-major" so that it's clear this is just an unpacking thing and not eg that spglib returns S instead of Stilde

src/external/spglib.jl
```diff
@@ -88,26 +102,41 @@ spglib_cell(lattice, atoms) = first(spglib_cell_atommapping_(lattice, atoms))
...
-    newlattice, newatoms
+    spg_lattice, newatoms
+end
+
+function spglib_get_stabilized_reciprocal_mesh(kgrid_size, rotations::Vector;
+                                               is_shift=Vec3(0, 0, 0),
+                                               is_time_reversal=false,
+                                               qpoints::Vector{Vec3{Float64}}=[Vec3(0.0, 0.0, 0.0)],
+                                               isdense=false)
+    spg_rotations = cat([Cint.(S)' for S in rotations]..., dims=3)
```
antoine-levitt, line 130: Same here add a comment

antoine-levitt:
Some nitpicking but apart from that good for me!

src/external/spglib.jl
louisponet: oh that's true. I'll change it back

src/external/spglib.jl
mfherbst: That's really a tough one. I think I would split it up in two expressions. There is anyway a lot happening in one line.

antoine-levitt: What we do in other places is 
```
x = [y
     for y in z]
```
which is not so bad I thought.

I don't particularly care about line length but Michael just really hates long lines because he has a very primitive editor ;-)

src/external/spglib.jl
mfherbst: I think that is a good idea. But again I would not not opting for too strong type constraints and rather explicitly convert to `Vector{Vec3{Float64}}` inside the function body. Just a `Vec3{Float64}.(qpoints)` should do, I'd say.

src/bzmesh.jl
mfherbst: Is this resolved @louisponet after the discussion yesterday evening?

src/external/spglib.jl
mfherbst: Fine from my end.

PR #267, opened 2020-06-23
Add hamiltonian_with_total_potential function

antoine-levitt:


src/terms/Hamiltonian.jl
```diff
@@ -126,3 +126,17 @@ function total_local_potential(ham::Hamiltonian)
...
+"""
+function hamiltonian_with_total_potential(ham::Hamiltonian, V)
+    @assert ham.basis.model.spin_polarization in (:none, :spinless)
+    Hamiltonian(ham.basis, hamiltonian_with_total_potential.(ham.blocks, Ref(V)))
+end
+function hamiltonian_with_total_potential(Hk::HamiltonianBlock, V)
+    @assert Hk.basis.model.spin_polarization in (:none, :spinless)
+    operators = [op for op in Hk.operators if !(op isa RealSpaceMultiplication)]
+    push!(operators, RealSpaceMultiplication(Hk.basis, Hk.kpoint, V))
+    HamiltonianBlock(Hk.basis, Hk.kpoint, operators, operators, Hk.scratch)
```
mfherbst, line 141: This skips doing the optimisation of operators. Are you sure this is a good idea?

mfherbst: Also why not make this a constructor?

antoine-levitt: It doesn't in this case, but I agree let's go through the constructor

antoine-levitt: It's pretty specific, I don't know if it should be a constructor

mfherbst: Well we still went for a constructor in [this case](https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/PlaneWaveBasis.jl#L371) which is pretty specific as well. I'm not saying constructor is the right way to go, but would not be the first specific constructor.

PR #268, opened 2020-06-24
Add nonvariatonal methods and total field for energies

antoine-levitt:


src/PlaneWaveBasis.jl
```diff
@@ -102,7 +102,7 @@ Base.eltype(basis::PlaneWaveBasis{T}) where {T} = T
         sizehint!(mapping, n_guess)
         sizehint!(Gvecs_k, n_guess)
```
mfherbst, line 103: What about the size hint for non-variational. Should be prod(fft_size) no?

src/energies.jl
```diff
@@ -22,7 +22,7 @@ Base.pairs(energies::Energies) = pairs(energies.energies)
 Base.iterate(energies::Energies) = iterate(energies.energies)
 Base.iterate(energies::Energies, state) = iterate(energies.energies, state)
 Base.haskey(energies::Energies, key) = haskey(energies.energies, key)
-
+Base.sum(energies::Energies) = sum(values(energies))
```
mfherbst, line 25: We had this before and I removed it. I really don't like this. For me energies is a Dict and a Dict iterates over pairs and not over the thing it maps to, so sum(energies) is not a simple value. I don't mind if you define a function `total` for this purpose.

src/PlaneWaveBasis.jl
antoine-levitt: Right now this is a hack, for non-variational there's better things to do than what we do now. The main use case is still (for now at least...) variational=true, so let's not bother optimizing for this

src/energies.jl
antoine-levitt: The fact that the pattern `sum(values(energies))` appears so much is a sign we should do something about it. I really don't like adding a generic `total` function: it makes little sense for DFTK to export such a generic word. And if you use another package that does this it conflicts. See, this is where we need OOP :-p We could maybe emulate this by overloading getproperty. Eg we could do energies.total, energies.Kinetic, etc. and there'd be nice tab completion at the REPL.

PR #272, opened 2020-06-30
Implement HybridMixing

mfherbst:


src/DFTK.jl
```diff
@@ -167,14 +168,11 @@ export apply_χ0
 include("postprocess/chi0.jl")
 
 function __init__()
-    # Use requires to only include eigen/diag_lobpcg_itsolve.jl once
-    # IterativeSolvers has been loaded (via a "using" or an "import")
+    # Use requires to only include fft_generic.jl once
```
antoine-levitt, line 171: Quote requires here

src/scf/mixing.jl
```diff
@@ -77,11 +77,86 @@ function mix(mixing::RestaMixing, basis, ρin::RealFourierArray, ρout::RealFour
...
     # take the correct DC component from ρout; otherwise the DC component never gets updated
     ρnext[1] = ρout[1]
     from_fourier(basis, ρnext)
 end
+
+
+@doc raw"""
+The model for the susceptibility is
+```math
+\chi_0(r, r') = (-LDOS(εF, r) + LDOS(εF, r) LDOS(εF, r') / DOS(εF))
```
antoine-levitt, line 93: Missing a delta, and you can remove one parenthesis

antoine-levitt: Also Unicode for the chi0

src/scf/mixing.jl
```diff
@@ -77,11 +77,86 @@ function mix(mixing::RestaMixing, basis, ρin::RealFourierArray, ρout::RealFour
...
+              + \sqrt{L(x)} IFFT \frac{C_0 G^2}{4π (1 - C_0 G^2 / k_F^2)} FFT \sqrt{L(x)}
+```
+where ``C_0 = 1 - ε_r`` and the same convention for parameters is used as in `RestaMixing`.
+Additionally there is the real-space localisation function `L(r)`.
+"""
+struct HybridMixing
+    α::Real
+    εr::Real
+    kF::Real
+    localisation::Function  # Function L(r), `r` is in fractional real-space coordinates
```
antoine-levitt, line 103: Zation

antoine-levitt: Maybe this is a good place for using this fancy macro to make structs have defaults

src/scf/mixing.jl
```diff
@@ -77,11 +77,86 @@ function mix(mixing::RestaMixing, basis, ρin::RealFourierArray, ρout::RealFour
...
+        δV = apply_kernel(basis, from_real(basis, δρ), ρ=ρin, rpa=mixing.rpa)
+
+        # Apply Resta term of χ0
+        loc_δV = apply_sqrtL(δV).fourier
+        resta_loc_δV =  @. C0 * kF^2 * Gsq / 4T(π) / (kF^2 - C0 * Gsq) * loc_δV
+        Jδρ .-= apply_sqrtL(from_fourier(basis, resta_loc_δV)).real
+
+        # Apply LDOS term of χ0
+        if ldos !== nothing
+            Jδρ .-= (-ldos .* δV.real .+ sum(ldos .* δV.real) .* ldos ./ sum(ldos))
```
antoine-levitt, line 148: Normalization is non obvious here, can you put it twice with the dvol in the sums?

src/scf/mixing.jl
```diff
@@ -77,11 +77,86 @@ function mix(mixing::RestaMixing, basis, ρin::RealFourierArray, ρout::RealFour
...
+        end
+
+        # Poor man's zero DC component before return
+        vec(Jδρ .-= sum(Jδρ) / length(Jδρ))
```
antoine-levitt, line 152: It's not poor man, it's correct

antoine-levitt: But if we do it here we probably should also do it to the input one, let's discuss

src/terms/terms.jl
```diff
@@ -99,9 +99,13 @@ function compute_kernel(basis::PlaneWaveBasis{T}; kwargs...) where {T}
     end
     k
 end
-function apply_kernel(basis, dρ; kwargs...)
+
+function apply_kernel(basis, dρ; rpa=false, kwargs...)
```
antoine-levitt, line 103: Caps for RPA?

src/scf/mixing.jl
mfherbst: Not sure I get what you mean here. What's dvol?

src/terms/terms.jl
mfherbst: Hmmm we don't really have all-caps kwargs so far and I'm not so much a fan of it, actually.

src/scf/mixing.jl
antoine-levitt: the integration volume, so that these sums are integrals

src/terms/terms.jl
antoine-levitt: We usually use caps for abbrevations (eg DFT, LDA - except in libxc), so we might just as well do it here also no?

PR #273, opened 2020-07-01
Bugfix: apply_kernel without functionals

mfherbst:


src/terms/xc.jl
```diff
@@ -79,7 +79,7 @@ function apply_kernel(term::TermXc, dρ::RealFourierArray; ρ::RealFourierArray,
     basis = term.basis
     T = eltype(basis)
     @assert all(xc.family in (:lda, :gga) for xc in term.functionals)
-    isempty(term.functionals) && return false  # Strong zero
+    isempty(term.functionals) && return nothing  # Strong zero
```
antoine-levitt, line 82: Uncomment "strong zero" then

Issue #215, opened 2020-04-29
Optimize fft_grid computation


antoine-levitt:
On a test (4x1x1 supercell of silicon), our prod(fft_size) is about 25% higher than abinit's. Abinit uses a more clever iterative algorithm (see `getng` function)

antoine-levitt:
This is *not* fixed by using an explicit search for the rectangular box of the sphere of radius 2Gmax. I'm not completely sure how to do it without a quadratic algorithm.

antoine-levitt:
For reference here's the code for doing that

```
# returns the lengths of the bounding rectangle in reciprocal space
# that encloses the sphere of radius Gmax
function bounding_rectangle(lattice::AbstractMatrix{T}, Gmax; tol=1e-8, alg=:crude) where {T}
    Glims = [0, 0, 0]
    if alg == :crude || det(lattice) == 0
        Glims_ = [norm(lattice[:, i]) / 2T(π) * Gmax for i in 1:3]
        # Round up, unless exactly zero (in which case keep it zero in
        # order to just have one G vector for 1D or 2D systems)
        for i = 1:3
            if Glims_[i] != 0
                Glims[i] = ceil(Int, Glims_[i] .- tol)
            end
        end
    else
        Glims_crude = bounding_rectangle(lattice, Gmax; tol=tol, alg=:crude)
        recip_lattice = 2T(π)*inv(lattice')
        for G1 = -Glims_crude[1]:Glims_crude[1]
            for G2 = -Glims_crude[2]:Glims_crude[2]
                for G3 = -Glims_crude[3]:Glims_crude[3]
                    G = recip_lattice * [G1; G2; G3]
                    if norm(G) ≤ Gmax
                        Glims .= max.(Glims, abs.((G1, G2, G3)))
                    end
                end
            end
        end
    end
    Glims
end
```
It doesn't decrease significantly the size of the grid

antoine-levitt:
OK, I've implemented the exhaustive search algorithm. Eg for the 4x1x1 silicon with gamma-only sampling at Ecut=30, I get
(144, 35, 35) (old algorithm)
(147, 40, 40) (new algorithm)
Too bad I forgot to record abinit's values while I was at it the other day... @mfherbst do you happen to have any abinit output file I could compare to?

mfherbst:
No, but I'll run it quickly.

mfherbst:
So on my experiments (current master) I got:
```
[...]

PR #274, opened 2020-07-01
More precise FFT grid computation and FFT-related renaming

antoine-levitt:
Misc refactors and doc


antoine-levitt:
I sneaked in also a partial fix for https://github.com/JuliaMolSim/DFTK.jl/issues/215. It's partial because it's very slow. There should be a middle ground (a O(N) algorithm that is still more precise than the naive version), but I've been too lazy to figure it out, so in the meantime the slow version is there for convergence tests

antoine-levitt:
I also added 7 to the list of small primes based on https://discourse.julialang.org/t/how-to-set-the-length-of-fft/38885

antoine-levitt:
Even with Eric's optimization there's still a not insignificant overhead to this, so let's keep this optional. If we can be bothered to optimize it in the future we can make it the default

src/PlaneWaveBasis.jl
```diff
@@ -121,10 +121,29 @@ build_kpoints(basis::PlaneWaveBasis, kcoords) =
...
+    if variational
+        @assert Ecut > 0
+        if fft_size === nothing
+            fft_size = determine_fft_size(model, Ecut; supersampling=supersampling)
+        end
+    else
+        # ensure fft_size is provided, and other options are not set
+        @assert fft_size !== nothing
+        @assert supersampling == 2
+        @assert !optimize_fft_size
```
mfherbst, line 134: I know you like assertions, but since this is very user-facing I think a slightly more verbose error message is appropriate in my opinion.

src/PlaneWaveBasis.jl
```diff
@@ -121,10 +121,29 @@ build_kpoints(basis::PlaneWaveBasis, kcoords) =
 # This is the "internal" constructor; the higher-level one below should be preferred
 @timing function PlaneWaveBasis(model::Model{T}, Ecut::Number,
                                 kcoords::AbstractVector, ksymops, symops=nothing;
-                                fft_size=determine_grid_size(model, Ecut), variational=true) where {T <: Real}
-    @assert Ecut > 0
+                                fft_size=nothing, variational=true, optimize_fft_size=false, supersampling=2) where {T <: Real}
```
mfherbst, line 124: Definitely cut this line.

src/PlaneWaveBasis.jl
```diff
@@ -121,10 +121,29 @@ build_kpoints(basis::PlaneWaveBasis, kcoords) =
...
+        @assert fft_size !== nothing
+        @assert supersampling == 2
+        @assert !optimize_fft_size
+    end
     fft_size = Tuple{Int, Int, Int}(fft_size)
 
+    kpoints = build_kpoints(model, fft_size, kcoords, Ecut; variational=variational)
+
+    if variational && optimize_fft_size
+        fft_size = determine_fft_size_precise(model.lattice, Ecut, kpoints; supersampling=supersampling)
```
mfherbst, line 141: I'd rather have a common interface to both these methods and use a flag `fast` or `algorithm=:fast or :precise` to switch between both.

src/fft.jl
```diff
@@ -1,63 +1,121 @@
...
+    Glims = [0, 0, 0]
+    for i = 1:3
+        if Glims_[i] != 0
+            Glims[i] = ceil(Int, Glims_[i] .- tol)
+        end
+    end
+    Glims
+end
+
+_smallprimes() = [2, 3, 5, 7]
```
mfherbst, line 22: Why a function? Why not just `const SMALLPRIMES = ` why not keep it inside the function?

src/fft.jl
```diff
@@ -1,63 +1,121 @@
...
+    # Optimize FFT grid size: Make sure the number factorises in small primes only
+    if ensure_smallprimes
+        fft_size = nextprod.(Ref(_smallprimes()), fft_size)
+    end
+    fft_size
+end
+function determine_fft_size(model::Model, Ecut; kwargs...)
+    determine_fft_size(model.lattice, Ecut; kwargs...)
+end
+function lattice_diameter(latt)
```
mfherbst, line 51: writing it out as `lattice` certainly does not hurt. Also some empty lines  around the functions please

mfherbst: Also why not just call the function `diameter`?

src/fft.jl
```diff
@@ -1,63 +1,121 @@
...
+    diam = zero(eltype(latt))
+    # brute force search
+    for i in (-1, 0, 1)
+        for j in (-1, 0, 1)
+            for k in (-1, 0, 1)
```
mfherbst, line 56: Is this really needed? Isn't it clear that the versions with (1) and (-1) are always larger than the versions with any zero?

mfherbst: Also:
```julia
for vec in Vec3.(Iterators.product(-1:1, -1:1, -1:1))
    ...
end
```

src/fft.jl
mfherbst: Also I'm not too sure about 7. I excluded it because it does do considerably worse in performances, e.g.  http://www.fftw.org/speed/Pentium4-3.60GHz-icc/

src/fft.jl
```diff
@@ -1,63 +1,121 @@
...
+                diam = max(diam, norm(latt*[i; j; k]))
+            end
         end
     end
+    diam
+end
 
+## This uses a more precise and slower algorithm than the one above,
+## simply enumerating all G vectors and seeing where their difference
+## is. It needs the kpoints to do so.
```
mfherbst, line 66: Could be a docstring.

src/fft.jl
```diff
@@ -1,63 +1,121 @@
...
+function determine_fft_size_precise(lattice::AbstractMatrix{T}, Ecut, kpoints;
+                                    supersampling=2, ensure_smallprimes=true) where T
+    recip_lattice = 2T(π)*pinv(lattice')  # pinv in case one of the dimension is trivial
+    recip_diameter = lattice_diameter(recip_lattice)
+    Glims = [0, 0, 0]
+    # get the bounding rectangle the naive way
+    for kpt in kpoints
+        for G in G_vectors(kpt)
+            if norm(recip_lattice * (G+kpt.coordinate)) ≤ sqrt(2Ecut) - recip_diameter
```
mfherbst, line 75: space around +

src/fft.jl
```diff
@@ -1,63 +1,121 @@
...
+                # each of the 8 neighbors (in ∞-norm) also belongs to the grid
+                # so we can safely skip the search knowing at least one of them
+                # will have larger |G-Gp|.
+                # Savings with this trick are surprisingly small : for silicon, 50% at Ecut 30, 70% at Ecut 100
```
mfherbst, line 79: Too long a line.

src/fft.jl
```diff
@@ -1,63 +1,121 @@
...
+                for i = 1:3
+                    @inbounds Glims[i] = max(Glims[i], abs(G[i] - Gp[i]))
+                end
+            end
+        end
+    end
+    if supersampling != 2
+        # no guarantees there, we just do our best to satisfy the
+        # target supersampling ratio
+        Glims = round.(Int, supersampling ./ 2 .* Glims)
```
mfherbst, line 92: Just took me a bit to understand this mechanics. Could you add a comment why you compare supersampling to `2` and divide by `2` (in the sense that the double loop exactly covers the conventional case supersampling of 2, so if it's not that you just scale).

src/fft.jl
```diff
@@ -1,63 +1,121 @@
 import FFTW
 
+# returns the lengths of the bounding rectangle in reciprocal space
+# that encloses the sphere of radius Gmax
+function bounding_rectangle(lattice::AbstractMatrix{T}, Gmax; tol=1e-8) where {T}
+    # If |B G| ≤ Gmax, then
+    # |Gi| ≤ |e_i^T B^-1 B G| ≤ |B^-T e_i| Gmax = |A_i| Gmax
+    # with B the reciprocal lattice matrix and A_i the i-th column of the lattice matrix
+    Glims_ = [norm(lattice[:, i]) / 2T(π) * Gmax for i in 1:3]
```
mfherbst, line 9: I don't like this variable name. It's too similar to `Glims` and easy to overlook the `_`. Change one or the other.

src/fft.jl
```diff
@@ -1,63 +1,121 @@
...
+
+    # Round up, unless exactly zero (in which case keep it zero in
+    # order to just have one G vector for 1D or 2D systems)
+    Glims = [0, 0, 0]
+    for i = 1:3
+        if Glims_[i] != 0
```
mfherbst, line 15: Compare Glims_[1] to `tol` as well for consistency? ... not that it matters for practical cases

src/fft.jl
```diff
@@ -1,63 +1,121 @@
 import FFTW
 
+# returns the lengths of the bounding rectangle in reciprocal space
+# that encloses the sphere of radius Gmax
+function bounding_rectangle(lattice::AbstractMatrix{T}, Gmax; tol=1e-8) where {T}
+    # If |B G| ≤ Gmax, then
+    # |Gi| ≤ |e_i^T B^-1 B G| ≤ |B^-T e_i| Gmax = |A_i| Gmax
```
mfherbst, line 7: The first thing is an `=`, no?

src/PlaneWaveBasis.jl
antoine-levitt: This is not very user-facing: the user should not mess with these unless they know what they're doing. I can add some doc for these options though

antoine-levitt: This is a pretty internal thing for tests, it's not a recommended thing (yet). Once we understand more about these methods we can see

src/PlaneWaveBasis.jl
antoine-levitt: I had that, but they have different input arguments. This is not very finalized so I'd rather not solidify the API

src/PlaneWaveBasis.jl
```diff
@@ -121,10 +121,29 @@ build_kpoints(basis::PlaneWaveBasis, kcoords) =
...
+                                fft_size=nothing, variational=true, optimize_fft_size=false, supersampling=2) where {T <: Real}
+    if variational
+        @assert Ecut > 0
+        if fft_size === nothing
+            fft_size = determine_fft_size(model, Ecut; supersampling=supersampling)
+        end
+    else
+        # ensure fft_size is provided, and other options are not set
+        @assert fft_size !== nothing
+        @assert supersampling == 2
```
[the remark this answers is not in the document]
antoine-levitt: Unfortunately not: supersampling still implies a spherical basis set defined by |k+G|^2 < 2Ecut. The "non-variational" uses the same basis set (or rather, grid) for wavefunctions and densities/potentials

src/fft.jl
```diff
@@ -1,63 +1,121 @@
 import FFTW
 
+# returns the lengths of the bounding rectangle in reciprocal space
+# that encloses the sphere of radius Gmax
```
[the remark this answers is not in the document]
antoine-levitt: I know but I like indicating internal functions by their absence of docstrings (it distinguishes them when looking at the source)

src/fft.jl
antoine-levitt: http://www.fftw.org/speed/Pentium4-3.60GHz-icc/RADPSDV683-icc.3d.drxx.np2.png is not bad

antoine-levitt: It's used in two functions so can't leave it inside. const is OK, but I don't like shouting, so function is nice

src/fft.jl
antoine-levitt: Problem with `lattice` is that it increases the confusion between lattice and recip_lattice, but OK

src/fft.jl
antoine-levitt: super nice !

antoine-levitt: About the optimization: I don't know and I don't care, this stuff makes my head hurt and so when it's not performance critical I definitely don't want to optimize it

src/fft.jl
antoine-levitt: As above, it's experimental so I'm not docstringing it

src/fft.jl
antoine-levitt: This is typically the kind of places where I like no spaces, to indicate priority (even with parentheses). But OK

src/fft.jl
antoine-levitt: added a comment above

src/fft.jl
antoine-levitt: No, we just want to catch exactly zero here

src/fft.jl
antoine-levitt: yes

src/fft.jl
mfherbst: http://www.fftw.org/speed/Pentium4-3.60GHz-icc/RADPSDV683-icc.3d.scxx.np2.png and http://www.fftw.org/speed/Pentium4-3.60GHz-icc/RADPSDV683-icc.3d.dcxx.np2.png is and that's what we use.

antoine-levitt: It's pretty noisy but there doesn't seem to be a clear pattern of slowdown for factors of 7, so I would include 7 in there also. I don't particularly mind either way.

mfherbst: I do what ABINIT does, but if you think 7 does not hurt, then ok.

src/fft.jl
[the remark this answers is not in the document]
mfherbst: I see, but (a) we are far from consistent with that and (b) If you go through the effort of writing something, why not just make it accessible to the julia shell using the docstring system that has been created for exactly that purpose?

src/PlaneWaveBasis.jl
mfherbst: Then I would not documented it, but make a TODO next to the assertions that this should be converted into error messages once this has stabilised.

src/fft.jl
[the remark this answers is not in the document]
antoine-levitt: I know you use the julia shell documenter, but I don't like it as much as you do because it makes the source code much more verbose. So I'd rather we keep the docstrings for external-facing functions, which as an added bonus makes it easier to see which functions you're "supposed" to use and which you're not. We can try to make it more consistent if that's your opposition. I don't have a strong position on the use of underscores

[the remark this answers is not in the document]
mfherbst: Well but in this case it's 8 characters and two lines more: `"""\n` and `\n"""`. Sounds fair, no?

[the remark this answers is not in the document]
antoine-levitt: Yeah but again it makes it less clear that this is a helper function that you're not supposed to use from the outside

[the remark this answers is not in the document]
antoine-levitt: I'm OK with docstrings if we find another way to make that clear. I don't particularly like underscores but we could do that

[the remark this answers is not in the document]
mfherbst: How about not `export`ing it?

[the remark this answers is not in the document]
antoine-levitt: That doesn't help you when you're looking at the file. This information is far away in DFTK.jl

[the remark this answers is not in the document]
mfherbst: Yes that's a valid point. Regarding the underscores thing the tipping point for me is how many functions this would affect. If it's many (like most in DFTK) than I would not do that. If it's occasional (like 2-3 as many underscore functions as we have now) I would agree to that solution.

PR #275, opened 2020-07-06
Various small amendments (SCF printing, GGA functionals, mixing)

mfherbst:


src/Model.jl
```diff
@@ -171,6 +171,14 @@ function model_LDA(lattice::AbstractMatrix, atoms::Vector; kwargs...)
 end
 
 
+"""
+Build an PBE-GGA model from the specified atoms.
+"""
+function model_PBE(lattice::AbstractMatrix, atoms::Vector; kwargs...)
```
antoine-levitt, line 177: Good idea

src/scf/self_consistent_field.jl
```diff
@@ -1,3 +1,5 @@
+using Plots
```
antoine-levitt, line 1: Do we really want this? It'll massively increase load times...

antoine-levitt: Or use requires?

src/scf/self_consistent_field.jl
```diff
@@ -27,20 +29,38 @@ function next_density(ham::Hamiltonian;
...
+            error[error .== 0] .= NaN
+            extra = ifelse(:mark in keys(kwargs), (), (mark=:x, ))
+            plot!(plt, error; extra..., kwargs...)
+            display(plt)
+        else
+            push!(energies, info.energies.total)
+        end
+    end
+end
+
```
antoine-levitt, line 47: This file is starting to get pretty busy, move all convenience callbacks/isconverged/determine_diagtol/etc to `callbacks.jl`?

antoine-levitt: Also doc a bit this function: what does it plot?

src/terms/xc.jl
```diff
@@ -184,9 +184,13 @@ function DensityDerivatives(basis, max_derivative::Integer, ρ::RealFourierArray
...
-        tmp = G_to_r(basis, x)
-        check_real(tmp)
-        real(tmp)
+        # For even-length grids the element with largest negative integer coordinate
+        # in G_vectors has no complex-conjugated partner. In order to avoid having
+        # a spurious imaginary component in real-space, we set these elements to zero
+        iseven(basis.fft_size[1]) && (x[div(end, 2) + 1, :, :] .= 0)
+        iseven(basis.fft_size[2]) && (x[:, div(end, 2) + 1, :] .= 0)
+        iseven(basis.fft_size[3]) && (x[:, :, div(end, 2) + 1] .= 0)
+        check_real(G_to_r(basis, x))
```
antoine-levitt, line 193: You removed too much here (hopefully the tests will catch it)

src/scf/self_consistent_field.jl
mfherbst: Crime already committed: https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/postprocess/band_structure.jl#L2

src/scf/self_consistent_field.jl
mfherbst: good point.

src/scf/self_consistent_field.jl
antoine-levitt: Well that might be why it's so slow... Use requires there as well maybe?

antoine-levitt: `using Plots` is 7s here (1.5beta)

mfherbst: I thought about that, but the problem is that then you cannot plot bandstructures unless issue a `using Plots` first, which is a little counterintuitive.

antoine-levitt: I'd say 7s of load time in most cases is worth a bit of extra boilerplate. Ideally we could revisit this once (if?) load times becomes less annoying

mfherbst: Hmmm I have plots in my sysimage so for me it loads in 0.2 seconds ^^. No but ok I get the idea. I'll do that in a separate PR, though.

src/terms/xc.jl
mfherbst: Apparently not. But also I'm not sure what you mean. If `fft_size` is `12, y, z` then `(-6, n, m)` has no counterpart since `(6, -n, -m)` is not part of the grid, no matter what `n` and `m` are.

antoine-levitt: I mean you removed the actual computation of the IFFT

mfherbst: That's in line 193, also `check_real(x)` returns `real(x)`

antoine-levitt: Ah, sorry!

antoine-levitt: That logic (with the iseven) should go elsewhere. It's weird to "purify" only here: we should either do it everywhere (eg in density building) or nowhere at all. Maybe just ignore these coefficients in check_real?

mfherbst: I thought about putting the purification into check_real, but than it's really strange that a check function modifies. Maybe we can just ignore these elements. That would make sense, I guess.

mfherbst: Not as easy as I thought because check_real takes a real-space array only, but the zeroing is to be done in fourier space. A made a special function for it now.

PR #276, opened 2020-07-07
Some more reasonable defaults for number of bands and kgrids

mfherbst:


src/PlaneWaveBasis.jl
```diff
@@ -187,16 +187,22 @@ end
...
 crystal are used to reduce the number of ``k``-Points which are
 treated explicitly. In this case all guess densities and potential
 functions must agree with the crystal symmetries or the result is
 undefined.
 """
-function PlaneWaveBasis(model::Model, Ecut::Number; kgrid=[1, 1, 1], kshift=[0, 0, 0],
-                        use_symmetry=true, kwargs...)
+function PlaneWaveBasis(model::Model, Ecut::Number;
+                        kgrid=kgrid_size_from_minimal_spacing(model.lattice, 2π * 0.022),
+                        kshift=[ifelse(iseven(nk), 1/2, 0) for nk in kgrid],
```
antoine-levitt, line 203: ifelse is to force argument evaluation, use ternary operator instead

src/scf/self_consistent_field.jl
```diff
@@ -1,7 +1,15 @@
 using Plots
 include("scf_callbacks.jl")
 
-default_n_bands(model) = div(model.n_electrons, filled_occupation(model))
+function default_n_bands(model)
+    min_n_bands = div(model.n_electrons, filled_occupation(model))
+    if iszero(model.temperature)
```
antoine-levitt, line 6: why not == 0 ?

antoine-levitt: ternary here ?

src/PlaneWaveBasis.jl
mfherbst: Not that it matters but ok

PR #277, opened 2020-07-07
Add example about SCF callbacks

mfherbst:


antoine-levitt:
Nice. But can't we merge it with the custom solvers example?

mfherbst:
We could, but I like having different examples to show different concepts. In this case solvers is numerics and this one is more about debugging and extracting information. I don't think there is a problem with having too many examples, but I also don't particularly mind.

antoine-levitt:
Problem is that having too many examples will become unreadable eventually. This falls under the same category of "let's inspect the SCF solver" and could give a nicely consistent tutorial when merged, but I don't particularly mind either way.

mfherbst:
I see. I guess eventually we will need to go a bit more for topic and tutorial-oriented examples (like the bigger codes do), which will mean merging a few examples into a bigger one. I think we should see how it evolves and organically merge examples as we see use.

antoine-levitt:
ok

PR #278, opened 2020-07-09
Adding H2 geometry optimzation example

gkemlin:


gkemlin:
I could have added the optimization geometry on H2O, but it takes way too much time even on my computer, so I don't think it might be a good idea to put it in the examples as it will take too long to compute on the cloud.

antoine-levitt:
Yes, having examples that run quickly is important. What is taking so long here? Is it forces?

gkemlin:
Yes, that's it: computing forces along with optimization in R^9.

examples/geometry_optimization.jl
```diff
@@ -0,0 +1,69 @@
...
+function compute_scfres(x)
+    atoms = [H => [x[1:3]/a, x[4:6]/a]]
+    model = model_LDA(lattice, atoms)
+    basis = PlaneWaveBasis(model, Ecut; kgrid=kgrid)
+    scfres = self_consistent_field(basis, tol=tol/10, callback=info->nothing)
+end;
+
+# Then, we create two optimization routines:
+# - `fg!` is used to update the value of the objective function `F`, namely the energy, and its gradient `G`, here computed with the forces;
+# - `optimize_geometry` optimizes the energy using `LBFGS()` starting from $x_0 \in \mathbb{R}^6$ where the three first coordinates correspond to the position of the first atom, and the three last correspond to the position of the second atom.
```
antoine-levitt, line 30: Move that doc to the first place that convention is used (above)

examples/geometry_optimization.jl
```diff
@@ -0,0 +1,69 @@
...
+
+function fg!(F, G, x)
+    scfres = compute_scfres(x)
+    if G != nothing
+        grad = forces(scfres)
+        G .= -[grad[1][1]; grad[1][2]]
+    end
+    sum(values(scfres.energies))
+end
+function optimize_geometry(x0)
```
antoine-levitt, line 40: This function you only call once so just inline that

examples/geometry_optimization.jl
```diff
@@ -0,0 +1,69 @@
...
+# - `fg!` is used to update the value of the objective function `F`, namely the energy, and its gradient `G`, here computed with the forces;
+# - `optimize_geometry` optimizes the energy using `LBFGS()` starting from $x_0 \in \mathbb{R}^6$ where the three first coordinates correspond to the position of the first atom, and the three last correspond to the position of the second atom.
+
+function fg!(F, G, x)
+    scfres = compute_scfres(x)
+    if G != nothing
+        grad = forces(scfres)
+        G .= -[grad[1][1]; grad[1][2]]
+    end
+    sum(values(scfres.energies))
```
antoine-levitt, line 38: This you can now do as `scfres.energies.total`

examples/geometry_optimization.jl
```diff
@@ -0,0 +1,69 @@
...
+# Goedecker-Hutter-Teter table gives a bond length of 1.522 Bohr
+# (cf. the [ABINIT tutorial](https://docs.abinit.org/tutorial/base1/)).
+#
+# !!! note "Degrees of freedom"
+#     We used here a very general setting where we optimized on the 6 variables
+#     representing the position of the 2 atoms and it can be then easily extended
+#     to molecules with more atoms (such as $H_2O$). In the particular case
+#     of $H_2$, there is in fact only one unknown which is the bond length $d$,
+#     and we could have used 1D optimization algorithms to minimize an energy
+#     functional built just as a function of the variable $d$.
```
antoine-levitt, line 69: Not sure that note is that useful...

gkemlin: I don't know, this is how I did it first and then I moved on to something more general, but it might still be good to mention it.

gkemlin:
The test on Travis to generate doc is failing with the following error : 

```
┌ Error: error when executing notebook based on input file: `~/build/JuliaMolSim/DFTK.jl/examples/geometry_optimization.jl`
└ @ Literate ~/.julia/packages/Literate/5an7K/src/Literate.jl:594

ERROR: LoadError: LoadError: Cannot map the atom at position [-0.6234998046636102, 4.387574846033396e-6, -1.0048786234097246e-5] to another atom of the same element under the symmetry operation (Stilde, τtilde):
([1 0 0; 0 0 -1; 0 1 0], Rational{Int64}[0//1, 1//1, 40903//40904])
```

Do you know where it might come from ? Maybe we should not use symmetries ?

examples/geometry_optimization.jl
```diff
@@ -0,0 +1,84 @@
...
+# stored as a vector: `x[1:3]` represents the position of the
+# first atom and `x[4:6]` the position of the second.
+# We also update `ψ` and `ρ` for the next iteration.
+
+function compute_scfres(x)
+    atoms = [H => [x[1:3], x[4:6]]]
+    model = model_LDA(lattice, atoms)
+    basis = PlaneWaveBasis(model, Ecut; kgrid=kgrid)
+    global ψ, ρ
+    if ρ ==  nothing
```
mfherbst, line 36: Comparison for `nothing` with `isnothing` or `===`

examples/geometry_optimization.jl
```diff
@@ -0,0 +1,84 @@
...
+        ρ = guess_density(basis)
+    end
+    scfres = self_consistent_field(basis; ψ=ψ, ρ=ρ,
+                                   tol=tol/10, callback=info->nothing)
```
mfherbst, line 40: I'd put spaces around `/`

examples/geometry_optimization.jl
```diff
@@ -0,0 +1,84 @@
...
+    ψ = scfres.ψ
+    ρ = scfres.ρ
+    scfres
+end;
+
+# Then, we create the function we want to optimize: `fg!` is used to update the value of the objective function `F`, namely the energy, and its gradient `G`, here computed with the forces.
```
mfherbst, line 46: I'd like line breaks here.

examples/geometry_optimization.jl
```diff
@@ -0,0 +1,84 @@
...
+
+function fg!(F, G, x)
+    scfres = compute_scfres(x)
+    if G != nothing
+        grad = forces(scfres)
+        G .= -[grad[1][1]; grad[1][2]]
```
mfherbst, line 52: Why the minus? (say that forces are the *negative* gradient by definition)

examples/geometry_optimization.jl
```diff
@@ -0,0 +1,84 @@
...
+    scfres.energies.total
+end
+
+# Now, we can optimize on the 6 parameters `x = [x1, y1, z1, x2, y2, z2]` in
+# reduced coordinates, using `LBFGS()` and starting from `x0`. By
+# default, `optimize` traces the output of the optimization algorithm during the
+# iterations. Once we have the minimizer `xmin`, we compute the bond length in
+# cartesian coordinates.
+
+x0 = vcat(invlattice*[0., 0., 0.], invlattice*[1.4, 0., 0.])
```
mfherbst, line 63: Frankly I'd avoid the invlattice and just use `lattice \ [0,0,0]`

examples/geometry_optimization.jl
```diff
@@ -0,0 +1,84 @@
...
+    scfres = compute_scfres(x)
+    if G != nothing
+        grad = forces(scfres)
+        G .= -[grad[1][1]; grad[1][2]]
+    end
+    scfres.energies.total
+end
+
+# Now, we can optimize on the 6 parameters `x = [x1, y1, z1, x2, y2, z2]` in
+# reduced coordinates, using `LBFGS()` and starting from `x0`. By
```
mfherbst, line 58: What is `x0`? (A guess for the coordinates)

examples/geometry_optimization.jl
```diff
@@ -0,0 +1,84 @@
...
+# iterations. Once we have the minimizer `xmin`, we compute the bond length in
+# cartesian coordinates.
+
+x0 = vcat(invlattice*[0., 0., 0.], invlattice*[1.4, 0., 0.])
+xres = optimize(Optim.only_fg!(fg!), x0, LBFGS(),
+                Optim.Options(show_trace=true, f_tol=tol))
+xmin = Optim.minimizer(xres)
+dmin = norm(lattice*xmin[1:3] - lattice*xmin[4:6])
+sdmin = @sprintf "%.3f" dmin
+println("\nBond length minimizing the energy for Ecut=$(Ecut): ", sdmin, " Bohr");
```
mfherbst, line 69: I'd put it in one line, like `@printf "\nOptimal bond length for Ecut=%.2f: %.3f Bohr\n" Ecut dmin`

examples/geometry_optimization.jl
```diff
@@ -0,0 +1,84 @@
...
+
+# We used here very rough parameters to generate the example,
+# but setting `Ecut` to 10 Ha yields a bond length of 1.523 Bohr.
```
mfherbst, line 72: Why the "but"? Just "Setting `Ecut` to 10 Ha ..."

examples/geometry_optimization.jl
mfherbst: I agree I think that's not super useful, but I don't think it hurts either. I'd leave it and maybe shorten it a bit. E.G. just say that we could be more clever and  use only the internal degrees of freedom, which in this case is just the bond length.

examples/geometry_optimization.jl
```diff
@@ -0,0 +1,84 @@
+# # Geometry optimization
+#
+# We use the DFTK and Optim packages in this example to compute numerically the bond length
```
mfherbst, line 3: "compute numerically" sounds strange to a chemist, better: "to optimise the bond length" or "to find the minimal-energy bond length"

examples/geometry_optimization.jl
```diff
@@ -0,0 +1,84 @@
...
+
+kgrid = [1, 1, 1]       # k-Point grid
+Ecut = 5                # kinetic energy cutoff in Hartree
+tol = 1e-8              # tolerance for the optimization routine
+a = 10                  # lattice constant in Bohr
+lattice = a * Diagonal(ones(3))
+invlattice = inv(lattice)        # matrix to go from cartesian to reduced coordinates
+H = ElementPsp(:H, psp=load_psp("hgh/lda/h-q1"));
+
+# We define a blochwave and a density to be used as global variables where we
```
mfherbst, line 19: I'd emphasise the "why": "... as global variables, such that we can transfer the solution from one iteration to another".

examples/geometry_optimization.jl
mfherbst: What is LBFGS?  (mention it is the default minisation algo in Optim)

antoine-levitt:
Oh that's a fun error. I'm guessing our check uses tolerances different from the one spglib uses?

mfherbst:
Ah yes the issue is that spglib uses a tolerance of 1e-5 to find symmetries, but than we check in DFTK whether these symmetries are correct. If you add some leg freedom for numerical errors in https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/external/spglib.jl#L90  (like a factor 2 or 5 say) than this error should not show up.

examples/geometry_optimization.jl
```diff
@@ -52,33 +57,32 @@ function fg!(F, G, x)
...
-sdmin = @sprintf "%.3f" dmin
-println("\nBond length minimizing the energy for Ecut=$(Ecut): ", sdmin, " Bohr");
+@printf "\nOptimal bond length for Ecut=%.2f: %.3f Bohr\n" Ecut dmin
 
-# We used here very rough parameters to generate the example,
-# but setting `Ecut` to 10 Ha yields a bond length of 1.523 Bohr.
-# As a comparison, ABINIT with the exact same lattice constant, Ecut = 10,
+# We used here very rough parameters to generate the example and
+# setting `Ecut` to 10 Ha yields a bond length of 1.523 Bohr.
+# As a comparison, ABINIT with the exact same lattice constant, `Ecut = 10`,
```
mfherbst, line 78: Typo here: ABINIT yields the same bond length

examples/geometry_optimization.jl
```diff
@@ -52,33 +57,32 @@ function fg!(F, G, x)
...
 #
 # !!! note "Degrees of freedom"
 #     We used here a very general setting where we optimized on the 6 variables
-#     representing the position of the 2 atoms and it can be then easily extended
+#     representing the position of the 2 atoms and it can be easily extended
 #     to molecules with more atoms (such as ``H_2O``). In the particular case
-#     of ``H_2``, there is in fact only one unknown which is the bond length ``d``,
-#     and we could have used 1D optimization algorithms to minimize an energy
-#     functional built just as a function of the variable ``d``.
+#     of ``H_2``, we could use only one degree of freedom which, in this case,
```
mfherbst, line 87: I would really mention that it is just the *internal* degree of freedom that you need to consider.

examples/geometry_optimization.jl
gkemlin: Do we mention that the parameters on ABINIT are the same or do we just write "ABINIT yields the same bond length" ?

mfherbst: Maybe just add `, which [agrees with ABINIT](link to tutorial).` to the first sentence and leave the rest. It's implicitly clear that the same setup is used.

gkemlin: yeah, sounds good :+1:

examples/geometry_optimization.jl
```diff
@@ -74,15 +74,12 @@ dmin = norm(lattice*xmin[1:3] - lattice*xmin[4:6])
...
-# (cf. the [ABINIT tutorial](https://docs.abinit.org/tutorial/base1/)).
+# setting `Ecut` to 10 Ha yields a bond length of 1.523 Bohr,
+# which [agrees with ABINIT](https://docs.abinit.org/tutorial/base1/).
 #
 # !!! note "Degrees of freedom"
 #     We used here a very general setting where we optimized on the 6 variables
 #     representing the position of the 2 atoms and it can be easily extended
 #     to molecules with more atoms (such as ``H_2O``). In the particular case
-#     of ``H_2``, we could use only one degree of freedom which, in this case,
+#     of ``H_2``, we could the internal degree of freedom which, in this case,
```
mfherbst, line 84: removed too much insert: `use only` :smile:.

gkemlin: haha, yes, sorry :smile:

PR #279, opened 2020-07-15
Correct for subtle numerical issue when finding Fermi level

mfherbst:


src/occupation.jl
```diff
@@ -58,6 +58,14 @@ function find_occupation(basis::PlaneWaveBasis{T}, energies;
                              Roots.Bisection(), atol=eps(T))
     end
 
+    if !isapprox(compute_n_elec(εF), n_electrons)
+        # For insulators it can happen that bisection stops in a final interval (a, b) where
+        # `compute_n_elec(a) ≈ n_electrons` and `compute_n_elec(b) > n_electrons`, but where
+        # the returned `(a+b)/2` is rounded to `b`, such that `εF` gives a too
+        # large electron count. To make sure this is not the case, make εF a little smaller.
+        εF -= eps(εF)
```
antoine-levitt, line 66: Can you add an assert here to make sure everything is nice at that point?

mfherbst: That's checked again down below and an error is thrown if not.

antoine-levitt: OK!

PR #280, opened 2020-07-20
Add timings to chi0 and mixing

mfherbst:


src/symmetry.jl
```diff
@@ -179,7 +179,7 @@ end
 
 # Accumulates the symmetrized versions of the density ρin into ρout (in Fourier space).
 # No normalization is performed
-function accumulate_over_symops!(ρaccu, ρin, basis, symops, Gs)
+function accumulate_over_symops!(ρaccu, ρin, basis, symops, Gs=G_vectors(basis))
```
antoine-levitt, line 182: Just remove this argument

src/postprocess/chi0.jl
```diff
@@ -117,11 +121,12 @@ Returns the change in density `δρ` for a given `δV`. Drop all non-diagonal te
...
-                  droptol=0,
-                  sternheimer_contribution=true,
-                  temperature=ham.basis.model.temperature,
-                  cgtol=1e-6)
+@timing function apply_χ0(ham, ψ, εF, eigenvalues, δV::RealFourierArray;
+                          droptol=0,
+                          sternheimer_contribution=true,
+                          temperature=ham.basis.model.temperature,
+                          cgtol=1e-6,
+                          verbose=false)
```
antoine-levitt, line 129: A bit tricky here: it's not clear that the verbosity is in the Sternheimer solve. I would either not expose this, change its name, or have a sternheimer_kwargs argument. In fact we should maybe do this more (having a "X_kwargs" field)

mfherbst:
@antoine-levitt Let me know if you have comments on the parallelisation notes

PR #284, opened 2020-07-23
time per SCF step

antoine-levitt:


src/PlaneWaveBasis.jl
```diff
@@ -80,6 +80,23 @@ Base.show(io::IO, basis::PlaneWaveBasis) =
...
 Base.eltype(basis::PlaneWaveBasis{T}) where {T} = T
 
+"""
+Returns a very (very) rough estimate of the time per SCF step (in seconds),
+assuming FFTs are the limiting operation.
+"""
+function estimate_time_per_SCF(basis::PlaneWaveBasis)
+    # TODO pretty print this instead of just seconds.
+    # There's nothing in julia stdlib for this apparently...
+    # on my laptop (2020) a 128^3 FFT takes 24ms
```
mfherbst, line 90: I tried the code example on various machines (laptop, clusters etc) and the spread this takes is pretty wide (between 30 and 50ms), so I'm not sure hard-coding a value here is the right way to go. How do you think about actually measuring this in-place?

antoine-levitt: Yeah I thought about that but it would mean either depending on benchmarktools or relying on a pretty noisy number. I'm fine hard-coding 30 or so: this is only to get an order of magnitude anyway
(so my laptop is great basically?)

mfherbst: Seems so. I changed it to 30 and moved it to a different file (I really don't think this fits into PlaneWaveBasis, because it makes assumptions about the scf)

antoine-levitt: Pwbasis is a bit crowded so I agree, but top-level is a bit too much for this. Utils/?

mfherbst: ok.

PR #285, opened 2020-07-24
Refactor chi0 computation

mfherbst:
The threading needs to be checked, also it's double parallelising (since H application inside Sternheimer is threaded), so I think I'll rework to just thread the explicit contributions (inner loop).

**Update:** Threading does not seem to be worth it. Removed for now, PR just refactors.

Depends on #286

### DIFF
--- a/src/postprocess/chi0.jl
+++ b/src/postprocess/chi0.jl
@@ -156,7 +156,6 @@ returns `3` extra bands, which are not converged by the eigensolver
     # δρ = ∑_nk (f'n δεn |ψn|^2 + 2Re fn ψn* δψn - f'n δεF |ψn|^2
     δρ_fourier = zeros(complex(T), size(δV))
     for ik = 1:length(basis.kpoints)
-        # One δρk accumulator per thread
         δρk = zero(δV)
         for n = 1:size(ψ[ik], 2)
             add_response_from_band!(δρk, n, ham.blocks[ik], eigenvalues[ik], ψ[ik],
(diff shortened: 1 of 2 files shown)


src/postprocess/chi0.jl
```diff
@@ -155,58 +151,79 @@ is false, only compute excitations inside the provided orbitals.
...
-        dos = DOS(εF, basis, eigenvalues, temperature=temperature)
+        dos  = DOS(εF, basis, eigenvalues, temperature=temperature)
         δρ .+= ldos .* dot(ldos, δV) .* dVol ./ dos
     end
 
     from_real(basis, δρ .* normδV)
 end
+
+
+function add_response_from_band_!(δρk, hamk, εk, ψk, n, εF, δV,
```
antoine-levitt, line 185: OK the underscore madness has to stop. Do we use underscores before the function name, after the function name, or not at all?

antoine-levitt: Add a very short doc as to what this function is doing. Maybe also put the "n" argument a bit before so that it's clearer it's a crucial input

src/postprocess/chi0.jl
```diff
@@ -155,58 +151,79 @@ is false, only compute excitations inside the provided orbitals.
...
-            end
+        hamk = ham.blocks[ik]
+        εk = eigenvalues[ik]
+        ψk = ψ[ik]
+
+        # One δρk accumulator per thread
+        δρk = [zero(δV) for tid = 1:Threads.nthreads()]
+        Threads.@threads for n = 1:size(ψ[ik], 2)
+            tid = Threads.threadid()
+            add_response_from_band_!(δρk[tid], hamk, εk, ψk, n, εF, δV,
```
antoine-levitt, line 162: Nice splitting, makes the structure clearer. Just inline the hamk, epsk, psik.

PR #286, opened 2020-07-24
Polish dielectric example and improve chi0 computation

mfherbst:


src/postprocess/chi0.jl
```diff
@@ -180,7 +186,7 @@ is false, only compute excitations inside the provided orbitals.
                 # ∑_{n,m != n} (fn-fm)/(εn-εm) ρnm <ρmn|δV>
                 ρnm = conj(ψnk_real) .* ψmk_real
                 weight = dVol * dot(ρnm, δV)
-                δρk .+= real(ratio .* weight .* ρnm)
+                δρk .+= (n == m ? 1 : 2) * real(ratio .* weight .* ρnm)
```
antoine-levitt, line 189: Can you add a comment either here or in the loop above? `for m = n:nbands` is very very easy to confuse with `for m=1:nbands`

.travis.yml
```diff
@@ -19,7 +19,7 @@ script: |  # Install dependencies and test with coverage
 
     # Some optional python packages via conda
     using Conda
-    for pkg in ["ase", "matplotlib =3.2.2"]
+    for pkg in ["nomkl", "ase", "matplotlib =3.2.2"]
```
antoine-levitt, line 22: Add a comment for why this is necessary?

src/postprocess/chi0.jl
antoine-levitt: Also fuse the multiplication

examples/dielectric.jl
```diff
@@ -19,18 +20,19 @@ model = model_LDA(lattice, atoms, symmetry=:off)
 basis = PlaneWaveBasis(model, Ecut; kgrid=kgrid)
 scfres = self_consistent_field(basis, tol=1e-14)
 
-# Apply ε = 1 - K χ0
-function eps_fun(dv)
-    dv = reshape(dv, size(scfres.ρ.real))
-    dρ = apply_χ0(scfres.ham, scfres.ψ, scfres.εF, scfres.eigenvalues, from_real(basis, dv))
-    Kdρ = apply_kernel(basis, dρ; ρ=scfres.ρ)
-    vec(dv - Kdρ.real)
+# Apply ε = 1 - χ0 (vc + fxc)
```
antoine-levitt, line 23: Eps is the dielectric operator, and applies to potentials conventionally. Call that epsT or something?

examples/dielectric.jl
```diff
@@ -41,20 +43,23 @@ function arnoldi(f, x0, howmany; tol=1e-4, maxiter=30)
...
+        is_converged = (N ≥ howmany && all(normr[1:howmany] .< tol)
+                        && all(normr[end-howmany:end] .< tol))
+        if is_converged || (N ≥ maxiter)
+            return (λ=ew, X=Vr, AX=AVr, residual_norms=normr)
         end
     end
 end
-howmany = 5
-arnoldi(eps_fun, vec(randn(size(scfres.ρ.real))), howmany)
+arnoldi(eps_fun, vec(randn(size(scfres.ρ.real))); howmany=5)
```
antoine-levitt, line 65: We should replace that by krylovkit's native implementation anyway now that they have the "greedy" option. @gkemlin is doing that, we'll just merge later

src/scf/self_consistent_field.jl
```diff
@@ -130,7 +133,7 @@ Solve the Kohn-Sham equations with a SCF algorithm, starting at ρ.
     # Callback is run one last time with final state to allow callback to clean up
     info = (ham=ham, basis=basis, energies=energies, converged=fpres.converged,
             ρ=ρout, eigenvalues=eigenvalues, occupation=occupation, εF=εF,
-            n_iter=n_iter, ψ=ψ, stage=:finalize)
+            n_iter=n_iter, ψ=ψ, diagonalization=info.diagonalization, stage=:finalize)
```
antoine-levitt, line 136: Diagonalization_info? We probably need a more consistent naming for nested named tuples. Eg args for input and info for output, so this could be diag_info and the input parameters diag_args

PR #287, opened 2020-07-31
Start a list of lecture notes and recordings

mfherbst:
Just thought it would be nice to collect this somewhere on our website. Any thoughts?


antoine-levitt:
Cool! Maybe also add a list of reference material? Eg the book by Martin and Lin/Lu?

mfherbst:
I had this idea as well, but wasn't sure. I'll do so right away.

PR #288, opened 2020-08-08
Integrate with Aqua.jl 

mfherbst:


antoine-levitt:
Seems useful! Why not test all?

mfherbst:
Because it does not work as of now.

PR #303, opened 2020-09-11
Speedup forces evaluation

antoine-levitt:
I did not microoptimize, just did the stupid stuff. It's definitely in the realm of reasonableness now (ie faster than an SCF step)


antoine-levitt:
The reason why it was so stupidly slow is that it did P C P' psi, and multiplications are left associative.

antoine-levitt:
Test failure is a weird dependency thing...

antoine-levitt:
This is now fast enough to not matter too much. The code is still asymptotically suboptimal (ie we scale like the cost of an energy computation times the number of atoms, and we should be able to "reverse-diff" it to the cost of a single energy computation; I wonder if people do this in other codes). For 60 Si atoms it was about one SCF step, so let's not bother for now. CC @gkemlin

src/terms/local.jl
```diff
@@ -104,13 +104,22 @@ end
...
-                                        ./ sqrt(unit_cell_volume)
-                                        for (iG, G) in enumerate(G_vectors(term.basis))))
+                forces[iel][ir] = _force(term.basis, ρ.fourier, unit_cell_volume, form_factors, r)
         end
     end
     forces
 end
+
+# function barrier to work around type instability
+function _force(basis, ρfour, unit_cell_volume, form_factors, r)
```
mfherbst, line 114: Can you give it a slightly longer name and also `ρfour`, seriously?

mfherbst (approved):
Nice catch.

PR #304, opened 2020-09-20
Add support for Hamiltonians modelling Anyons

antoine-levitt:
Also includes a couple of misc fixes and improvements. I tested it and it works, but I'm not including it as an example yet


antoine-levitt:
Code validated and working (for now at least...), good to merge (some of the other fixes are good to have more generally). I also tweaked the parameters of GPE2D to validate it against my old paper, so I'm reasonable confident there's no big mistake in the formulas.
I'll add an anyon example once things settle down.

antoine-levitt:
@mfherbst OK to merge when tests pass? I'd like to start from this for the g_vectors_cart stuff

mfherbst:
I did not take a look so far. I didn't know you were done.

antoine-levitt:
I wrote "good to merge" above. I know you like the WIP tags and draft PRs, but I never remember to do it when I create the PR, I'll try to remember next time

mfherbst:
Ups my bad, I overlooked that comment, sorry.

src/Model.jl
```diff
@@ -65,17 +65,19 @@ The `symmetry` kwarg can be:
...
                smearing=nothing,
                spin_polarization=:none,
                symmetry=:auto
-               ) where {T <: Real}
-
+               ) where {Tlatt <: Real}
+    # if the lattice was passed as integers, use doubles instead
+    T = Tlatt <: Integer ? Float64 : Tlatt
     lattice = Mat3{T}(lattice)
+    temperature = T(temperature)
```
mfherbst, line 80: This can be solved much more cleanly by adding a method that specialises for integer lattices.

src/RealFourierArray.jl
```diff
@@ -39,10 +40,22 @@ end
...
+        # arr[1] is G=0, arr[1] is G=1, arr[N] is G=-1.
+        # So 1 -> 1, 2 -> N, ..., N -> 2
+        reflect(i, N) = i == 1 ? 1 : N-i+2
+        # Eg if N = 3, we get [0 1 -1], if N = 4, we get [0 1 2 -1] => we check to div(N+1,2)
+        for i = 1:div(basis.fft_size[1]+1, 2)
+            for j = 1:div(basis.fft_size[2]+1, 2)
+                for k = 1:div(basis.fft_size[3]+1, 2)
+                    err = abs(fourier_part[i, j, k] -
+                              conj(fourier_part[reflect(i, end), reflect(j, end), reflect(k, end)]))
+                    err > sqrt(eps(real(T))) && error()
```
mfherbst, line 55: Error message please

src/RealFourierArray.jl
```diff
@@ -10,6 +10,7 @@ import Base.eltype, Base.size, Base.length
 ## TODO this always allocates both real and fourier parts, which might be a bit wasteful
 ## TODO implement broadcasting. This seems to be non-trivial, so I just implement a few simple methods
 ## TODO implement compressed representation in Fourier space of real arrays
+## TODO generalize to AbstractArray
```
mfherbst, line 13: Where do you need that to be an Array? Can you add a comment why it cannot be generalised yet?

src/densities.jl
```diff
@@ -79,3 +79,23 @@ be one coefficient matrix per k-Point.
     count = sum(length(basis.ksymops[ik]) for ik in 1:length(basis.kpoints))
     from_fourier(basis, sum(ρaccus) / count)
 end
+
+"""
+Computes the *probability* (not charge) current, ∑ fn Im(ϕn* ∇ϕn)
+"""
+function compute_current(basis::PlaneWaveBasis, ψ::AbstractVector,
```
mfherbst, line 86: I would rename it to `compute_probability_current` because it is a very odd use case and definitely not what I would expect from the name.

mfherbst: Why do you explicitly type \psi and occupation?

src/densities.jl
```diff
@@ -79,3 +79,23 @@ be one coefficient matrix per k-Point.
...
+                         occupation::AbstractVector)
+    @assert all(symop -> length(symop) == 1, basis.ksymops) == 1  # TODO lift this
```
mfherbst, line 88: I think we should have a function for that that checks whether symmetry is on for a model or the basis.

src/densities.jl
```diff
@@ -79,3 +79,23 @@ be one coefficient matrix per k-Point.
     count = sum(length(basis.ksymops[ik]) for ik in 1:length(basis.kpoints))
     from_fourier(basis, sum(ρaccus) / count)
 end
+
+"""
+Computes the *probability* (not charge) current, ∑ fn Im(ϕn* ∇ϕn)
```
mfherbst, line 84: Why is this in this file?

src/densities.jl
```diff
@@ -79,3 +79,23 @@ be one coefficient matrix per k-Point.
...
+"""
+function compute_current(basis::PlaneWaveBasis, ψ::AbstractVector,
+                         occupation::AbstractVector)
+    @assert all(symop -> length(symop) == 1, basis.ksymops) == 1  # TODO lift this
+    current = [zeros(eltype(basis), basis.fft_size) for α = 1:3]
+    for (ik, kpt) in enumerate(basis.kpoints)
+        for (n, ψnk) in enumerate(eachcol(ψ[ik]))
+            ψnk_real = G_to_r(basis, kpt, ψnk)
+            for α = 1:3
+                dαψnk = [im*(basis.model.recip_lattice * (G+kpt.coordinate))[α] for G in G_vectors(kpt)] .* ψnk
```
mfherbst, line 94: Shall we have a helper function for taking derivatives?

src/terms/Hamiltonian.jl
```diff
@@ -81,7 +81,6 @@ function LinearAlgebra.mul!(Hψ, H::Hamiltonian, ψ)
 end
 # need `deepcopy` here to copy the elements of the array of arrays ψ (not just pointers)
 Base.:*(H::Hamiltonian, ψ) = mul!(deepcopy(ψ), H, ψ)
-
```
mfherbst, line 84: That empty line was on purpose.

src/terms/Hamiltonian.jl
```diff
@@ -90,8 +89,21 @@ Base.:*(H::Hamiltonian, ψ) = mul!(deepcopy(ψ), H, ψ)
...
+        ret = []
+        for a in arr
+            if a isa RealFourierOperator
+                push!(ret, a)
+            else
+                push!(ret, a...)
+            end
+        end
+        ret
+    end
```
mfherbst, line 104: Why do you need this? Notice that `[1...]` and `[(1, 2)...]` does what you want.

src/densities.jl
mfherbst, line 94: Line too long

src/densities.jl
```diff
@@ -79,3 +79,23 @@ be one coefficient matrix per k-Point.
...
+                dαψnk_real = G_to_r(basis, kpt, dαψnk)
+                current[α] .+= @. basis.kweights[ik] * occupation[ik][n] * imag(conj(ψnk_real) * dαψnk_real)
```
mfherbst, line 96: Line too long

src/terms/operators.jl
```diff
@@ -111,16 +111,16 @@ Matrix(op::NonlocalOperator) = op.P * op.D * op.P'
...
+    basis::PlaneWaveBasis{T}
+    kpoint::Kpoint{T}
+    Apot  # Apot[α][i,j,k] is the A field in (cartesian) direction α
 end
 @timing_seq "apply MagneticFieldOperator" function apply!(Hψ, op::MagneticFieldOperator, ψ)
     # TODO this could probably be better optimized
     for α = 1:3
         all(op.Apot[α] .== 0) && continue
-        pα = [(G[α] + op.kpoint.coordinate[α]) for G in G_vectors(op.kpoint)]
+        pα = [(op.basis.model.recip_lattice*(G + op.kpoint.coordinate))[α] for G in G_vectors(op.kpoint)]
```
mfherbst, line 123: Add a space around `*` and line too long.

src/terms/terms.jl
```diff
@@ -41,15 +48,9 @@ include("ewald.jl")
...
 include("entropy.jl")
 include("magnetic.jl")
-
-# breaks_symmetries on a term builder answers true if this term breaks
-# the symmetries of the lattice/atoms (in which case kpoint reduction
-# is invalid)
-breaks_symmetries(term_builder::Any) = false
 breaks_symmetries(term_builder::Magnetic) = true
-breaks_symmetries(term_builder::ExternalFromReal) = true
-breaks_symmetries(term_builder::ExternalFromFourier) = true
```
mfherbst, line 51: Actually I liked it more as a block ... first include, then extend.

test/hamiltonian_consistency.jl
```diff
@@ -6,17 +6,13 @@ include("testcases.jl")
     using Random
     Random.seed!(0)
 
-    function test_consistency_term(term; rtol=1e-3, atol=1e-8, ε=1e-8)
-        testcase = silicon
-        Ecut = 10
-        lattice = testcase.lattice
+    function test_consistency_term(term; rtol=1e-3, atol=1e-8, ε=1e-8, kgrid=[1, 2, 3], lattice=silicon.lattice, Ecut=10)
```
mfherbst, line 9: Line too long.

test/hamiltonian_consistency.jl
```diff
@@ -32,21 +28,21 @@ include("testcases.jl")
         ρ_trial = compute_density(basis, ψ_trial, occupation)
 
         @assert length(basis.terms) == 1
-        E0, ops = ene_ops(basis.terms[1], ψ, occupation; ρ=ρ)
-        E1, _ = ene_ops(basis.terms[1], ψ_trial, occupation; ρ=ρ_trial)
-        diff = (E1 - E0)/ε
+        E0, ham = energy_hamiltonian(basis, ψ, occupation; ρ=ρ)
+        E1, _ = energy_hamiltonian(basis, ψ_trial, occupation; ρ=ρ_trial)
+        diff = (E1.total - E0.total)/ε
```
mfherbst, line 33: Why do you need this? Because for the anyons an energy term may return two hamiltonian terms?

src/terms/anyonic.jl
```diff
@@ -0,0 +1,91 @@
...
+            A2[iG] = +2T(π) * G[1] / G2 * ρ.fourier[iG] * im
+        end
+    end
+    Areal = [from_fourier(basis, A1).real,
+             from_fourier(basis, A2).real,
+             zeros(T, basis.fft_size)]
+
+    # 2β (-i∇)⋅A + β^2 |A|^2
+    ops_energy = [MagneticFieldOperator(basis, basis.kpoints[1], 2β .* Areal),
+                  RealSpaceMultiplication(basis, basis.kpoints[1], β^2 .* (abs2.(Areal[1]) .+ abs2.(Areal[2])))]
```
mfherbst, line 59: Line too long.

src/terms/anyonic.jl
```diff
@@ -0,0 +1,91 @@
...
+#     so for consistency the added kinetic energy must have a scaling of 2
+# Energy = <u, ((-i∇+βA)^2 + V) u>
+# where ∇∧A = 2π ρ, ∇⋅A = 0 => A = x⟂/|x|² ∗ ρ
+# H = (-i∇+βA)² + V - 2β x⟂/|x|² ∗ (βAρ + J)
+#   = -Δ + 2β (-i∇)⋅A + β²|A|^2 - 2β x⟂/|x|² ∗ (βAρ + J)
+# where only the first three terms "count for the energy", and where
+# J = 1/(2i) (u* ∇u - u ∇u*)
+
+struct Anyonic
+    β
```
mfherbst, line 13: What is beta?

src/terms/anyonic.jl
```diff
@@ -0,0 +1,91 @@
+# Ref https://arxiv.org/pdf/1901.10739.pdf
+# This term does not contain the kinetic energy, which must be added separately
+# /!\ They have no 1/2 factor in front of the kinetic energy,
+#     so for consistency the added kinetic energy must have a scaling of 2
+# Energy = <u, ((-i∇+βA)^2 + V) u>
+# where ∇∧A = 2π ρ, ∇⋅A = 0 => A = x⟂/|x|² ∗ ρ
```
mfherbst, line 6: Can you write `x^⟂` just to make clear the `⟂` is not an operator, but the orthogonal component.

src/terms/anyonic.jl
```diff
@@ -0,0 +1,91 @@
...
+# H = (-i∇+βA)² + V - 2β x⟂/|x|² ∗ (βAρ + J)
```
mfherbst, line 7: Space before / after +

src/terms/anyonic.jl
```diff
@@ -0,0 +1,91 @@
+# Ref https://arxiv.org/pdf/1901.10739.pdf
+# This term does not contain the kinetic energy, which must be added separately
+# /!\ They have no 1/2 factor in front of the kinetic energy,
+#     so for consistency the added kinetic energy must have a scaling of 2
+# Energy = <u, ((-i∇+βA)^2 + V) u>
```
mfherbst, line 5: Space before / after +

src/terms/anyonic.jl
```diff
@@ -0,0 +1,91 @@
...
+    Areal = [from_fourier(basis, A1).real,
+             from_fourier(basis, A2).real,
+             zeros(T, basis.fft_size)]
+
+    # 2β (-i∇)⋅A + β^2 |A|^2
+    ops_energy = [MagneticFieldOperator(basis, basis.kpoints[1], 2β .* Areal),
+                  RealSpaceMultiplication(basis, basis.kpoints[1], β^2 .* (abs2.(Areal[1]) .+ abs2.(Areal[2])))]
+
+    # Now compute effective local potential - 2β x⟂/|x|² ∗ (βAρ + J)
+    current = compute_current(basis, ψ, occ)
```
mfherbst, line 62: You might as well call it `J` to be consistent with your formula.

src/terms/anyonic.jl
```diff
@@ -0,0 +1,91 @@
...
+    eff_current = [current[α].real .+
+                   β .* ρ.real .* Areal[α] for α = 1:2]
+    eff_current_fourier = [from_real(basis, eff_current[α]).fourier for α = 1:2]
```
mfherbst, line 65: `r_to_G`?

src/terms/anyonic.jl
```diff
@@ -0,0 +1,91 @@
...
+    # eff_pot = - 2β x⟂/|x|² ∗ eff_current
+    # => ∇∧eff_pot = -4πβ eff_current
+    # => eff_pot(G1, G2) = -4πβ i eff_current(G1, G2) * [-G2;G1;0]/(G1^2 + G2^2)
+    eff_pot_fourier = zeros(complex(T), basis.fft_size)
+    for (iG, Gred) in enumerate(G_vectors(basis))
+        G = basis.model.recip_lattice * Gred
+        G2 = sum(abs2, G)
+        if G2 != 0
+            eff_pot_fourier[iG] += +4T(π)*β * G[2] / G2 * eff_current_fourier[1][iG] * im
```
mfherbst, line 74: I would keep the order in the formula (put the im after β)

src/terms/anyonic.jl
```diff
@@ -0,0 +1,91 @@
+# Ref https://arxiv.org/pdf/1901.10739.pdf
```
mfherbst, line 1: I did not check the formulas here ...

mfherbst (commented):
I'm not completely sure about the design. On the one hand what you have in the Anyonic term is the equivalent of the kinetic energy for Anyons, no? Then it misses the "normal" kinetic energy. On the other hand it actually adds more than one term. Could you not just each of them as a separate term?

I have not completely read the paper (and so I misunderstand what you do). It just feels weird to me that you need to break the pattern.

antoine-levitt:
Basically the operator is (p+A)^2 where A is given self-consistently by rho. I expand the square which gives p^2 + 2 Ap + A^2 (A and p commute because div A = 0) and I compute it like that. Anyonic does 2Ap + A^2 but not p^2, so it doesn't include the kinetic energy. There's no real choice but to have more than one operator (or we have some CompositeOperators, which is inefficient and nasty to handle).

src/Model.jl
antoine-levitt: You're right, somehow that didn't come to me

src/RealFourierArray.jl
antoine-levitt: You know how I feel about this kind of internal error message: it's not more informative than a stacktrace. But fine :-p

src/RealFourierArray.jl
antoine-levitt: It definitely can, I've just been too lazy to figure out the correct dispatch here. In any case we'll have other troubles when we do get to non-Array types, so we'll see this when we need to

src/densities.jl
antoine-levitt: It's pretty clear no? It's the current of electrons (equal to minus the charge current). It's the one that's consistent with the electronic charge rho. I can call it electronic current if you want.
What do you mean by explicitly type?

src/densities.jl
antoine-levitt: I'm fine with this, it's not like we do it a lot, and it's all when we're too lazy to figure out the right expressions

src/densities.jl
antoine-levitt: It's pretty similar to a density, it's a "current density"... I don't want to create a `current.jl` file just for this.

src/densities.jl
antoine-levitt: It's pretty clear what it does. We do need a simpler way to do this though.
```
                dαψnk = [im*(G+kpt.coordinate_cart)[α] for G in G_vectors_cart(kpt)] .* ψnk
```
?

src/terms/Hamiltonian.jl
antoine-levitt: I need it to flatten the list (when operators return more than one term). I'm all ears if you have a better solution

src/densities.jl
antoine-levitt: will be taken care of by the _cart

src/terms/operators.jl
antoine-levitt: Also will be taken care of by the _cart

src/terms/terms.jl
antoine-levitt: ?

test/hamiltonian_consistency.jl
antoine-levitt: Yes, and it's cleaner

src/terms/anyonic.jl
antoine-levitt: it's not really supposed to be read without the paper

src/terms/anyonic.jl
antoine-levitt: not going to bother. These are implementation comments recalling the formulas in the (obscure) paper. If it keeps you up all nights by all means go ahead and format it

src/terms/anyonic.jl
antoine-levitt: We're not really consistent with this in the code... I always have trouble with r_to_G expecting complex numbers

src/terms/anyonic.jl
antoine-levitt: I did :-)

src/RealFourierArray.jl
antoine-levitt: Fixed it, needed it for passing tests anyway

mfherbst: Sure I agree we are not ready for non-array types, I just wondered why you decided to make it less abstract.

src/densities.jl
mfherbst: ok

src/densities.jl
mfherbst: I think more files are better. In my head it is definitely a different concept.

src/RealFourierArray.jl
antoine-levitt: For performance. It killed the for loops operating on RFAs

src/densities.jl
mfherbst: Like applying the gradient? I agree. It's used quite a bit for the XC stuff as well. Let's wait until I get to do the spin.

src/densities.jl
antoine-levitt: It kind of belongs with DOSes as "density-like things we might want to compute". I can put it there in postprocess

src/terms/Hamiltonian.jl
mfherbst: It seems it somehow should, because this is horrible. But what I came up with was no better.

src/densities.jl
mfherbst: which is not there yet. So than make a TODO to remind yourself to fix it later.

src/terms/operators.jl
mfherbst: same as above.

src/terms/terms.jl
mfherbst: I do not like your restructuring. I think it's clearer to have the `breaks_symmetries` as a block below the includes.

test/hamiltonian_consistency.jl
mfherbst: ok.

src/terms/anyonic.jl
mfherbst: hmm ok.

src/densities.jl
antoine-levitt: I'll do it soon, and I'll just grep for lattice so it won't get forgotten

src/densities.jl
mfherbst: Why you make them `AbstractVector` instead of `Any`

mfherbst: For the function: Maybe I misunderstood. Is this the generic expression for computing the electronic current density for bloch waves?

src/terms/Hamiltonian.jl
antoine-levitt: Yeah I know... haven't found anything more clever either

src/densities.jl
antoine-levitt: Yes: https://en.wikipedia.org/wiki/Probability_current

src/Model.jl
```diff
@@ -74,8 +74,9 @@ function Model(lattice::AbstractMatrix{T};
                spin_polarization=:none,
                symmetry=:auto
                ) where {T <: Real}
-
+    # if the lattice was passed as integers, use doubles instead
```
mfherbst, line 77: This comment is obsolete.

src/Model.jl
```diff
@@ -134,6 +135,7 @@ function Model(lattice::AbstractMatrix{T};
     Model{T}(lattice, recip_lattice, unit_cell_volume, recip_cell_volume, d, n_electrons,
              spin_polarization, T(temperature), smearing, atoms, terms, symops)
 end
+Model(lattice::AbstractMatrix{T};kwargs...) where {T <: Integer} = Model(Float64.(lattice); kwargs...)
```
mfherbst, line 138: Space after `;`

src/RealFourierArray.jl
```diff
@@ -18,10 +18,12 @@ A structure to facilitate manipulations of an array of real-space type T, in bot
 and fourier space. Create with `from_real` or `from_fourier`, and access
 with `A.real` and `A.fourier`.
 """
-mutable struct RealFourierArray{T <: Real}
+mutable struct RealFourierArray{T <: Real,
+                                Treal <: AbstractArray{T, 3},
```
mfherbst, line 22: `Treal` ... not sure. Have not used that to mean `real(T)` at some place? How about `RealArray` and `ComplexArray`?

src/terms/terms.jl
antoine-levitt: I did it this way because it's attached to each kind of term. When I added the new term I forgot to add the break_symmetry annotation, so I figured it would be clearer this way

src/densities.jl
mfherbst: Your name in god's ear as we say in german ;).

src/terms/terms.jl
mfherbst: Ok I see, but it depends what you're after. If you are reading the code and you want to understand what `breaks_symmetries` is doing, than it's better the old way. Question is what happens more often.

src/terms/anyonic.jl
mfherbst: Well I kind of meant at each occurrence of `x⟂` ...

src/terms/terms.jl
antoine-levitt: Less possibility of bugs trumps (very minor) decrease in readability in this case I would say

src/terms/anyonic.jl
mfherbst: Yes that's true, but I think we should decide. What you do here feels a bit convoluted, because you are only doing it for running the `r_to_G`. What is the issue with `r_to_G(complex(xyz))`?

src/terms/anyonic.jl
mfherbst: Ok. I'll do it ...

src/terms/terms.jl
mfherbst: Ok, but then at least make a space after each "include and extend" block

src/terms/anyonic.jl
antoine-levitt: Again, it's an implementation comment of a pretty obscure feature... but fine. It doesn't drive you mad that I have sometiems ^2 and sometimes ²? :-p (I tried and figured out it was too annoying to type and didn't work in julia so gave up)

src/RealFourierArray.jl
```diff
@@ -19,11 +19,11 @@ and fourier space. Create with `from_real` or `from_fourier`, and access
 with `A.real` and `A.fourier`.
 """
 mutable struct RealFourierArray{T <: Real,
-                                Treal <: AbstractArray{T, 3},
-                                Tcomplex <: AbstractArray{Complex{T}, 3}}
+                                T_real_array <: AbstractArray{T, 3},
+                                T_complex_array <: AbstractArray{Complex{T}, 3}}
```
mfherbst, line 23: We have decided to camel-case types.

antoine-levitt:
Tests fail on macos because of some dep thing, is this because of your recent changes to the deps?

mfherbst:
It might have something to do with the `nomkl`, but I wonder why it only happens in this branch. I've read about a PyCall bug where the julia process does not look in the proper place for compiled parts of python libraries. Could have something to do with that.

src/PlaneWaveBasis.jl
```diff
@@ -24,9 +26,13 @@ end
...
 
 """
-The list of G vectors of a given `basis` or `kpoint`.
+The list of G vectors of a given `basis` or `kpoint`, in reduced coordinates.
 """
 G_vectors(kpt::Kpoint) = kpt.G_vectors
+"""
+The list of G vectors of a given `basis` or `kpoint`, in cartesian coordinates.
+"""
+G_vectors_cart(kpt::Kpoint) = (kpt.model.recip_lattice * G for G in G_vectors(kpt))
```
mfherbst, line 35: I don't like having another function for this. I would prefer `G_vectors(abc, cartesian=true)` or `G_vectors(abs, :cartesian)`

src/PlaneWaveBasis.jl
```diff
@@ -12,8 +12,10 @@ More generally, a kpoint is a block of the Hamiltonian;
 eg collinear spin is treated by doubling the number of kpoints.
 """
 struct Kpoint{T <: Real}
+    model::Model{T}
     spin::Symbol                     # :up, :down, :both or :spinless
     coordinate::Vec3{T}              # Fractional coordinate of k-Point
+    coordinate_cart::Vec3{T}         # Cartesian coordinate of k-Point
```
mfherbst, line 18: cart =>cartesian ?

src/PlaneWaveBasis.jl
antoine-levitt: What is the point apart from being more verbose?

src/PlaneWaveBasis.jl
antoine-levitt: I actually think coordinate is too verbose :p

src/PlaneWaveBasis.jl
mfherbst: I don't like to abbreviate it as `cart` and it is one function (ok weak point).

src/PlaneWaveBasis.jl
mfherbst: But cartesian is shorter, so that's ok, no ;)

src/PlaneWaveBasis.jl
antoine-levitt: Why? cart and coord are standard, unmistakeable abbreviations. The one function I would buy if there was more than two coordinates sytems, but it just seems more trouble than not

mfherbst: ok fine, convinced.

mfherbst:
I know what's causing the failing tests. In MacOS it now installs nomkl in conda from main, https://travis-ci.com/github/JuliaMolSim/DFTK.jl/jobs/390324148#L677, but it used to use conda forge https://travis-ci.com/github/JuliaMolSim/DFTK.jl/jobs/389804416#L676 which it still does in linux https://travis-ci.com/github/JuliaMolSim/DFTK.jl/jobs/390324147#L784.

PR #305, opened 2020-09-21
Renames to bring in line with paper and interval-related fixes

mfherbst:


src/PlaneWaveBasis.jl
```diff
@@ -135,7 +135,7 @@ build_kpoints(basis::PlaneWaveBasis, kcoords) =
     # operations are inverse to each other). The convention we want is
     # ψ(r) = sum_G c_G e^iGr / sqrt(Ω)
     # so that the ifft is normalized by 1/sqrt(Ω). It follows that the
-    # fft must be normalized by sqrt(Ω)/length
+    # fft must be normalized by sqrt(Ω/length)
```
antoine-levitt, line 138: ?

src/scf/mixing.jl
```diff
@@ -136,15 +140,15 @@ end
         δρ = devec(x)
         Jδρ = copy(δρ)
 
-        # Apply Kernel (just vc for RPA and (vc + fxc) if not RPA)
+        # Apply Kernel (just vc for RPA and (vc + K_{xc}) if not RPA)
```
antoine-levitt, line 143: refer to the paper somewhere?

mfherbst: Right at the top?

antoine-levitt: Yeah for instance, or in the doc of hybrid (or both)

src/PlaneWaveBasis.jl
mfherbst: ups ... not sure how that happened.

PR #310, opened 2020-09-23
Make HybridMixing the default

mfherbst:


antoine-levitt:
Good idea. The hybrid mixing has not really been optimized, setting it to default will make us notice it if it messes up

mfherbst:
Also it's plainly the better default ;).

antoine-levitt:
Only useful for kind of large systems so let's have the simpler one by default.

Issue #307, opened 2020-09-22
Symmetry Issue with HCP Lattice


umbriquse:
The following code is attempting to obtain the self consistent field for an HCP lattice for Platinum. The error that is produced can be suppressed by setting the symmetry option for the Model to `:off`.

 I don't understand why an HCP lattice throws this type of error, or why a lattice has to be a Unitary rotation matrix in order to use symmetry operations.

# Code
```
using DFTK
using Unitful, UnitfulAtomic

latticeConstant = 10
α = [1, 1, 1.633]
HCPLattice = 0.5 * [
    [2 0 0];
    [1 √3 0];
    [0 0 2]
] .* α

element = ElementPsp(:Pt, psp = load_psp(list_psp(:Pt; functional = "lda")[2].identifier))
temp = austrip(300u"K")
atoms = [element => [[0., 0, 0], ones(3)/3]]

cutoffEnergy = 10.0
kGrid = ones(Int,3) * 4

lattice = latticeConstant * HCPLattice

model = model_LDA(lattice, atoms; temperature = temp)
basis = PlaneWaveBasis(model, cutoffEnergy; kgrid = kGrid)
scfGroundState = self_consistent_field(basis)
@info scfGroundState.energies
```

# Error
```
ERROR: LoadError: spglib returned non-unitary rotation matrix
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] spglib_get_symmetry(::SArray{Tuple{3,3},Float64,2,9}, ::Array{Pair{ElementPsp,Array{Array{Float64,1},1}},1}; tol_symmetry::Float64) at /Users/jasonlehto/.julia/packages/DFTK/dtzTz/src/external/spglib.jl:82
 [3] symmetry_operations(::SArray{Tuple{3,3},Float64,2,9}, ::Array{Pair{ElementPsp,Array{Array{Float64,1},1}},1}; tol_symmetry::Float64, kcoords::Nothing) at /Users/jasonlehto/.julia/packages/DFTK/dtzTz/src/symmetry.jl:51
 [4] symmetry_operations at /Users/jasonlehto/.julia/packages/DFTK/dtzTz/src/symmetry.jl:49 [inlined]
[...]

PR #314, opened 2020-09-24
Fix spglib wrapper

antoine-levitt:
Fixes https://github.com/JuliaMolSim/DFTK.jl/issues/307


src/external/spglib.jl
```diff
@@ -116,7 +113,7 @@ function spglib_standardize_cell(lattice::AbstractArray{T}, atoms; correct_symme
     # Ask spglib to standardize the cell (i.e. find a cell, which fits the spglib conventions)
     num_atoms = ccall((:spg_standardize_cell, SPGLIB), Cint,
       (Ptr{Cdouble}, Ptr{Cdouble}, Ptr{Cint}, Cint, Cint, Cint, Cdouble),
-      spg_lattice, spg_positions, spg_numbers, length(spg_numbers), Cint(primitive),
+      copy(spg_lattice'), spg_positions, spg_numbers, length(spg_numbers), Cint(primitive),
```
mfherbst, line 116: This won't work, because spglib modifies the lattice in-place.

mfherbst (commented):
Did you test this with the failing case?

mfherbst:
Cool. Thanks for doing that. I'll add a test for the failing case in #307 before merging.

src/external/spglib.jl
antoine-levitt: Good catch!

antoine-levitt:
Great! (I hate tests)

mfherbst:
Who doesn't, but I have the pipeline after all ...

mfherbst:
Failing test on MacOS is the same issue as #304. I added a fix there ... let's hope that cuts it.

PR #316, opened 2020-09-25
Setup bors

mfherbst:
Modifications to setup [bors](https://bors.tech/).


antoine-levitt:
So we just do `bors r+` on PRs we approve of now?

PR #327, opened 2020-10-07
Collinear spin for GGA functionals

mfherbst:
- [X] silicon no ρspin works
- [X] silicon random ρspin relaxes to proper non-spin-broken solution
- [X] Hamiltonian consistency works
- [x] Agreement with ABINIT on spin-polarised iron
- [x] Tests


antoine-levitt:
>  Hamiltonian consistency works

I'm so happy we wrote this test, it's super useful.

mfherbst:
Hooray! Got it to work after all ...

src/external/etsf_nanoquanta.jl
```diff
@@ -146,18 +154,29 @@ load_basis(folder; kwargs...) = load_basis(Float64, folder; kwargs...)
...
+    if n_spin == 1
+        @assert basis.model.spin_polarization == :none
+        ρtot_real = Array{T}(folder.den["density"][1, :, :, :, 1])
+        @assert basis.fft_size == size(ρ_real)
+
+        return from_real(basis, ρtot_real), nothing
+    else
+        @assert basis.model.spin_polarization == :collinear
+        ρtot_real = Array{T}(folder.den["density"][1, :, :, :, 1])
+        ρα_real = Array{T}(folder.den["density"][1, :, :, :, 2])
```
antoine-levitt, line 175: They really have rhotot and rhoalpha ?!?! Crazy people.

antoine-levitt:
LGTM!

antoine-levitt:
Looks like SCF convergence is pretty good, even in the spin polarized case.

antoine-levitt:
About 15 iterations. Dielectric eigenvalues should be very interesting here!

src/external/etsf_nanoquanta.jl
mfherbst: Yes ... that surprised me as well.

antoine-levitt:
So playing around with the XC scaling factor it does look like a pitchfork (I haven't done any fancy continuation, just trying several random values), at about a scaling of .35. It would maybe be better to only scale the spin-coupling exchange interaction, without touching the XC of each spin independently. Is this something that can be done easily with libxc?

mfherbst:
Not so easily because libxc internally computes `\zeta = (ρ_α - ρ_β) / (ρ_α + ρ_β)` and that you would need to scale without scaling `ρ_α` and `ρ_β` itself.

If that's an issue for our investigations we might as well code up PBE ourselves. It's still reasonable

antoine-levitt:
You don't really want to scale ζ itself, you'd want to scale the contribution to the energy coming from ζ. The correlation energy is not separable though... So I guess scaling ζ itself is the simplest way. We can hack this by scaling the rhospin taken as input to the xc functional.

mfherbst:
But that's the point. It does not take ρspin. It takes ρα and ρβ.

antoine-levitt:
Yeah but you can always fudge the rho_a and rho_b to have what you want, it's just a change of variables. Essentially I think there are two ways to do it: the energy is F(rhotot, rhospin), so we can do F(rhotot, alpha*rhospin) or (1-alpha) F(rhotot, 0) + alpha F(rhotot, rhospin). The second one is probably better (because you deform energies instead of densities), and simpler to implement. To implement we'd simply need another xc term that doesn't take into account the rhospin, and then you'd just have two XC terms, with different weights.

mfherbst:
Yes that last thing sounds like a reasonable solution.

PR #328, opened 2020-10-09
Kernels for collinear spin

mfherbst:
Kernels for Hartree, power-nonlinearity and LDA functionals for collinear spin.

### DIFF
--- a/src/postprocess/chi0.jl
+++ b/src/postprocess/chi0.jl
@@ -126,7 +126,7 @@ sufficiently converged. By default the `self_consistent_field` routine of `DFTK`
 returns `3` extra bands, which are not converged by the eigensolver
 (see `n_ep_extra` parameter). These should be discarded before using this function.
 """
-@timing function apply_χ0(ham, ψ, εF, eigenvalues, δV::RealFourierArray;
+@timing function apply_χ0(ham, ψ, εF, eigenvalues, δV::RealFourierArray, δVβ=nothing;
                           droptol=0,
                           sternheimer_contribution=true,
                           temperature=ham.basis.model.temperature,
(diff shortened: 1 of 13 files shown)


src/postprocess/chi0.jl
```diff
@@ -126,7 +126,7 @@ sufficiently converged. By default the `self_consistent_field` routine of `DFTK`
 returns `3` extra bands, which are not converged by the eigensolver
 (see `n_ep_extra` parameter). These should be discarded before using this function.
 """
-@timing function apply_χ0(ham, ψ, εF, eigenvalues, δV::RealFourierArray;
+@timing function apply_χ0(ham, ψ, εF, eigenvalues, δV::RealFourierArray, δVβ=nothing;
```
antoine-levitt, line 129: that's pretty ugly, but we'll fix that with the array stuff hopefully

src/terms/terms.jl
```diff
@@ -80,42 +80,67 @@ end
...
-    nothing  # by default, no kernel
-end
-
 function compute_kernel(basis::PlaneWaveBasis{T}; kwargs...) where {T}
-    k = zeros(T, prod(basis.fft_size), prod(basis.fft_size))
+    n_spin = basis.model.n_spin_components
+    kernel = zeros(T, n_spin * prod(basis.fft_size), n_spin * prod(basis.fft_size))
     for term in basis.terms
-        k .+= compute_kernel(term; kwargs...)
+        term == false && continue
```
antoine-levitt, line 106: If we're not going to use the "strong zero" property, might as well return nothing there, no?

antoine-levitt (commented):
Cool!

src/postprocess/chi0.jl
mfherbst: Agree 100%. I just wanted to get it all to work first.

PR #330, opened 2020-10-10
Add simplistic checkpointing to DFTK

### LINKED ISSUE #298: Simplistic checkpointing callback
While we should probably wait until #75 and #33 are done to do checkpointing properly, at least a simplistic checkpointing (ρ and \psi) function using the callback mechanism could be quite useful.

mfherbst:
Based on JLD2 (which is again maintained, yeah!). Should be reasonably robust.

Closes #298 (for now).

- [x] Simplistic `load_scfres` / `save_scfres` (with no guarantees yet)
- [x] Simplistic checkpointing callback
- [x] Example
- [x] Test


antoine-levitt:
I don't understand why you don't just save the struct?

mfherbst:
Because you can't serialise pointers.

antoine-levitt:
Is that really true? If it was you couldn't serialize anything. I'll check when I'm at a computer. Storage wise if it fits in ram it fits on disk. The most storage intensive should be the psi and the things stored in the hamiltonian (everything else is linear scaling with the number of atoms), which you can just nullify if you don't want to use them. Worth a try at least.

mfherbst:
The psi is important for checkpointing. JLD2 explicitly bails out with pointers (like in the FFTW objects or in the Libxc functional), that's the first thing I tried. As of now PlaneWaveBasis cannot be serialised out of the box. So that's the easiest solution.

antoine-levitt:
Ah you mean C pointers, ok I see. Can't it just ignore it?

mfherbst:
I haven't found an easy way to do it. Custom serialisation for custom types is kind of in the making.

antoine-levitt:
I opened an issue at JLD2 to see what they think about this C pointer issue.

mfherbst:
Go ahead if you want, but I'm not sure what can be done about this. If you just drop the pointers you have an inconsistent state (where you could easily provoke a segfault), but you can't store them either, because they likely are invalid when you deserialise. I think the only solution is on a case by case basis using the knowledge of what the pointers mean.

antoine-levitt:
See discussion at https://github.com/JuliaIO/JLD2.jl/issues/251. Yes I agree the context is needed, but this should be doable with a custom serializer. But the best solution in the meantime is to just write a serializer for PlaneWaveBasis that just ignores the plan and recreates them on deserialization. Let's do that (if one of us can be bothered) once the dust settles at JLD2.

mfherbst:
Agreed.

PR #331, opened 2020-10-11
Use Unitful in key constructors

### LINKED ISSUE #297: Integrate with unitful in key constructors
It probably would make life for setting up DFTK calculations a little easier if we integrated with Unitful in key constructors for quantities such as:
- Temperature
- Energy cutoff
- Lattice vectors
- Kgrid density

Continued discussion from #296

berquist:
Closes #297 

This is my first real external Julia contribution, so if I'm not doing something idiomatic, please let me know.

Progress:
- [x] temperature
- [x] energy cutoffs
- [x] lattice vectors
- [x] kgrid density


mfherbst:
Hi @berquist. Thanks for giving this a go. I'll review your PR right away!

test/occupation.jl
```diff
@@ -46,7 +48,7 @@ end
     # See that the electron count still works if we add temperature
     Ts = (0, 1e-6, .1, 1.0)
     for T in Ts, meth in DFTK.Smearing.smearing_methods
-        model = Model(silicon.lattice; n_electrons=silicon.n_electrons, temperature=T, smearing=meth())
+        model = Model(silicon.lattice; n_electrons=silicon.n_electrons, temperature=T * 1u"Eh_au", smearing=meth())
```
mfherbst, line 51: That's exactly the sort of type annotations we wanted to avoid.

From my point of view units are a convenience feature and strictly "opt-in". The idea is that we use atomic units internally in the code and so if nothing is specified it should just default to being atomic.

src/Model.jl
```diff
@@ -80,13 +96,13 @@ function Model(lattice::AbstractMatrix{T};
                atoms=[],
                magnetic_moments=[],
                terms=[],
-               temperature=T(0.0),
+               temperature::Union{T, Unitful.Quantity}=T(0.0),
```
mfherbst, line 99: I would not type-annotate this. If someone uses a very strange type, it'll just crash in the conversion.

src/Model.jl
```diff
@@ -1,5 +1,21 @@
...
+function uconvert(a::Unitful.EnergyUnits, x::Unitful.Quantity{T,Unitful.𝚯,U}) where {T <: Real, U <: Unitful.Units}
+    uconvert(a, Unitful.k * x)
+end
+
+# Convert a value to an energy in atomic units.
+# If given a number, assume it is already in atomic units (pass through directly).
+function to_energy(x::Unitful.Quantity)::Real
+    ustrip(uconvert(UnitfulAtomic.Eh_au, x))
+end
+to_energy(x::Real)::Real = x
```
mfherbst, line 17: Nice idea, but actually no need to be so complicated. Take a look at the `austrip` function from `UnitfulAtomic`. I have never used it, but it seems to do pretty much what we want.

src/Model.jl
```diff
@@ -1,5 +1,21 @@
 # Contains the physical specification of the model
 
+using Unitful
+import Unitful.uconvert
```
mfherbst, line 4: I typically do `import Unitful: uconvert` instead.

src/Model.jl
```diff
@@ -1,5 +1,21 @@
...
+using UnitfulAtomic
+
+# Convert a temperature to an energy via the Boltzmann constant.
+function uconvert(a::Unitful.EnergyUnits, x::Unitful.Quantity{T,Unitful.𝚯,U}) where {T <: Real, U <: Unitful.Units}
```
mfherbst, line 8: This is a bit over-typed (in Julia you try to be as loose on specifying the types as possible). You can simply do `where {T, U}` and that's it.

src/Model.jl
```diff
@@ -1,5 +1,21 @@
...
+    uconvert(a, Unitful.k * x)
+end
+
+# Convert a value to an energy in atomic units.
+# If given a number, assume it is already in atomic units (pass through directly).
+function to_energy(x::Unitful.Quantity)::Real
```
mfherbst, line 14: Here annotating the `::Real` is actually pretty bad. The reason is that if you pass the function a concrete type (let's say a float annotated by the unit) the compiler will infer that the return value is a Float64 and directly compile the function that way. If you however annotate the return type as `::Real` (where `Real` is a supertype), than this *forces* to return the generic type, which in general a lot slower, since it cannot be stored on the stack for example. In this case it should not matter, but there is no reason to put it there, so just don't. That's less writing *and* faster code :smile:.

When to use generic types and when not one needs a little experience, but for starters (especially for people coming from the C++ world like you seem to): Don't put types and only put them if the compiler complains or you really need to!

test/units.jl
```diff
@@ -0,0 +1,16 @@
+using Test
+using Unitful
+using UnitfulAtomic
+using DFTK: uconvert, to_energy
+
+@testset "Check Unitful conversions" begin
+    @test uconvert(Unitful.J, 300u"K") == 4.141947e-21u"J"
+    @test uconvert(UnitfulAtomic.Eh_au, 300u"K") == 0.0009500434690331574u"Eh_au"
+end
```
mfherbst, line 9: That does not belong here. It does not check DFTK, but it checks unitful ;)

test/units.jl
```diff
@@ -0,0 +1,16 @@
+using Test
+using Unitful
+using UnitfulAtomic
+using DFTK: uconvert, to_energy
```
mfherbst, line 4: uconvert is not from DFTK, it's from unitful, so `import Unitful: uconvert`.

antoine-levitt:
Hi and thank you very much for doing this! I'm concerned this is type piracy though, because you define methods that don't belong to us on types that don't belong to us. We want to identify temperatures and energies, but some people might not, won't that break their code? The way I would do it is to have a `to_au` (our own) that we could then overload with units, and there we can do our identification of temperature with energy.

mfherbst (commented):
Nice idea. I think in this case you should try to build more directly on the functions from UnitfulAtomic. I think they have some stuff that suits pretty much our problem.

Regarding the Julia style I made a few annotations on things I noticed.

mfherbst:
> The way I would do it is to have a to_au (our own) that we could then overload with units, and there we can do our identification of temperature with energy.

No need to do this ourselves. UnitfulAtomic has `austrip` and if that misses something we can PR it there, I think.

antoine-levitt:
Ah our comments crossed :) OK but will austrip do the temp to energy conversion? The au unit of temperature is still Kelvin, right?

mfherbst:
Yes, of course :smile: (first thing I checked):
```julia
abs(austrip(1u"K") - 1 * DFTK.units.K) < 1e-9
abs(austrip(1u"eV") - 1 * DFTK.units.eV) < 1e-9
```

antoine-levitt:
Right so if we just do austrip and pass it 300K it'll just convert it to 300 and DFTK will do 300Ha

mfherbst:
What? How did you arrive to that conclusion? Just look at my code ... not at all. It'll do the right thing.

antoine-levitt:
Oh right, I see. That's a bit weird, but it's what we want so indeed everything is good!

src/fft.jl
```diff
@@ -44,7 +44,7 @@ function determine_fft_size(lattice::AbstractMatrix{T}, Ecut;
     fft_size
 end
 function determine_fft_size(model::Model, Ecut; kwargs...)
-    determine_fft_size(model.lattice, Ecut; kwargs...)
+    determine_fft_size(model.lattice, to_energy(Ecut); kwargs...)
```
mfherbst, line 47: That's pretty low-level in the code. At that point I think we should assume that Ecut has already been converted to the proper value.

src/fft.jl
```diff
@@ -63,6 +63,7 @@ end
 #      It should be merged with build_kpoints somehow
 function determine_fft_size_precise(lattice::AbstractMatrix{T}, Ecut, kpoints;
                                     supersampling=2, ensure_smallprimes=true) where T
+    Ecut = to_energy(Ecut)
```
mfherbst, line 66: Same here.

src/PlaneWaveBasis.jl
```diff
@@ -229,10 +233,11 @@ treated explicitly. In this case all guess densities and potential
 functions must agree with the crystal symmetries or the result is
 undefined.
 """
-function PlaneWaveBasis(model::Model, Ecut::Number;
+function PlaneWaveBasis(model::Model,
+                        Ecut::Union{Number, Unitful.Quantity{T,D,Unitful.EnergyUnits}};
```
mfherbst, line 237: Same here.

src/PlaneWaveBasis.jl
```diff
@@ -119,10 +121,12 @@ build_kpoints(basis::PlaneWaveBasis, kcoords) =
     build_kpoints(basis.model, basis.fft_size, kcoords, basis.Ecut)
 
 # This is the "internal" constructor; the higher-level one below should be preferred
-@timing function PlaneWaveBasis(model::Model{T}, Ecut::Number,
+@timing function PlaneWaveBasis(model::Model{T},
+                                Ecut::Union{Number, Unitful.Quantity{T,D,Unitful.EnergyUnits}},
```
mfherbst, line 125: That feels again a bit strongly typed in my opinion. How about just `Union{Number, Quantity}` (I suppose if you are `using Unitful` that `Quantity` is already in the main scope and thus you don't need the `Unitful.`, but check this).

mfherbst (commented):
Thanks for the next set of patches :smile:. Some quick comments ...

src/common/units.jl
```diff
@@ -21,14 +18,8 @@ E.g. `unit_to_au(:eV)` returns the conversion factor from electron volts to Hart
...
-    uconvert(a, Unitful.k * x)
-end
-
 # Convert a value to an energy in atomic units.
 # If given a number, assume it is already in atomic units (pass through directly).
-function to_energy(x::Unitful.Quantity)::Real
-    ustrip(uconvert(UnitfulAtomic.Eh_au, x))
+function to_energy(x)
+    austrip(x)
 end
```
mfherbst: There is not really much point in having this function any more is there? But maybe I miss something.

src/PlaneWaveBasis.jl
berquist: I don't understand this one: it looks like the highest-level constructor for the plane wave basis and seems to be called explicitly in most of the examples. (I understand how the comment below about type specificity applies here.)

src/Model.jl
```diff
@@ -85,8 +88,8 @@ function Model(lattice::AbstractMatrix{T};
                spin_polarization=default_spin_polarization(magnetic_moments),
                symmetries=default_symmetries(lattice, atoms, magnetic_moments, terms, spin_polarization),
                ) where {T <: Real}
-    lattice = Mat3{T}(lattice)
-    temperature = T(temperature)
+    lattice = austrip.(Mat3{T}(lattice))
```
antoine-levitt, line 91: I think you want to strip before converting to Mat3{T} here or it won't work

src/Model.jl
```diff
@@ -85,8 +88,8 @@ function Model(lattice::AbstractMatrix{T};
...
+    temperature = austrip(temperature)
```
antoine-levitt, line 92: We still want it to end up a T, so T(austrip())

src/Model.jl
berquist: It seems to work fine:
```julia
julia> m = [1.1 2.2 3.3; 4.4 5.5 6.6; 7.7 8.8 9.9]
3×3 Array{Float64,2}:
 1.1  2.2  3.3
 4.4  5.5  6.6
 7.7  8.8  9.9

julia> ms = Mat3(m)
3×3 SArray{Tuple{3,3},Float64,2,9} with indices SOneTo(3)×SOneTo(3):
 1.1  2.2  3.3
 4.4  5.5  6.6
 7.7  8.8  9.9

julia> msa = ms * 1u"angstrom"
3×3 SArray{Tuple{3,3},Quantity{Float64,𝐋,Unitful.FreeUnits{(Å,),𝐋,nothing}},2,9} with indices SOneTo(3)×SOneTo(3):
 1.1 Å  2.2 Å  3.3 Å
 4.4 Å  5.5 Å  6.6 Å
 7.7 Å  8.8 Å  9.9 Å

julia> austrip.(msa)
3×3 SArray{Tuple{3,3},Float64,2,9} with indices SOneTo(3)×SOneTo(3):
  2.0787    4.1574   6.2361
  8.31479  10.3935  12.4722
 14.5509   16.6296  18.7083
```

antoine-levitt:
```
julia> austrip.(Mat3{Float64}(randn(3,3)*1u"angstrom"))
ERROR: DimensionError:  and ha are not dimensionally compatible.
```

.github/workflows/runexamples.yaml
```diff
@@ -12,9 +12,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-python@v1
+      - uses: actions/setup-python@v2
         with:
-          version: '3.x'
+          python-version: 3.8.6
```
mfherbst, line 17: I would not pin version 3.8 explicitly here.

Project.toml
```diff
@@ -63,6 +65,8 @@ Roots = "1"
 SpecialFunctions = "0.10"
 StaticArrays = "0.12"
 TimerOutputs = "0.5.6"
+Unitful = "1.4"
```
mfherbst, line 68: Any reason why this needs to be specifically 1.4 ?

src/Model.jl
mfherbst: Also I am not sure this constructor actually works as we intent. If `lattice` is a Array{Quantity} than `T` will be bound to `Quantity`, which is not `<: Real`, so this constructor anyway does not apply. I think I'd solve this by having another constructor `Model(lattice::AbstractArray{Q}; kwargs...) where {Q <: Quantity}` which does the step to strip away the units.

src/PlaneWaveBasis.jl
```diff
@@ -241,6 +246,7 @@ function PlaneWaveBasis(model::Model, Ecut::Number;
         # store in symmetries the set of kgrid-preserving symmetries
         symmetries = symmetries_preserving_kgrid(model.symmetries, kcoords)
     end
+    Ecut = austrip(Ecut)
     PlaneWaveBasis(model, Ecut, kcoords, ksymops, symmetries; kwargs...)
```
mfherbst, line 250: You might as well merge those two lines.

src/PlaneWaveBasis.jl
```diff
@@ -229,10 +232,12 @@ treated explicitly. In this case all guess densities and potential
 functions must agree with the crystal symmetries or the result is
 undefined.
 """
-function PlaneWaveBasis(model::Model, Ecut::Number;
+function PlaneWaveBasis(model::Model,
+                        Ecut::Union{Number, Quantity};
                         kgrid=kgrid_size_from_minimal_spacing(model.lattice, 2π * 0.022),
                         kshift=[iseven(nk) ? 1/2 : 0 for nk in kgrid],
                         use_symmetry=true, kwargs...)
+    kgrid = austrip.(kgrid)
```
mfherbst, line 240: Hmm. I don't think this makes sense. `kgrid` is in either case just a number. (In fact just a vector of three integers). I suppose you wanted to add units to the kgrid spacing, right? That should go to the `kgrid_size_from_minimal_spacing` function.

src/PlaneWaveBasis.jl
```diff
@@ -229,10 +232,12 @@ treated explicitly. In this case all guess densities and potential
 functions must agree with the crystal symmetries or the result is
 undefined.
 """
-function PlaneWaveBasis(model::Model, Ecut::Number;
+function PlaneWaveBasis(model::Model,
```
mfherbst, line 235: I'm not sure this is the best place to add units. This is kind of the low-level constructor, which is only supposed to be used internally by DFTK. I think the better place would be the high-level constructor at

https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/PlaneWaveBasis.jl#L232

In fact at that level I think it would be fair to just drop the type annotation `::Number` and pass whatever you get straight to `austrip` before passing it on to this constructor.

.github/workflows/runexamples.yaml
mfherbst: Ok. I see you did that to avoid 3.9. Could you do "3.8" instead ... just to be the least specific necessary.

src/PlaneWaveBasis.jl
berquist: This is the same constructor you're linking to.

src/PlaneWaveBasis.jl
berquist: That means that if you don't use the default argument, you won't be able to pass an explicit value for the kgrid with units.

antoine-levitt: Kgrid is int, there's no units there

src/PlaneWaveBasis.jl
mfherbst: Indeed you are right. My bad.

With respect to the type annotation of `Ecut`, maybe a bit more of my reasoning: In general in Julia one wants to avoid "strange type unions" (i.e. ones which involve to some extent unrelated types) and `Union{Number, Quantity}` is to me one of these cases. The general argument is that if you start needing these `Union`s this points to your interface being more generic anyway and therefore typically there is no need to annotate the type. So that's why my second remark was to suggest dropping the type annotation for `Ecut`. ;)

src/PlaneWaveBasis.jl
mfherbst: Yes sorry, that's what I meant. Only the `kgrid_size_from_minimal_spacing` (which takes a kgrid spacing as second argument) needs to get some extension to support units.

src/Model.jl
```diff
@@ -85,8 +88,7 @@ function Model(lattice::AbstractMatrix{T};
                spin_polarization=default_spin_polarization(magnetic_moments),
                symmetries=default_symmetries(lattice, atoms, magnetic_moments, terms, spin_polarization),
                ) where {T <: Real}
-    lattice = Mat3{T}(lattice)
```
mfherbst, line 88: This line is still needed, actually to do the conversion to the type we use internally in DFTK. That's why some of the tests  fail.

src/PlaneWaveBasis.jl
berquist: I still don't understand this, but I'll change it.

mfherbst:
LGTM!

Last thing that is missing from my point of view is some sort of tests. Since the changes are a bit all over the place, I think the best would be if you added some unit-related testing to some of the existing tests for the respective objects you modified. Basically just check it strips away the units properly and does the right conversion (especially with temperature that should be checked). 

Another thing which would be awesome would be if you added an example or integrated the units into the existing examples to make people aware of the new feature. I'd say the tutorial, Temperature and metallic systems or Creating supercells with pymatgen are good places for this. But don't worry too much about it ... we can also just do it at a later point.

test/forces.jl
```diff
@@ -1,10 +1,12 @@
 using DFTK
 using Test
+using Unitful
+using UnitfulAtomic
 include("testcases.jl")
 
 @testset "Forces on semiconductor (using total energy)" begin
     function energy(pos)
-        Ecut = 5
+        Ecut = 5 * u"hartree"
```
mfherbst, line 9: Hmm. I think at this place this is not needed because this test is really testing the forces and not the Ecut / grid setup.

mfherbst (commented):
Nice work. We're almost there with the tests. Just some minor remarks from my end.

docs/src/guide/tutorial.jl
```diff
@@ -27,13 +27,22 @@
...
 lattice = a / 2 * [[0 1 1.];  # Silicon lattice vectors
                    [1 0 1.];  # specified column by column
                    [1 1 0.]]
 
+# By default, all numbers passed as arguments are assumed to be in atomic units.
+# Quantities such as temperature, energy cutoffs, lattice vectors, and the
+# k-point grid spacing can optionally be annotated with Unitful units, which are
+# automatically converted to the atomic units used internally. For Unitful
+# package documentation, see https://painterqubits.github.io/Unitful.jl/stable/
+# and https://github.com/sostock/UnitfulAtomic.jl.
```
mfherbst, line 44: Does that directly render as a clickable link in the html? I think I would still shorten it to sth like:
```
For more details on this see the [Unitful package documentation](https://painterqubits.github.io/Unitful.jl/stable/) and
the [UnitfulAtomic.jl package](https://github.com/sostock/UnitfulAtomic.jl).
```

docs/src/guide/tutorial.jl
```diff
@@ -42,8 +51,8 @@ atoms = [Si => [ones(3)/8, -ones(3)/8]]
 
 ## 2. Select model and basis
 model = model_LDA(lattice, atoms)
-kgrid = [4, 4, 4]  # k-point grid (Regular Monkhorst-Pack grid)
-Ecut = 7           # kinetic energy cutoff in Hartree
+kgrid = [4, 4, 4]      # k-point grid (Regular Monkhorst-Pack grid)
+Ecut = 190.5u"eV"  # kinetic energy cutoff
```
mfherbst, line 55: Just make sure that the hashes are aligned ;) looks a little nicer.

examples/collinear_magnetism.jl
```diff
@@ -20,9 +22,9 @@ atoms = [Fe => [zeros(3)]];
 # To get the ground-state energy we use an LDA model and rather moderate
 # discretisation parameters.
 
-kgrid = [3, 3, 3]  # k-point grid (Regular Monkhorst-Pack grid)
-Ecut = 15          # kinetic energy cutoff in Hartree
-model_nospin = model_LDA(lattice, atoms, temperature=0.01)
+kgrid = [3, 3, 3]       # k-point grid (Regular Monkhorst-Pack grid)
+Ecut = 15 * u"hartree"  # kinetic energy cutoff
+model_nospin = model_LDA(lattice, atoms, temperature=3150u"K")
```
mfherbst, line 27: I think I would add a comment here that this is about 0.01 Hartree, because using Kelvin here is not that usual (but it's good to show that it works!)

examples/geometry_optimization.jl
```diff
@@ -7,12 +7,14 @@ using DFTK
...
+using UnitfulAtomic
 
 kgrid = [1, 1, 1]       # k-Point grid
-Ecut = 5                # kinetic energy cutoff in Hartree
+Ecut = 5u"hartree"      # kinetic energy cutoff
 tol = 1e-8              # tolerance for the optimization routine
-a = 10                  # lattice constant in Bohr
-lattice = a * Diagonal(ones(3))
+a = 10 * u"bohr"        # lattice constant
+lattice = austrip(a) * Diagonal(ones(3))
```
mfherbst, line 17: Do you need to austrip here? I think this should work with the units too shouldn't it?

examples/geometry_optimization.jl
```diff
@@ -52,7 +54,7 @@ end;
 
 function fg!(F, G, x)
     scfres = compute_scfres(x)
-    if G != nothing
+    if G !== nothing
```
mfherbst, line 57: Good catch!

examples/geometry_optimization.jl
```diff
@@ -71,7 +73,7 @@ xres = optimize(Optim.only_fg!(fg!), x0, LBFGS(),
                 Optim.Options(show_trace=true, f_tol=tol))
 xmin = Optim.minimizer(xres)
 dmin = norm(lattice*xmin[1:3] - lattice*xmin[4:6])
-@printf "\nOptimal bond length for Ecut=%.2f: %.3f Bohr\n" Ecut dmin
+@printf "\nOptimal bond length for Ecut=%.2f: %.3f Bohr\n" austrip(Ecut) dmin
```
mfherbst, line 76: No need to strip here ... the printing with units looks fine.

examples/collinear_magnetism.jl
```diff
@@ -9,8 +9,10 @@
 # First we setup BCC iron without spin polarization
 # using a single iron atom inside the unit cell.
 using DFTK
+using Unitful
+using UnitfulAtomic
 
-a = 5.42352  # Bohr
+a = 5.42352 * u"bohr"
```
mfherbst, line 15: You're not too much a fan of leaving off the `*` :smile:. Fair enough.

mfherbst (commented):
Thanks for keeping the steam to finish this up. Only a few more nits from my end.

mfherbst:
To fix the failing documentation test, just add Unitful and UnitfulAtomic here: https://github.com/JuliaMolSim/DFTK.jl/blob/master/docs/Project.toml

antoine-levitt:
Thanks for doing this! I'm not actually too sure about some of the example changes. I think it makes sense to talk about it in the tutorial, but after that we just drop it, no?

docs/src/guide/tutorial.jl
```diff
@@ -27,13 +27,22 @@
...
+using UnitfulAtomic
 
 ## 1. Define lattice and atomic positions
-a = 10.26  # Silicon lattice constant in Bohr
+a = 5.431u"angstrom" # Silicon lattice constant
 lattice = a / 2 * [[0 1 1.];  # Silicon lattice vectors
                    [1 0 1.];  # specified column by column
                    [1 1 0.]]
 
+# By default, all numbers passed as arguments are assumed to be in atomic units.
```
antoine-levitt, line 39: (Hartree, Bohr)

mfherbst:
Not so sure. I think now that we support it, it should become our default to unit-annotate in examples to avoid confusion and make them easier to understand for newcomers. If people choose to not do this in their own research for lazyness, that's fine, but units are clearly a constant source of confusion in these types of calculations and at least in the examples we should make it as clear as possible.

docs/src/guide/tutorial.jl
antoine-levitt: It's very unusual to use eV for Ecuts. I would leave it as `7` and add "in Hartree. Can also be specified with the Unitful package, see above`

examples/collinear_magnetism.jl
antoine-levitt: Yes, I would also leave it in Hartree. I think we don't need to recall here that it can be specified with Unitfuls, it's well covered in the tutorial

examples/geometry_optimization.jl
antoine-levitt: Indeed we don't really want the user to know about austrip

docs/src/guide/tutorial.jl
mfherbst: That is not true. eV is the default unit for ASE.

antoine-levitt:
Hm, that's a good question. I think we still need to impart the idea that we use atomic units, because all our internals and output is in AU (which we agreed we're not going to change), and we only provide unitful support as a convenience for the input. So giving the user the option is good, but it should still be clear that this is convenience only and AU are used by default, so I would do all the tutorials except the first one in AU. It's not laziness not to do it, it depends what your goal is, and if you never interact with something that is in different units there's no reason to do it.

examples/collinear_magnetism.jl
mfherbst: Fair enough. I agree it might be a bit too much here (sorry @berquist for asking you to do this earlier).

mfherbst:
I'm not fully convinced because ...

> if you never interact with something that is in different units there's no reason to do it.

... that is bound to happen in this field.

docs/src/guide/tutorial.jl
antoine-levitt: No DFT code I know works in eV for Ecuts, and for good reason: it's not the natural energy scale to use for this, because Ecut has to be big compared to the characteristic kinetic energy of the electrons, which is much more naturally in Hartree than in eV (as seen here). OTOH things people commonly report as *outputs*, eg band plots or such, are about differences of energies that are much smaller than that, and for this Hartree makes little sense (and indeed we plot them in eV)

antoine-levitt:
Very much depends. Atomic coordinates you probably want to use angstroms I agree, but Ecuts and temperature are convergence parameters which don't interact with anything. In any case, we need users to be very much aware of AU so they don't get confused when they input atomic coordinates in angstroms but get forces in hartree/bohr

docs/src/guide/tutorial.jl
mfherbst: > No DFT code I know works in eV for Ecuts

[VASP](https://www.vasp.at/wiki/index.php/ENCUT) does and CASTEP, too. Probably there are some more.  
I completely agree that eV is a stupid unit for this, but you'd be surprised, but it's not that uncommon as you think. Many people in the physics community think in terms of eV for cutoffs from the interactions I had. We should definitely have this in the tutorial to make people aware.

mfherbst:
Ok the forces aspect is a good point. Ok, let's remove the units from all examples, but keep them in the tutuorial. I would still point out 

https://github.com/JuliaMolSim/DFTK.jl/blob/078ab54590ae20501646fcc2cfae34201eccbfaa/examples/metallic_systems.jl#L32

as an exception, where I would keep it, because it is not uncommon to use Angström here.

docs/src/guide/tutorial.jl
antoine-levitt: Oh wow I had no idea, sorry! That's super surprising, all the talk about convergence I've seen is in Hartree. I still don't think we should have it as the default (it's not a habit we want to encourage), but yeah OK it makes a lot of sense to be super explicit. Maybe something like `Ecut = 7 # in Hartree, equivalent to eg `Ecut = 190.5u"eV"``?

antoine-levitt:
OK that makes sense

docs/src/guide/tutorial.jl
mfherbst: Hmm, perhaps more like:
```julia
Ecut = 7  # in Hartree
## Ecut = 190.5u"eV"  # Could also use eV to specify cutoffs
```

mfherbst:
Documentation failed because of a tiny detail (you also need to add the packages there explicitly to the Project.toml). Good to go if tests pass.

antoine-levitt:
Great! Thanks a lot @berquist, sorry for the long review!

PR #332, opened 2020-10-12
Pointers for contributors

antoine-levitt:


mfherbst:
Cool. I'm not sure we want to completely remove the contributing section in the docs though. I think some sort of pointer should still be there.

antoine-levitt:
It's duplicate no? The README is our homepage, not the docs, so people are going to see this much more easily.

mfherbst:
I don't think duplication hurts there. I mean to push that argument to the extreme one could say: Why even have the docs? People can just browse the examples folder, no? :smile: Just put a short form there. If you want one sentence on the index page of the docs is good enough for me, too.

antoine-levitt:
OK, done

docs/src/index.md
```diff
@@ -48,7 +48,7 @@ codes such as [Abinit](https://www.abinit.org/),
       DFTK-based calculator implementation for ASE.
     - Read data in [ETSF Nanoquanta](https://doi.org/10.1016/j.commatsci.2008.02.023) format.
 
-Missing a feature? Look for an open issue or [create a new one](https://github.com/JuliaMolSim/DFTK.jl/issues).
+Missing a feature? Look for an open issue or [create a new one](https://github.com/JuliaMolSim/DFTK.jl/issues). Want to contribute? See the information [here](https://github.com/JuliaMolSim/DFTK.jl#contributing).
```
mfherbst, line 51: Here is not a good name for a link (too short and meaningless).

antoine-levitt: Oh come on it's a link, this is as clear as can be. Feel free to change it if you don't like it.

mfherbst: ok will do

PR #333, opened 2020-10-14
Extend apply_chi0 and compute_chi0 to collinear spin

mfherbst:


src/postprocess/chi0.jl
```diff
@@ -131,53 +166,77 @@ returns `3` extra bands, which are not converged by the eigensolver
...
 
     # Add variation wrt εF
     if temperature > 0
         dVol = basis.model.unit_cell_volume / prod(basis.fft_size)
-        ldos = LDOS(εF, basis, eigenvalues, ψ, temperature=temperature)
+        ldos = [LDOS(εF, basis, eigenvalues, ψ, temperature=temperature, spins=[σ])
+                for σ in 1:n_spin]
         dos  = DOS(εF, basis, eigenvalues, temperature=temperature)
-        δρ .+= ldos .* dot(ldos, δV) .* dVol ./ dos
+        if n_spin == 1
```
antoine-levitt, line 225: can be factorized by first computing the delta eps_f (spin-dependent part)

antoine-levitt (commented):
The (V, Vbeta) thing is unfortunate, something to be tackled later

src/postprocess/chi0.jl
mfherbst: Sorry, I don't get what you mean.

antoine-levitt: delta eps_f = ...; nspin == 2 && delta eps_f += ...; delta rho .+= ldos .* delta eps_f .* ...

mfherbst: Ah I see ... indeed makes it a bit nicer to read.

PR #336, opened 2020-10-19
Refactor mixing to take ρtot and ρspin

mfherbst:


src/scf/mixing.jl
```diff
@@ -4,14 +4,30 @@ import Base: @kwdef
...
-# The interface is `mix(m; basis, ρin, ρout, kwargs...) -> ρnext`
+
+@doc raw"""
+Simple mixing: ``J^{-1} ≈ α``
+"""
+@kwdef struct SimpleMixing
+    α::Real = 0.8
+end
+@timing "mixing Simple" function mix(mixing::SimpleMixing, ::PlaneWaveBasis, δFs...; kwargs...)
+    @assert length(δFs) > 0 && !isnothing(δFs[1])
```
antoine-levitt, line 23: This assert feels unnecessary. I'm not a huge fan of the dFs... args instead of just passing an array, but we should discuss the API later anyway

antoine-levitt: Otherwise LGTM!

mfherbst: Me neither, actually. It's just to make it work. If you move to a more sensible datastructure the assert will go. It's just that I don't mess up (but it'll err anyway I suppose)

PR #341, opened 2020-10-23
Refactor chi0 models out of HybridMixing

mfherbst:
A bit of refactoring to make it easier to play with different χ0 models inside HybridMixing. Also implements "exact" χ0 for mixing. Happy to discuss alternatives.

### DIFF
--- a/src/scf/chi0models.jl
+++ b/src/scf/chi0models.jl
@@ -25,7 +25,7 @@ function (::LdosModel)(basis; eigenvalues, ψ, εF, kwargs...)
     # Catch cases that will yield no contribution
     iszero(basis.model.temperature) && return nothing
     ldos = [LDOS(εF, basis, eigenvalues, ψ, spins=[σ]) for σ in 1:n_spin]
-    if maximum(maximum(abs, ldos[σ]) for σ in 1:n_spin) < eps(eltype(ldos))
+    if maximum(maximum(abs, ldos[σ]) for σ in 1:n_spin) < eps(eltype(basis))
         return nothing
     end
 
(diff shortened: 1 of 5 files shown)


antoine-levitt (commented):
Looks about right. I would maybe just have used plain old functions instead of structs, but it doesn't change much probably

src/scf/chi0models.jl
```diff
@@ -0,0 +1,106 @@
...
+
+    function apply!(δρ, δV)
+        # χ0δV[1] is total, χ0δV[2] is spin
+        χ0δV = apply_χ0(ham, ψ_cvg, εF, eigenvalues_cvg, δV...; χ0.kwargs_apply_χ0...)
+        if basis.model.n_spin_components == 1
+            δρ[1] .+= χ0δV[1].real
+        else
+            δρ[1] .+= (χ0δV[1].real .+ χ0δV[2].real) ./ 2
+            δρ[2] .+= (χ0δV[1].real .- χ0δV[2].real) ./ 2
+        end
```
antoine-levitt, line 103: OK by now we have three cohabitating conventions in the code: (rhoalpha, rhobeta), (rhotot, rhodiff) and rho[:,:,:,ispin]. Can't we cut it down a bit, preferably to just the second?

mfherbst: I agree I think we should, but as it turns out I'm not sure the second is the best. It's not the subject of this PR, but I start to get the feeling that internally using `ρ[:, :, :, σ]` and wrapping that in a convenience datastructure (or just providing convenience functions to get `total` and `spin`) might be best.

antoine-levitt: That makes sense, if we decide not to go for a special array type

mfherbst: Even then. Whether you "wrap" and AbstractArray or a SpecialArray (which is really a list of 2 arrays) should not matter.

PR #342, opened 2020-10-23
Read data from files using ASE

mfherbst:


mfherbst:
@gkemlin We talked about this at some point before if I remember correctly. This actually saves you some effort when loading structures from CIF files to use them in DFTK: Just call `load_lattice` and `load_atoms` on the file paths directly.

PR #347, opened 2020-10-30
MPI for distribution over k-Points

antoine-levitt:
```
antoine@beta ~ $ mpirun -np 4 ~/julia/bin/julia /home/antoine/.julia/dev/DFTK/examples/silicon.jl 
n     Energy            Eₙ-Eₙ₋₁     ρout-ρin   Diag
---   ---------------   ---------   --------   ----
  1   -7.926717728172         NaN   2.00e-01    4.4 
n     Energy            Eₙ-Eₙ₋₁     ρout-ρin   Diag
---   ---------------   ---------   --------   ----
  1   -7.926773511107         NaN   2.00e-01    4.4 
n     Energy            Eₙ-Eₙ₋₁     ρout-ρin   Diag
---   ---------------   ---------   --------   ----
  1   -7.926671367140         NaN   2.00e-01    4.4 
  2   -7.931891825405   -5.17e-03   3.00e-02    2.3 
  2   -7.931890288259   -5.12e-03   2.99e-02    2.2 
  3   -7.931982420939   -9.06e-05   4.14e-03    1.0 
  2   -7.931888841102   -5.22e-03   3.00e-02    2.3 
  3   -7.931981941904   -9.17e-05   4.15e-03    1.0 
n     Energy            Eₙ-Eₙ₋₁     ρout-ρin   Diag
---   ---------------   ---------   --------   ----
  1   -7.926755623924         NaN   2.00e-01      3   -7.931982028106   -9.32e-05   4.18e-03    1.0 
4.5 
  4   -7.931992983616   -1.06e-05   5.98e-04    2.0 
  4   -7.931992941880   -1.09e-05   6.03e-04    2.0 
  4   -7.931992977753   -1.10e-05   6.14e-04    2.1 
  5   -7.931993452107   -4.68e-07   8.58e-05    1.9 
  5   -7.931993452526   -5.11e-07   8.14e-05    2.0 
  5   -7.931993452045   -4.74e-07   8.51e-05    1.9 
  2   -7.931890032521   -5.13e-03   2.99e-02    2.2 
  6   -7.931993462970   -1.09e-08   1.31e-05    2.4 
  6   -7.931993462901   -1.04e-08   1.62e-05    2.4 
  3   -7.931981803475   -9.18e-05   4.15e-03    1.0 
  6   -7.931993462879   -1.08e-08   1.60e-05    2.3 
  7   -7.931993463234   -2.64e-10   2.58e-06    2.2 
  7   -7.931993463230   -3.29e-10   3.36e-06    2.1 
  4   -7.931992985203   -1.12e-05   6.12e-04    2.1 
  7   -7.931993463229   -3.50e-10   3.92e-06    2.2 
  5   -7.931993452211   -4.67e-07   8.37e-05    1.9 
  6   -7.931993462931   -1.07e-08   1.45e-05    2.2 
  7   -7.931993463233   -3.01e-10   3.27e-06    2.3 
```
Still a few things to fix, but not bad for two hours of work!


antoine-levitt:
So, I'm going to unconditionally use MPI code paths, and don't provide a way for the user to disable MPI. If you don't want MPI, just don't run under mpirun.

antoine-levitt:
OK it seemed too good to be true and it was: I wasn't actually running with multiple processors. Now it's giving me stupid results, as it should. Now to debug...

antoine-levitt:
Huh, it actually works. Trying my luck at CI now...

antoine-levitt:
There's a race (apparently in threading) with Double32. I have no idea how to debug, giving up for now

antoine-levitt:
Direct minimization is very tricky to make MPI-compatible, because for that we need reductions *inside* Optim, so the only way is a custom MPI array type. I disabled it for now.

antoine-levitt:
Tests actually almost work, and I did not even cheat too much! The only thing that still fails is the tests in run_scf_and_compare, but honestly I'm not sure what's going on there, with the n_rounds and so on. @mfherbst could you take a look? There's also a bug in the forces, I'll have a look.

Also it complains that I didn't add MPI to compat, but honestly this thing of having to add explicit versions is just annoying. I would just assume MPI.jl won't break anything (which it probably won't because this is pretty straightforward), and change it if and when they do break stuff, but I'm guessing you won't agree with that @mfherbst :-)

antoine-levitt:
Forces bug is a fun one: displacements were generated randomly... and so of course were different from processor to processor. One more thing to look out for with MPI! Another fun gotcha is that convergence criteria need to be exactly the same for all processes, or else processes will not be in sync and just hang. Didn't take any special precaution for that (quantities we test on are equal up to machine precision between all processes, so the likelyhood is small)

antoine-levitt:
I did some basic scaling tests. It seems to work well, in the sense that timing doesn't flag anything suspicious as I increase the processor count (on my crappy laptop). Exact timings are much too noisy to give a sensible result (they depend on the initial wavefunctions). Testing script:

```
using DFTK
using FFTW
BLAS.set_num_threads(1)
FFTW.set_num_threads(1)

a = 10.263141334305942  # Lattice constant in Bohr
lattice = a / 2 .* [[0 1 1.]; [1 0 1.]; [1 1 0.]]
atoms = [Si => [ones(3)/8, -ones(3)/8]];

pystruct = pymatgen_structure(lattice, atoms)
pystruct.make_supercell([3, 1, 1])
lattice = load_lattice(pystruct)
atoms = [Si => [s.frac_coords for s in pystruct.sites]];

model = model_LDA(lattice, atoms)
basis = PlaneWaveBasis(model, 30, kgrid=(1, 8, 8))

scfres = self_consistent_field(basis, tol=Inf);
DFTK.reset_timer!(DFTK.timer)
@time scfres = self_consistent_field(basis, tol=1e-5);
#-
scfres.energies
DFTK.mpi_master() && display(DFTK.timer)
```
Should do that on larger systems.

antoine-levitt:
This is ready to merge from my end modulo the run_scf_and_compare and the compat

antoine-levitt:
Regarding printing and debug: the least annoying solution I've found is to call `mpi_master() || (redirect_stdout(); redirect_stderr())` at the beginning of a script to disable all printing except on process 0.

mfherbst:
Cool! I'll take a closer look tomorrow. Regarding printing: Does Julia's default logging module offer something more clever? Perhaps it makes sense to integrate with that for our printing? Also the SCF printing can be easily disabled inside the callback.

antoine-levitt:
We can just prefix our printing with `if mpi_master()`. I'm not yet sure this is the way to go. Let's see once we gain a bit more experience with it.

src/common/mpi.jl
```diff
@@ -0,0 +1,49 @@
...
+    # TODO look more closely at interaction between MPI and threads
+    MPI.Initialized() || MPI.Init_thread(MPI.ThreadLevel(3))
+end
+
+"""
+Number of processors used in MPI. Can be called without ensuring initialization.
+"""
+mpi_nprocs() = mpi_nprocs(MPI.COMM_WORLD)
+mpi_nprocs(comm) = (mpi_ensure_initialized(); MPI.Comm_size(comm))
+mpi_master() = (mpi_ensure_initialized(); MPI.Comm_rank(MPI.COMM_WORLD) == 0)
```
mfherbst, line 19: I would go with the Julia standard and call this `is_mpi_master()`

src/common/mpi.jl
```diff
@@ -0,0 +1,49 @@
...
+function mpi_ensure_initialized()
+    # TODO look more closely at interaction between MPI and threads
+    MPI.Initialized() || MPI.Init_thread(MPI.ThreadLevel(3))
+end
+
+"""
+Number of processors used in MPI. Can be called without ensuring initialization.
+"""
+mpi_nprocs() = mpi_nprocs(MPI.COMM_WORLD)
+mpi_nprocs(comm) = (mpi_ensure_initialized(); MPI.Comm_size(comm))
```
mfherbst, line 18: We have underscores after the `n`, so `mpi_n_procs`?

src/common/mpi.jl
```diff
@@ -0,0 +1,49 @@
...
+mpi_master() = (mpi_ensure_initialized(); MPI.Comm_rank(MPI.COMM_WORLD) == 0)
+
+mpi_sum(arr, comm::MPI.Comm) = MPI.Allreduce(arr, +, comm)
+mpi_sum!(arr, comm::MPI.Comm) = MPI.Allreduce!(arr, +, comm)
+mpi_min(arr, comm::MPI.Comm) = MPI.Allreduce(arr, min, comm)
+mpi_min!(arr, comm::MPI.Comm) = MPI.Allreduce!(arr, min, comm)
+mpi_max(arr, comm::MPI.Comm) = MPI.Allreduce(arr, max, comm)
+mpi_max!(arr, comm::MPI.Comm) = MPI.Allreduce!(arr, max, comm)
+mpi_average(arr, comm::MPI.Comm) = mpi_sum(arr, comm) ./ mpi_nprocs(comm)
+mpi_average!(arr, comm::MPI.Comm) = (mpi_sum!(arr, comm); arr ./= mpi_nprocs(comm))
```
mfherbst, line 28: Could you improve the alignment of the `=` and the brackets a bit better that makes it more readable

src/common/mpi.jl
```diff
@@ -0,0 +1,49 @@
...
+mpi_nprocs(comm) = (mpi_ensure_initialized(); MPI.Comm_size(comm))
+mpi_master() = (mpi_ensure_initialized(); MPI.Comm_rank(MPI.COMM_WORLD) == 0)
+
+mpi_sum(arr, comm::MPI.Comm) = MPI.Allreduce(arr, +, comm)
+mpi_sum!(arr, comm::MPI.Comm) = MPI.Allreduce!(arr, +, comm)
+mpi_min(arr, comm::MPI.Comm) = MPI.Allreduce(arr, min, comm)
+mpi_min!(arr, comm::MPI.Comm) = MPI.Allreduce!(arr, min, comm)
+mpi_max(arr, comm::MPI.Comm) = MPI.Allreduce(arr, max, comm)
+mpi_max!(arr, comm::MPI.Comm) = MPI.Allreduce!(arr, max, comm)
+mpi_average(arr, comm::MPI.Comm) = mpi_sum(arr, comm) ./ mpi_nprocs(comm)
```
mfherbst, line 27: That function is really surprising to me, why is this needed? The problem I have is that could be confused with `mpi_mean` as the pendant to `mean`, which would do something else.

src/common/mpi.jl
```diff
@@ -0,0 +1,49 @@
...
+mpi_average!(arr, comm::MPI.Comm) = (mpi_sum!(arr, comm); arr ./= mpi_nprocs(comm))
+
+"""
+Splits N work units between the processes and returns the slice of 1:N handled by the current process
+"""
+function mpi_split_work(comm, N)
+    nprocs = mpi_nprocs()
+    @assert nprocs <= N
```
mfherbst, line 35: Unless you catch this somewhere else, I think an `error` would be better here advising the user what to do.

src/common/mpi.jl
```diff
@@ -0,0 +1,49 @@
...
+    N_per_proc = div(N, nprocs, RoundUp)
+    ibeg = my_rank * N_per_proc + 1
+    iend = (my_rank+1) * N_per_proc
+    if iend > N
+        iend = N  # last process is slacking off.
+        # This will result in work distribution like (eg) 3 3 3 1: it's not bad but 3 3 2 2 might be a bit better
+        # TODO optimize this better. Eg see
+        # https://stackoverflow.com/questions/15658145/how-to-share-work-roughly-evenly-between-processes-in-mpi-despite-the-array-size
+    end
+    # println("Process $(my_rank+1)/$(nprocs) computing $(ibeg:iend)")
```
mfherbst, line 47: debug leftover.

src/common/mpi.jl
```diff
@@ -0,0 +1,49 @@
+# Convenience functions for working with MPI
+using MPI
```
mfherbst, line 2: If you prepend everything by `mpi_` I wonder if it would not be better to put this into a separate module?

src/terms/nonlocal.jl
```diff
@@ -80,13 +81,13 @@ end
 
                         dHψSk = P * (C * (dPdR' * ψSk))
                         for iband = 1:size(ψ[ik], 2)
-                            @views fr[α] -= (occ[ik][iband] / tot_red_kpt_number
+                            @views fr[α] -= (occ[ik][iband] / tot_red_kpt_number * basis.model.n_spin_components
```
mfherbst, line 84: Too long a line.

docs/src/guide/parallelization.md
```diff
@@ -32,10 +31,8 @@ inside this datastructure, any timing measurement should first reset
 this timer before running the calculation of interest.
 
 For example to measure the timing of an SCF:
-```@example parallelisation
-import TimerOutputs
-
-TimerOutputs.reset_timer!(DFTK.timer)
+```@example parallelization
+DFTK.reset_timer!(DFTK.timer)
```
mfherbst, line 35: This works? Takes me by surprise.

docs/src/guide/parallelization.md
```diff
@@ -62,8 +59,18 @@ as a breakdown over individual routines.
     unless you set `DFTK_TIMING` to `"all"`. In this case you must not use
     Julia threading (see section below) or otherwise undefined behaviour results.
 
+## MPI
+DFTK uses MPI to distribute on kpoints only at the moment. This should be the most performant method of parallelization: if you have kpoints, start by disabling all threading and use this. Simply follow the instructions on [MPI.jl](https://github.com/JuliaParallel/MPI.jl) and run DFTK under MPI:
```
mfherbst, line 63: Too long a line.

docs/src/guide/parallelization.md
```diff
@@ -62,8 +59,18 @@ as a breakdown over individual routines.
...
+```
+mpiexecjl -np 16 julia myscript.jl  # use mpiexecjl (see MPI.jl docs) or make sure your mpiexec is the one used by julia
+```
+
+Issues and workarounds:
+- Printing is garbled, as usual with MPI. You can use `mpi_master() || (redirect_stdout(); redirect_stderr())` at the top of your script to disable printing on all processes but one.
```
mfherbst, line 69: Too long a line.

src/PlaneWaveBasis.jl
```diff
@@ -53,15 +54,19 @@ struct PlaneWaveBasis{T <: Real}
...
-    # kweights[ik] = length(ksymops[ik]) / sum(length(ksymops[ik]) for ik=1:Nk)
+    # BZ integration weights, summing up to model.n_spin_components
     kweights::Vector{T}
     # ksymops[ik] is a list of symmetry operations (S,τ)
     # mapping to points in the reducible BZ
     ksymops::Vector{Vector{SymOp}}
 
+    comm_k::MPI.Comm  # communicator for the kpoints distribution
+    krange_thisproc::Vector{Int}  # indices of kpoints treated explicitly by this
+                                  # processor in the global kcoords array
```
mfherbst, line 68: thisproc? Not so sure ... how about `krange_local` That's typically used in the MPI jargon.

Also I would not access this field directly but go via a function. That will make it easier if we generalise to other computational backends later (but we can also refactor than of course).

src/PlaneWaveBasis.jl
```diff
@@ -81,6 +86,9 @@ struct PlaneWaveBasis{T <: Real}
     # Independent of the `use_symmetry` option
     symmetries::Vector{SymOp}
 end
```
mfherbst, line 88: Empty line.

src/PlaneWaveBasis.jl
```diff
@@ -81,6 +86,9 @@ struct PlaneWaveBasis{T <: Real}
...
+# prevent broadcast on pwbasis
+import Base.Broadcast.broadcastable
```
mfherbst, line 90: I see why you do this, but I tend to prefer imports at the top of the file.

src/PlaneWaveBasis.jl
```diff
@@ -113,7 +121,7 @@ Base.eltype(::PlaneWaveBasis{T}) where {T} = T
         end
     end
 
-    vcat(kpoints_per_spin...)
+    vcat(kpoints_per_spin...) # put all spin up first, then all spin down
```
mfherbst, line 124: extra space before #

src/PlaneWaveBasis.jl
```diff
@@ -177,23 +186,47 @@ build_kpoints(basis::PlaneWaveBasis, kcoords) =
...
+
+    # Build kpoint-specific basis. Notice that this also builds index mapping from the k-point-specific basis
+    # to the global basis and thus the fft_size needs to be final at this point.
+    kpoints = build_kpoints(model, fft_size, kcoords, Ecut; variational=variational)
+
+    # kpoints is now possibly twice the size of kcoords. Double it for spin
+    if model.n_spin_components == 2
+        ksymops = vcat(ksymops, ksymops)
+        krange_thisproc = vcat(krange_thisproc, nkpt .+ krange_thisproc)
+    end
```
mfherbst, line 215: This function keeps getting longer. Any chance on refactoring something into another function?

src/PlaneWaveBasis.jl
```diff
@@ -244,6 +277,14 @@ function PlaneWaveBasis(model::Model, Ecut::Number;
     PlaneWaveBasis(model, Ecut, kcoords, ksymops, symmetries; kwargs...)
 end
 
+"""
+Sum an array over kpoints, taking weights into account
+"""
+function weighted_ksum(basis, arr)
```
mfherbst, line 283: The argument order is a bit inconsistent compared to for example `mpi_sum`, where the thing acted upon is the first arg and the communcator is second.

src/PlaneWaveBasis.jl
```diff
@@ -244,6 +277,14 @@ function PlaneWaveBasis(model::Model, Ecut::Number;
...
+    res = sum(basis.kweights .* arr)
+    mpi_sum(res, basis.comm_k)
+end
```
mfherbst, line 286: Maybe also add a function `weighted_ksum(f::Function, arr, basis)` which first applies `f` to `arr`?

src/common/mpi.jl
```diff
@@ -0,0 +1,49 @@
...
+end
+
+"""
+Number of processors used in MPI. Can be called without ensuring initialization.
+"""
+mpi_nprocs() = mpi_nprocs(MPI.COMM_WORLD)
+mpi_nprocs(comm) = (mpi_ensure_initialized(); MPI.Comm_size(comm))
+mpi_master() = (mpi_ensure_initialized(); MPI.Comm_rank(MPI.COMM_WORLD) == 0)
+
+mpi_sum(arr, comm::MPI.Comm) = MPI.Allreduce(arr, +, comm)
```
mfherbst, line 21: For these type of functions julia often has versions `mpi_sum(f::Function, arr, comm)`, which I think could be quite useful to have as well (and you can just generate them with a macro).

mfherbst:
Todo for @mfherbst:
- [ ] Add MPI-parallel test to travis
- [ ] Fix tests

src/common/mpi.jl
antoine-levitt: I like prefixing everything with `mpi`, makes it easier on the eye to find out which parts are MPI commands. I'd go with `mpi_is_master` but both this and `is_mpi_master` makes it sound like we're asking whether MPI is master, which is weird

src/common/mpi.jl
antoine-levitt: Depends; eg julia has nthreads. I think I like `nprocs` better

src/common/mpi.jl
antoine-levitt: It's useful eg in the forces test. More generally if we want to make sure some arrays are synced exactly (as opposed to up to machine precision) this is a good way to do it. I don't get your comment: we can call it `mean` if you prefer

antoine-levitt: The nice thing is that `average` is a verb, so `average!` makes sense

src/common/mpi.jl
antoine-levitt: This is very internal, there's no user.

src/common/mpi.jl
antoine-levitt: I think that's the kind of comment that's useful to leave. That's an internal function pretty hidden, so I'm fine with it aesthetically

src/common/mpi.jl
antoine-levitt: We did that for smearing and I'm not sure I like how that turned out. We definitely don't want to confuse mpi_sum with sum (they are pretty different)

docs/src/guide/parallelization.md
antoine-levitt: Yup. I guess because DFTK uses TimerOutputs, which exports it, so reset_timer! is "known" in the DFTK namespace

docs/src/guide/parallelization.md
antoine-levitt: That's OK for docs, no?

src/PlaneWaveBasis.jl
antoine-levitt: if you look at pwbasis and see `krange_local` you might not think it's a parallelization thing. This makes it more explicit. Generalization let's tackle this when we get to it

src/PlaneWaveBasis.jl
antoine-levitt: That doesn't really make sense here: better to have it locally where it's used

src/PlaneWaveBasis.jl
antoine-levitt: I know, it's getting out of hand... But really it's one of the rare cases where long functions are OK: it's just initializing a lot of things in sequence. We can split it up but it reduce readability rather than the reverse

src/PlaneWaveBasis.jl
antoine-levitt: Yeah I wondered a lot about that. We generally put `basis` first in a lot of our functions, so I tried to be consistent with that.

src/PlaneWaveBasis.jl
antoine-levitt: We don't really use it?

src/common/mpi.jl
antoine-levitt: We could, but I'd like to keep the MPI statement as simple as possible, so that if you just read the code without wanting to think about MPI you can just ignore them

src/common/mpi.jl
mfherbst: `mpi_on_master` or `on_mpi_master`?

src/common/mpi.jl
mfherbst: yes but we use the convention `n_` in many places. (n_iter, n_spin_components, n_electrons, n_k, ...)

mfherbst: Also julia has set_num_threads ... so it's not perfectly consistent either. It's just driving me mad to rember where it is which way. At least in our code I would like consistency.

src/common/mpi.jl
mfherbst: At this point for sure, but if that's the only place where you check, than I think erroring at this place could be very cryptic.

src/common/mpi.jl
mfherbst: My problem is that `mpi_sum` / `mpi_max` / `mpi_min` does basically the same thing as `sum`, `max` and `min` as julia functions. There is also `mean` and so why not call this `mpi_mean` to agree with the pattern?

src/common/mpi.jl
mfherbst: It's dead code. Either you make it switchable with a flag (verbose or something) or you remove it. Or you make it an `@debug`.

src/common/mpi.jl
mfherbst: That's true. Also agree with the smearing aspect actually ...

docs/src/guide/parallelization.md
mfherbst: Yes indeed, but for some reason I never thought about that.

src/PlaneWaveBasis.jl
mfherbst: Hmm. I'm not convinced ... for Model there turned out to be a lot of simplification possible last time I worked on the initialiser.  Can you make a TODO. I'll take a look next time I have to work on PlaneWaveBasis.

src/common/mpi.jl
mfherbst: ok that I agree with.

src/PlaneWaveBasis.jl
mfherbst: Yes that's true, so should we do that in mpi_sum with the `comm` as well?

(That's one of the disadvantages of multiple dispatch)

src/common/mpi.jl
antoine-levitt: is that really better than mpi_master? It's clear what this does

src/PlaneWaveBasis.jl
mfherbst: Sure, but only since you have not thought about it. Actually it could be used in *all* instances where you employ `weighted_ksum` instead of the constructs you have right now (and it could even avoid allocations by passing it right through to `sum`).

src/occupation.jl
```diff
@@ -31,16 +31,18 @@ function find_occupation(basis::PlaneWaveBasis{T}, energies;
     # If temperature is zero, (εi-εF)/T = ±∞.
     # The occupation function is required to give 1 and 0 respectively in these cases.
     compute_occupation(εF) = [filled_occ * Smearing.occupation.(smearing, (ε .- εF) ./ temperature) for ε in energies]
-    compute_n_elec(εF) = sum(basis.kweights .* sum.(compute_occupation(εF)))
+    compute_n_elec(εF) = weighted_ksum(basis, sum.(compute_occupation(εF)))
 
-    if filled_occ*sum(basis.kweights .* length.(energies)) < n_electrons - sqrt(eps(T))
+    if filled_occ*weighted_ksum(basis, length.(energies)) < n_electrons - sqrt(eps(T))
```
mfherbst, line 36: spaces around `*`

docs/src/guide/parallelization.md
mfherbst: Not in my world ... docs are code, too and should be easy to read in an editor ... just do your emacs magic to break the line and I'm happy :smile:

src/common/mpi.jl
antoine-levitt: These things are always going to be inconsistent, for good reason: naming things is hard, if you impose consistency it becomes impossible. I would just relax and accept it :-) For instance, my preferred spelling would be `niter`, `n_spin_components`, `n_electrons`, `nk`, etc. Readability trumps consistency. In this case I'd sync nprocs on nthreads

src/common/mpi.jl
antoine-levitt: We really want to avoid too much similarity between mpi_sum and sum. I thought about overloading sum to avoid mpi_sum, eg something like `sum(comm, arr)`. I decided against it: we want to really separate mpi calls and highlight them as such. So we can do mean if you prefer, but `mean!(arr)` is weird

src/common/mpi.jl
antoine-levitt: nope, we check it pwbasis and have another code path in this case

src/common/mpi.jl
antoine-levitt: I removed it. It's just the kind of thing that's annoying to have to retype to debug... Flag doesn't make sense, and we don't really have a good convention for debug info.

docs/src/guide/parallelization.md
antoine-levitt: Just fix your vim already :-) but ok done

src/PlaneWaveBasis.jl
antoine-levitt: I agree we should simplify it, especially the variational bit

src/PlaneWaveBasis.jl
antoine-levitt: I thought about it also (and in fact had it in a previous version), but when you do `mpi_sum!` it makes more sense to have arr as the first argument, and also the first argument is the one you really want to look at when you read the line of code, so it makes more sense to put it first. We have `basis` as our OOP-style exception.

src/PlaneWaveBasis.jl
antoine-levitt: Well if we use it we add it. Otherwise it's dead code

src/common/mpi.jl
mfherbst: Ok fair enough, that's true.

src/common/mpi.jl
mfherbst: ok that's a good point. Fair enough ... I give in!

src/common/mpi.jl
mfherbst: Well that's what `@debug` is for ... which we do not use so far.

src/PlaneWaveBasis.jl
mfherbst: ok.

src/common/mpi.jl
mfherbst: Sure but it exists in `Statistics`. I don't know why they used `mean` instead of `average`, but I prefer to stick to it (and use it here).

src/PlaneWaveBasis.jl
mfherbst: yeah ok. Still not happy about the name, but I also have no better.

src/PlaneWaveBasis.jl
mfherbst: Well in some sense that's my point: We can use it sensibly, but fair enough. I might give it a try and see what it looks like, though.

antoine-levitt: I agree we could probably replace some of the reduce code we have (eg in the energy computations) with a do block using this. It avoids explicit mpi calls.

src/common/mpi.jl
antoine-levitt: ok let's go with that

mfherbst:
Ok cool. I'll take over from here.

mfherbst:
Any strong opinion on renaming `comm_k` to `comm_kpts` ... just because later we might have `comm_bands` and then that name is nicer, I think.

antoine-levitt:
Yup that's good

mfherbst:
I decided against extending the weigted_ksum function because it cannot be used for e.g. the DOS in its current stage (you somehow need to limit the range by spin) and I could not be bothered to think about that now.

antoine-levitt:
Thanks for the CI work!!

mfherbst:
I would write "my pleasure", but that would be a lie this time ;). At least it works now :smile:

antoine-levitt:
Yeah, it's super annoying, but I have to say it's satisfying to know we have automatic tests with MPI! It's also a burden off future developments: you don't have to worry too much about MPI, because if you forgot an mpi_sum CI will tell you!

mfherbst:
Yes I completely agree ... definitely worth the effort in the long run.

And the nice thing is if MPI does not work for someone you at least have one case in the CI where you know exactly the steps to follow to get it to work. Like a self-documenting proof of concept.

mfherbst:
Just to conclude and document: What does not work with MPI for now is :
- IntervalArithmetic
- DoubleFloats
- direct minimisation
- bandstructure computations

Anything else?

antoine-levitt:
The main trouble I had with MPI is that if you do `mpiexec julia ...` and your MPI doesn't match julia's, it just silently doesn't do MPI.  We should probably revisit the instructions for parallelization; let's do that once the dust of the threading settles.

I think your list is about right. Anything that requires the full scfres.

mfherbst:
see the travis :smile: That tells you how to do it

antoine-levitt:
Yeah I know, that's what I ended up doing, just took me a while. That's the kind of things we could put in docs

mfherbst:
Indeed, that's true.

antoine-levitt:
See also https://github.com/JuliaParallel/MPI.jl/issues/430

PR #352, opened 2020-11-02
Optimize and parallelize density computation

antoine-levitt:


src/terms/Hamiltonian.jl
```diff
@@ -87,7 +87,7 @@ Base.:*(H::Hamiltonian, ψ) = mul!(deepcopy(ψ), H, ψ)
 # (eg the density ρ)
 @timing function energy_hamiltonian(basis::PlaneWaveBasis, ψ, occ; kwargs...)
     # it: index into terms, ik: index into kpoints
-    ene_ops_arr = [ene_ops(term, ψ, occ; kwargs...) for term in basis.terms]
+    @timing "ene_ops" ene_ops_arr = [ene_ops(term, ψ, occ; kwargs...) for term in basis.terms]
```
mfherbst, line 90: Should be consistent with https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/terms/local.jl#L98. One way or the other ... I don't care.

src/densities.jl
```diff
@@ -26,7 +32,7 @@ function compute_partial_density(basis, kpt, ψk, occupation)
     end
 
     # FFT and return
-    r_to_G(basis, ρk_real)
+    r_to_G!(ρ, basis, complex.(ρk_real))
```
mfherbst, line 35: You don't need a `complex.` here just `complex` does it (either way a copy).

mfherbst (commented):
Nice. ~~You forgot to update the LDOS routine.~~

**Update:** Ah no sorry that's a different layer.

src/terms/Hamiltonian.jl
antoine-levitt: I don't understand what you mean here

src/densities.jl
antoine-levitt: good catch

src/terms/Hamiltonian.jl
mfherbst: Ah sorry ... wrong annotation. I meant the individual `ene_ops` functions for each term where you use a different convention than for the timing label to the respective force terms.

PR #353, opened 2020-11-02
More detailed installation instructions

mfherbst:


docs/src/guide/installation.md
```diff
@@ -1,15 +1,23 @@
...
+[downloading the Julia binaries](https://julialang.org/downloads/)
+and following the [Julia installation instructions](https://julialang.org/downloads/platform/).
+At least **Julia 1.4** is required for DFTK.
+
+Afterwards you can install DFTK
+[like any other package](https://julialang.github.io/Pkg.jl/v1/getting-started/)
+in Julia. For example run in your Julia REPL terminal:
+```julia
+using Pkg
+pkg"add DFTK"
```
antoine-levitt, line 13: I think we want to encourage people to use `]`, no? That's more idiomatic and more pleasant.

docs/src/guide/installation.md
```diff
@@ -1,15 +1,23 @@
...
 ```
-or if you like to be fully up to date:
+which will install the latest DFTK release.
+Alternatively (if you like to be fully up to date) install the master branch:
 ```
-] add DFTK#master
+using Pkg
+pkg"add DFTK#master"
```
antoine-levitt, line 19: We can add something like "if you want to start working on DFTK, you can do `]dev DFTK` which will check out a development copy in (on linux) ~/.julia/dev/DFTK"

docs/src/guide/installation.md
mfherbst: Actually it's less clear for beginners what this means in my experience. A person who knows Julia will not read this anyway.

docs/src/guide/installation.md
mfherbst: Good idea. I'll add a few "developer's notes".

docs/src/guide/installation.md
antoine-levitt: I'd write "type `]` to get into package mode, then type `add DFTK`". I thought we had that at some point

mfherbst: Yes we did. But people in other packages are moving away from this, too, see  http://docs.juliaplots.org/latest/install/ or https://github.com/JuliaPy/PyCall.jl which I would say represent amongst the first Julia packages people will encounter.

mfherbst: I am ok with `Pkg.add("DFTK")` as well if you prefer that.

antoine-levitt: Ok for Pkg.add then. pkg"" is just too magical. It's just annoying to have to `import Pkg`.

PR #354, opened 2020-11-02
Avoid some more allocations in Hamiltonian apply

antoine-levitt:
I still don't understand why I don't get good threading scaling...

### DIFF
--- a/src/PlaneWaveBasis.jl
+++ b/src/PlaneWaveBasis.jl
@@ -377,7 +377,6 @@ end
     mul!(f_real, basis.ipFFT, f_real)
 
     # Truncate
-    fill!(f_fourier, 0)
     f_fourier .= view(f_real, kpt.mapping)
 end
 
--- a/src/PlaneWaveBasis.jl
+++ b/src/PlaneWaveBasis.jl
@@ -378,7 +378,7 @@ end
 
     # Truncate
     fill!(f_fourier, 0)
-    f_fourier[:] = f_real[kpt.mapping]
+    f_fourier .= view(f_real, kpt.mapping)
 end
 
 """
--- a/src/terms/Hamiltonian.jl
+++ b/src/terms/Hamiltonian.jl
@@ -31,6 +31,7 @@ end
 @views @timing "Hamiltonian multiplication" function LinearAlgebra.mul!(Hψ::AbstractArray,
                                                                         H::HamiltonianBlock,
                                                                         ψ::AbstractArray)
+    # Special-case of DFT Hamiltonian: go to a fast path
     if length(H.optimized_operators) == 3 &&
         any(o -> o isa FourierMultiplication, H.optimized_operators) &&
         any(o -> o isa RealSpaceMultiplication, H.optimized_operators) &&
@@ -96,7 +97,9 @@ Base.:*(H::HamiltonianBlock, ψ) = mul!(similar(ψ), H, ψ)
     end
 
     # Apply the nonlocal operator
-    @timing "nonlocal" apply!((fourier=Hψ, real=nothing), nonlocal_op, (fourier=ψ, real=nothing))
+    @timing "nonlocal" begin
+        apply!((fourier=Hψ, real=nothing), nonlocal_op, (fourier=ψ, real=nothing))
+    end
 
     Hψ
 end
--- a/src/terms/Hamiltonian.jl
+++ b/src/terms/Hamiltonian.jl
@@ -26,13 +26,18 @@ struct Hamiltonian
     blocks::Vector{HamiltonianBlock}
 end
 
+
 # Loop through bands, IFFT to get ψ in real space, loop through terms, FFT and accumulate into Hψ
-# Do this band by band to conserve memory
-# As an optimization we special-case nonlocal operators to apply them
-# instead on the full block and benefit from BLAS3
 @views @timing "Hamiltonian multiplication" function LinearAlgebra.mul!(Hψ::AbstractArray,
                                                                         H::HamiltonianBlock,
                                                                         ψ::AbstractArray)
+    if length(H.optimized_operators) == 3 &&
+        any(o -> o isa FourierMultiplication, H.optimized_operators) &&
+        any(o -> o isa RealSpaceMultiplication, H.optimized_operators) &&
+        any(o -> o isa NonlocalOperator, H.optimized_operators)
+        return mul!_fast(Hψ, H, ψ)
+    end
+
     basis = H.basis
     kpt = H.kpoint
     nband = size(ψ, 2)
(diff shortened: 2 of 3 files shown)


src/terms/operators.jl
```diff
@@ -133,7 +133,7 @@ Matrix(op::MagneticFieldOperator) = error("Not implemented")
 function optimize_operators_(ops)
     RSmults = [op for op in ops if op isa RealSpaceMultiplication]
     isempty(RSmults) && return ops
-    nonRSmults = [op for op in ops if !(op isa RealSpaceMultiplication)]
+    nonRSmults = [op for op in ops if !(op isa RealSpaceMultiplication) && !(op isa NoopOperator)]
```
mfherbst, line 136: I think the logic is clearer if you just do this as the first line of the function:
```
ops = filter(op -> !(op isa NoopOperator), ops)
```

src/PlaneWaveBasis.jl
```diff
@@ -377,8 +377,7 @@ end
     mul!(f_real, basis.ipFFT, f_real)
 
     # Truncate
-    fill!(f_fourier, 0)
-    f_fourier[:] = f_real[kpt.mapping]
+    f_fourier .= view(f_real, kpt.mapping)
```
mfherbst, line 380: This should be documented, since now if f_fourier contains garbage (is not fully allocated to zero), then the resulting data will be garbage. Did you check we don't use the property that f_fourier can be undefined as input to this function?

src/terms/Hamiltonian.jl
```diff
@@ -26,13 +26,18 @@ struct Hamiltonian
...
-# instead on the full block and benefit from BLAS3
 @views @timing "Hamiltonian multiplication" function LinearAlgebra.mul!(Hψ::AbstractArray,
                                                                         H::HamiltonianBlock,
                                                                         ψ::AbstractArray)
+    if length(H.optimized_operators) == 3 &&
+        any(o -> o isa FourierMultiplication, H.optimized_operators) &&
+        any(o -> o isa RealSpaceMultiplication, H.optimized_operators) &&
+        any(o -> o isa NonlocalOperator, H.optimized_operators)
+        return mul!_fast(Hψ, H, ψ)
+    end
```
mfherbst, line 39: This is not very Julian. What I would do is have a special type, e.g.
```
Base.@kwdef struct DftOperators
fourier_multiplication
real_space_multiplication
nonlocal_operator
end
```
and then pass `mul!` over to a four-argument version that also has the `H.optimized_operators` as an argument. Than you can dispatch on that to implement the special `mul!` function. The check for that case (and the introduction of that special type) can be done in `optimise_operators_`

mfherbst (commented):
Hmmm ... it starts to get a little ugly. Does this help a lot?

src/terms/Hamiltonian.jl
```diff
@@ -51,27 +56,50 @@ end
...
+
+        G_to_r!(ψ_real, basis, kpt, ψ[:, iband])
+        ψ_real .*= real_op.potential
+        r_to_G!(Hψ[:, iband], basis, kpt, ψ_real)  # overwrites ψ_real
+        Hψ[:, iband] .+= fourier_op.multiplier .* ψ[:, iband]
+    end
     end
 
+    # Apply the nonlocal operator
+    @timing "nonlocal" apply!((fourier=Hψ, real=nothing), nonlocal_op, (fourier=ψ, real=nothing))
```
mfherbst, line 99: I would write this as 3 lines, it looks a bit crowded.

antoine-levitt:
> Hmmm ... it starts to get a little ugly. Does this help a lot?

It avoids a couple of allocs. What I like with this is that the standard (fallback) version gets simpler (no special-casing of nonlocal). Then we have a generic version, and a very DFT-specialized version, which I think makes sense.

src/terms/operators.jl
antoine-levitt: good idea

src/PlaneWaveBasis.jl
antoine-levitt: No, it's OK: this line actually fills the whole of `f_fourier` (f_fourier is a subset of f_real)

src/terms/Hamiltonian.jl
antoine-levitt: That just adds more indirection, more code and more complexity. I like the current hack better: it's clear what's going on here: there's a generic path, and a fast optimized path.

antoine-levitt:
> It avoids a couple of allocs

In particular, the only alloc now is in the BLAS nonlocal terms. There's no allocations on the FFT grid.

src/PlaneWaveBasis.jl
mfherbst: Ah sorry that's on the kpoint-specific FFT than it's ok (and I wonder how we missed that until now).

src/terms/Hamiltonian.jl
mfherbst: The part I don't like is that you manually do multiple dispatch here. Also these 6 lines of check code are not exactly fun to read. Also `mul!_fast` is much less meaningful than having `mul!( ... , ::DftOperators, ...)` somewhere. And code complexity? The type can be defined pretty locally in the same file. I think when it comes to number of code lines its no longer that what you have now it's probably even shorter and you don't have to do the weird extraction stuff like `https://github.com/JuliaMolSim/DFTK.jl/pull/354/files/5cfbb2a4f3261093c23a7b1978fa9e5da0d75de0#diff-c88cddd6ac7d48004a07a50ffcaf0f3877f04610fd6858f13a23cccae8035d06R77` again, which is also complex code.

antoine-levitt: It's a hack. We don't want to institutionalize hacks by promoting them to have data structures, the simplest they are the better. I thought about having an extra `eligible_for_fast_path` or so function, but it's not really better...

mfherbst: Ok, I agree with that. My main point is that I disapprove of these 6 lines of explicit check code and that weird extraction routine and the `mul!_fast` as a name. What I propose solves all of these. I'm ok to not introduce the type if we come up with a more elegant solution to solve these aspects.

Re `eligible_for_fast_path` ... at some point you will have to check and I think a good point is when you construct the Hamiltonian and anyway get your hands dirty with the operators by calling `optimise_operators` don't you think? So why not determine that info there and somehow pass it on?

antoine-levitt: I pushed something. I'm not sure I like it too much, it's more code, but if you prefer we can keep it.

mfherbst (commented):
I don't like this, but I think it offers for a compromise we would both agree with, namely:

Just add `mul! = hblock_mul!(H\psi, H, fast_hblock(H), \psi)` and than just specialise this function for the case where the second arg is nothing or when it is a named tuple.

src/terms/Hamiltonian.jl
```diff
@@ -70,15 +66,22 @@ end
...
 # Minimizes the number of allocations
-@views function mul!_fast(Hψ::AbstractArray,
-                          H::HamiltonianBlock,
-                          ψ::AbstractArray)
-    fourier_op = only(filter(o -> o isa FourierMultiplication, H.optimized_operators))
-    real_op = only(filter(o -> o isa RealSpaceMultiplication, H.optimized_operators))
-    nonlocal_op = only(filter(o -> o isa NonlocalOperator, H.optimized_operators))
-
+@views function fast_hblock_mul!(Hψ::AbstractArray,
+                                 fast_hblock,
```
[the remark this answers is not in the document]
antoine-levitt: > Just add mul!(H\psi, H, \spi) = hblock_mul!(H\psi, H, fast_hblock(H), \psi) and than just specialise this function for the case where the second arg is nothing or when it is a named tuple.

I'd rather not do that, because that enshrines the hack... Right now it's clear that the official function is mul!, and that it internally does some hacks which are implementation details. Whereas your version hides that behind one more level of dispatch. Just because julia has dispatch doesn't mean we have to use it :p We can add underscores if you want to make that more explicit?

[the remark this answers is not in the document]
mfherbst: Nah all right. I'm not happy with this, since it feels an awful lot like we are reimplementing a language feature. As I said to me such kind of specialisation to give special code paths for special cases is exactly what multiple dispatch is for. Sure you don't have to use it, but structure and patterns aid thinking ... could be there is more to be gained elsewhere. But whatever ... it's a hack first of all and probably not worth discussing endlessly.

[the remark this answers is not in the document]
antoine-levitt: Structure is good for organizing large things. For implementation of small things, plain old C/fortran-style is better I think.

antoine-levitt:
Don't merge yet, I've figured out we don't need Hpsi_real anymore

antoine-levitt:
OK done

src/terms/Hamiltonian.jl
```diff
@@ -39,16 +41,12 @@ end
...
-        tid = Threads.threadid()
-        ψ_real = H.scratch.ψ_reals[tid]
-        Hψ_real = H.scratch.Hψ_reals[tid]
-        Hψ_fourier = Hψ_fouriers[tid]
+    Hψ_fourier = similar(Hψ[:, 1])
+    ψ_real = zeros(complex(T), basis.fft_size...)
+    Hψ_real = zeros(complex(T), basis.fft_size...)
 
+    # take ψi, IFFT it to ψ_real, apply each term to Hψ_fourier and Hψ_real, and add it to Hψ
+    for iband = 1:nband
```
mfherbst, line 49: So no more threading over bands for the fallback implementation?

antoine-levitt: I wanted an implementation as simple as possible, so either no threads or no preallocation. It doesn't matter too much, the optimized implementation covers pretty much everything we're interested in right now; we'll revisit when we try out new things

mfherbst: Just wanted to make sure this was intentional. Indeed there are few model which come to my mind which are not covered by the fast implementation and where speed should truly matter.

Issue #349, opened 2020-11-01
Threading


antoine-levitt:
Our threading is a bit all over the place right now.

Linalg threading is automatically taken care of, and can be improved by installing MKL.

I think FFT threading sucks more than it should, and I'm not sure what we can do about that, except maybe do benchmarks, compare to a C version, and report upstream. A good thing to try is to use MKL FFT (see https://github.com/JuliaMath/FFTW.jl#mkl) and compare performance. If it improves, it means that FFTW's (= julia's) threading sucks.

Already if we do the above two well, I believe we should scale relatively well. Julia threading by itself should be unnecessary (and even possibly counterproductive). Maybe just on a few big array operations (densities and potentials): to be determined by profiling once we take care of FFT threading. For those we should either do the loop ourselves or use a package such as https://github.com/Jutho/Strided.jl.

As a guess, I would say we can use that to get pretty good intra-node scaling. With this plus MPI for kpoints, I would be surprised if we can't get >50% efficiency on relatively large computations (eg 10 nodes each with 16 cores), which should be enough to tackle relatively big problems.

antoine-levitt:
So I did some very stupid benchmarking. 8x1x1 supercell of silicon, kerker mixing, Ecut 30, no kpoints. There is some variation in the number of hamiltonian applies, but not enough to affect the results by more than 10%. I don't use julia threads, only FFTW and BLAS.

With standard FFT, OpenBLAS: 1 thread 26s, 2 threads 19.5s
With MKL FFT, BLAS: 1 thread 22.5s, 2 threads 17.3s

So MKL is slightly faster, but no significant impact on threading scaling.

So it looks like FFTs just don't scale. Looking at individual timings, even pure linalg functions do not scale perfectly (although better than FFTs). This is pretty surprising to me. So either my laptop sucks, or threading sucks. :-(

Based on this it seems julia-level threading might be the way to go after all? If julia and BLAS threading are not active at the same (which I think is already the case right now), we can just set FFT threads to 1, julia and BLAS to n. Essentially the only thing missing is a `@threads` on the bands loop in `compute_partial_density`. Then we benchmark and take a look.

Also we should look at kernels (GEMM, FFT) with sizes corresponding to test cases we care about and microbenchmark them to look at scalings.

antoine-levitt:
[...]

PR #356, opened 2020-11-03
Use unnormalized FFT plans in Hamiltonian applies

antoine-levitt:
Sorry about the merge, I wasn't able to do it with a rebase


antoine-levitt:
At this point, the code is perfectly aligned on the reproducer from https://github.com/JuliaMolSim/DFTK.jl/issues/349#issuecomment-721070997

src/PlaneWaveBasis.jl
```diff
@@ -67,10 +67,15 @@ struct PlaneWaveBasis{T <: Real}
...
     # Plans for forward and backward FFT
+    # These plans follow DFTK conventions (see above)
     opFFT  # out-of-place FFT plan
     ipFFT  # in-place FFT plan
     opIFFT
     ipIFFT
+    opFFT_unnormalized   # unnormalized plans
+    ipFFT_unnormalized
+    opBFFT_unnormalized  # "backward" FFT in FFTW terminology: no normalization at all
+    ipBFFT_unnormalized
```
mfherbst, line 78: This starts to become a zoo. Why don't we do the normalisation in a post-step as a `./= norm` ?

mfherbst: Alternatively some sort of struct / named tuple(s) to hold these guys would be helpful.

antoine-levitt: I don't understand what you mean. This is directly used in the G_to_r! functions, so should be precomputed.

antoine-levitt: We could, and have a function to construct it. Can't be bothered for now though.

mfherbst: No I meant after the G_to_r! you use the normalisation as a post-step if normalisation is needed. So we no longer have any "scaled plans". That was my suggestion.

antoine-levitt: Hm, I see. I think it's nicer to have these normalized plans following DFTK conventions and the unnormalized ones as an optimization

mfherbst: Ok then at least split them up into two blocks of four to make it clear that the "these follow DFTK's convention" only apply to the first four blocks.

mfherbst:
So I suppose after this it's time to rerun my scripts and see if the graphs have changed?

antoine-levitt:
Yup! See comment in the other thread

antoine-levitt:
Still won't get perfect scaling, but at this point this is very much diminishing returns. There are still tricks one could conceivably play but they are nasty and possibly won't improve things. I think we're at the hardware limit here.

antoine-levitt:
Something is going very wrong with the code in fft_generic. It looks like before the line
```
    ipFFT *= sqrt(model.unit_cell_volume) / length(ipFFT)	
```
was hitting
```
*(p::DummyInplace, X) = p.fft * X
```
How can that have worked?! Any idea what we should do?

mfherbst:
> How can that have worked?! 

I don't remember the details, but I suppose the wrapping is done by AbstractFFTs? Just provide wrappers for what is needed in `fft_generic` ?

antoine-levitt:
No but I mean it was very buggy before, so hopefully we didn't use it...

antoine-levitt:
So it's probably still buggy, but at least it should pass CI...

antoine-levitt:
Hopefully somebody will pick up generic ffts again and we'll be able to drop these ugly workarounds

PR #359, opened 2020-11-04
Density screening in XC term

mfherbst:
Most of the code change is to pass a custom `xc` object to `model_DFT`

### DIFF
--- a/src/terms/xc.jl
+++ b/src/terms/xc.jl
@@ -6,13 +6,22 @@ Exchange-correlation term, defined by a list of functionals and usually evaluate
 """
 struct Xc
     functionals::Vector{Symbol}  # Symbols of the functionals (Libxc.jl / libxc convention)
-    scaling_factor::Real         # to scale by an arbitrary factor (useful for exploration)
+    scaling_factor::Real         # Scales by an arbitrary factor (useful for exploration)
+
+    # Density cutoff for XC computation: Below this value gridpoint counts as zero
+    # `nothing` implies that Libxc defaults are used (different for every functional)
+    density_threshold::Union{Nothing,Float64}
 end
-Xc(symbols::Vector; scaling_factor=1) = Xc(convert.(Symbol, symbols), scaling_factor)
 Xc(symbols::Symbol...; kwargs...) = Xc([symbols...]; kwargs...)
+function Xc(symbols::Vector; scaling_factor=1, density_threshold=nothing)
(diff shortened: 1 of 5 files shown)


src/terms/xc.jl
```diff
@@ -6,13 +6,22 @@ Exchange-correlation term, defined by a list of functionals and usually evaluate
...
-Xc(symbols::Vector; scaling_factor=1) = Xc(convert.(Symbol, symbols), scaling_factor)
 Xc(symbols::Symbol...; kwargs...) = Xc([symbols...]; kwargs...)
+function Xc(symbols::Vector; scaling_factor=1, density_threshold=nothing)
+    Xc(convert.(Symbol, symbols), scaling_factor, density_threshold)
+end
 
 function (xc::Xc)(basis)
     functionals = Functional.(xc.functionals; n_spin=basis.model.n_spin_components)
+    if !isnothing(xc.density_threshold)
+        setproperty!.(functionals, :density_threshold, Ref(xc.density_threshold))
```
antoine-levitt, line 23: Feels a bit too magical (also property/field always confuses me), can you write down a loop?

src/terms/xc.jl
```diff
@@ -6,13 +6,22 @@ Exchange-correlation term, defined by a list of functionals and usually evaluate
 """
 struct Xc
     functionals::Vector{Symbol}  # Symbols of the functionals (Libxc.jl / libxc convention)
-    scaling_factor::Real         # to scale by an arbitrary factor (useful for exploration)
+    scaling_factor::Real         # Scales by an arbitrary factor (useful for exploration)
+
+    # Density cutoff for XC computation: Below this value gridpoint counts as zero
+    # `nothing` implies that Libxc defaults are used (different for every functional)
+    density_threshold::Union{Nothing,Float64}
```
antoine-levitt, line 13: can you quickly say what the default looks like? Is it 0 or a non-zero value?

mfherbst: It's not zero but super close (1e-34, 1e-24, ...) I'm not sure where the values come from. Could be some numerical saveguard against NaNs I suppose?

antoine-levitt: OK so numerical zero basically, but no "cheating" (ie it's not 1e-4 or what)

mfherbst: Indeed. It looked like completely reasonable default values, but where raising them a bit could improve performance.

PR #360, opened 2020-11-04
Small performance improvements in eigensolver and nonlocal apply

antoine-levitt:
Yeah, when I get bored, I micro-optimize unimportant stuff. This particular bottleneck-chasing session was with setting supersampling=1, which amplifies operations that normally don't matter so much. There are tons more but they would be much more annoying to optimize. It shouldn't matter really because they are lower-complexity for larger systems, but it's fun

### DIFF
--- a/src/eigen/lobpcg_hyper_impl.jl
+++ b/src/eigen/lobpcg_hyper_impl.jl
@@ -117,7 +117,7 @@ end
             success = false
         end
         invR = inv(R)
-        X = X*invR # we do not use X/R because we use invR next
+        rmul!(X, invR) # we do not use X/R because we use invR next
 
         normest(M) = maximum(abs.(diag(M))) + norm(M - Diagonal(diag(M)))
 
@@ -150,7 +150,7 @@ end
 
 # Randomize the columns of X if the norm is below tol
 function drop!(X, tol=2eps(real(eltype(X))))
-    dropped = []
+    dropped = Vector{Int}()
     for i=1:size(X,2)
         n = norm(@views X[:,i])
         if n <= tol
@@ -298,7 +298,7 @@ end
         end
 
         ### Compute new residuals
-        new_R = new_AX .- new_BX * Diagonal(λs)
+        new_R = new_AX .- new_BX .* λs'
         # it is actually a good question of knowing when to
         # precondition. Here seems sensible, but it could plausibly be
         # done before or after
--- a/src/eigen/lobpcg_hyper_impl.jl
+++ b/src/eigen/lobpcg_hyper_impl.jl
@@ -150,7 +150,7 @@ end
 
 # Randomize the columns of X if the norm is below tol
 function drop!(X, tol=2eps(real(eltype(X))))
-    dropped = Vector{Int}()
+    dropped = Int[]
     for i=1:size(X,2)
         n = norm(@views X[:,i])
         if n <= tol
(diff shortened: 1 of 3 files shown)


src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -150,7 +150,7 @@ end
 
 # Randomize the columns of X if the norm is below tol
 function drop!(X, tol=2eps(real(eltype(X))))
-    dropped = []
+    dropped = Vector{Int}()
```
mfherbst, line 153: Just use `Int[]`. I've started to prefer that over the long version.

src/eigen/preconditioners.jl
```diff
@@ -66,5 +66,5 @@ end
 (Base.:*)(P::PreconditionerTPA, R) = mul!(copy(R), P, R)
 
 function precondprep!(P::PreconditionerTPA, X)
-    P.mean_kin = [real(dot(x, P.kin .* x)) for x in eachcol(X)]
+    P.mean_kin = [sum(real.(conj.(x) .* P.kin .* x)) for x in eachcol(X)]
```
mfherbst, line 69: Why is that faster? `dot` should just be a BLAS call no?

src/terms/nonlocal.jl
```diff
@@ -30,11 +30,9 @@ end
 
     E = zero(T)
     for (ik, k) in enumerate(basis.kpoints)
-        for iband = 1:size(ψ[1], 2)
-            ψnk = @views ψ[ik][:, iband]
-            Pψnk = term.ops[ik].P' * ψnk
-            E += basis.kweights[ik] * occ[ik][iband] * real(dot(Pψnk, term.ops[ik].D * Pψnk))
-        end
+        Pψ = term.ops[ik].P' * ψ[ik] # nproj x nband
```
mfherbst, line 33: Extra space before `#`.

src/terms/nonlocal.jl
```diff
@@ -30,11 +30,9 @@ end
...
+        band_enes = dropdims(sum(conj.(Pψ) .* (term.ops[ik].D*Pψ), dims=1), dims=1)
```
mfherbst, line 34: Space around `*`.

mfherbst: I don't think you need the `dropdims` because in the next line you could just do `.* occ[ik]'`. Also I'd put the `real` in the line computing the `band_enes` (and I would call it `PDPψ` not `band_enes`)

src/eigen/lobpcg_hyper_impl.jl
antoine-levitt: Cool, did not know

src/eigen/preconditioners.jl
antoine-levitt: Putting the `real` part in should in theory allow more optimizations, and it's 2x less storage (real instead of cplx). It's just exposing to julia what it needs to be in theory faster, I did not benchmark to see if it was actually faster

src/terms/nonlocal.jl
antoine-levitt: Why? It is the band energies

antoine-levitt: dropdims is just simpler because otherwise you have to think about sizes

mfherbst: But now I wonder why you don't play the same trick with the real as above at this instance?

src/eigen/preconditioners.jl
mfherbst: good point.

src/terms/nonlocal.jl
antoine-levitt: very good point!

PR #363, opened 2020-11-06
Harmonise and prepend compute_ for exported names

mfherbst:
The `forces` function has a slightly too generic name. It happened to me a few times that I had a variable of this sort defined when using DFTK, which caused issues. I propose to rename it to `compute_forces`.

Other renames:
- [x] `forces` -> `compute_forces`
- [x] `determine_grid_size` -> `compute_grid_size`
- [x] `find_fermi_level` -> `compute_fermi_level`
- [x] `find_occupation` -> `compute_occupation`
- [x] `DOS` / `LDOS` / `NOS` -> `compute_dos`
- [x] `compute_forces_cart`


antoine-levitt:
Hm. That's a good question of how we should call these things. We can go with `compute`, that's already what we do for the density and chi0. OTOH DOS/LDOS etc we call them just that. So do we want to have `compute_` everywhere?

mfherbst:
Probably. I also started to dislike that `DOS` is in captial, because it's the only place where we do this. I think `compute_dos` and `compute_ldos` are good alternatives. The other "prefix" could be `get_` ... that's what ASE uses for example.

antoine-levitt:
`get` feels like an accessor, but maybe it's an old OOP reflex and should be discarded... regarding caps I'm not sure, it's an acronym

mfherbst:
so is XC and we decided to stick to the naming convention with it (TermXc, Xc). I find it just surprising.

mfherbst:
Yes that's exactly my balance as well.

antoine-levitt:
Yes, maybe we should have gone with XC in fact... But we can go full un-capitalized if you prefer.

antoine-levitt:
so `compute_dos`?

mfherbst:
It's one of these things where everyone does something else ... let's just go lowercase (we have plot_dos now as well actually).

mfherbst:
yes I think that's fine

antoine-levitt:
OK. Then let's rename determine_grid_size as well, it's been bugging me for a while

antoine-levitt:
I'm still not completely sold on `compute` but I don't have a better alternative yet...

mfherbst:
> Then let's rename determine_grid_size as well,

To `compute_grid_size`?

antoine-levitt:
Yeah. So compute or get... I think get could work also actually. Let's leave this to marinate for a few days if you're not in any hurry

mfherbst:
Not really.

antoine-levitt:
There are *a lot* of functions that could be called `get_X` or `compute_X` and are just called `X`. The only places where we differ from this right now is determine_grid_size, compute_density, compute_chi0, compute_kernel. I think we should only deviate from this when there's a good reason to. chi0 and kernel we do to distinguish from apply. Grid_size and density this is because there are several grid sizes and densities in the code, and so it would not be completely clear what `grid_size()` is without looking at the arguments closely. Others, including forces, are much clearer and so I would tend to keep them... What's the problem with name shadowing exactly? It's pretty easy to avoid, and you can always use DFTK.forces if you want a local variable named forces.

mfherbst:
I typically use short, concrete names for variables, so in a script I would use `scfres` to hold the SCF results, `dos` to hold the DOS or `forces` to hold the forces, however I cannot write `forces = forces(scfres)` and this just bugs me.

Similarly I find `DOS` is just a very bad function name. In both cases it does not need to be `compute_forces` or `compute_dos` if you don't like it, but the current names ... I don't really like them.

mfherbst:
Also we should probably have a `compute_forces_cart` function, where the forces are returned in cartesian coords.

antoine-levitt:
What don't you like about DOS? The problem about adding `compute_` is that why should we stop there? Look at eg elements.jl, hamiltonian.jl, etc.

mfherbst:
I don't care about `compute_`. What I care about is that these functions block nice short names I would like to have available in Main for variables in user scripts. Sure I can use `DFTK.` every time, but that does not completely cut it, because you don't want to write that every time. And if you e.g. use `forces = DFTK.forces(scfres)` in a function than all of a sudden the the meaning of `forces` depends on context, which is super annoying.

For example KryolvKit exports `basis`, which we use quite frequently in functions and scripts as a variable for `PlaneWaveBasis` objects. When I write a function to do computations and I write `basis.bla` than Julia will not complain at compile time that `basis` is undefined (if I forgot a `basis = scfres.basis`) and it will only crash at runtime. If the SCF takes a day to complete until it gets to this spot that is really, really annoying. With `forces` in particular the same thing is bound to happen. (Might have happened to me, but I cannot remember to be fair.). Probably `dos` and `ldos` are better, but the same line of arguments apply and I'd prefer to  "avoid" these as exported names as well.

antoine-levitt:
OK that's a good point, we should distinguish between exported and unexported names then. For unexported we use the short version unless good reason not to. Non-capitalized, single word exported names: forces, DOS, LDOS, NOS. For these we add `compute_`? I think I'm fine with that, my fear was that it would spread. So we do that and harmonize compute_ to find_fermi_level, find_occupation and determine_grid_size. OK?

mfherbst:
I agree with your fear and with your suggestions. I'll do the changes when I get to it.

Yes the distinction exported / not exported is indeed important. For the latter short names are ok for me.

mfherbst:
Our `export` policy is something we should maybe revisit.

I think generally we export too much. For example I don't think the `mpi_sum`-type functions should not be exported. They are really "our" wrapper around `MPI.jl` and that's not what I would expect a code like DFTK to provide. If someone finds them useful  they might as well import it explicitly. Also many algorithm details (like the selection occupation functions) I would consider internal details, but there it's a bit trickier because selecting them is part of the SCF interface.

antoine-levitt:
yeah I agree

mfherbst:
I kept `mpi_nprocs` and `mpi_master` for now because they are actually really useful in scripts, but one might disagree there.

Issue #351, opened 2020-11-02
Reduced supersampling can lead to `Input array not real`


mfherbst:
Reproducer:

```julia
using DFTK      
using PyCall      
                         
Ecut  = 30      
kgrid = [5, 5, 5]    
supersampling = 1.5    
    
cobalt  = pyimport("ase.build").bulk("Co")    
lattice = load_lattice(cobalt)    
Co      = ElementPsp(:Co, psp=load_psp("hgh/lda/co-q9"))    
atoms   = [Co => load_atoms(cobalt)[1][2]]    
magnetic_moments = [Co => [5, 5]]    
    
model = model_LDA(lattice, atoms, magnetic_moments=magnetic_moments, temperature=0.01,    
                  smearing=Smearing.Gaussian(), spin_polarization=:collinear)    
fft_size = determine_fft_size(model, Ecut, supersampling=supersampling)    
basis    = PlaneWaveBasis(model, Ecut, kgrid=kgrid, fft_size=fft_size)     
ρspin    = guess_spin_density(basis, magnetic_moments)    
scfres   = self_consistent_field(basis, ρspin=ρspin)   
```

Traceback:
```
 [1] error(::String) at ./error.jl:33
 [2] from_fourier(::PlaneWaveBasis{Float64}, ::Array{Complex{Float64},3}; check_real::Bool) at /packages/DFTK/hYYPc/src/RealFourierArray.jl:57
 [3] from_fourier at /packages/DFTK/hYYPc/src/RealFourierArray.jl:45 [inlined]
 [4] macro expansion at /packages/DFTK/hYYPc/src/densities.jl:90 [inlined]
 [5] (::DFTK.var"#319#325"{PlaneWaveBasis{Float64},Array{Array{Complex{Float64},2},1},Array{Array{Float64,1},1}})() at /packages/TimerOutputs/dVnaw/src/TimerOutput.jl:237
 [6] timeit(::DFTK.var"#319#325"{PlaneWaveBasis{Float64},Array{Array{Complex{Float64},2},1},Array{Array{Float64,1},1}}, ::TimerOutputs.TimerOutput, ::String) at /packages/TimerOutputs/dVnaw/src/TimerOutput.jl:285
 [7] compute_density(::PlaneWaveBasis{Float64}, ::Array{Array{Complex{Float64},2},1}, ::Array{Array{Float64,1},1}) at /packages/TimerOutputs/dVnaw/src/TimerOutput.jl:236
[...]

PR #364, opened 2020-11-06
Lowpass density for symmetry

antoine-levitt:
Fix https://github.com/JuliaMolSim/DFTK.jl/issues/351


antoine-levitt:
Well done CI! Should be good now.

mfherbst:
Yes good that we have it.

PR #369, opened 2020-11-16
Update parallelisation documentation

mfherbst:


docs/src/guide/parallelization.md
```diff
@@ -52,60 +52,109 @@ as a breakdown over individual routines.
...
-and large numbers of threads is not great,
-such that by default threading in FFTW is even disabled.
-
-The **recommended setting** for FFT threading with DFTK
-is therefore to only use moderate number of FFT threads,
-something like ``2`` or ``4`` and for smaller calculations
-disable FFT threading completely.
-To **enable parallelization of FFTs** (which is by default disabled),
-use
+## Options for parallelisation
```
antoine-levitt, line 61: still ization I'm afraid :-p

docs/src/guide/parallelization.md
```diff
@@ -52,60 +52,109 @@ as a breakdown over individual routines.
...
+At the moment DFTK offers two ways to parallelise a calculation,
+firstly shared-memory parallelism using threading
+and secondly (possibly distributed) multiprocessing using MPI
```
antoine-levitt, line 64: It's always distributed between processes, whether it's also distributed across machines is, from the conceptual point of view, irrelevant

docs/src/guide/parallelization.md
```diff
@@ -52,60 +52,109 @@ as a breakdown over individual routines.
...
+
+1. To ensure the thread setup in DFTK is consistent with the environment
+   under which DFTK runs
+   add the following to your script running the DFTK calculation:
+   ```julia
+   using DFTK
+   setup_threading()
+   ```
+   This will ensure that all threading is disabled
+   (which unfortunately is not the default).
```
antoine-levitt, line 103: I don't know if I would call it unfortunate. What's unfortunate is that OpenBLAS sucks.

antoine-levitt: I would not talk about threads at all in this section (except for saying don't use any) and leave it for the next one

docs/src/guide/parallelization.md
```diff
@@ -52,60 +52,109 @@ as a breakdown over individual routines.
...
+
+2. Run Julia in parallel using the `mpiexecjl` wrapper script from MPI.jl:
+   ```sh
+   mpiexecjl -np 16 julia -t 1 myscript.jl
+   ```
+   In this `-np 16` tells MPI to use 16 processes and `-t 1` tells Julia
+   to use one thread only.  
+   Notice that we use `mpiexecjl` to automatically select the `mpiexec`
+   compatible with the MPI version used by Julia.
```
antoine-levitt, line 112: used by MPI.jl

docs/src/guide/parallelization.md
```diff
@@ -52,60 +52,109 @@ as a breakdown over individual routines.
...
+   ```
+   This will ensure that all threading is disabled
+   (which unfortunately is not the default).
+
+2. Run Julia in parallel using the `mpiexecjl` wrapper script from MPI.jl:
+   ```sh
+   mpiexecjl -np 16 julia -t 1 myscript.jl
+   ```
+   In this `-np 16` tells MPI to use 16 processes and `-t 1` tells Julia
+   to use one thread only.
```
antoine-levitt, line 110: -t 1 is the default and we've already told people not to use threads before, no need to resay it here

docs/src/guide/parallelization.md
```diff
@@ -52,60 +52,109 @@ as a breakdown over individual routines.
...
+!!! warning "MPI-based parallelism is experimental"
+    Even though MPI-based parallelism shows the better scaling it is still
+    experimental and some routines (e.g. band structure and direct minimization)
+    are not compatible with it yet.
+
+## Thread-based parallelism
+Threading in DFTK currently happens on multiple layers
+distributing the workload
+over different ``k``-Points, bands or within
+an FFT or BLAS call between processors.
```
antoine-levitt, line 129: don't talk about processors here. Processors is orthogonal to threads, no need to confuse people with it

docs/src/guide/parallelization.md
```diff
@@ -52,60 +52,109 @@ as a breakdown over individual routines.
...
+At its current stage our scaling for thread-based parallelism
+is worse compared MPI-based and therefore the parallelism
+described here should
+only be used if no other option exists.
```
antoine-levitt, line 133: already said above

docs/src/guide/parallelization.md
mfherbst: Aaaaaaaarghhhhhh! Ok. I'll fix it tomorrow :smile:.

docs/src/guide/parallelization.md
```diff
@@ -52,60 +52,109 @@ as a breakdown over individual routines.
...
+To use thread-based parallelism proceed as follows:
+
+1. Ensure that threading is properly setup inside DFTK by adding
+   to the script running the DFTK calculation:
+   ```julia
+   using DFTK
+   setup_threading()
+   ```
```
antoine-levitt, line 141: which disables FFT threading and sets the number of BLAS threads to the number of Julia threads

src/common/threading.jl
```diff
@@ -0,0 +1,10 @@
+import FFTW
+using LinearAlgebra
+
+function setup_threading(;n_fft=1, n_blas=Threads.nthreads())
+    n_julia = Threads.nthreads()
+    FFTW.set_num_threads(n_fft)
+    BLAS.set_num_threads(n_blas)
+    mpi_master() && @info "Threading setup" n_fft n_blas n_julia
```
antoine-levitt, line 8: That's a bit barebones. Either print what these numbers are ("n_fft=$n_fft, n_blas=$n_blas, n_julia=$n_julia") or don't write anything

src/common/threading.jl
```diff
@@ -0,0 +1,10 @@
...
+    nothing
```
antoine-levitt, line 9: Why?

antoine-levitt (commented):
Cool, thanks for doing this!

docs/src/guide/parallelization.md
mfherbst: Ok I worked around it by adding another helper function `disable_threading`.

docs/src/guide/parallelization.md
mfherbst: Ok.

docs/src/guide/parallelization.md
mfherbst: That's a leftover from the old docs :smile:

docs/src/guide/parallelization.md
mfherbst: You are assuming people actually have read from top to bottom. I don't think they will.

src/common/threading.jl
mfherbst: But this is exacty what happens

antoine-levitt: Oh I didn't know that, cool!

PR #370, opened 2020-11-17
OLD : First steps towards an interface to wannier90

LaurentVidal95:
I left a few comments for your attention at the beginning of the code.

The two routines to be called by the user start with dftk2wan.
The rest are hidden functions to read the nnkp file first, then write the eig file, mmn file and finaly amn file.


src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+
+"""
+    Create a .win file for Wannier90, compatible with the system studied with DFTK.
+    Options have to be compeleted by hand.
+"""
+function dftk2wan_win_file(prefix::String,basis::PlaneWaveBasis,scfres,num_wann;
+                           kgrid = [4,4,4], #can we retrieve this from scfres or basis ? Feel this should not be an argument.
+                           bands_plot = false,
+                           num_iter = 500,
+                           num_print_cycles = 50
```
antoine-levitt, line 29: You can use kwargs... here. Then do something like (untested)
```
for (key, value) in kwargs
    write(f, string(key) * " = " * string(value))
```
Saves you code, and works for every option you want to give wannier90

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+    end    
+
+    ## Write in file   
+    open("$prefix.win","w") do f
+
+        write(f,"! $prefix.win file generated by DFTK at "*string(now())*"\n"^2)
+        
+        ## General parameters
+        num_bands = size(scfres.ψ[1],2)
+        write(f,"num_bands "*(" "^9)*"= $num_bands"*"\n")
```
antoine-levitt, line 44: do you realy need these empty spaces ?

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+
+        # Unit cell block
+        write(f,"begin unit_cell_cart"*"\n"*"bohr"*"\n")
+        unit_cell = transpose(basis.model.lattice)         #unit cell vectors are in raws in Wannier90.
+        for i in 1:3
+            for j in 1:3
+                write(f,"$(unit_cell[i,j]) ")
+            end
+            write(f,"\n")
+        end
```
antoine-levitt, line 66: can't you just write unit_cell here?

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+        write(f,"!"^20*" Complete with optional parameters :"*"\n"^2)
+
+
+
+        ## System
+        write(f,"!"^20*" System"*"\n"^2)
+
+        # Unit cell block
+        write(f,"begin unit_cell_cart"*"\n"*"bohr"*"\n")
+        unit_cell = transpose(basis.model.lattice)         #unit cell vectors are in raws in Wannier90.
```
antoine-levitt, line 60: are units OK here?

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+            for i in 1:size(path[1],1)-1
+                ## write path section A -> B
+                A,B = path[1][i],path[1][i+1]
+                # kpoints are writen in the file with normalized coordinates
+                k_A = round.(DFTK.normalize_kpoint_coordinate.(get!(labels,A,1)), digits = 5)
+                k_B = round.(DFTK.normalize_kpoint_coordinate.(get!(labels,B,1)), digits = 5)
+                # Wannier90 convention for the Γ point is G
+                if A == "\\Gamma"
+                    A = "G"
+                end
```
antoine-levitt, line 106: use a function to do this, similar to what we do in detexify_kpoint. Eg you can have a `detexify_kpoint(string, unicode=true)` that returns `Γ` if unicode == true, and `G` if false

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+            write(f,"bands_plot = T"*"\n"^2)
+        end
+        
+        # Mp grid
+        if kgrid[1]*kgrid[2]*kgrid[3] !== size(scfres.ψ,1)
+            error("Given kgrid doesn't match the kgrid used for scf calculation")
+        end
+        write(f,"mp_grid : $(kgrid[1]) $(kgrid[2]) $(kgrid[3])"*"\n")
+        if kgrid == [1,1,1]
+            write(f,"gamma_only      = true"*"\n")
```
antoine-levitt, line 122: did you test this? (also you can just do `true\n"`)

antoine-levitt: I'm asking because these tools usually have different code paths for gamma_only

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+        write(f,"begin kpoints"*"\n")
+        for i in 1:size(scfres.ψ,1)
+            for x in basis.kpoints[i].coordinate
+                write(f,"$x ")
+            end
+            write(f,"\n")
+        end
+        write(f,"end kpoints"*"\n"^2)
+
+        @info "File generated.",f
```
antoine-levitt, line 137: Not really worth it to tell the user that: if the function returns without crashing, it's proof enough :-)

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+    end
+    
+    
+end
+
+
+"""
+Read the .nnkp file provided by the preprocessing routine of Wannier90 (i.e. "wannier90.x -pp prefix")
+Returns: 
+1) the array 'nn_kpts' of k points, they respective neirest neighbours and associated shifing vectors (non zero if the neighbour of is located in another cell).
```
antoine-levitt, line 147: nearest

antoine-levitt: also we use US english (much to Michael's dismay) so neighbors

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+    ln = [string(l) for l in eachline(file)]
+    close(file)
+
+
+    ## Extract nnkp block 
+    i_nn_kpts = findall(x-> endswith(x,"nnkpts"),ln)                               #Indices of the first and last line of the nnkpts block
+    @assert size(i_nn_kpts,1) == 2
+    char_to_vec(line) = [parse.(Int,x) for x in split(line,' ',keepempty=false)]
+    
+    nn_num = parse(Int64,ln[i_nn_kpts[1]+1])                                       #number of neighbours per k points
```
antoine-levitt, line 170: parse(Int) as above

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+
+    # reshape so that one line gives all infos about one projection
+    n_projs = parse(Int,only(popfirst!(raw_projs)))
+    @assert(n_projs == size(raw_projs,1)/2)
+    raw_projs = reshape(raw_projs,(2,n_projs))
+
+    # Parse in the format proj = [ [[center],[l,mr,r],[z_axis],[x_axis], α], ... ]
+    projs = []
+    for j in 1:n_projs
+        center = [parse(Float64,x) for x in raw_projs[1,j][1:3]]
```
antoine-levitt, line 188: you can do `parse.(Float64, raw_projs[1,j][1:3])`

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+    
+    nn_kpts,nn_num,projs    
+end
+
+
+
+"""
+Take scf_res, the result of a self_consistent_field calculation and writes eigenvalues (in eV) in a format readable by Wannier90.
+"""
+function generate_eig_file(prefix::String,scf_res)
```
antoine-levitt, line 204: is there something we can generally about eigenvalue offsets? Does Wannier90 dictate a particular convention?

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+    Ha_to_eV = 1/(unit_to_au(:eV))
+
+    eigvalues = scf_res.eigenvalues
+    k_size = size(eigvalues,1)
+    n_bands = size(eigvalues[1],1)
```
antoine-levitt, line 209: don't necessarily define things if you're just going to use them once

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+@doc raw"""
+ Computes the matrix ``[M^{k,b}]_{m,n} = \langle u_{m,k} | u_{n,k+b} \rangle`` for given k, kpb = k+b.
+ 
+ Return in the format of a (n_bands^2, 2) matrix, where each line is one overlap, and columns are real and imaginary parts.
+
+ K_shift is the "shifting" vector, correction due to the periodicity conditions imposed on k -> ψ_k. 
+ It is non zero if kpb is taken in another unit cell of the reciprocal lattice.
+ We use here that : ``u_{n(k + K_shift)}(r) = e^{-i*\langle K_shift,r \rangle} u_{nk}``
+"""
+function overlap_Mmn_k_kpb(pw_basis::PlaneWaveBasis,ψ,k,kpb,n_bands; K_shift = [0,0,0])
```
antoine-levitt, line 234: spaces between arguments (here and elsewhere)

antoine-levitt: Also we usually call vectors in the reciprocal lattice G vectors

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+    Mkb = zeros(Float64,n_bands*n_bands,2)
```
antoine-levitt, line 235: spaces (here and in other places). I personally don't mind so much but Michael will kill you :-)

antoine-levitt: Float64 is redundant here

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+    current_line = 0 #Manual count necessary to match the format of .mmn files.
+
+    for n in 1:n_bands
+        for m in 1:n_bands
+            
+            ovlp = 0im
```
antoine-levitt, line 241: careful here, it's a type-instability. Try doing `ovlp = 0im; ovlp += randn(ComplexF64)`: ovlp changes type. One reason for bad performance (probably not the only one though)

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+            
+            ## Extract Fourier coeffs and corresponding vectors in reciprocal lattice
+            Gk_coeffs = ψ[k][:,m]
```
antoine-levitt, line 244: When you slice columns like this, have a `@views` before function definition: it'll avoid making a copy.

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+            map_fourier_modes = []
+            for G1 in Gk_vec
+                for G2 in Gkpb_vec
+                    if  G1 == G2
+                        iG1 = only(findall(x-> x==G1,Gk_vec))
+                        iG2 = only(findall(x-> x==G2,Gkpb_vec))
+                        push!(map_fourier_modes,[iG1,iG2])
+                    end
+                end
+            end
```
antoine-levitt, line 259: Can you split this block off to a function? Both good style and easier to improve perf.

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+            ovlp = 0im
+            
+            ## Extract Fourier coeffs and corresponding vectors in reciprocal lattice
+            Gk_coeffs = ψ[k][:,m] 
+            Gk_vec = G_vectors(pw_basis.kpoints[k])
+            Gkpb_coeffs = ψ[kpb][:,n]
+            Gkpb_vec = [ G - K_shift for G in G_vectors(pw_basis.kpoints[kpb]) ] # Don't forget the shift, see the DOC block
+            
+            ## Search for common Fourier modes, corresponding indices are written in map_fourier_mode
+            map_fourier_modes = []
```
antoine-levitt, line 250: Also a type instability

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+                        iG2 = only(findall(x-> x==G2,Gkpb_vec))
+                        push!(map_fourier_modes,[iG1,iG2])
+                    end
+                end
+            end
+            
+            ## Compute the overlap for mn
+            for (i,j) in map_fourier_modes
+                ovlp += conj(Gk_coeffs[i])*Gkpb_coeffs[j]
+            end
```
antoine-levitt, line 264: Yeah all this is super slow, lots of improvements to be made. Once you go through this round of review let's make a skype call and I'll show you.

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+
+    ## Write file
+    open("$prefix.mmn","w") do f
+        write(f,"Generated by DFTK at ",string(now()),"\n")
+        write(f,"$n_bands  $k_size  $nn_num"*"\n")
+        
+        for i_nnkp in 1:size(nn_kpts,1) #Loop over all (k_points, neirest_neighbour, shif_vector)
+            #Write the label of the matrix
+            k,kpb,shift = read_nn_kpts(i_nnkp)
+            write(f,string(k)*"  "*string(kpb)*"  "*string(shift[1])*"  "*string(shift[2])*"  "*string(shift[3])*"\n")
```
antoine-levitt, line 299: You can use string interpolation here: "$k  $kpb   $(shift[1])` etc

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+    push!(angular_parts, [-5,2] => (θ,φ) -> (1/√6)*s + (1/√2)*px - (1/√12)*dz2 +  (1/2)*dx2_y2) #sp3d2-2
+    push!(angular_parts, [-5,3] => (θ,φ) -> (1/√6)*s - (1/√2)*py - (1/√12)*dz2 -  (1/2)*dx2_y2) #sp3d2-3
+    push!(angular_parts, [-5,4] => (θ,φ) -> (1/√6)*s + (1/√2)*py - (1/√12)*dz2 -  (1/2)*dx2_y2) #sp3d2-4
+    push!(angular_parts, [-5,5] => (θ,φ) -> (1/√6)*s - (1/√2)*pz + (1/√3)*dz2)                  #sp3d2-5
+    push!(angular_parts, [-5,6] => (θ,φ) -> (1/√6)*s + (1/√2)*pz + (1/√3)*dz2)                  #sp3d2-6
+    
+ 
+
+    radial_parts,angular_parts
+end
```
antoine-levitt, line 440: Well I was hoping to avoid doing all this but if you've already done it... :D It's *very* easy to introduce bugs in tables like these (ask Michael, we had a fun bug in the spherical harmonics), which is why I'm always a bit suspicious...

antoine-levitt: I haven't read the code very attentively but if you want to do them in real space, you have to periodize them, no? Eg if you have an atom at the corner of the cell it'll miss most of the function.

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+    Remember that : proj = [ [center], [quantum numbers], [z_axis], [x_axis], α ]
+"""
+function guess_win(proj)
+    center = proj[1]
+    l,mr,r = proj[2]
+    α = proj[5]
+
+    radial_parts,angular_parts = dict_AOs(α)
+    #TODO take into account non-canonical basis
+    @assert(proj[3] == [0.00,0.00,1.00]) #For now we limit ourselves to the canonical basis
```
antoine-levitt, line 456: For comparisons you can just do `[0, 0, 1]`: `[0.0, 0.0, 1.0] == [0, 0, 1]` is true

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+        if kgrid == [1,1,1]
+            write(f,"gamma_only      = true"*"\n")
+        else
+            write(f,"\n")
+        end
+        
+        # kpoints bloc
+        write(f,"begin kpoints"*"\n")
+        for i in 1:size(scfres.ψ,1)
+            for x in basis.kpoints[i].coordinate
```
mfherbst, line 130: Again ... join probably helps here.

src/external/dftk_to_wannier90.jl
mfherbst: And we break lines if they are longer than say 90 characters, much to Anoine's dismay.

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+"""
+function read_nnkp_file(prefix::String)
+
+    ## Check the presence of the file
+    if !isfile("$prefix.nnkp")
+        error("file $prefix.nnkp not found.")   
+    end
+
+    ## Read the file
+    file = open("$prefix.nnkp")
```
mfherbst, line 159: Don't do it like this, use the `do` ... `end` syntax, because this automatically closes the file once you are done. Also I would avoid reading the complete file into the `ln` array ... in case something goes wrong you might end up running out of memory or it just takes forever. I'd just try to loop over the file and extract the relevant blocks or process it on the fly.

src/external/dftk_to_wannier90.jl
mfherbst, line 204: Our convention for this variable name is `scfres` (no underscore)

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+    nn_kpts,nn_num,projs    
+end
+
+
+
+"""
+Take scf_res, the result of a self_consistent_field calculation and writes eigenvalues (in eV) in a format readable by Wannier90.
+"""
+function generate_eig_file(prefix::String,scf_res)
+    Ha_to_eV = 1/(unit_to_au(:eV))
```
mfherbst, line 205: Just use `1/units.eV` or do this directly in the loop below.

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+
+    eigvalues = scf_res.eigenvalues
+    k_size = size(eigvalues,1)
+    n_bands = size(eigvalues[1],1)
+
+    #write file
+    open("$prefix.eig","w") do f
+        for k in 1:k_size
```
mfherbst, line 213: Try to use the `enumerate` iterator for such things. It gives you an index and the object at the same time, which is typically much nicer to read.

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+
+
+"""
+Read the .nnkp file provided by the preprocessing routine of Wannier90 (i.e. "wannier90.x -pp prefix")
+Returns: 
+1) the array 'nn_kpts' of k points, they respective neirest neighbours and associated shifing vectors (non zero if the neighbour of is located in another cell).
+2) the number 'nn_num' of neighbours per k point. 
+2) the array 'projs' of projections. Each line contains the informations for one projection :  [center],[quantum numbers: l,mr,r],[ref. z_axis],[ref. x_axis],α = Z/a
+"""
+function read_nnkp_file(prefix::String)
```
mfherbst, line 151: I don't really know the format of the nnkp files, but this parser code looks a little complicated at some parts and likely can be made a little easier to read, but without understanding the file format there is little I can say. We should talk about it when we get closer to merge this.

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+    file = open("$prefix.nnkp")
+    @info "Reading nnkp file", file 
+    ln = [string(l) for l in eachline(file)]
+    close(file)
+
+
+    ## Extract nnkp block 
+    i_nn_kpts = findall(x-> endswith(x,"nnkpts"),ln)                               #Indices of the first and last line of the nnkpts block
+    @assert size(i_nn_kpts,1) == 2
+    char_to_vec(line) = [parse.(Int,x) for x in split(line,' ',keepempty=false)]
```
mfherbst, line 168: That's a mouthful! What does it do?

src/external/dftk_to_wannier90.jl
mfherbst: I generally would not recommend doing this before the function (unless you know everything is working). Better do it selectively with an `@view` instead.

src/external/dftk_to_wannier90.jl
mfherbst: I think this is an instance where a `map` call to a product iterator could be helpful.

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+
+"""
+Iteratively use the preceeding function on each k and kpb to generate the whole .mmn file.
+"""
+function generate_mmn_file(prefix::String,pw_basis::PlaneWaveBasis,ψ, nn_kpts, nn_num)
+    ## general parameters
+    n_bands = size(ψ[1],2)
+    k_size = length(ψ)
+    
+    progress = Progress(size(nn_kpts,1),desc = "Computing Mmn overlaps : ")
```
mfherbst, line 286: For kwargs we don't use spaces, so: `desc="Compute ..."`

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+    
+    # Small function for the sake of clarity
+    read_nn_kpts(n) = nn_kpts[n][1],nn_kpts[n][2],nn_kpts[n][3:end]
+
+    ## Write file
+    open("$prefix.mmn","w") do f
```
mfherbst, line 292: Yeah! Nice.

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+    end
+
+end
+
+
+"""
+Given a k point, provide the indices of the corresponding G_vectors in the general FFT grid.
+Used in 'A_k_matrix_gaussian_guesses'.
+"""
+function map_k_G_vectors(pw_basis::PlaneWaveBasis,k,fft_grid)
```
mfherbst, line 316: Don't we have a function for that (index_G_vectors)?

src/external/dftk_to_wannier90.jl
mfherbst: Also this is a big code duplication with src/common/spherical_harmonics.jl. We can definitely refactor this and unify the implementations.

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+    g_n(r,θ,φ) = R(r)*Θ(θ,φ)
+
+    g_n
+    
+end
+
+"""
+Convert cartesian to spherical coordinates.
+"""
+function cart_to_spherical(cart_coord)
```
mfherbst, line 472: There should be no need to do that if you use the spherical harmonics implementation we already have that should be implicit.

mfherbst (commented):
Wow. Nice work. I think there are quite a few place where one can simplify the code (which will avoid bugs and make it easier to understand).

Also two bigger aspects to think about:
- This file is too long and should probably be split up (which will come more naturally when we refactor later, so don't worry about it too much)
- Perhaps it is a good idea to try and call into wannier90 directly from DFTK, because than you might be in more control of the workflow. Not immediately important, but useful for later (to make this easier to use by users).

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+# It is bugged and may not be usefull but I left it there.
+
+## TODO
+# 1 - check if kgrid can be removed from dftk2wan_win_file arguments
+# 2 - add handling of the SCDM routine as an alternative guess
+# 3 - optimize the "generate_mmn" routine
+
+
+"""
+    Create a .win file for Wannier90, compatible with the system studied with DFTK.
```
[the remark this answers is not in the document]
antoine-levitt: I would not bother too much with this for now... The file workflow makes it easier to interop with other tools.

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+                k_A = round.(DFTK.normalize_kpoint_coordinate.(get!(labels,A,1)), digits = 5)
+                k_B = round.(DFTK.normalize_kpoint_coordinate.(get!(labels,B,1)), digits = 5)
+                # Wannier90 convention for the Γ point is G
+                if A == "\\Gamma"
+                    A = "G"
+                end
+                if B == "\\Gamma"
+                    B = "G"
+                end
+                write(f,"$A $(k_A[1]) $(k_A[2]) $(k_A[3]) $B $(k_B[1]) $(k_B[2]) $(k_B[3])"*"\n")
```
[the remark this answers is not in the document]
antoine-levitt: > Antoine's comment is obsolete in the recent DFTK version, where high_symmetry_kpath should directly return Γ.

But that's the point, he wants G and not Γ, no?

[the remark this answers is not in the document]
antoine-levitt: Nice trick with the `join`

src/external/dftk_to_wannier90.jl
antoine-levitt: Not sure I agree here: yes in theory it's better to minimize passes over the data, but if it's easier to just read the whole thing to search within it, just do it like that. nnkp should be relatively small files (eg compared with mmn which are huge), so no issue here

src/external/dftk_to_wannier90.jl
LaurentVidal95: I tried but the output is not a string and string(unit_cell) writes "[line 1; line 2; line 3]".
I managed to change that with the @sprintf macro however. I think it is a bit better but there is still one "for" loop over the lines of the unit_cell.

line_sprintf(l) = @sprintf("%.6f %.6f %.6f \n",l[1],l[2],l[3])
for i in 1:3
      write(f,line_sprintf(unit_cell[i,:]))
end

src/external/dftk_to_wannier90.jl
LaurentVidal95: They are here because I entered the lattice vectors in cartesian coordinates and in bohr. 
But that's a good point I thought this was the standard way to do in DFTK but if there are other possible units and coordinates we should take this into account. For now it only works with Bohr + cartesian.

src/external/dftk_to_wannier90.jl
mfherbst: I would just do:
```
for uc in eachcol(unit_cell)
    write(f, uc..., "\n")
end
```

src/external/dftk_to_wannier90.jl
mfherbst: No in DFTK lattice vectors are always Bohr (and in columns).

src/external/dftk_to_wannier90.jl
LaurentVidal95: Indeed works fine !

src/external/dftk_to_wannier90.jl
LaurentVidal95: In fact it works without "gamma_only" so I removed this line (I thought it was necessary) !

src/external/dftk_to_wannier90.jl
LaurentVidal95: I did not find anything in the user guide but I'll look for this elsewhere. When I did the test on entangled bands, I didn't correct the shift and I could have the good results though. So I thought I could just go along with the shift.

src/external/dftk_to_wannier90.jl
LaurentVidal95: I see : ovlp = 0im gives a Complex{Int64}. As I understand it doing ovlp = zero(ComplexF64) should be better.

antoine-levitt: Yup. Typically we try to never put floating point types explicitly but try to guess it from the environment (eg eltype(basis)). Although since you're going to write it to a file, might as well do F64 explicitly

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+# 2 - add handling of the SCDM routine as an alternative guess
+# 3 - optimize the "generate_mmn" routine
+
+
+"""
+    Create a .win file for Wannier90, compatible with the system studied with DFTK.
+    Options have to be compeleted by hand.
+"""
+function dftk2wan_win_file(prefix::String,basis::PlaneWaveBasis,scfres,num_wann;
+                           kgrid = [4,4,4], #can we retrieve this from scfres or basis ? Feel this should not be an argument.
```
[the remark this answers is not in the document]
LaurentVidal95: The win file asks for the k grid in the specific format : 
mp_grid : 4 4 4 (or other grids)
That's the only reason I ask for the kgrid. But if it is possible to retrieve the "4 4 4" in some way, or if we decide not to generate a win file after all this will be obsolete !

src/external/dftk_to_wannier90.jl
LaurentVidal95: I think you're right and it is called..... "index_G_vectors" !

src/external/dftk_to_wannier90.jl
LaurentVidal95: Hello Michael, the high_symmetry_kpath still gives me "\\Gamma"  labels, so it seems that I didn't update DFTK the right way.
There's something I don't understand because I installed DFTK via the package manager, but we updated the folder I obtained latter via the fork. For me these two things are not the same DFTK so shouldn't I update from pkg directly ? I think I'm missing something here.

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,605 @@
...
+                           num_iter = 500,
+                           num_print_cycles = 50
+                           )
+    
+    ## Check kgrids
+    if (prod(kgrid) != size(scfres.ψ,1))
+        error("The given kgrid doesn't match the one used for the scf calculation")
+    end    
+
+    ## Write in file
```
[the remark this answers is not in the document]
LaurentVidal95: I intentionally put the comments for major blocs with two "#" and subcomments with a single "#" but I can easily change that if it doesn't meet dftk conventions !

src/external/dftk_to_wannier90.jl
mfherbst: I'm not exactly sure what you did, but Julia allows you to have multiple versions of the same package. Maybe the (recently updated) docs https://juliamolsim.github.io/DFTK.jl/stable/guide/installation/#Developer-setup help you? You basically have to tell Julia somehow that you want to use the code of your fork. So either you use `julia --project=/some/path/` with the path to your fork or `julia --project` in a shell with working directory `/some/path` or you use the `dev` method outlined in the manual.

src/external/dftk_to_wannier90.jl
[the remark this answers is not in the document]
mfherbst: Please do. In my experience if one feels these sort of "multilevel" comments are needed it's a sign that you ought better split up your one big function into smaller ones. That than automatically takes care to show the logical steps of an algorithm the way the `##` comments would do.

src/external/dftk_to_wannier90.jl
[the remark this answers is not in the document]
LaurentVidal95: > > Antoine's comment is obsolete in the recent DFTK version, where high_symmetry_kpath should directly return Γ.
> 
> But that's the point, he wants G and not Γ, no?

Indeed Wannier90 only understands G !

examples/dfk2wan.jl
```diff
@@ -51,29 +48,80 @@ model = model_PBE(lattice,atoms)
...
+                  dis_num_iter      =  120,
+                  dis_mix_ratio     = "1d0")
 
-# Then, we perfom a preproprocessing task via Wannier90 to generate
+# We then perfom a preproprocessing task via Wannier90 to generate
 # the `Si.nnkp` file. This is simply done by :
 
-PathToWannier90/wannier90.x -pp Si
+#md #```@raw bash
+#md #$ PathToWannier90/wannier90.x -pp Si
```
mfherbst, line 73: There is now https://github.com/JuliaBinaryWrappers/wannier90_jll.jl, so you can (soon) just install and use Wannier90 for the Julia repl and from Julia scripts. E.g. for this you would:
```
using wannier90_jll
wannier90() do exe
    run(`$exe -pp Si`)
end
```

antoine-levitt: Oh wow that's cool. Did you do this?

mfherbst: Yes.

antoine-levitt: Haha, week-ends I do performance optimizations, you do devops, something is wrong with us ;-) So basically you did this https://github.com/JuliaPackaging/Yggdrasil/pull/2153/files and then it works all by itself? Neat!

examples/dfk2wan.jl
```diff
@@ -0,0 +1,149 @@
...
+wannier90() do exe
+    run(`$exe $prefix`)
+end
+
+
+# Note that it can be useful to compute several `.amn` files to test different
+# guesses. The user can simply disable the generation of a
+# specific file with the keyword arguments `write_mnn = false`,
+# `write_amn = false` or `write_eig = false` in
+# `dftk2wan_wannierization_files`.
```
antoine-levitt, line 141: might as well put these keywords explicitly in the call to wannierization_files: this way it's clear what it does and there's no need to document it explicitly

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,415 @@
...
+using Printf
+
+## TODO
+# add handling of the SCDM routine as an alternative guess
+
+include("win_guess_utils.jl")
+
+"""
+    Create a .win file for Wannier90, compatible with the system studied with DFTK.
+    Options have to be compeleted by hand.
```
antoine-levitt, line 13: completed

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,415 @@
...
+    Parameters can also be added as kwargs argument : e.g. num_iter = 500.
+"""
+function dftk2wan_win_file(prefix::String, basis::PlaneWaveBasis, scfres,
```
antoine-levitt, line 16: same as above, just take scfres

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,415 @@
...
+                           )
+    
+    # Write in file   
+    open("$prefix.win", "w") do f
+
+        write(f,"! $prefix.win file generated by DFTK at $(now())"*"\n"^2)
+        
+        # Required parameters
+        num_bands = size(scfres.ψ[1],2)
+        write(f,@sprintf("%-20s %s %5s \n","num_bands","=",num_bands))
```
antoine-levitt, line 29: do you really have to be this careful with formating? Can't you just do `write(f, "num_bands=$num_bands")`?

src/external/dftk_to_wannier90.jl
antoine-levitt: also I'm not sure why this is necessary?

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,415 @@
...
+        write(f,"\n")
+        
+        # Optional parameters in kwargs
+        for (key,value) in kwargs
+            write(f,@sprintf("%-20s %s   %-30s \n",key,"=",value))
+        end
+        write(f,"\n"^2)
+
+        # Messages for the user
+        write(f,"!"^20*" Complete with optional parameters :"*"\n"^2)
```
antoine-levitt, line 40: remove?

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,415 @@
...
+
+        # System
+        write(f,"!"^20*" System"*"\n"^2)
+
+        # Unit cell block
+        write(f,"begin unit_cell_cart"*"\n"*"bohr"*"\n")
+        unit_cell = transpose(basis.model.lattice) # cell vectors are in raws in Wannier90
```
antoine-levitt, line 47: rows

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,415 @@
...
+    
+end
+
+
+"""
+Read the .nnkp file provided by the preprocessing routine of Wannier90 
+(i.e. "wannier90.x -pp prefix")
+Returns: 
+1) the array 'nn_kpts' of k points, they respective nearest neighbors and associated
+   shifing vectors (non zero if the neighbour of is located in another cell).
```
antoine-levitt, line 124: neighbor

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,415 @@
...
+    return nothing
+    
+end
+
+
+"""
+Read the .nnkp file provided by the preprocessing routine of Wannier90 
+(i.e. "wannier90.x -pp prefix")
+Returns: 
+1) the array 'nn_kpts' of k points, they respective nearest neighbors and associated
```
antoine-levitt, line 123: their

examples/dfk2wan.jl
```diff
@@ -0,0 +1,149 @@
...
+# the number of wanted MLWFs; 8 in our case. The keyword `bands_plot = true`
+# adds the ``k`` point path to the `.win` file, needed to plot the bands structure.
+# We also add as keyword arguments the parameters of the disentanglement
+# procedure. Any other argument can be added as such.
+
+prefix = "Si"
+dftk2wan_win_file(prefix, basis, scfres, 8;
+                  bands_plot=true, num_print_cycles=50, num_iter=500,
+                  dis_win_max       = "17.185257d0",
+                  dis_froz_max      =  "6.8318033d0",
```
antoine-levitt, line 66: is this eV or Hartree? What's the reference?

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,415 @@
...
+    end
+    return nothing
+end
+
+
+@doc raw"""
+ Computes the matrix ``[M^{k,b}]_{m,n} = \langle u_{m,k} | u_{n,k+b} \rangle`` 
+ for given k, kpb = k+b.
+ 
+ Return in the format of a (n_bands^2, 2) matrix, where each line is one overlap,
```
antoine-levitt, line 196: not up to date

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,415 @@
...
+    k = basis.kpoints[ik]; kpb = basis.kpoints[ikpb] #renaming for clarity
+    map_fourier_modes = [ (iGk,DFTK.index_G_vectors(basis, kpb, Gk+G_shift))
+                          for (iGk,Gk) in enumerate(G_vectors(k))
+                          if !isnothing(DFTK.index_G_vectors(basis,kpb,Gk+G_shift)) ]
+
+    # Compute overlaps
+    for n in 1:n_bands
+        for m in 1:n_bands
+            # Select the coefficient in right order
+            Gk_coeffs = @view ψ[ik][[i[1] for i in map_fourier_modes],m]
```
antoine-levitt, line 219: you're doing this `[i[1] for i in map_fourier_modes]` at each step, which is slow. Better to precompute it (which suggests that map_fourier_modes is not the right structure anyway)

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,415 @@
...
+            Mkb[m,n] = dot(Gk_coeffs,Gkpb_coeffs)
+        end
+    end    
+
+    Mkb
+end
+
+
+"""
+Iteratively use the preceeding function on each k and kpb to generate the whole .mmn file.
```
antoine-levitt, line 230: eding

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,415 @@
...
+Compute the matrix ``[A_k]_{m,n} = \langle ψ_m^k | g^{\text{per}}_n \rangle`` 
+
+``g^{per}_n`` are periodized gaussians whose respective centers are given as an
+ (n_bands,1) array [ [center 1], ... ].
+Centers are to be given in lattice coordinates.
+
+The dot product is computed in the Fourier space. 
+
+Given an orbital ``g_n``, the periodized orbital is defined by :
+ ``g^{per}_n=  \sum\limits_{R in lattice} g_n( ⋅ - R)``.
```
antoine-levitt, line 276: if you want to do proper latex do proper latex (ie \cdot, \in and {\rm lattice}

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,415 @@
...
+
+``g^{per}_n`` are periodized gaussians whose respective centers are given as an
+ (n_bands,1) array [ [center 1], ... ].
+Centers are to be given in lattice coordinates.
+
+The dot product is computed in the Fourier space. 
+
+Given an orbital ``g_n``, the periodized orbital is defined by :
+ ``g^{per}_n=  \sum\limits_{R in lattice} g_n( ⋅ - R)``. 
+``g^{per}_n`` is not explicitly computed. Its Fourier coefficient at any G
```
antoine-levitt, line 277: it is explicitly computed, just not in real space

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,415 @@
...
+ ``g^{per}_n=  \sum\limits_{R in lattice} g_n( ⋅ - R)``. 
+``g^{per}_n`` is not explicitly computed. Its Fourier coefficient at any G 
+is given by the value of the Fourier transform of ``g_n`` in G.
+"""
+function A_k_matrix_gaussian_guesses(basis::PlaneWaveBasis, ψ, k::Integer,
+                                     n_bands::Integer, n_wann::Integer;
+                                     centers = [], coords = "", projs = [])
+
+    # associate a center with the fourier transform of the corresponding gaussian
+    guess_fourier(center) = xi ->  exp(-im*dot(xi,center) - dot(xi,xi)/4)
```
antoine-levitt, line 285: we typically call the Fourier variable `q`

antoine-levitt: also this kind of function returning function is usually a big performance gotcha, and can be hard to read. Might as well do `guess_fourier(center, q) = ...`

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,415 @@
...
+"""
+function A_k_matrix_gaussian_guesses(basis::PlaneWaveBasis, ψ, k::Integer,
+                                     n_bands::Integer, n_wann::Integer;
+                                     centers = [], coords = "", projs = [])
+
+    # associate a center with the fourier transform of the corresponding gaussian
+    guess_fourier(center) = xi ->  exp(-im*dot(xi,center) - dot(xi,xi)/4) 
+
+    # Select G_vectors and the scalar product in reduced or cartesian coordinates
+    if coords == "reduced"
```
antoine-levitt, line 288: I'd rather avoid this and always work in reduced

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,415 @@
...
+        # "2π .*" so that the dot product is in reduced coordinates
+        G_vec = (2π) .*[G for (iG,G) in enumerate(G_vectors(basis))] 
+    else
+        G_vec =[G for (iG,G) in enumerate(G_vectors_cart(basis))]
+    end
+    
+    # Indices of the Fourier modes of the bloch states in the general FFT_grid for given k
```
antoine-levitt, line 295: Bloch

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,415 @@
...
+    index = [DFTK.index_G_vectors(basis,G) for G in G_vectors(basis.kpoints[k])]                    
+
+    A_k = zeros(Complex,(n_bands,n_wann))
+
+    # Compute A_k
+    for n in 1:n_wann
+        fourier_gn = guess_fourier(basis.model.lattice*transpose(centers[n]))
+        # functions are l^2 normalized in Fourier, in DFTK conventions.
+        norm_gn_per = norm(fourier_gn.(G_vec),2)
```
antoine-levitt, line 304: well normalizations are probably wrong anyway, so don't bother too much...

antoine-levitt: Although OK yeah that's probably good, keep it

src/external/dftk_to_wannier90.jl
```diff
@@ -0,0 +1,415 @@
...
+        # Fourier coeffs of gn_per for G_vectors in common with ψm
+        coeffs_gn_per = fourier_gn.(G_vec[index])./ norm_gn_per     
+        # Compute overlap
+        for m in 1:n_bands
+            coeffs_ψm = ψ[k][:,m]
+            A_k[m,n] = dot(coeffs_ψm,coeffs_gn_per)
+        end  
+    end
+    
+    A_k
```
antoine-levitt, line 315: We usually just write Ak without underline

src/external/win_guess_utils.jl
```diff
@@ -0,0 +1,145 @@
...
+    (l == 2) && return m_d[mr]  # d
+    (l == 3) && return m_f[mr]  # f
+    error("Quantum numbers are not matching any implemented
+                orbital (s,p,d,f)")
+end
+
+@doc raw"""
+    Gives the analytic expression of the integral 
+    ``I_l(G) = \int_{\mathbb{R}^+} r^(l+2) exp(-r^2/2) j_l(|G|r)dr``
+    as given in arXiv:1908.07374v2
```
antoine-levitt, line 29: pretty sure that's standard and there's no need to go into such a specific reference, but OK to keep it as it is. Add reference to the specific equation (since that paper is about a much more complicated problem)

src/external/win_guess_utils.jl
```diff
@@ -0,0 +1,145 @@
...
+    arg_ylm = -Gcart ./ Gnorm
+
+    # Computes the phase prefactor due to center ≠ [0,0,0]
+    phase_prefac = exp(-im*dot(Gcart,center))
+    
+    if l ≥ 0  # s,p,d or f
+        m = retrieve_proper_m(l,mr)
+        return (phase_prefac *
+                (4π*im^l)*DFTK.ylm_real(l,m,arg_ylm) * intR_l(l,Gcart) )
+    else      # hybrid orbitals
```
antoine-levitt, line 70: hm, I'm pretty reluctan to have this because it's very hard to mess it up and we haven't tested them all... I'd rather not have the feature than have it with possible bugs. Unless you're very sure these are correct, can you either comment this block out or print a warning that these formulas are not properly tested and the user should use them at their own risk?

antoine-levitt (commented):
Looks nice, great job!

antoine-levitt:
Probably doesn't work with MPI: in the MPI case each process only sees a subset of the kpoint data. Add a check to not run the tests in that case (look how it's done in other tests) and have an assert to fail wannierization rather than do incorrect things (see eg direct_minimization.jl)

examples/dfk2wan.jl
```diff
@@ -0,0 +1,149 @@
...
+
+# !!! note "Launching Wannier90 from Julia file or shell"
+#     All Wannier90 calls can be done through julia, with
+#     the help of the package wannier90_jll. The previous
+#     command is written as follows
+
+using wannier90_jll
+wannier90() do exe
+    run(`$exe -pp Si`)
+end
```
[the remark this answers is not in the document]
mfherbst: Also if this is the workflow, why is there no function in DFTK that does the whole lot? Is there a reason why the user has to "see" the individual steps?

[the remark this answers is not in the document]
mfherbst: >  Why can't that just be wannier90("-pp Si")?

Because the `wannier90_jll` actually messes with the environment variables to ensure that wannier90 and dependent libraries are found. So if we do not want to hide this call inside DFTK than this is the way it has to be done.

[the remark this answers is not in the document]
antoine-levitt: The wannier90 workflow is sufficiently messy and error-prone that it's good to leave the black box open here. Also we might want to replace parts of wannier90 (like all of it) incrementally, so let's not commit ourselves to something final just yet

mfherbst:
Not sure how close we are to merging this. I have a few more nits regarding syntax, file structure etc to bikeshed, but I have not yet put it down (because it belongs to the pass when the code structure is there). On a quick glance I agree it's getting into shape.

examples/dfk2wan.jl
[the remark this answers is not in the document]
mfherbst: I see. Ok. Still perhaps we might want to have a nice support function that hides at least the wannier90_jll ... don't know just a suggestion.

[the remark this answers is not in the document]
antoine-levitt: I don't understand. What's keeping me from defining `wannier90(opts::String) = wannier90(exe -> run(`$exe $opts`)`?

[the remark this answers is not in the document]
antoine-levitt: meh github doesn't like nested backticks, but you understand what I mean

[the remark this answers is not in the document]
LaurentVidal95: In fact I chose to work that way so the user can control the win file and add what's needed. I think we choose:
 1) to hide all the nnkp creation and creation of amn, mmn, eig, but every options has to be called as kwargs.
 2) keep the process split in several steps to be able to control them individually. 

Since I didn't use much Wannier90 I don't quite realize how badly the calculation can turn if you cannot inspect every step. Nevertheless, with wannier90__jll the option 1 can be done very easily I think.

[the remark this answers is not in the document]
LaurentVidal95: Also if we choose the option 1) I can remove all the formating process of the win file to reduce the length of the parser (even if I have to admit I love well parsed files).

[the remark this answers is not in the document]
mfherbst: Oh in DFTK nothing, but that is not the right way to do it ;). It should be:
```
wannier90(opts::Cmd) = wannier90_jll.wannier90(exe -> run(`$exe $opts`)
```
To do with shell quoting :smile:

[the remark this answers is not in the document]
antoine-levitt: Yeah right, try a file name with spaces and see how wannier90 likes it :-p Anyway, I don't know but this command thing just feels like a very roundabout API for running a command.

mfherbst:
By the way: For some reason the CI hangs until the maximum runtime of 360 minutes here! There is definitely something wrong in your example (e.g. is wannier waiting indefinitely for input or just the wannierisation takes forever). Please check this before you push again, because we get these CI minutes for free ...

examples/dfk2wan.jl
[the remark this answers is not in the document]
mfherbst: Well it's low-level. That's the point of the jll packages. If you don't like it ... write a wrapper! That's the point.

antoine-levitt:
hanging waiting for input consumes no CPU so the planet is safe at least ! But yeah I'm also in the habit of pushing a bit carelessly, forgetting that this whole thing is run each time

mfherbst:
I think that's fine as long it's doing sth useful, but here clearly is broken and as an act of courtesy I would not overdo it.

examples/dfk2wan.jl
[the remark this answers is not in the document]
antoine-levitt: I don't understand what's going here, but that stuff passes me over the head. If it's needed, might as well hide it : by defining a DFTK.run_wannier90(options) that calls `split` on the input and passes it to the wannier90_jll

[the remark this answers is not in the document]
antoine-levitt: To clarify above: I'm not bitching about the wannier90_jll, which I think is awesome, I'm bitching about why we have to pass a closure to run a command line program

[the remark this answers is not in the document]
mfherbst: I also think we misunderstand. Just to be certain: You mean in the example or in general to run the jll?

I'm 0% against having a `run_wannier90` function that takes some args, processes them to the extent needed and then runs wannier.

src/external/dftk_to_wannier90.jl
LaurentVidal95: Yes I initially wanted to take the first and second column of a table but I don't know if it is feasible with Tuples.
Still I will take this out of the loop !

src/external/win_guess_utils.jl
LaurentVidal95: Ok I just left the one we tested for this commit ! That is the already tested s, p, d, and f and the sp3. If still strange I will cut it in the next one.

antoine-levitt:
I'm reading through the PR and making small fixes along the way, please take a look at them to see what I did and check I haven't broken anything. Some remarks:

- typically we call q=G+k, and so the function eval_fourier_orbital and intR_l would more naturally take a q as argument. It is strange to me that Ak_matrix_win_guesses uses the G vectors and not the k+G, can you make sure that's not a bug?
- Ideally, eval_fourier_orbital could be merged with the routines that compute the Fourier components of the projectors in the psp codes... Maybe for a later PR.
- The function Ak_matrix_win_guesses should take as input psik and the kpoint instead of the whole of psi and basis
- Continuing from that principle, it is good practice that a function only takes as much info as it needs to do its job. So for instance intR_l should take as input q = norm(G+k) rather than a vector.
- Spacing is inconsistent, I fixed some of it
- Can you rename win_guess_utils to w90_guess_utils? Or just merge it into dftk_to_wannier90.
- You have a lot of trailing whitespace. I don't particularly care but in general it is a good practice to not have any: look into configuring your editor for this.
- We don't usually type input parameters unless for clarity or for dispatch.

Otherwise it looks good. I haven't done a full review because I don't have the time right now, but if it works and is tested we should probably merge it. If I have time at some point I will revisit this whole thing by merging my old wannier code, and then I might change a few things.

LaurentVidal95:
Hello everyone !

Sorry for the multiple commit. The INRIA laptop had automatically changed my git username for commits so I fixed that but I had to push again apparently.

I forgot to mention in the commit that I added a function "run_wannier90" that wraps all the wannierization procedure, using Michael's wannier90_jll library. I also fixed some of the point mentioned above by Antoine.
It is maybe possible to change the structure of the code so that the Ak generation routines don't take the whole basis as an argument if it is nicer that way ! I did not do it for the moment !

Regarding the documentation, I didn't manage to fix the use of python libraries (ase and pymatgen) which is broken for me (it doesn't work on my INRIA laptop but is working fine on my personnal one...), but more importantly my example generates files. So every time I try to compile the doc I end up with all the files in the doc/build/example folder ; I don't know if this is a problem.

Bests,
Laurent

PR #377, opened 2020-11-18
Anyon optimization

antoine-levitt:
@mfherbst since you're interested in isolated systems, expanded description: for anyons, we have an isolated system (with confining potential), and have to solve a system ∇∧A = 2π ρ, div A = 0, so it's like rho acts like a source of magnetic potential instead of electric potential (anyons are weird). The same problem as with electrostatic occurs: if you have a nonzero mass for rho, then A is long-range, and that makes it super slow to converge with system size. So I do an "Ewald-like" trick and compute the long-range part explicitly, which makes it converge *a lot* faster (basically it's impossible to get to reasonable convergence without this trick). This is what I meant when discussing tricks for isolated systems; eg if you have an isolated ion in DFT probably you want to do something like this? People might already be doing this...

### DIFF
--- a/examples/anyons.jl
+++ b/examples/anyons.jl
@@ -1,0 +1,34 @@
+# # Anyons
+
+# We solve the almost-bosonic anyon model of https://arxiv.org/pdf/1901.10739.pdf
+
+using DFTK
+using StaticArrays
+using Plots
+
+# Unit cell. Having one of the lattice vectors as zero means a 2D system
+a = 14
+lattice = a .* [[1 0 0.]; [0 1 0]; [0 0 0]];
+
+# Confining scalar potential
+pot(x, y, z) = ((x - a/2)^2 + (y - a/2)^2)
+
+# Parameters
+Ecut = 50
+n_electrons = 1
+β = 140
+
+terms = [Kinetic(2),
+         ExternalFromReal(X -> pot(X...)),
+         Anyonic(β)
+]
+model = Model(lattice; n_electrons=n_electrons,
+              terms=terms, spin_polarization=:spinless)  # "spinless electrons"
+# basis = PlaneWaveBasis(model, Ecut, kgrid=(1, 1, 1), fft_size=DFTK.determine_grid_size(lattice, Ecut, supersampling=1.2))
+basis = PlaneWaveBasis(model, Ecut, kgrid=(1, 1, 1); fft_size=DFTK.determine_fft_size(lattice, Ecut, supersampling=1.2))
+scfres = direct_minimization(basis, tol=1e-14)  # Reduce tol for production
+E = scfres.energies.total
+s = 2
+E11 = π/2 * (2(s+1)/s)^((s+2)/s) * (s/(s+2))^(2(s+1)/s) * E^((s+2)/s) / β
+println("e(1,1)/(2π)= ", E11/(2π))
+display(heatmap(scfres.ρ.real[:, :, 1], c=:blues))
--- a/examples/anyons.jl
+++ b/examples/anyons.jl
@@ -16,16 +16,15 @@ pot(x, y, z) = ((x - a/2)^2 + (y - a/2)^2)
 # Parameters
 Ecut = 50
 n_electrons = 1
-β = 140
+β = 5
 
 terms = [Kinetic(2),
          ExternalFromReal(X -> pot(X...)),
-         Anyonic(β)
+         Anyonic(1, β)
 ]
 model = Model(lattice; n_electrons=n_electrons,
               terms=terms, spin_polarization=:spinless)  # "spinless electrons"
-# basis = PlaneWaveBasis(model, Ecut, kgrid=(1, 1, 1), fft_size=DFTK.determine_grid_size(lattice, Ecut, supersampling=1.2))
-basis = PlaneWaveBasis(model, Ecut, kgrid=(1, 1, 1); fft_size=DFTK.determine_fft_size(lattice, Ecut, supersampling=1.2))
+basis = PlaneWaveBasis(model, Ecut, kgrid=(1, 1, 1); fft_size=DFTK.compute_fft_size(lattice, Ecut, supersampling=1.2))
 scfres = direct_minimization(basis, tol=1e-14)  # Reduce tol for production
 E = scfres.energies.total
 s = 2
--- a/examples/anyons.jl
+++ b/examples/anyons.jl
@@ -24,10 +24,11 @@ terms = [Kinetic(2),
 ]
 model = Model(lattice; n_electrons=n_electrons,
               terms=terms, spin_polarization=:spinless)  # "spinless electrons"
-basis = PlaneWaveBasis(model, Ecut, kgrid=(1, 1, 1); fft_size=DFTK.compute_fft_size(lattice, Ecut, supersampling=1.2))
+basis = PlaneWaveBasis(model, Ecut, kgrid=(1, 1, 1);
+                       fft_size=DFTK.compute_fft_size(lattice, Ecut, supersampling=1.2))
 scfres = direct_minimization(basis, tol=1e-14)  # Reduce tol for production
 E = scfres.energies.total
 s = 2
 E11 = π/2 * (2(s+1)/s)^((s+2)/s) * (s/(s+2))^(2(s+1)/s) * E^((s+2)/s) / β
-println("e(1,1)/(2π)= ", E11/(2π))
+println("e(1,1) / (2π)= ", E11 / (2π))
 display(heatmap(scfres.ρ.real[:, :, 1], c=:blues))
--- a/src/eigen/preconditioners.jl
+++ b/src/eigen/preconditioners.jl
@@ -33,7 +33,8 @@ mutable struct PreconditionerTPA{T <: Real}
 end
 
 function PreconditionerTPA(basis::PlaneWaveBasis{T}, kpt::Kpoint; default_shift=1) where T
-    kin = Vector{T}([sum(abs2, G + kpt.coordinate_cart)
+    scaling = only([t for t in basis.model.term_types if t isa Kinetic]).scaling_factor
+    kin = Vector{T}([scaling * sum(abs2, G + kpt.coordinate_cart)
                      for G in G_vectors_cart(kpt)] ./ 2)
     @assert basis.model.spin_polarization in (:none, :collinear, :spinless)
     PreconditionerTPA{T}(basis, kpt, kin, nothing, default_shift)
--- a/src/terms/anyonic.jl
+++ b/src/terms/anyonic.jl
@@ -9,6 +9,67 @@
 # where only the first three terms "count for the energy", and where
 # J = 1/(2i) (u* ∇u - u ∇u*)
 
+
+# for numerical reasons, we solve the equation ∇∧A = 2π ρ in two parts:
+# ∇∧A = 2π(ρ-ρref + ρref)
+# where ρref is a gaussian centered at the origin and with the same integral as ρ (=M)
+# ρref(x) = M exp(-1/2 (x/σ)^2) / (σ^3 (2π)^3/2)
+
+function ρref_real(x::T, y::T, M, σ) where {T <: Real}
+    r = hypot(x, y)
+    M * exp(-1/2 * (r/σ)^2) / (σ^2 * (2T(π))^(2/2))
+end
+
+function magnetic_field_produced_by_ρref(x::T, y::T, M, σ) where {T <: Real}
+    # The solution of ∇∧A = 2π ρref is ϕ(r) [-y;x] where ϕ satisfies
+    # ∇∧A = 2ϕ + r ϕ' => rϕ' + 2 ϕ = 2π ρref
+    # Wolfram alpha (after a bit of coaxing) says the solution to
+    # rϕ' + 2ϕ = C exp(-α x^2)
+    # is
+    # 1/x^2 (c1 - C exp(-α x^2)/2α), which coupled with smoothness at 0 gives
+    # C/(2α)/x^2*(1 - exp(-α x^2))
+    r = hypot(x, y)
+    r == 0 && return magnetic_field_produced_by_ρref(1e-8, 0.0, M, σ) # hack
+
+    α = 1/(2*σ^2)
+    C = 2T(π)*M/(σ^2 * (2T(π))^(2/2))
+    r = hypot(x, y)
+    ϕ = 1/2*C/α/r^2*(1-exp(-α*r^2))
+    ϕ * @SVector[-y, x]
+end
+
+function make_div_free(basis::PlaneWaveBasis{T}, A) where {T}
+    out = [zeros(complex(T), basis.fft_size), zeros(complex(T), basis.fft_size)]
+    A_fourier = [from_real(basis, A[α]).fourier for α = 1:2]
+    for (iG, G) in enumerate(G_vectors(basis))
+        vec = [A_fourier[1][iG], A_fourier[2][iG]]
+        G = [G[1], G[2]]
+        if iG != 1
+            out[1][iG], out[2][iG] = vec - (G'vec)*G/sum(abs2, G)
+        else
+            out[1][iG], out[2][iG] = vec
+        end
+    end
+    [from_fourier(basis, out[α]).real for α = 1:2]
+end
+
+function _test_ρref()
+    x = 1.23
+    y = -1.8
+    ε = 1e-8
+    M = 2.31
+    σ = 1.81
+    dAdx = (magnetic_field_produced_by_ρref(x+ε, y, M, σ) - magnetic_field_produced_by_ρref(x, y, M, σ))/ε
+    dAdy = (magnetic_field_produced_by_ρref(x, y+ε, M, σ) - magnetic_field_produced_by_ρref(x, y, M, σ))/ε
+    curlA = dAdx[2] - dAdy[1]
+    divA = dAdx[1] + dAdy[2]
+    @assert norm(curlA - 2π*ρref_real(x, y, M, σ)) < 1e-4
+    @assert abs(divA) < 1e-6
+end
+_test_ρref()
+
+
+
 struct Anyonic
     β
 end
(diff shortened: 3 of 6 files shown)


mfherbst:
Ah I see. I'm not sure if people are doing it like this, but the corrections given the literature are clearly related to Ewald-like problems, so I would not be surprised if there is a relation.

antoine-levitt:
This is good to merge from my point of view btw, I won't do any further changes to DFTK itself for this project.

antoine-levitt:
(probably)

examples/anyons.jl
```diff
@@ -0,0 +1,33 @@
...
+n_electrons = 1
+β = 5
+
+terms = [Kinetic(2),
+         ExternalFromReal(X -> pot(X...)),
+         Anyonic(1, β)
+]
+model = Model(lattice; n_electrons=n_electrons,
+              terms=terms, spin_polarization=:spinless)  # "spinless electrons"
+basis = PlaneWaveBasis(model, Ecut, kgrid=(1, 1, 1); fft_size=DFTK.compute_fft_size(lattice, Ecut, supersampling=1.2))
```
mfherbst, line 27: This line is too long. Just assign `fft_size` first.

examples/anyons.jl
```diff
@@ -0,0 +1,33 @@
...
+scfres = direct_minimization(basis, tol=1e-14)  # Reduce tol for production
+E = scfres.energies.total
+s = 2
+E11 = π/2 * (2(s+1)/s)^((s+2)/s) * (s/(s+2))^(2(s+1)/s) * E^((s+2)/s) / β
```
mfherbst, line 31: This is super hard to read. Some more spaces could really help here.

examples/anyons.jl
```diff
@@ -0,0 +1,33 @@
...
+println("e(1,1)/(2π)= ", E11/(2π))
```
mfherbst, line 32: Again ... spaces around operators.

src/eigen/preconditioners.jl
```diff
@@ -33,7 +33,8 @@ mutable struct PreconditionerTPA{T <: Real}
 end
 
 function PreconditionerTPA(basis::PlaneWaveBasis{T}, kpt::Kpoint; default_shift=1) where T
-    kin = Vector{T}([sum(abs2, G + kpt.coordinate_cart)
+    scaling = only([t for t in basis.model.term_types if t isa Kinetic]).scaling_factor
+    kin = Vector{T}([scaling * sum(abs2, G + kpt.coordinate_cart)
```
mfherbst, line 37: If you are doing that you might as well re-use the `kinetic_energies` of the `TermKinetic`.

src/terms/anyonic.jl
```diff
@@ -2,14 +2,60 @@
 # This term does not contain the kinetic energy, which must be added separately
 # /!\ They have no 1/2 factor in front of the kinetic energy,
 #     so for consistency the added kinetic energy must have a scaling of 2
-# Energy = <u, ((-i∇ + βA)^2 + V) u>
+# Energy = <u, ((-i hbar ∇ + βA)^2 + V) u>
```
mfherbst, line 5: Use `ħ`?

src/terms/anyonic.jl
```diff
@@ -2,14 +2,60 @@
...
+# for numerical reasons, we solve the equation ∇∧A = 2π ρ in two parts: A = A_SR + A_LR, with
+# ∇∧A_SR = 2π(ρ-ρref)
+# ∇∧A_LR = 2πρref
+# where ρref is a gaussian centered at the origin and with the same integral as ρ (=M)
+# ρref(x) = M exp(-1/2 (x/σ)^2) / (σ^3 (2π)^3/2)
+# This way, ρ-ρref has zero mass, and A_SR is shorter range. The long-range part is computed explicitly
+
+function ρref_real(x::T, y::T, M, σ) where {T <: Real}
+    r = hypot(x, y)
+    M * exp(-1/2 * (r/σ)^2) / (σ^2 * (2T(π))^(2/2))
```
mfherbst, line 21: The `2/2` is a bit strange. I think I would assign `dim = 2` and than use that variable to make that clear. Also be careful with type stability here `1/2` will always be Float64.

src/terms/anyonic.jl
```diff
@@ -2,14 +2,60 @@
...
+    # 1/x^2 (c1 - C exp(-α x^2)/2α), which coupled with smoothness at 0 gives
+    # C/(2α)/x^2*(1 - exp(-α x^2))
+    r = hypot(x, y)
+    r == 0 && return magnetic_field_produced_by_ρref(1e-8, 0.0, M, σ) # hack
+
+    α = 1/(2*σ^2)
+    C = 2T(π)*M/(σ^2 * (2T(π))^(2/2))
+    r = hypot(x, y)
+    ϕ = 1/2*C/α/r^2*(1-exp(-α*r^2))
+    ϕ * @SVector[-y, x]
```
mfherbst, line 39: `dim` variable and better spacing.

src/terms/anyonic.jl
```diff
@@ -2,14 +2,60 @@
...
+end
+
+function make_div_free(basis::PlaneWaveBasis{T}, A) where {T}
+    out = [zeros(complex(T), basis.fft_size), zeros(complex(T), basis.fft_size)]
+    A_fourier = [from_real(basis, A[α]).fourier for α = 1:2]
+    for (iG, G) in enumerate(G_vectors(basis))
+        vec = [A_fourier[1][iG], A_fourier[2][iG]]
+        G = [G[1], G[2]]
+        if iG != 1
+            out[1][iG], out[2][iG] = vec - (G'vec)*G/sum(abs2, G)
```
mfherbst, line 49: Better spacing.

src/terms/anyonic.jl
```diff
@@ -2,14 +2,60 @@
...
+    r == 0 && return magnetic_field_produced_by_ρref(1e-8, 0.0, M, σ) # hack
+
+    α = 1/(2*σ^2)
+    C = 2T(π)*M/(σ^2 * (2T(π))^(2/2))
+    r = hypot(x, y)
+    ϕ = 1/2*C/α/r^2*(1-exp(-α*r^2))
+    ϕ * @SVector[-y, x]
+end
+
+function make_div_free(basis::PlaneWaveBasis{T}, A) where {T}
```
mfherbst, line 42: A comment for what this function does: Perhaps just the formula written out more nicely.

src/terms/anyonic.jl
```diff
@@ -42,26 +107,28 @@ function ene_ops(term::TermAnyonic, ψ, occ; ρ, kwargs...)
...
     A1 = zeros(complex(T), basis.fft_size)
     A2 = zeros(complex(T), basis.fft_size)
+    ρ_fourier = ρ.fourier  # unpack before hot loop for perf
+    ρref_fourier = term.ρref.fourier
     for (iG, G) in enumerate(G_vectors_cart(basis))
         G2 = sum(abs2, G)
         if G2 != 0
-            A1[iG] = +2T(π) * G[2] / G2 * ρ.fourier[iG] * im
-            A2[iG] = -2T(π) * G[1] / G2 * ρ.fourier[iG] * im
+            A1[iG] = +2T(π) * G[2] / G2 * (ρ_fourier[iG]-ρref_fourier[iG]) * im
```
mfherbst, line 115: Space around "-"

src/terms/anyonic.jl
```diff
@@ -2,14 +2,60 @@
...
 # where only the first three terms "count for the energy", and where
-# J = 1/(2i) (u* ∇u - u ∇u*)
+# J = 1/(2i) hbar (u* ∇u - u ∇u*)
+
+# for numerical reasons, we solve the equation ∇∧A = 2π ρ in two parts: A = A_SR + A_LR, with
+# ∇∧A_SR = 2π(ρ-ρref)
+# ∇∧A_LR = 2πρref
+# where ρref is a gaussian centered at the origin and with the same integral as ρ (=M)
+# ρref(x) = M exp(-1/2 (x/σ)^2) / (σ^3 (2π)^3/2)
+# This way, ρ-ρref has zero mass, and A_SR is shorter range. The long-range part is computed explicitly
```
mfherbst, line 17: Line too long.

src/terms/anyonic.jl
```diff
@@ -22,17 +68,36 @@ function (A::Anyonic)(basis)
...
+function TermAnyonic(basis::PlaneWaveBasis{T}, hbar, β) where {T}
+    # compute correction magnetic field
+    ρref = zeros(T, basis.fft_size)
+    Aref = [zeros(T, basis.fft_size), zeros(T, basis.fft_size)]
+    M = basis.model.n_electrons  # put M to zero to disable
+    σ = 2
+    for (ir, r) in enumerate(r_vectors(basis))
+        rcart = basis.model.lattice * (r - @SVector[.5, .5, .0])
+        ρref[ir] = ρref_real(rcart[1], rcart[2], M, σ)
+        Aref[1][ir], Aref[2][ir] = magnetic_field_produced_by_ρref(rcart[1], rcart[2], M, σ)
```
mfherbst, line 90: Line too long.

test/anyons.jl
```diff
@@ -0,0 +1,18 @@
...
+
+@testset "Anyons" begin
+    # Test that the magnetic field satisfies ∇∧A = 2π ρref, ∇⋅A = 0
+    x = 1.23
+    y = -1.8
+    ε = 1e-8
+    M = 2.31
+    σ = 1.81
+    dAdx = (DFTK.magnetic_field_produced_by_ρref(x+ε, y, M, σ) - DFTK.magnetic_field_produced_by_ρref(x, y, M, σ))/ε
+    dAdy = (DFTK.magnetic_field_produced_by_ρref(x, y+ε, M, σ) - DFTK.magnetic_field_produced_by_ρref(x, y, M, σ))/ε
```
mfherbst, line 13: Lines too long.

test/hamiltonian_consistency.jl
```diff
@@ -70,5 +70,5 @@ let
     Apot(x, y, z) = .2 * [y - a/2, -(x - a/2), 0]
     Apot(X) = Apot(X...)
     test_consistency_term(Magnetic(Apot);  kgrid=[1, 1, 1], lattice=[a 0 0; 0 a 0; 0 0 0], Ecut=20)
-    test_consistency_term(DFTK.Anyonic(1); kgrid=[1, 1, 1], lattice=[a 0 0; 0 a 0; 0 0 0], Ecut=20)
+    test_consistency_term(DFTK.Anyonic(2, 3.2); kgrid=[1, 1, 1], lattice=[a 0 0; 0 a 0; 0 0 0], Ecut=20)
```
mfherbst, line 73: Line breaks.

examples/anyons.jl
antoine-levitt: It's an ugly formula, spaces won't make it less ugly (in contrary, in my opinion...)

src/eigen/preconditioners.jl
antoine-levitt: yes but we don't know the index of kpt in basis. That's a bit of a problem with our API for the kpoint datastructure. Hard to fix it without a backref from kpoint to the basis, which you didn't want

src/terms/anyonic.jl
antoine-levitt: too easy to confuse with h

examples/anyons.jl
mfherbst: I generally disagree. Spacing and subexpressions generally help. Just try a little to come up with something better.

src/terms/anyonic.jl
antoine-levitt: it's just the formula for a normalized gaussian, assigning dim = 2 would just be a waste of space I think...

src/eigen/preconditioners.jl
mfherbst: Well you could look that up ... but arguably that's no cleaner.

src/terms/anyonic.jl
mfherbst: Come on it's one line!

src/terms/anyonic.jl
antoine-levitt: The name is self-explanatory, no? I could write a comment, but the comment is going to be "make a given vector field divergence-free", which does not actually add more information. I did add a comment on why this was necessary

mfherbst: Just write the formula in nice physics instead of G's and ffts.

src/terms/anyonic.jl
antoine-levitt: I generally agree with your line breaks when it helps readability by breaking up complex expressions, but here there's no natural way to break

test/anyons.jl
antoine-levitt: Frankly I'm not going to bother with style like that in tests

src/terms/anyonic.jl
mfherbst:
```
res = magnetic_field_produced_by_ρref(rcart[1], rcart[2], M, σ)
Aref[1][ir], Aref[2][ir] = res
```

test/anyons.jl
mfherbst: Frankly I do. It's a classical broken windows thing. Test code is code and should be as readable as the main code. Either we care about formatting or we don't. If we do we also care in tests.

examples/anyons.jl
antoine-levitt: How would it help here? I'm copying the formula from https://arxiv.org/pdf/1901.10739.pdf, I think it's the nicest way to render this (particularly ugly and not meant to be read) formula

src/eigen/preconditioners.jl
antoine-levitt: Yeah, I'm wondering now why we don't have a field in kpoint with the basis and the index of this kpoint in the basis?

examples/anyons.jl
mfherbst:
```suggestion
ratio = (s+1) / s
rplus = (s+2) / s
E11 = π/2 * (2ratio)^rplus * rplus^(2ratio) * E^rplus / β
```

mfherbst: I think that's much clearer.

src/eigen/preconditioners.jl
mfherbst: Because circular pointers are bad. They kick you in the back eventually. I'm fine with storing the index, though.

src/terms/anyonic.jl
antoine-levitt: you can't seriously think this is more readable than the previous version?

test/anyons.jl
antoine-levitt: well about formatting you know my answer ;-) I agree that we should treat tests seriously, but I'm already having a very hard time caring about long lines in main code, so in tests...

antoine-levitt: also for this one line breaking is pretty bad... I'll do it for the other one

examples/anyons.jl
antoine-levitt: No way. These ratio and rplus names and the formula are actively misleading : they suggest meaning that just isn't there. It's just a stupid formula coming from a stupid integral. Again you're not really supposed to actually read this formula, it's just for checking that the code works.

src/terms/anyonic.jl
mfherbst: I think it is, but fair enough.

antoine-levitt: I'm trying, but I just can't understand your objection to long lines. I've seen your terminal, you have enough space to display them, and so does everybody I've seen. I do agree that long lines are very often a sign of bad code, like long functions and a large number of indentation levels, but it's not bad *per se*, is it? Like here, it's just because the function name is very long

mfherbst: The part where it is bad per se is that it makes it harder to read independent from the fact that it is code.

Basically it's the same reason why papers and books are typically printed in two columns if they are on A4. When lines get too long the eye can't follow and you have to read the same thing more often or you miss part of the info. The recommended number of characters on a line is something between 70 and 90. (Arguably this is 92, so it's actually ok, but we should really aim below 90 chars to make sure that it does not become 150 in the end.)

test/anyons.jl
mfherbst: Talking about line breaks ... this is one of the cases where I really would insist. It's easy to miss the ε because the line is so long. Nicer is:
```suggestion
    dAdx = (  DFTK.magnetic_field_produced_by_ρref(x + ε, y, M, σ)
            - DFTK.magnetic_field_produced_by_ρref(x,     y, M, σ)) / ε
    dAdy = (  DFTK.magnetic_field_produced_by_ρref(x, y + ε, M, σ)
            - DFTK.magnetic_field_produced_by_ρref(x, y,     M, σ)) / ε
```
which also makes the structure of what you are doing a lot clearer.

src/terms/anyonic.jl
mfherbst: Also my experience is that accepting long lines sometimes stops one too early from trying to make code formatted in a nicer way.

examples/anyons.jl
mfherbst: ok. I tried.

src/terms/anyonic.jl
antoine-levitt: This is hugely different for unstructured text (where the unit is the word or sentence and lines have no specific meaning) and for imperative code (where the unit is the line of code). In the case of code, it is much more important to preserve the integrity of the line of code (like in the example above where breaking the line breaks the natural flow of the assignment). What you're saying could plausibly be true for more functional languages which are less linear in their form, but not for us.

antoine-levitt: Yeah I agree with the fact that long lines are a sign to be suspicious, but it doesn't mean it's unconditionally bad

test/anyons.jl
antoine-levitt: I don't really agree, but it's test so I'm less willing to fight, committed your suggestion (thanks to github's new API)

src/terms/anyonic.jl
mfherbst: If the get too long they are unconditionally bad. Probably not in this example though, that I can agree upon.

PR #378, opened 2020-11-19
Use JLD2 custom serialisation

mfherbst:
Not exactly simpler, but at least more flexible.

### DIFF
--- a/src/jld2io.jl
+++ b/src/jld2io.jl
@@ -22,47 +22,6 @@ function ScfSaveCheckpoints(filename="dftk_scf_checkpoint.jld2"; keep=false, ove
     end
 end
 
-
-struct PlaneWaveBasisSerialisation{T <: Real}
-    model
-    Ecut
-    kcoords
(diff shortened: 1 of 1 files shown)


antoine-levitt:
Yeah I think I prefered the old version :-p
So, I know you're not going to like it, but assume we allow ourselves piracy, and overload JLD2 serialization for FFTW. Then that simplifies things quite a bit, no?

mfherbst:
Well for FFTW and Libxc for starters and then It implies storing quite some stuff that is big and can be computed (e.g. the potential arrays for each term). Since JLD2 files tend to be quite big for the calculations we are doing already (and storage on some of the clusters we work on is running low) I don't think this is a wise choice.

One advantage of the new version is that we can also store basis objects to a JLD2 file independent of load_scfres / save_scfres, which actually could be useful.

antoine-levitt:
OK yeah I agree, let's go with this then!

PR #385, opened 2020-11-27
Store kgrid and kshift if available

mfherbst:
@LaurentVidal95 who requested this.

### DIFF
--- a/src/PlaneWaveBasis.jl
+++ b/src/PlaneWaveBasis.jl
@@ -62,7 +62,7 @@ struct PlaneWaveBasis{T <: Real}
     # ksymops[ik] is a list of symmetry operations (S,τ)
     # mapping to points in the reducible BZ
     ksymops::Vector{Vector{SymOp}}
-    # Monkhorst-Pack grid used to generate the k-Points, nothing if unknown
+    # Monkhorst-Pack grid used to generate the k-Points or nothing for custom k-Points
     kgrid::Union{Nothing,Vec3{Int}}
     kshift::Union{Nothing,Vec3{T}}
 
--- a/src/PlaneWaveBasis.jl
+++ b/src/PlaneWaveBasis.jl
@@ -62,7 +62,7 @@ struct PlaneWaveBasis{T <: Real}
     # ksymops[ik] is a list of symmetry operations (S,τ)
     # mapping to points in the reducible BZ
     ksymops::Vector{Vector{SymOp}}
-    # Monkhorst-Pack grid used to generate the k-Points or nothing for custom k-Points
+    # Monkhorst-Pack grid used to generate the k-Points, or nothing for custom k-Points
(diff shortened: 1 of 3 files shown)


src/PlaneWaveBasis.jl
```diff
@@ -62,6 +62,9 @@ struct PlaneWaveBasis{T <: Real}
     # ksymops[ik] is a list of symmetry operations (S,τ)
     # mapping to points in the reducible BZ
     ksymops::Vector{Vector{SymOp}}
+    # Monkhorst-Pack grid used to generate the k-Points, nothing if unknown
```
antoine-levitt, line 65: not really unknown, more "custom", non-MP set of kpoints?

mfherbst: yes, that's a good point.

src/PlaneWaveBasis.jl
```diff
@@ -138,7 +141,8 @@ build_kpoints(basis::PlaneWaveBasis, kcoords) =
 @timing function PlaneWaveBasis(model::Model{T}, Ecut::Number,
                                 kcoords::AbstractVector, ksymops, symmetries=nothing;
                                 fft_size=nothing, variational=true,
-                                optimize_fft_size=false, supersampling=2) where {T <: Real}
+                                optimize_fft_size=false, supersampling=2,
+                                kgrid=nothing, kshift=nothing) where {T <: Real}
```
antoine-levitt, line 145: this is pretty confusing that this constructor has both kgrid, kshift and kcoords, ksymops. But we probably need to fix this constructor someday ayway...

src/PlaneWaveBasis.jl
```diff
@@ -62,6 +62,9 @@ struct PlaneWaveBasis{T <: Real}
     # ksymops[ik] is a list of symmetry operations (S,τ)
     # mapping to points in the reducible BZ
     ksymops::Vector{Vector{SymOp}}
+    # Monkhorst-Pack grid used to generate the k-Points or nothing for custom k-Points
```
antoine-levitt, line 65: missing comma (makes it look like "nothing" is the complement of "generate")

src/PlaneWaveBasis.jl
mfherbst: Yes. I feel for the "K-Point grid and symmetry" we might actually need to introduce a datastructure. It's becoming out of hand all these k-arguments. And the meaning of the order of the k-Points inside and outside PlaneWaveBasis has diverged anyways since the MPI and spin stuff has been implemented.

antoine-levitt: Yep, we have an issue for that: https://github.com/JuliaMolSim/DFTK.jl/issues/235

mfherbst: Right I remember now.

PR #386, opened 2020-11-27
Make pwbasis constructor less of a mess (hopefully)

antoine-levitt:
Main thing is putting our mess in fft.jl :-)


src/PlaneWaveBasis.jl
```diff
@@ -143,63 +143,16 @@ build_kpoints(basis::PlaneWaveBasis, kcoords) =
...
-    # ψ(r) = sum_G c_G e^iGr / sqrt(Ω)
-    # so that the ifft is normalized by 1/sqrt(Ω). It follows that the
-    # fft must be normalized by sqrt(Ω) / length
-    ipFFT = ipFFT_unnormalized * (sqrt(model.unit_cell_volume) / length(ipFFT_unnormalized))
-    opFFT = opFFT_unnormalized * (sqrt(model.unit_cell_volume) / length(opFFT_unnormalized))
-    ipIFFT = inv(ipFFT)
-    opIFFT = inv(opFFT)
 
-    # Compute kpoint information
+    #### Compute kpoint information
```
mfherbst, line 148: I'm not a fan of this multiple `####` thing. It's only opening the room for all sorts of messes (Is it 3 or 4 `#`, how about 2 is that a more important comment than `#`. I think with the changes you made there is no real need for this. And either a function is short enough that the structure is clear or you ought to split it up.

src/PlaneWaveBasis.jl
```diff
@@ -143,63 +143,16 @@ build_kpoints(basis::PlaneWaveBasis, kcoords) =
...
     @assert length(kcoords) == length(ksymops)
     if symmetries === nothing
         # TODO instead compute the group generated by with ksymops, or
         # just retire this constructor. Not critical because this
         # should not be used in this context anyway...
         symmetries = vcat(ksymops...)
     end
-
```
mfherbst, line 202: Why remove these lines? I think the help to structure what is happening.

src/PlaneWaveBasis.jl
```diff
@@ -217,15 +170,30 @@ build_kpoints(basis::PlaneWaveBasis, kcoords) =
         @assert mpi_sum(length(krange_thisproc), mpi_comm) == n_kpt
         @assert !isempty(krange_thisproc)
     end
-
```
mfherbst, line 220: I also would not remove this

src/PlaneWaveBasis.jl
```diff
@@ -217,15 +170,30 @@ build_kpoints(basis::PlaneWaveBasis, kcoords) =
...
 
-    # Build kpoint-specific basis. Notice that this also builds index mapping from the k-point-specific basis
-    # to the global basis and thus the fft_size needs to be final at this point.
+    #### Setup fft_size and plans
+    fft_size = compute_final_fft_size(model, fft_size, Ecut, supersampling,
+                                      variational, optimize_fft_size, kcoords)
+    fft_size = mpi_max(fft_size, mpi_comm)
+    (ipFFT_unnormalized,  opFFT_unnormalized,
+     ipBFFT_unnormalized, opBFFT_unnormalized) = build_fft_plans(T, fft_size)
+    # The FFT interface specifies that fft has no normalization, and
```
mfherbst, line 182: And here I actually would add an empty line.

src/PlaneWaveBasis.jl
```diff
@@ -217,15 +170,30 @@ build_kpoints(basis::PlaneWaveBasis, kcoords) =
         @assert mpi_sum(length(krange_thisproc), mpi_comm) == n_kpt
         @assert !isempty(krange_thisproc)
     end
-
     kcoords = kcoords[krange_thisproc]
     ksymops = ksymops[krange_thisproc]
 
-    # Build kpoint-specific basis. Notice that this also builds index mapping from the k-point-specific basis
-    # to the global basis and thus the fft_size needs to be final at this point.
+    #### Setup fft_size and plans
```
mfherbst, line 176: Multiple hashes.

src/fft.jl
```diff
@@ -100,6 +99,51 @@ function compute_fft_size_precise(lattice::AbstractMatrix{T}, Ecut, kpoints;
...
+            "For a variational method, Ecut should be less than the maximal kinetic " *
+            "energy the grid supports ($max_E)"
+        )
+    else
+        # ensure fft_size is provided, and other options are not set
+        # TODO make proper error messages when the interface gets a bit cleaned up
+        @assert fft_size !== nothing
+        @assert supersampling == 2
+        @assert !optimize_fft_size
+    end
```
mfherbst, line 137: Empty line afterwards.

src/fft.jl
```diff
@@ -100,6 +99,51 @@ function compute_fft_size_precise(lattice::AbstractMatrix{T}, Ecut, kpoints;
...
 
+# Main entry point from pwbasis. Uses the above functions to find out
+# the correct fft_size according to user specification
+function compute_final_fft_size(model::Model{T}, fft_size, Ecut, supersampling,
+                                variational, optimize_fft_size, kcoords) where {T}
+    if variational
+        @assert Ecut > 0
+        if fft_size === nothing
+            fft_size = compute_fft_size(model, Ecut;
+                                        supersampling=supersampling)
```
mfherbst, line 110: If you want we can drop the supersampling parameter in this function (and in the PlaneWaveBasis constructors). I have not really been using it in the wild and always called `compute_fft_size` explicitly. I think this is fine, because it does represent an advanced use case.

src/PlaneWaveBasis.jl
antoine-levitt: I think I agree. I put them before the refactor but now there's less need

antoine-levitt:
The more I read this the less I understand it. I need to think it over more.

mfherbst:
Sure no rush. Then I make a release without it.

src/fft.jl
antoine-levitt: I think let's keep it for now

antoine-levitt:
This one makes the logic slightly more clear. This is still a bit messy, but good enough for now

src/fft.jl
```diff
@@ -121,20 +119,22 @@ function compute_final_fft_size(model::Model{T}, fft_size, Ecut, supersampling,
...
         max_E = sum(abs2, model.recip_lattice * floor.(Int, Vec3(fft_size) ./ 2)) / 2
         Ecut > max_E && @warn(
             "For a variational method, Ecut should be less than the maximal kinetic " *
             "energy the grid supports ($max_E)"
         )
     else
-        # ensure fft_size is provided, and other options are not set
-        # TODO make proper error messages when the interface gets a bit cleaned up
+        # ensure no other options are set
         @assert fft_size !== nothing
```
mfherbst, line 133: This assert can never be false and could actually be misleading with the comment line you put above it.

src/fft.jl
```diff
@@ -101,15 +101,13 @@ end
...
-                                        supersampling=supersampling)
-        end
-
+function validate_or_compute_fft_size(model::Model{T}, fft_size, Ecut, supersampling,
+                                      variational, optimize_fft_size, kcoords) where {T}
+    # compute if not provided
+    if fft_size === nothing
+        @assert variational
+        fft_size = compute_fft_size(model, Ecut;
+                                    supersampling=supersampling)
```
mfherbst, line 110: Might as well skip the line break here.

src/fft.jl
```diff
@@ -100,6 +99,51 @@ function compute_fft_size_precise(lattice::AbstractMatrix{T}, Ecut, kpoints;
...
+        if optimize_fft_size
+            # We build a temporary set of kpoints here
+            # This gymnastics is because build_kpoints builds index
+            # mapping from the k-point-specific basis to the global
+            # basis and thus the fft_size needs to be final at kpoint
+            # construction time
+            fft_size = Tuple{Int, Int, Int}(fft_size)
+            kpoints_temp = build_kpoints(model, fft_size, kcoords, Ecut; variational=variational)
```
mfherbst, line 118: This line is a bit long.

src/fft.jl
```diff
@@ -100,6 +99,51 @@ function compute_fft_size_precise(lattice::AbstractMatrix{T}, Ecut, kpoints;
...
+    end
+
+    # TODO generic FFT is kind of broken for some fft sizes
+    #      ... temporary workaround, see more details in fft_generic.jl
+    fft_size = next_working_fft_size.(T, fft_size)
+    fft_size = Tuple{Int, Int, Int}(fft_size)
+    fft_size
+end
+
+
```
mfherbst, line 145: Three empty lines is a bit too many I'd say.

mfherbst (commented):
Looks much better.

src/fft.jl
antoine-levitt: Indeed, good catch

antoine-levitt:
Should be good now (I actually did the review a few days back but pushed to the wrong branch)

antoine-levitt:
CI seems out because of a python thing?

mfherbst:
Ok that error in the docs generation keeps me puzzled ... can't be bothered to investigate that now, sorry ...

mfherbst:
Ok seems like the bug's not on us. I'm going to merge this nevertheless for now.

PR #387, opened 2020-11-27
Enable codecov and remove Travis completely

mfherbst:
A bit annoying since coveralls upload is still broken, but at least we have codecov instead.


antoine-levitt:
Cool to have this to check if we want but can we remove the spam message that pops up on every PR?

mfherbst:
100% agree.

Issue #147, opened 2020-03-11
Support VTK output


mfherbst:
Use [WriteVTK](https://github.com/jipolanco/WriteVTK.jl) to dump VTK files with densities, structures etc.

Dsantra92:
Can I try to solve this ?

mfherbst:
Yes sure. Go ahead!

PR #392, opened 2020-12-17
Write SCF results to VTK files

Dsantra92:
Closes Issue : #147

Functions added : 
- EnergyToVTK
- DensityToVtk
- WavesToVTK
- OccupationToVTK
- EigenValuesToVTK


src/external/VTK.jl
```diff
@@ -0,0 +1,79 @@
...
+Stores the Energy structure in a VTK image file and returns the name of the file.
+"""
+function EnergyToVTK(VTKFileName::String, energy::Energies{<:Number})
+
+    vtkfile = vtk_grid(VTKFileName, 1, 1)
+    for i in keys(energy)
+        vtkfile[i] = energy.energies[i]
+    end
+    out = vtk_save(vtkfile)
+    return out[1]
```
mfherbst, line 16: We don't use return if it's the last statement of a function.

src/external/VTK.jl
```diff
@@ -0,0 +1,79 @@
...
+
+Stores the energy density in a VTK image file and returns the name of the file.
+The density is stored in 3 submatrices : Real, Real components of Fourier, Imaginary
+components of Fourier.
+"""
+function DensityToVtk(VTKFileName::String, ρ::RealFourierArray{T, <: AbstractArray{T, 3}, 
+                                             <: AbstractArray{Complex{T}, 3}}) where T<:Real
+
+    out = vtk_write_array(VTKFileName,(ρ.real, Base.getproperty.(ρ.fourier, :re),
+                          Base.getproperty.(ρ.fourier, :im)),
```
mfherbst, line 30: `getproperty` is a bit magic for these sort of jobs. Better use `imag.(ρ.fourier)` and `real(...)`, that's much clearer.

src/external/VTK.jl
```diff
@@ -0,0 +1,79 @@
...
+    DensityToVtk(VTKFileName::String, ρ)
+
+Stores the energy density in a VTK image file and returns the name of the file.
+The density is stored in 3 submatrices : Real, Real components of Fourier, Imaginary
+components of Fourier.
+"""
+function DensityToVtk(VTKFileName::String, ρ::RealFourierArray{T, <: AbstractArray{T, 3}, 
+                                             <: AbstractArray{Complex{T}, 3}}) where T<:Real
+
+    out = vtk_write_array(VTKFileName,(ρ.real, Base.getproperty.(ρ.fourier, :re),
```
mfherbst, line 29: In many calculations there is also a non-zero `ρspin` that should be dumped to VTK.

src/external/VTK.jl
```diff
@@ -0,0 +1,79 @@
...
+
+Stores the wave values in a VTK image file and returns the name of the file.
+If the sizes of wave vectors for each each value is not equal, the wave matrix
+will have size = (e, w, n) where e = No of eigen values, w = max no of values for
+each eigen value, n =  no of Kpoints). The matrix is stored as two submatrix,
+one containing the real part and the other containing the imaginary part.
+"""
+function WavesToVTK(VTKFileName::String, wave)
+
+    mat_size = (maximum(x->size(x)[1], wave), size(wave[1])[2], length(wave))
```
mfherbst, line 57: When you are doing the combined `save_scfres` there is no need for this juggeling, because you will have the `basis` available that contains these lengths in a more convenient fashion :smile:.

src/external/VTK.jl
```diff
@@ -0,0 +1,79 @@
...
+    wave_mat = zeros(ComplexF64, mat_size)
+    @views for i = 1:length(wave)
```
mfherbst, line 59: Since this is doing IO (which is slow) I would not bother with `@views`.

mfherbst (commented):
Thanks for giving this a go @Dsantra92 and for contributing to DFTK! I left a few specific comment inline. And here is my more general feedback:

On a more general sight, I'm a bit hesitant of this approach for these reasons:
- If I get it right the functions you have coded each produce a separate VTK file for each of the quantities. The main scenario I have in mind for this feature is, however, that I do a calculation and I want to quickly dump the resulting density / wavefunction etc to a VTK grid for post-processing and analysis in paraview. For this purpose it is much more useful in my opinion to have all the data in a single VTK file. This should be possible using `vtk_grid`, e.g.
  ```julia
   xyz = zeros(3, gridsize...)
   for (idcs, r) in zip(CartesianIndices(gridsize), DFTK.r_vectors_cart(basis))
       xyz[:, Tuple(idcs)...] = r
   end
   vtkfile = vtk_grid(filename, xyz)
  ```
  and then adding densities and wavefunctions on this structured grid
 and energies and occupations as metadata.
- Many of the internal data structures of DFTK (e.g. the scfres) are not yet fully settled, meaning that if one builds these individual functions, it might end up being a lot of effort to keep supporting them as the datastructures change. While I agree that this is the best long-term solution for this feature, I think for the moment the more pragmatic approach of having a single top-level function `save_scfres(vtkfile, scfres)` similar to what is coded in `src/jld2io.jl`, which is user-facing and orchestrates the save (setting up the grid, dumping the bloch wave and the densities and adding the metadata) is the best solution. For that purpose I would also generalise https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/jld2io.jl#L38 such that it either generates a VTK file or a JLD2 file depending on the file extension.

Dsantra92:
Thansk @mfherbst  for taking time to review this pr. I will be be adding the requested changes in a day or two's time.

src/external/VTK.jl
Dsantra92: Yes I have noticed that convention in the package. Julia does automatically considers the last line as return value, but adding a return value makes the code more readable.

src/DFTK.jl
```diff
@@ -149,6 +149,12 @@ include("external/load_from_python.jl")
 include("external/load_from_file.jl")
 include("external/ase.jl")
 include("external/pymatgen.jl")
+include("external/VTK.jl")
+export EnergyToVTK
+export DensityToVtk
+export WavesToVTK
+export OccupationToVTK
+export EigenValuesToVTK
```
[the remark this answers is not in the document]
Dsantra92: Not adding hard dependency is reasonable. I will be using Requires in this case.

src/external/VTK.jl
```diff
@@ -0,0 +1,79 @@
+# Uses WriteVTK.jl to convert various structures and matrices to VTk file format
+using WriteVTK
+
+"""
+    EnergyToVTK(VTKFileName::String, energy::Energies{::Number})
+
+Stores the Energy structure in a VTK image file and returns the name of the file.
+"""
+function EnergyToVTK(VTKFileName::String, energy::Energies{<:Number})
```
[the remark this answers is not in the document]
Dsantra92: I noticed some code as in [this function](https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/external/etsf_nanoquanta.jl#L19) which were in Camel Case.
Apart from changing the naming convention to Pascal type, you wanted some changes in the function parameter types. I am not sure about how much generalization would be good here. You mentioned DFTK structures are meant to changed in the future so annotating types like `AbstractMatrix` for matrix (like densities) will be okay? or, should I not annotate types for non-structures?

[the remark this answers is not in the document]
Dsantra92: Or maybe if we are putting all these in a single VTK(as you mentioned) using single function we won't need such type specifications anymore.

src/external/VTK.jl
mfherbst: I see that point, but IMHO that's debatable, especially in more functional languages like Julia, that often feature short functions. We decided to skip the returns, much inline with the Julia standard library.

src/external/VTK.jl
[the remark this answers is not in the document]
mfherbst: Yes the exception to the snail-case rule in DFTK (much as in base Julia) are constructor-like functions, such as the one you are pointing out.

As a general rule and guideline, I would not put types at first unless you need it to dispatch. In a later pass (when the structure of the code is more clear and thus so are the public interfaces), then I tend to add the types. So in line with that I would follow your second comment and not worry that much about the types for the redesign.

mfherbst:
> I will be be adding the requested changes in a day or two's time

No rush. Take your time, we're happy for your help in either case!

src/external/VTK.jl
Dsantra92: Cool, I will go with no return statements.

src/external/VTK.jl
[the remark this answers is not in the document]
Dsantra92: I am almost done with new set of changes, I just have a question. If I am not wrong, the  wave function will be stored in real space rather than the Fourier space. The waves in real space have imaginary components, (You mentioned that imaginary parts are more difficult to interpret visually) I am finding it difficult to decide whether to store the modulus of the number or the real part(trimming out the imaginary part).

mfherbst:
Hi @Dsantra92. Sorry for my delay in getting back to you. I'm currently on vacations. I'll review your newest contributions next week!

src/external/VTK.jl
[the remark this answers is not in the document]
mfherbst: Sorry that I ignored your question for so long. Yes, the way you are doing it now the wave function is stored in real space. More precisely only the real part of the real-space version of the wave function is stored (due to your `real.( ...  )`), which is an issue as the imaginary part also contains useful information (the wave function is complex-valued).

Be a little careful with the terminology here: Unfortunately "real" has two different meanings in this context. There is the real part of the complex-valued wave function and then there is the real-space and the fourier-space (or reciprocal-space) representation of said wave function. Both of which are again complex-valued. So in fact both in real space and in reciprocal space you have a real and an imaginary part and both are needed.

src/external/VTK.jl
```diff
@@ -0,0 +1,49 @@
...
+
+The function takes in the VTK filename and the scfres structure and stores into a VTK file.
+Grid Values:
+ρ_real -> real values of scfers.ρ
+ψreal_i_j -> Real values of Bloch waves in real space, where i and j are Kpoint and EigenVector indexes respectively
+ρspin -> Real of ρspin are stored if ρspin in present
+MetaData:
+Energy and occupation.
+"""
+function save_scfres(VTKFileName::AbstractString, scfres)
```
mfherbst, line 15: Be careful here: This function has the identical signature to https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/jld2io.jl#L38, so this will cause trouble. I think what I would do is make this function as well as https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/jld2io.jl#L25 a three-argument function, where the third argument (`format`) is a `Val`-type: [Val](https://docs.julialang.org/en/v1/base/base/#Base.Val) (e.g. `Val(:jld2)` and `Val(:vtk)`)). Then you can make a two-argument version of the function which can then use the file extension to either call into https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/jld2io.jl#L25 or into the function you code here. Of course that generic `save_scfres(jld::JLD2.JLDFile, scfres::NamedTuple)` needs to be always available to DFTK, so needs to be in a file that is always loaded (not with Requires). With a generic three-argument version of `save_scfres` you can catch cases for formats that are not implemented and give users a graceful error message.

src/external/VTK.jl
```diff
@@ -0,0 +1,49 @@
...
+    # Initialzing the VTK Grid
+    basis = scfres.basis
+    basis_r_co = r_vectors_cart(basis)
+    gridsize = size(basis_r_co)
+    grid = zeros(3, gridsize...)
```
mfherbst, line 20: Just use `basis.fft_size`.

src/external/VTK.jl
```diff
@@ -0,0 +1,49 @@
...
+ψreal_i_j -> Real values of Bloch waves in real space, where i and j are Kpoint and EigenVector indexes respectively
+ρspin -> Real of ρspin are stored if ρspin in present
+MetaData:
+Energy and occupation.
+"""
+function save_scfres(VTKFileName::AbstractString, scfres)
+    # Initialzing the VTK Grid
+    basis = scfres.basis
+    basis_r_co = r_vectors_cart(basis)
+    gridsize = size(basis_r_co)
```
mfherbst, line 19: No need to define these variables (shorter code is better code!)

src/external/VTK.jl
mfherbst, line 15: Also: We use snail case for variable names, so `vtk_file_name` (or just `filename`).

src/external/VTK.jl
```diff
@@ -0,0 +1,49 @@
...
+    grid = zeros(3, gridsize...)
+    for (idcs, r) in zip(CartesianIndices(gridsize), basis_r_co)
+        grid[:, Tuple(idcs)...] = r
+    end
+    vtkfile = vtk_grid(VTKFileName, grid)
+    
+    # Storing the Bloch Waves in Real space
+    for i in 1:length(basis.kpoints)
+        for j in 1:7
```
mfherbst, line 28: Why `7`? Don't rely on there being `7` bands :smile:.

src/external/VTK.jl
```diff
@@ -0,0 +1,49 @@
...
+            vtkfile["ψreal_$(i)_$(j)"] = real.(G_to_r(basis, basis.kpoints[i], scfres.ψ[i][:,j]))
+        end
+    end
+
+    # Storing the Real component of density
+    vtkfile["ρ_real"] = scfres.ρ.real
```
mfherbst, line 34: Just call it `ρ`. The density has no imaginary component in real space.

src/external/VTK.jl
```diff
@@ -0,0 +1,49 @@
...
+    
+    # Storing ρspin if it is present.
+    isnothing(scfres.ρspin)  || (vtkfile["ρspin_real"] = scfres.ρspin.real)
+    
+    # Storing the energy components
+    for i in keys(scfres.energies)
```
mfherbst, line 40: `i` is more for integers. Better call the loop variable `key`.

src/external/VTK.jl
```diff
@@ -0,0 +1,49 @@
...
+        vtkfile["Energy_$i"] = scfres.energies[i]
+    end
+
+    # Storing the Occupation as a matrix
+    occupation_mat = hcat(scfres.occupation...)
+    vtkfile["Occupation"] = occupation_mat
```
mfherbst, line 46: Just make this one line. Also stay with lower-case keys.

mfherbst (commented):
Thanks for rewriting the function @Dsantra92. I added some comments.

While you are on it: I think it would be nice to also store the eigenvalues, the Fermi level as metadata to the calculation.

src/external/VTK.jl
Dsantra92: I forgot to add stashed changes . My bad. 😞 . I have update the function in the recent commit.

src/external/VTK.jl
Dsantra92: Sounds good. Although the functions behave correctly when only one of `JLD2` or `WriteVTK` is called up, what you suggest will be a lot cleaner way to do things. 
P.S. - Thanks for the `Val` tip, didn't know such a dispatch hack existed.

Dsantra92: I took some liberty to use 2 argument `save_scfres` definition in `dummy_definitions.jl` and since it was no longer containing only dummy definitions I renamed it to `save_load_scfres.jl` . I have also added some `MethodError` catches to provide more meaningful error phrases to the user.

src/save_load_scfres.jl
```diff
@@ -0,0 +1,42 @@
...
+
+Save an `scfres` obtained from `self_consistent_field` to a JLD2 or VTK file depending on the extension.
+
+!!! warning "No compatibility guarantees"
+    No guarantees are made with respect to this function at this point.
+    It may change incompatibly between DFTK versions or stop working / be removed
+    in the future.
+"""
+function save_scfres(filename::AbstractString, scfres::NamedTuple)
+    fs = split(filename,".")
```
mfherbst, line 22: Space after commas.

src/external/VTK.jl
```diff
@@ -0,0 +1,58 @@
...
+
+"""
+    save_scfres(filename, scfres, Val(:vtk))
+
+The function takes in the VTK filename and the scfres structure and stores into a VTK file.
+
+Grid Values:
+
+- ρ -> Density in real space
+- ψreal\\_i\\_j -> Real values of Bloch waves in real space, where i and j are Kpoint and EigenVector indexes respectively
```
mfherbst, line 13: Line too long: We try to keep lines below 90 char (but we are not super strict about it, so a few chars more are ok.)

src/save_load_scfres.jl
mfherbst: Also there is `splitext`, which is for this purpose.

src/save_load_scfres.jl
```diff
@@ -0,0 +1,42 @@
...
+    @assert length(fs) != 1 "No file extension mentioned"
+    
+    if(fs[end] == "vts")
```
mfherbst, line 25: I would use `Val(Symbol(fs[end]))` directly to dispatch. If you then have a generic `save_scfres(filename, scfres, format)` you can use that to emit error messages.

src/external/VTK.jl
```diff
@@ -0,0 +1,58 @@
...
+    grid = zeros(3, basis.fft_size...)
+    for (idcs, r) in zip(CartesianIndices(basis.fft_size), r_vectors_cart(basis))
+        grid[:, Tuple(idcs)...] = r
+    end
+    vtkfile = vtk_grid(filename, grid)
+    
+    # Storing the Bloch Waves in Real space
+    for i in 1:length(basis.kpoints)
+        for j in 1:size(scfres.ψ[1])[2]
+            vtkfile["ψ_real_$(i)_$(j)"] = real.(G_to_r(basis, basis.kpoints[i], scfres.ψ[i][:,j]))
```
mfherbst, line 32: Can you add a `k` and `band` or something to the label. Otherwise you always have to remember which index is the band and which is the kpoint.

src/external/VTK.jl
```diff
@@ -0,0 +1,58 @@
...
+    # Storing the energy components
+    for key in keys(scfres.energies)
+        vtkfile["energy_$key"] = scfres.energies[key]
+    end
+    
+    # Storing the Fermi level 
+    vtkfile["fermi_level"] =  scfres.εF
+
+    # Storing the EigenValues as a matrix
+    vtkfile["eigen_values"] = hcat(scfres.eigenvalues...)
```
mfherbst, line 52: no underscore.

src/external/VTK.jl
```diff
@@ -0,0 +1,58 @@
...
+    # Storing ρspin if it is present.
+    isnothing(scfres.ρspin)  || (vtkfile["ρspin_real"] = scfres.ρspin.real)
+    
+    # Storing the energy components
+    for key in keys(scfres.energies)
+        vtkfile["energy_$key"] = scfres.energies[key]
+    end
+    
+    # Storing the Fermi level 
+    vtkfile["fermi_level"] =  scfres.εF
```
mfherbst, line 49: Just use `εF` ... that's completely fine.

src/external/VTK.jl
```diff
@@ -0,0 +1,58 @@
...
+            vtkfile["ψ_real_$(i)_$(j)"] = real.(G_to_r(basis, basis.kpoints[i], scfres.ψ[i][:,j]))
+            vtkfile["ψ_imag_$(i)_$(j)"] = imag.(G_to_r(basis, basis.kpoints[i], scfres.ψ[i][:,j]))
+        end
+    end
+
+    # Storing the  density in real space
+    vtkfile["ρ"] = scfres.ρ.real
+    
+    # Storing ρspin if it is present.
+    isnothing(scfres.ρspin)  || (vtkfile["ρspin_real"] = scfres.ρspin.real)
```
mfherbst, line 41: I'd also call it `ρspin`.

mfherbst (commented):
Also I think we should have some sort of test for this routine to ensure it does not randomly break. 
I'm unaware of a Julia package being able to read VTK files unfortunately, but how about just storing a VTK file (without checking it) ... at least that ensures the code does not use outdated stuff.

src/save_load_scfres.jl
Dsantra92: I have not defined the generic three argument function. So I am creating the three argument function in the `save_load_scfres.jl`, where I am dumping the error methods.

src/save_load_scfres.jl
Dsantra92: Yes that will be a better choice.

src/external/VTK.jl
Dsantra92: I am going with with `band` and `k` so i don't think I will need such a long docstring anyway.

src/external/VTK.jl
Dsantra92: I inferred the name  convention from `jld2io.jl` ,nvm changing it.

src/save_load_scfres.jl
mfherbst: Sure, that's fine. You can just put the generic fallback to `save_load_scfres.jl` as well.

src/external/VTK.jl
mfherbst: Well, both versions are fine, I guess. Most importantly it just has to be consistent between lines 41 and 38.

src/save_load_scfres.jl
Dsantra92: I am curious about this decision. If both ways give us the same results(here errors) then why can do we care about another function dispatch? Is there some special advantage of dispatch here or is it just  a design thing?

mfherbst: Because it avoids explicitly writing the `if`s, which saves you about 10 lines of code, I'd say ... so yes, it's a design thing.

src/external/VTK.jl
```diff
@@ -0,0 +1,58 @@
...
+    for (idcs, r) in zip(CartesianIndices(basis.fft_size), r_vectors_cart(basis))
+        grid[:, Tuple(idcs)...] = r
+    end
+    vtkfile = vtk_grid(filename, grid)
+    
+    # Storing the Bloch Waves in Real space
+    for i in 1:length(basis.kpoints)
+        for j in 1:size(scfres.ψ[1])[2]
+            vtkfile["ψ_real_k$(i)_band$(j)"] = real.(G_to_r(basis, basis.kpoints[i], scfres.ψ[i][:,j]))
+            vtkfile["ψ_imag_k$(i)_band$(j)"] = imag.(G_to_r(basis, basis.kpoints[i], scfres.ψ[i][:,j]))
```
mfherbst, line 33: Much better. Now can you use iband and ik as iteration variables (a convention we use throughout the code) and also assign the G_to_r expression to a variable since this avoids computing it twice (quite costly) and also shortens the lines below the limit.

src/external/VTK.jl
```diff
@@ -0,0 +1,58 @@
...
+
+- ρ -> Density in real space
+- ψreal\\_k(i)\\_band(j) -> Real values of Bloch waves in real space
+- ψimag\\_k(i)\\_band(j) -> Imaginary values of Bloch waves in real space
+- ρspin -> Real value of ρspin are stored if ρspin in present
+
+MetaData:
+Energy , EigenValues, Fermi Level and occupation.
+"""
+function save_scfres(filename::AbstractString, scfres::NamedTuple, format::Val{Symbol(".vts")})
```
mfherbst, line 20: Just `::Val{:vts}`.

src/jld2io.jl
```diff
@@ -35,7 +35,7 @@ function save_scfres(jld::JLD2.JLDFile, scfres::NamedTuple)
 
     jld
 end
-function save_scfres(file::AbstractString, scfres::NamedTuple)
+function save_scfres(file::AbstractString, scfres::NamedTuple, format::Val{Symbol(".jld")})
```
mfherbst, line 38: Same, Val{:jld}.

test/save_load_scfres.jl
```diff
@@ -0,0 +1,37 @@
...
+kgrid = [3, 3, 3]
+Ecut = 15
+magnetic_moments = [Fe => [4, ]];
+model = model_LDA(lattice, atoms, magnetic_moments=magnetic_moments, temperature=0.01)
+basis = PlaneWaveBasis(model, Ecut; kgrid=kgrid)
+ρspin = guess_spin_density(basis, magnetic_moments)
+scfres = self_consistent_field(basis, tol=1e-6, ρspin=ρspin, mixing=KerkerMixing());
+
+@test_throws ErrorException save_scfres("MyVTKfile.vts",scfres)
+@test_throws ErrorException save_scfres("MyVTKfile.jld",scfres)
```
mfherbst, line 22: Some other tests actually load JLD2 and you should not rely on this being the only use of WriteVTK either (tests should be decoupled).

test/save_load_scfres.jl
```diff
@@ -0,0 +1,37 @@
...
+using DFTK
+
+a = 5.42352
+lattice = a / 2 * [[-1  1  1];
+                   [ 1 -1  1];
+                   [ 1  1 -1]]
+Fe = ElementPsp(:Fe, psp=load_psp("hgh/lda/Fe-q8.hgh"))
+atoms = [Fe => [zeros(3)]];
+kgrid = [3, 3, 3]
+Ecut = 15
```
mfherbst, line 14: This is going to be much to expensive too run on the CI for such a simple thing to test for. Pick something cheaper like good old silicon.

test/save_load_scfres.jl
```diff
@@ -0,0 +1,37 @@
...
+@test_throws ErrorException save_scfres("MyVTKfile.jld",scfres)
+@test_throws ErrorException save_scfres("MyVTKfile.random",scfres)
+@test_throws ErrorException save_scfres("MyVTKfile",scfres)
+
+using WriteVTK
+using JLD2
+
+mktempdir() do tmpdir
+    cd(tmpdir)
+    @test save_scfres("MyVTKfile.vts",scfres) == "MyVTKfile.vts"
```
mfherbst, line 31: Spacing

src/save_load_scfres.jl
```diff
@@ -0,0 +1,36 @@
...
+
+function save_scfres(filename::AbstractString, scfres::NamedTuple, format)
+    if format == Val(Symbol(".vts"))
+        error("Package WriteVTK needs to be imported before using this function.")
+    elseif format == Val(Symbol(".jld"))
+        error("Package JLD2 needs to be imported before using this function.")
+    else
+        error("$(splitext(filename)[end])extension is currently unsupported. The currently supported 
+        extensions are .vts(for VTK file)  and .jld(for JLD file).")
+    end
```
mfherbst, line 35: I'd be less specific: Just say generically that the extension is not supported, which could indicate that a package (e.g. JLD2 or WriteVTK) is not yet loaded.

src/save_load_scfres.jl
```diff
@@ -0,0 +1,36 @@
...
+
+Save an `scfres` obtained from `self_consistent_field` to a JLD2 or VTK file depending on the extension.
+
+!!! warning "No compatibility guarantees"
+    No guarantees are made with respect to this function at this point.
+    It may change incompatibly between DFTK versions or stop working / be removed
+    in the future.
+"""
+function save_scfres(filename::AbstractString, scfres::NamedTuple)
+    fs = splitext(filename)
```
mfherbst, line 22:
```suggestion
    _, ext = splitext(filename)
```

src/save_load_scfres.jl
```diff
@@ -0,0 +1,36 @@
...
+    @assert length(fs) != 1 "No file extension mentioned"
```
mfherbst, line 23:
```suggestion
    isempty(ext) && error("Provided filename misses an extension.")
```

src/save_load_scfres.jl
```diff
@@ -0,0 +1,36 @@
...
+    save_scfres(filename, scfres, Val(Symbol(fs[end])))
```
mfherbst, line 24:
```suggestion
    save_scfres(filename, scfres, Val(Symbol(ext[2:end])))
```

test/save_load_scfres.jl
Dsantra92: What exactly do we mean by decouple here? A test should run different from runtests.jl ? Or can we just shift this test to the first( this can cause problems in future I guess.)

src/external/VTK.jl
```diff
@@ -27,10 +27,11 @@ function save_scfres(filename::AbstractString, scfres::NamedTuple, format::Val{S
     vtkfile = vtk_grid(filename, grid)
     
     # Storing the Bloch Waves in Real space
-    for i in 1:length(basis.kpoints)
-        for j in 1:size(scfres.ψ[1])[2]
-            vtkfile["ψ_real_k$(i)_band$(j)"] = real.(G_to_r(basis, basis.kpoints[i], scfres.ψ[i][:,j]))
-            vtkfile["ψ_imag_k$(i)_band$(j)"] = imag.(G_to_r(basis, basis.kpoints[i], scfres.ψ[i][:,j]))
+    for ik in 1:length(basis.kpoints)
+        for iband in 1:size(scfres.ψ[1])[2]
+            ψ = G_to_r(basis, basis.kpoints[ik], scfres.ψ[ik][:,iband])
```
mfherbst, line 32: That variable name is a little misleading (it's not the full wave function). Better call it `ψnk_real`

src/save_load_scfres.jl
```diff
@@ -19,18 +19,12 @@ Save an `scfres` obtained from `self_consistent_field` to a JLD2 or VTK file dep
     in the future.
 """
 function save_scfres(filename::AbstractString, scfres::NamedTuple)
-    fs = splitext(filename)
-    @assert length(fs) != 1 "No file extension mentioned"
-    save_scfres(filename, scfres, Val(Symbol(fs[end])))
+    _, ext = splitext(filename)
+    isempty(ext)  && error("Provided filename misses extension")
```
mfherbst, line 23: Only one space before `&&` (could have been my bad ... sorry)

src/save_load_scfres.jl
```diff
@@ -19,18 +19,12 @@ Save an `scfres` obtained from `self_consistent_field` to a JLD2 or VTK file dep
...
-    if format == Val(Symbol(".vts"))
-        error("Package WriteVTK needs to be imported before using this function.")
-    elseif format == Val(Symbol(".jld"))
-        error("Package JLD2 needs to be imported before using this function.")
-    else
-        error("$(splitext(filename)[end])extension is currently unsupported. The currently supported 
-        extensions are .vts(for VTK file)  and .jld(for JLD file).")
-    end
+    error("The extension is not supported, which could indicate that a package 
+    (e.g. JLD2 or WriteVTK) is not yet loaded.")
```
mfherbst, line 29: I think this keeps the line-break in the string. I'd better do
```suggestion
    error("The extension is not supported, which could indicate that a package" *
          "(e.g. JLD2 or WriteVTK) is not yet loaded.")
```

test/save_load_scfres.jl
```diff
@@ -1,27 +1,27 @@
-# This test is based on example examples/collinear_magnetism.jl
+# This test is based on example docs/src/guide/tutorial.jl
 # Import statements are not all included in front to test for 
 # error statements in dispatch
```
mfherbst: I'd put that in line 26.

test/save_load_scfres.jl
```diff
@@ -1,27 +1,27 @@
-# This test is based on example examples/collinear_magnetism.jl
+# This test is based on example docs/src/guide/tutorial.jl
```
mfherbst: I don't think this is really needed.

test/save_load_scfres.jl
```diff
@@ -1,27 +1,27 @@
...
-magnetic_moments = [Fe => [4, ]];
-model = model_LDA(lattice, atoms, magnetic_moments=magnetic_moments, temperature=0.01)
+a = 5.431u"angstrom"
+lattice = a / 2 * [[0 1 1.];
+                   [1 0 1.];
+                   [1 1 0.]]
+Si = ElementPsp(:Si, psp=load_psp("hgh/lda/Si-q4"))
+atoms = [Si => [ones(3)/8, -ones(3)/8]]
+model = model_LDA(lattice, atoms)
+kgrid = [4, 4, 4]
```
mfherbst, line 16: Just on k-Point is sufficient. Does not need to be accurate at all.

test/save_load_scfres.jl
```diff
@@ -1,27 +1,27 @@
...
-kgrid = [3, 3, 3]
-Ecut = 15
-magnetic_moments = [Fe => [4, ]];
-model = model_LDA(lattice, atoms, magnetic_moments=magnetic_moments, temperature=0.01)
+a = 5.431u"angstrom"
+lattice = a / 2 * [[0 1 1.];
+                   [1 0 1.];
+                   [1 1 0.]]
+Si = ElementPsp(:Si, psp=load_psp("hgh/lda/Si-q4"))
+atoms = [Si => [ones(3)/8, -ones(3)/8]]
```
mfherbst, line 14: Actually in tests we tend to use the `testcases.jl` file, which stores some of this data once and for all ... see the other tests for how to do that.

test/save_load_scfres.jl
```diff
@@ -1,27 +1,27 @@
...
+model = model_LDA(lattice, atoms)
+kgrid = [4, 4, 4]
+Ecut = 7
 basis = PlaneWaveBasis(model, Ecut; kgrid=kgrid)
-ρspin = guess_spin_density(basis, magnetic_moments)
-scfres = self_consistent_field(basis, tol=1e-6, ρspin=ρspin, mixing=KerkerMixing());
+scfres = self_consistent_field(basis, tol=1e-8)
```
mfherbst, line 19: No need for such a low tolerance.

test/save_load_scfres.jl
mfherbst: Yes combining the tests is fine, but this won't help. The problem is that you rely that the tests are executed in a particular order (because no test may load `JLD2` before this test runs, otherwise it will fail ... try it!), but there is no guarantee the test order will stay as it is in the future.

src/external/VTK.jl
Dsantra92: I kept the name  `ψ_real` to keep the naming consistent with the next few lines.

test/save_load_scfres.jl
Dsantra92: `save_load_scfres` will run a completely separate test? I am a little hesitant to do that because ideally `test/runtests.jl` should run all tests in the package. We can use a package [Safetests.jl](https://github.com/YingboMa/SafeTestsets.jl) as mentioned in a [discourse thread](https://discourse.julialang.org/t/best-practices-for-julia-unit-testing/30858/3) but then I would not bother adding another package for one error test which can be done separately. Another thing we can do is add a `@testset` before `DFTK`, but that will be a major design issue. I am confused.

src/external/VTK.jl
```diff
@@ -0,0 +1,59 @@
...
+    # Storing ρspin if it is present.
+    isnothing(scfres.ρspin)  || (vtkfile["ρspin"] = scfres.ρspin.real)
+    
+    # Storing the energy components
+    for key in keys(scfres.energies)
+        vtkfile["energy_$key"] = scfres.energies[key]
+    end
+    
+    # Storing the Fermi level 
+    vtkfile["εF"] =  scfres.εF
```
mfherbst, line 50: Got a space too much here :smile:
```suggestion
    vtkfile["εF"] = scfres.εF
```

src/external/VTK.jl
```diff
@@ -0,0 +1,59 @@
...
+
+    # Storing the EigenValues as a matrix
+    vtkfile["eigenvalues"] = hcat(scfres.eigenvalues...)
+
+    # Storing the Occupation as a matrix
+    vtkfile["occupation"] = hcat(scfres.occupation...)
+
+    vtk_save(vtkfile)[1]
```
mfherbst, line 58: Now that I think about it, I think I would use the `only` function here, because you always expect to only get one element in return.

src/external/VTK.jl
```diff
@@ -0,0 +1,59 @@
...
+        grid[:, Tuple(idcs)...] = r
+    end
+    vtkfile = vtk_grid(filename, grid)
+    
+    # Storing the Bloch Waves in Real space
+    for ik in 1:length(basis.kpoints)
+        for iband in 1:size(scfres.ψ[1])[2]
+            ψ_real = G_to_r(basis, basis.kpoints[ik], scfres.ψ[ik][:,iband])
+            vtkfile["ψ_real_k$(ik)_band$(iband)"] = real.(ψ)
+            vtkfile["ψ_imag_k$(ik)_band$(iband)"] = imag.(ψ)
```
mfherbst, line 34: Sorry to be asking for another change on this ... names are always tricky. I think actually `ψ_k$(ik)_band$(iband)_real` and  `ψ_k$(ik)_band$(iband)_imag` is actually better, because this makes it more clear that this is the real and the imaginary part of `ψ[ik][:,iband]`. I think that's also what confused you here.

src/external/VTK.jl
```diff
@@ -0,0 +1,59 @@
...
+Grid Values:
+
+- ρ -> Density in real space
+- ψreal\\_k(i)\\_band(j) -> Real values of Bloch waves in real space
+- ψimag\\_k(i)\\_band(j) -> Imaginary values of Bloch waves in real space
+- ρspin -> Real value of ρspin are stored if ρspin in present
+
+MetaData:
+Energy , EigenValues, Fermi Level and occupation.
+"""
```
mfherbst, line 19: Since we use documenter for the docstrings you can put actual LaTeX inside, see https://juliadocs.github.io/Documenter.jl/stable/man/latex/

src/external/VTK.jl
mfherbst: Actually ... it's more the other way round. The symbols stored in the VTK file are a bit confusing. The reason is that `ψ` refers to all orbitals in DFTK and we use `ψk` for the orbitals at a particular `k`-Point (so `ψ[ik]`) and `ψnk` for `ψ[ik][:, iband]`.

test/save_load_scfres.jl
mfherbst: Yes, sorry to be confusing. Indeed `runtests.jl` should run all tests. But it could happen that in future restructuring of the test suite we change the order of the includes in `runtests.jl`. The problem is that your test assumes that e.g. `JLD2` has not yet been imported with a `using` statement in any of the testcases running before your testcase. That might not be true in the future when we might restructure the test suite, so you should not make this assumption in the test.

mfherbst: And I would not add Safetests.jl as another dependency actually. I don't think it's needed.

Dsantra92: On another note I feel the test for errors for non loaded modules are redundant. If they are not loaded there will be obvious errors and `@test_throws` does not test for which error whether it is MethodError or any user defined Error. Therefore the first two lines serves no purpose as of such and rest two tests can be integrated easily in a normal structure.

src/external/VTK.jl
mfherbst: Cool! Thanks!

src/save_load_scfres.jl
```diff
@@ -11,11 +11,20 @@ function load_scfres end
 """
     save_scfres(filename, scfres)
 
-Save an `scfres` obtained from `self_consistent_field` to a JLD2 file.
+Save an `scfres` obtained from `self_consistent_field` to a JLD2 or VTK file depending on the extension.
```
mfherbst, line 14: Can you break that line in two?

test/save_load_scfres.jl
```diff
@@ -0,0 +1,25 @@
...
+using WriteVTK
+using JLD2
+include("testcases.jl")
+
+Si = ElementPsp(14, psp=load_psp(silicon.psp))
+atoms = [Si => silicon.positions]
+model = model_LDA(silicon.lattice, atoms)
+kgrid = [1, 1, 1]
+Ecut = 7
+basis = PlaneWaveBasis(model, Ecut; kgrid = kgrid)
```
mfherbst, line 11: For kwargs we don't put spaces around the equals, so
```suggestion
basis = PlaneWaveBasis(model, Ecut; kgrid=kgrid)
```

test/save_load_scfres.jl
```diff
@@ -0,0 +1,25 @@
...
+scfres = self_consistent_field(basis)
+
+@test_throws ErrorException save_scfres("MyVTKfile.random", scfres)
+@test_throws ErrorException save_scfres("MyVTKfile", scfres)
+
+mktempdir() do tmpdir
+    cd(tmpdir)
+    @test save_scfres("MyVTKfile.vts",scfres) == "MyVTKfile.vts"
+    @test isfile("MyVTKfile.vts") == true
+    save_scfres("MyVTKfile.jld",scfres)
```
mfherbst, line 21: Again a small space issue :smile::
```suggestion
    save_scfres("MyVTKfile.jld", scfres)
```

test/save_load_scfres.jl
```diff
@@ -0,0 +1,25 @@
...
+Ecut = 7
+basis = PlaneWaveBasis(model, Ecut; kgrid = kgrid)
+scfres = self_consistent_field(basis)
+
+@test_throws ErrorException save_scfres("MyVTKfile.random", scfres)
+@test_throws ErrorException save_scfres("MyVTKfile", scfres)
+
+mktempdir() do tmpdir
+    cd(tmpdir)
+    @test save_scfres("MyVTKfile.vts",scfres) == "MyVTKfile.vts"
```
mfherbst, line 19: Put an extra space after a `,` to increase readability:
```suggestion
    @test save_scfres("MyVTKfile.vts", scfres) == "MyVTKfile.vts"
```

test/save_load_scfres.jl
```diff
@@ -0,0 +1,25 @@
...
+    @test isfile("MyVTKfile.vts") == true
+    save_scfres("MyVTKfile.jld",scfres)
+    @test isfile("MyVTKfile.jld") == true
```
mfherbst, line 22: Could you add cleanup code that removes those files again (just to ensure that the environment is clean in case the test is rerun).

mfherbst (commented):
Nice work! Almost there, just a few small nits from my end.

test/save_load_scfres.jl
mfherbst: I agree. I think the way you have it right now is completely fine.

mfherbst:
By the way to fix the missing WriteVTK in the tests, you need to add it to the [Project.toml](https://github.com/JuliaMolSim/DFTK.jl/blob/master/Project.toml#L83) in the same way as it is done for JLD2.

src/external/VTK.jl
```diff
@@ -0,0 +1,59 @@
+import WriteVTK: vtk_grid, vtk_save
+
+# Uses WriteVTK.jl to convert scfres structure to VTk file format
+
+@doc raw"""
+    save_scfres(filename, scfres, Val(:vtk))
```
mfherbst, line 6: Not in sync with the function definition, maybe just scratch this line completely ... it tends to become out of sync super quickly.

src/external/VTK.jl
```diff
@@ -0,0 +1,59 @@
...
+    grid = zeros(3, basis.fft_size...)
+    for (idcs, r) in zip(CartesianIndices(basis.fft_size), r_vectors_cart(basis))
+        grid[:, Tuple(idcs)...] = r
+    end
+    vtkfile = vtk_grid(filename, grid)
+    
+    # Storing the Bloch Waves in Real space
+    for ik in 1:length(basis.kpoints)
+        for iband in 1:size(scfres.ψ[1])[2]
+            ψnk_real = G_to_r(basis, basis.kpoints[ik], scfres.ψ[ik][:,iband])
```
mfherbst, line 32: Spacing after `,`.

src/external/VTK.jl
```diff
@@ -0,0 +1,59 @@
...
+    # Storing the Fermi level 
+    vtkfile["εF"] = scfres.εF
+
+    # Storing the EigenValues as a matrix
+    vtkfile["eigenvalues"] = hcat(scfres.eigenvalues...)
+
+    # Storing the Occupation as a matrix
+    vtkfile["occupation"] = hcat(scfres.occupation...)
+
+    only(vtk_save(vtkfile))
```
mfherbst, line 58: Actually in agreement with the JLD2 version, we should probably return the `vtkfile` from this function.

src/save_load_scfres.jl
```diff
@@ -11,11 +11,21 @@ function load_scfres end
...
 !!! warning "No compatibility guarantees"
     No guarantees are made with respect to this function at this point.
     It may change incompatibly between DFTK versions or stop working / be removed
     in the future.
 """
-function save_scfres end
+function save_scfres(filename::AbstractString, scfres::NamedTuple)
+    _, ext = splitext(filename)
+    isempty(ext) && error("Provided filename misses extension")
+    save_scfres(filename, scfres, Val(Symbol(ext[2:end])))
```
mfherbst, line 25: I'd add a comment that you dispatch on the extension.

mfherbst (approved):
Looking really good already! Just a few more small nits.

src/external/VTK.jl
Dsantra92: `vtkfile` unlike `jldfile` is not the actual file. It is actually a structure of arrays waiting to dumped to a '.vts' file. Returning that will no make sense.

mfherbst: Yeah ... kind of true. Ok. Let's leave it as is.

mfherbst:
The failing test is because writing stuff to disk is not yet supported in MPI-parallel runs. Just add an error (like for JLD2) to the VTK function and disable the test for MPI test runs (see the JLD2 test).

mfherbst:
Thanks @Dsantra92 !

Dsantra92:
My pleasure @mfherbst

PR #398, opened 2020-12-30
Windows support

### LINKED ISSUE #357: Windows support
Somebody asked me about this by email. Do we plan to do it? What sort of timeline? FFTW, MPI, Libxc all support windows, so it should not be too hard to do it in DFTK as well.

mfherbst:
Closes #357.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -24,7 +24,7 @@ jobs:
         include:
           - {version: '1.4',   os: ubuntu-latest,  payload: serial}
           - {version: '1.4',   os: macOS-latest,   payload: serial}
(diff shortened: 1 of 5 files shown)


antoine-levitt:
Nice! Ran into a family windows computer over the break? :p

Is there an easy way to run some CI less frequently than others? Most of what we do is platform independent so it seems like a waste to test each PR on all platforms.

mfherbst:
> Nice! Ran into a family windows computer over the break? :p

Yeah. Kind of.

> Is there an easy way to run some CI less frequently than others? Most of what we do is platform independent so it seems like a waste to test each PR on all platforms.

Yes, I have thought the same. Not sure how difficult it is, but probably one can implement some sort of flag / command to switch to "full CI" via a PR comment.

Issue #397, opened 2020-12-28
PSP Files For Missing Elements


umbriquse:
Just found that some elements are not available for use in the list_pspfile() function. Would it be possible to use a psp file found online? I came across a series of [psp files that were generated using the Troullier-Martins method](https://www.abinit.org/sites/default/files/PrevAtomicData/psp-links/lda_tm.html). I'm curious how appropriate they would be in regards to simulation, and if they could be incorporated into the DFTK database of elements.

antoine-levitt:
We only support Goedecker pseudopotentials right now; supporting others is https://github.com/JuliaMolSim/DFTK.jl/issues/2. It's not especially hard, one just needs to write a parser and add a simple radial integration routine to accomodate numerical rather than analytical radial parts, but we never got around to it. Regarding accuracy, I think the old Troullier-Martins pseudopotentials are pretty bad, and the more modern ONCVPSP pseudos are better, but I don't have much experience with pseudopotentials so couldn't say for sure.

umbriquse:
Ok, so what I plan on doing is to create a function(s) that parses the psp files that were generated by the Troullier-Martins method. Should I base this function on the "parse_hgh_file()" function? Maybe even use type dispatch to differentiate the two?

antoine-levitt:
That'd be awesome! So there are three parts to this : first, create a PspTM (or perhaps PspNormConserving since there's nothing specific to TM) which implements the same methods as PspHgh (we might then create an AbstractPsp type from which these two would derive but I'd rather wait until we can better see what the interface should be). Once this is done you should be able to create it manually and test it. The second part is writing a wrapper that reads a psp file and creates a PspTM struct. This is a bit more tricky because there are many formats around. Here I would search the internet for the most convenient format (pspnc, XML, upsp...) and implement that. I believe there are converters around. The third part is figuring out an "external" API and changing the list/load psp functions ; that we should do at the end.

Let us know if you need any help!

mfherbst:
Closing this in favour of the more specific #2 and #134. Feel free to add specific issues for the other mentioned file formats to keep track.

Regarding availability of pseudo definitions, I think psp8 and UPF are promising as ABINIT and QE have tons of files for a large list of elements / functionals.

antoine-levitt:
So from the formula side I think what you need is to compute the integral of a radial function specified on a grid against a bessel function j(q r), for several q. How do people do that, just simple sums? No special tricks eg for high q?

[...]

Issue #134, opened 2020-02-25
Implement ONCVPSP (Norm-converving Vanderbilt PSPs)


mfherbst:
Related to #2

azadoks:
Partially done in #741 .

mfherbst:
I'm going to call this done for now. We still miss features, but the key things are there.

antoine-levitt:
> pseudos where the local potential angular momentum is defined (i.e. not -1)

Is that common? Shouldn't be too hard

> fully-relativistic pseudos

This means Dirac equation? Do people really do plane waves with that?

azadoks:
> > pseudos where the local potential angular momentum is defined (i.e. not -1)
> 
> 
> 
> Is that common? Shouldn't be too hard

Not super common, and it shouldn't be very hard to handle.


> > fully-relativistic pseudos
> 
> 
> 
> This means Dirac equation? Do people really do plane waves with that?

Yes, they're from a solution of the Dirac equation. AFAIK, they're needed for spin-orbit coupling.

antoine-levitt:
> Yes, they're from a solution of the Dirac equation. AFAIK, they're needed for spin-orbit coupling

OK I see, the PSP code solves the Dirac equation, but the PW code solves for a C^2 valued wavefunction, which has the same Hamiltonian as usual plus a sigma . (nabla V x p) term. I think it's just an info on how the PSP was generated,  but there's nothing more to be done at the PSP level in this case - at least I don't see any relevant field in the UPF format (just PP_SEMILOCAL but pseudos are usually given in nonlocal form anyway).

azadoks:
>  but there's nothing more to be done at the PSP level in this case

I don't think this is exactly true; the data for the non-local projectors and the pseudo-waves _are_ different, and the way fully-relativistic pseudos are treated in UPF and ABINIT's PSP8 is different enough that they're incompatible without some major finagling.

antoine-levitt:
The data is different but the pseudo-pw code (ie the pseudopotential part of the hamiltonian) is the same? Or what do you mean? If you have references please add them here https://github.com/JuliaMolSim/DFTK.jl/issues/526, @ClementineBarat might be interested in implementing this at some point

azadoks:
[...]

PR #399, opened 2020-12-31
WIP: Troullier-Martins pseudopotential parser

umbriquse:
So this is my progress as of now. Sorry it has taken this long, I had no idea how pseudopotentials worked.

The parser is fully working, I am now stuck on where to go from here.


umbriquse:
Looking at #397 @antoine-levitt suggested to

> compute the integral of a radial function specified on a grid against a bessel function j(q r), for several q

But this format provides radial values for the pseudopotentials, and the first and second projector values. So I'm a bit confused as to which to use and also what this function is intended to do.

antoine-levitt:
Great, thanks for this and happy new year! Yes indeed 15 and 19 are what I was thinking about. The psp file gives you the radial part on a grid of the local and nonlocal potentials (or possibly the KB form of the nonlocal pot directly, I'm not sure), and DFTK uses the matrix elements in a plane wave basis (psphgh.jl has both real and recip space for completeness, but only the recp space expressions are actually used in practice - at least for now). These matrix elements involve Fourier transforms of radial times spherical harmonics, which in turn are expressed as integrals of the radial part against bessel functions (see https://juliamolsim.github.io/DFTK.jl/dev/advanced/useful_formulas/). There are some scattered docs in the psphgh and in local.jl and nonlocal.jl (in terms/), if anything is unclear please let me know and I can explain or add docs.

antoine-levitt:
To be clearer : since you're talking about projectors it means it's directly in KB form, the link with the TM paper is that the projectors are phiPP * Vnonloc (modulo normalisations and trickiness like that).

antoine-levitt:
After looking at the code in nonlocal.jl it appears we have been pretty careful about separating generic pseudopotentialness and specific Hgh (@mfherbst to thank for that!), so essentially you just have to use a new type instead of psphgh and add the missing methods that pop up.

mfherbst:
Hi @umbriquse and happy new year!

Thanks for giving this a go! Getting the HGHs to work was tedious enough so we never really bothered going beyond it, but TM pseudos are pretty wide-spread so definitely good to have them.

Indeed to get a new pseudo to work in DFTK you only need to slot it into the `psp` variable of the `ElementPsp` struct and make sure that all the required functions are implemented for `PspTM`. To be precise that is
- `eval_psp_projection_radial`  (KB Projector in Fourier space, i.e. equation 19)
- `eval_psp_energy_correction` (Correction of the ewald energy due to the PP ... have not found the expression on first sight in TM paper, but cannot be hard to compute it)
- `eval_psp_local_real` (local part in real space, currently unused, but should still be implemented for later)
- `eval_psp_local_fourier` (local part in fourier space the Vlocal(G - G') perhaps modulo different normalisation)

antoine-levitt:
Indeed it's as simple as that. We also use `psp.h` and `psp.lmax` but that should be it.

mfherbst:
Ah yes, that's true, but those you well need for the KB form anyway.

antoine-levitt:
I'll make a PR adding an abstract supertype and outlining what needs to be implemented, hopefully that'll clarify things.

mfherbst:
Good idea. When researching this it occurred to me as well that this could be useful.

antoine-levitt:
Hm, should we merge ElementPsp and PspHgh (ie have PspHgh inherit Element directly)?

mfherbst:
I'm not sure because you might want to have an element of one sort but combine it deliberately with the pseudo of another element. If element and pseudo are kind of the same thing that might become confusing.

antoine-levitt:
Why would you do something like that?

antoine-levitt:
The extra indirection is a bit confusing so it'd be good to get rid of it.

mfherbst:
Because you want to play with the number of valence electrons and the charge?

Sure give it a try. But what I like about the current design is that you can just say `ElementPsp(psp=load_psp(identifier))` and this does the right thing for every psp identifier (once it's implemented). So you don't need to specify the Psp type twice (once in the identifier and once in the type of element you use). That would be my main point of concern for the design if you merge both things.

antoine-levitt:
> Because you want to play with the number of valence electrons and the charge?

I'm not sure it makes sense. Let's discuss it sometimes

> Sure give it a try. But what I like about the current design is that you can just say ElementPsp(psp=load_psp(identifier)) and this does the right thing for every psp identifier (once it's implemented). So you don't need to specify the Psp type twice (once in the identifier and once in the type of element you use). That would be my main point of concern for the design if you merge both things.

That's surface API, it shouldn't interfere with the design of the structs. It can be done by an external function (or constructor), no? I'm not going to do it for now, but something to keep in mind.

mfherbst:
> It can be done by an external function (or constructor), no? 

I'm sure it can be done somehow. It's just something to keep in mind because types are to some extend more static than strings ... and the `list_psp` / `load_psp` stuff pops up at quite some places in user code, so we should invest some thought how changes can be least painful. Let's see when you have a design.

antoine-levitt:
Yeah I'm not too confortable around the load/listpsp stuff, so I'll let you handle it if you're OK with that

mfherbst:
Sure, no problem.

antoine-levitt:
I think the comment `p(q) = ∫_{R+} r^2 p(r) j_l(q r) dr` in eval_psp_projection_radial is missing a 4\pi (like for the local potential). Do you agree?

mfherbst:
Yes, see [tests](https://github.com/JuliaMolSim/DFTK.jl/blob/master/test/PspHgh.jl#L129)

antoine-levitt:
Nice, thanks!

antoine-levitt:
Done: https://github.com/JuliaMolSim/DFTK.jl/pull/401

antoine-levitt:
@umbriquse OK so with the new API in https://github.com/JuliaMolSim/DFTK.jl/pulls?q=is%3Apr+is%3Aclosed you just need to create a new subtype of NormConservingPsp and implement the documented methods.

umbriquse:
I'm going to need someone to check over the 'eval_psp_projection_radial' function. I'm not confident that I've got it 100% correct.

antoine-levitt:
I think you're still working off the old code without the refactoring PR, eg eval_psp_projection_radial has changed name to eval_psp_projector_fourier I think.

antoine-levitt:
From a very cursory glance it looks like it should be doing the right thing. I think a very good starting point to compare is the `@testset "Numerical integration to obtain fourier-space projectors" begin` in tests/PspHgh.jl (which should also be extended to the pair (eval_psp_local_real, eval_psp_local_fourier)). After that, once you're confident the _fourier and _real are consistent, I would just try to set up a silicon computation, compare against another DFT and hope for the best!

umbriquse:
Alright,  I'll start working on that then.

mfherbst:
>  compare against another DFT and hope for the best!

I think ABINIT has TM pseudos, which has the advantage that we have parses for their densities (i.e. you could converge the SCF in abint and really check the resulting SCF state term by term) ... but I would not do this detailed comparison unless you really need it for debugging, but good to have it in the back as an option.

umbriquse:
@antoine-levitt From memory ABINIT has since discontinued the use of TM pseudos due to their inaccuracy.

antoine-levitt:
Well the nice thing is that the functionality should really be the same for TM and ONCVPSP so you might as well just read the ONCVPSP pseudos?

src/pseudo/PspTm.jl
```diff
@@ -0,0 +1,162 @@
...
+    firstProjectorVals::Vector{Vector{Float64}}     # Radial First projection functions for each angular momentum
+    secondProjectorVals::Vector{Vector{Float64}}    # If numProjectorFctns permits, the second projection functions for each angluar momentum
+    identifier::String                              # String identifying the PSP
+    description::String                             # Descriptive string
+end
+
+function parse_tm_file(path; identifier="")
+    file = readlines(path)
+    description = file[1]
+
```
mfherbst, line 31: Just a general remark: In my experience debugging these parsers is a pain, especially if it's been a while since you wrote it. We had multiple subtle bugs in the HGH parser that only popped up a few months after I wrote it ... and until then you forget a lot. 

So to make it easier for our future selves: Now that you understand the format, I'd strongly suggest to comment in detail what stuff you are expecting to see in the TM file in each parsing step, what assumptions you are making about its structure and what you parse in each line.

I'd also avoid to try and be "clever" and use a lot of assertive programming. E.g. lines 32 and 34 are quite a mouthful and additionally might silently throw away fields beyond `6` without checking there are none. I'm not an expert in the TM format, so this might be completely ligit, but definitely comment on what you expect.

test/parsePspTM.jl
```diff
@@ -0,0 +1,26 @@
+using Test
+using DFTK: parse_tm_file
+
+@testset "Check reading" begin
+    psp = parse_tm_file("/Users/jasonlehto/git/Dino.jl/dev/DFTK/data/psp/tm/69tm.pspnc")
```
mfherbst, line 5: I guess Dino is the name of your new package then :smile:.

mfherbst (commented):
Wow nice work collecting all this stuff ... truly is annoying work. I left a tiny note on the parser.

Also take a look at the coding conventions we use in other files once things work and you're ready to do a polishing round ... you differ a bit at places :smile:.

umbriquse:
I'm not familiar with the Hankel transform that is done in the PspHgh.jl test file. Does this transform work for any real projector function?

test/parsePspTM.jl
umbriquse: Yes, I believe that eventually it will be made public. Once some more polishing happens. It's kind of a mess right now.

antoine-levitt:
It's just what comes out of the second equation in https://juliamolsim.github.io/DFTK.jl/dev/advanced/useful_formulas/

umbriquse:
While setting up a Si lattice simulation I encountered an error. The setup is shown below and the error is shown last.
```
Si = ElementPsp(14,:Si,parse_tm_file("dev/DFTK/data/psp/tm/lda/si-q4.pspnc"))
lattice = austrip(0.5431u"nm") / 2 * [  [0 1 1.];
                                    [1 0 1.];
                                    [1 1 0.]]
atoms = [Si => [ones(3)/8, -ones(3)/8]]
model = model_LDA(lattice,atoms)
kgrid = [4,4,4]
Ecut = 20
basis = PlaneWaveBasis(model, Ecut; kgrid = kgrid)
scfres = self_consistent_field(basis)
@show scfres.energies.total
```

The TM psp doesn't have any h coefficients. I'm curious if I should just plant an empty h variable, or create a new `count_n_proj()` function that counts the projection functions in the psp.

```
ERROR: LoadError: type PspTm has no field h
Stacktrace:
 [1] getproperty(::PspTm, ::Symbol) at ./Base.jl:33
 [2] (::DFTK.var"#14#15"{PspTm})(::Int64) at ./none:0
 [3] MappingRF at ./reduce.jl:93 [inlined]
 [4] _foldl_impl at ./reduce.jl:58 [inlined]
 [5] foldl_impl(::Base.MappingRF{DFTK.var"#14#15"{PspTm},Base.BottomRF{typeof(Base.add_sum)}}, ::NamedTuple{(),Tuple{}}, ::UnitRange{Int64}) at ./reduce.jl:48
 [6] mapfoldl_impl(::typeof(identity), ::typeof(Base.add_sum), ::NamedTuple{(),Tuple{}}, ::Base.Generator{UnitRange{Int64},DFTK.var"#14#15"{PspTm}}) at ./reduce.jl:44
 [7] mapfoldl(::Function, ::Function, ::Base.Generator{UnitRange{Int64},DFTK.var"#14#15"{PspTm}}; kw::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at ./reduce.jl:160
 [8] mapfoldl(::Function, ::Function, ::Base.Generator{UnitRange{Int64},DFTK.var"#14#15"{PspTm}}) at ./reduce.jl:160
 [9] mapreduce(::Function, ::Function, ::Base.Generator{UnitRange{Int64},DFTK.var"#14#15"{PspTm}}; kw::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at ./reduce.jl:287
 [10] mapreduce(::Function, ::Function, ::Base.Generator{UnitRange{Int64},DFTK.var"#14#15"{PspTm}}) at ./reduce.jl:287
 [11] sum(::Function, ::Base.Generator{UnitRange{Int64},DFTK.var"#14#15"{PspTm}}) at ./reduce.jl:494
 [12] sum at ./reduce.jl:511 [inlined]
 [13] count_n_proj at /Users/jasonlehto/git/Dino.jl/dev/DFTK/src/pseudo/NormConservingPsp.jl:89 [inlined]
 [14] #16 at ./none:0 [inlined]
 [15] MappingRF at ./reduce.jl:93 [inlined]
 [16] _foldl_impl(::Base.MappingRF{DFTK.var"#16#17",Base.BottomRF{typeof(Base.add_sum)}}, ::Base._InitialValue, ::Array{Pair{PspTm,Array{SArray{Tuple{3},Float64,1,3},1}},1}) at ./reduce.jl:58
 [17] foldl_impl at ./reduce.jl:48 [inlined]
 [18] mapfoldl_impl(::typeof(identity), ::typeof(Base.add_sum), ::NamedTuple{(),Tuple{}}, ::Base.Generator{Array{Pair{PspTm,Array{SArray{Tuple{3},Float64,1,3},1}},1},DFTK.var"#16#17"}) at ./reduce.jl:44
 [19] mapfoldl(::Function, ::Function, ::Base.Generator{Array{Pair{PspTm,Array{SArray{Tuple{3},Float64,1,3},1}},1},DFTK.var"#16#17"}; kw::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at ./reduce.jl:160
 [20] mapfoldl at ./reduce.jl:160 [inlined]
 [21] mapreduce(::Function, ::Function, ::Base.Generator{Array{Pair{PspTm,Array{SArray{Tuple{3},Float64,1,3},1}},1},DFTK.var"#16#17"}; kw::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at ./reduce.jl:287
 [22] mapreduce at ./reduce.jl:287 [inlined]
 [23] sum at ./reduce.jl:494 [inlined]
 [24] sum at ./reduce.jl:511 [inlined]
 [25] count_n_proj at /Users/jasonlehto/git/Dino.jl/dev/DFTK/src/pseudo/NormConservingPsp.jl:92 [inlined]
 [26] build_projection_vectors_(::PlaneWaveBasis{Float64}, ::Array{Pair{PspTm,Array{SArray{Tuple{3},Float64,1,3},1}},1}, ::Kpoint{Float64}) at /Users/jasonlehto/git/Dino.jl/dev/DFTK/src/terms/nonlocal.jl:152
 [27] #201 at /Users/jasonlehto/git/Dino.jl/dev/DFTK/src/terms/nonlocal.jl:14 [inlined]
 [28] iterate at ./generator.jl:47 [inlined]
 [29] _collect(::Array{Kpoint{Float64},1}, ::Base.Generator{Array{Kpoint{Float64},1},DFTK.var"#201#204"{PlaneWaveBasis{Float64},Array{Pair{PspTm,Array{SArray{Tuple{3},Float64,1,3},1}},1}}}, ::Base.EltypeUnknown, ::Base.HasShape{1}) at ./array.jl:699
 [30] collect_similar(::Array{Kpoint{Float64},1}, ::Base.Generator{Array{Kpoint{Float64},1},DFTK.var"#201#204"{PlaneWaveBasis{Float64},Array{Pair{PspTm,Array{SArray{Tuple{3},Float64,1,3},1}},1}}}) at ./array.jl:628
 [31] map(::Function, ::Array{Kpoint{Float64},1}) at ./abstractarray.jl:2162
 [32] (::AtomicNonlocal)(::PlaneWaveBasis{Float64}) at /Users/jasonlehto/git/Dino.jl/dev/DFTK/src/terms/nonlocal.jl:13
 [33] macro expansion at /Users/jasonlehto/.julia/packages/TimerOutputs/ZmKD7/src/TimerOutput.jl:190 [inlined]
 [34] macro expansion at /Users/jasonlehto/git/Dino.jl/dev/DFTK/src/PlaneWaveBasis.jl:226 [inlined]
 [35] (::DFTK.var"#858#859"{Float64,Bool,Bool,Int64,Array{Int64,1},Array{Float64,1},Model{Float64},Int64})() at /Users/jasonlehto/.julia/packages/TimerOutputs/ZmKD7/src/TimerOutput.jl:237
 [36] timeit(::DFTK.var"#858#859"{Float64,Bool,Bool,Int64,Array{Int64,1},Array{Float64,1},Model{Float64},Int64}, ::TimerOutputs.TimerOutput, ::String) at /Users/jasonlehto/.julia/packages/TimerOutputs/ZmKD7/src/TimerOutput.jl:285
 [37] #PlaneWaveBasis#857 at /Users/jasonlehto/.julia/packages/TimerOutputs/ZmKD7/src/TimerOutput.jl:236 [inlined]
 [38] PlaneWaveBasis(::Model{Float64}, ::Int64; kgrid::Array{Int64,1}, kshift::Array{Float64,1}, use_symmetry::Bool, kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /Users/jasonlehto/git/Dino.jl/dev/DFTK/src/PlaneWaveBasis.jl:268
 [39] top-level scope at /Users/jasonlehto/git/Dino.jl/dev/DFTK/test/PspTM.jl:96
 [40] include(::String) at ./client.jl:457
 [41] top-level scope at REPL[1]:1
in expression starting at /Users/jasonlehto/git/Dino.jl/dev/DFTK/test/PspTM.jl:96
```

antoine-levitt:
There should be an h in the pseudopotential specification. The nonlocal term is sum_l,i1,i2 h[l][i1,i2] |pli1><pli2|

mfherbst:
I think it is hidden in the three fields `pseudoPotentials`, `firstProjectorVals` and `secondProjectorVals`, probably you just need to reorganise the expression involving them and then you can define such an `h`.

umbriquse:
I tested out the new psp and it's functional. The biggest issue with this psp is that the inverse functions of the local psp and the non-local psp do not have analytical functions. So the current files are very inaccurate, but are evaluated quick enough, just to show that the psp is functional.
```
n     Free energy       Eₙ-Eₙ₋₁     ρout-ρin   Diag
---   ---------------   ---------   --------   ----
  1   -415.4003818532         NaN   2.04e+01    25.0 
  2   -415.1978059239    2.03e-01   4.23e+00    9.1 
  3   -411.5964139299    3.60e+00   2.82e+00    45.0 
  4   -415.6349664097   -4.04e+00   7.44e-01    27.0 
  5   -415.8347856391   -2.00e-01   3.55e-01    10.2 
  6   -415.8583515221   -2.36e-02   1.99e-01    6.6 
  7   -415.8466844242    1.17e-02   1.92e-01    7.4 
  8   -415.8808276543   -3.41e-02   3.66e-02    6.3 
  9   -415.8820543371   -1.23e-03   1.22e-02    3.4 
 10   -415.8822390962   -1.85e-04   4.02e-03    6.0 
 11   -415.8822505754   -1.15e-05   2.49e-03    5.1 
 12   -415.8822508048   -2.29e-07   2.39e-03    4.0 
scfres.energies.total = -415.88225080480703
```
I would like to use the Hankel.jl package to speed up these hankel transformations, but there are some dependency issues. The dependency error is shown below.
Dependency ERROR:
```
(Dino) pkg> add Hankel
   Updating registry at `~/.julia/registries/General`
   Updating git-repo `https://github.com/JuliaRegistries/General.git`
  Resolving package versions...
ERROR: Unsatisfiable requirements detected for package FunctionZeros [b21f74c0]:
 FunctionZeros [b21f74c0] log:
 ├─possible versions are: 0.1.1 or uninstalled
 ├─restricted by compatibility requirements with Hankel [74863788] to versions: 0.1.1
 │ └─Hankel [74863788] log:
 │   ├─possible versions are: [0.1.0, 0.2.0-0.2.1, 0.3.0, 0.4.0, 0.5.0-0.5.1] or uninstalled
 │   └─restricted to versions * by an explicit requirement, leaving only versions [0.1.0, 0.2.0-0.2.1, 0.3.0, 0.4.0, 0.5.0-0.5.1]
 └─restricted by compatibility requirements with Roots [f2b01f46] to versions: uninstalled — no versions left
   └─Roots [f2b01f46] log:
     ├─possible versions are: [0.7.0-0.7.4, 0.8.0-0.8.1, 0.8.3-0.8.4, 1.0.0-1.0.7] or uninstalled
     └─restricted to versions 1 by DFTK [acf6eb54], leaving only versions 1.0.0-1.0.7
       └─DFTK [acf6eb54] log:
         ├─possible versions are: 0.2.4 or uninstalled
         └─DFTK [acf6eb54] is fixed to version 0.2.4
```

antoine-levitt:
Nice, that's a good first step! Now it would be interesting to check accuracy. You can compare with other pseudos just to get an idea if the results are reasonable or not, eg by computing the lattice constant of silicon with different psps and checking they're about the same. For real validation we should compare with eg abinit: the results with the same pseudo should be very very close.

Regarding the Hankel transform, have you checked how people do it in other codes? Hankel.jl fixes the grids both in real and recp spaces, so you'd need interpolation to convert between the two, but maybe it's still the fastest way? Also https://github.com/chrisbrahms/Hankel.jl/issues/14 is not very reassuring... Regarding the compat I'll let @mfherbst answer, but it looks from https://github.com/chrisbrahms/Hankel.jl/blob/master/Project.toml#L14 that hankel.jl is restricted to a very old version of FunctionZeros, so probably they need to update their requirements.

mfherbst:
Given it's current status I'm pretty critical of Hankel.jl. The package is not well maintained (recent commits are GHA updates, which have been recommended by the community for a couple of months) and its dependency FunctionZeros.jl does not seem mature. So if we can avoid it, I'd not introduce that dependency.

Regarding the compat issue, the issue is that FunctionZeros.jl and Roots.jl currently don't go well with each other in the version of Roots we use (1.0), so for now not much we can do on our side. It seems they fixed it in FunctionZeros master, though, so it will become available in the future.

Also I don't think the numerical quadrature as you do it now has such a big speed impact, since you only have to do it a single time per calculation per grid, in theory. Right now it's once per SCF step because of the way we build the hamiltonian, so should still be manageable. Can you time it to give a feel how urgent it is.

Another way to solve it is to avoid doing this for every SCF step. What I have in mind is to pass an "old hamiltonian" when building the new one ... in this way the terms that don't change can decide to forward some info from the old hamiltonian. @antoine-levitt It introduces some coupling, but should be manageable, don't you think? Also we will probably need sth like this for PAW anyway at some point.

antoine-levitt:
> Another way to solve it is to avoid doing the setup for every SCF step. What I have in mind is to pass an "old hamiltonian" when building the new one in energy_hamiltonian

There should be no need for this. The term is instantiated with the pwbasis, and can precompute everything it needs here. That's what is done right now with psphgh.

> Regarding the compat issue, the issue is that FunctionZeros.jl and Roots.jl currently don't go well with each other in the version of Roots we use (1.0), so for now not much we can do on our side. It seems they fixed it in FunctionZeros master, though, so it will at least work soonish.

Yeah that's what I hate with this compatibility thing. Unless you constantly merge PRs on your repo, your deps get out of date really quick and then all downstream packages get messed up. So if somebody abandons a repo, it's unusable even if the code itself would be perfectly valid.

mfherbst:
> There should be no need for this. The term is instantiated with the pwbasis, and can precompute everything it needs here. That's what is done right now with psphgh.

Ah yes you're absolutely right. I got confused for a sec. So indeed: The Hankel will only be done once per SCF per basis, so no need to worry about timing that much.

> So if somebody abandons a repo, it's unusable even if the code itself would be perfectly valid.

Yes but in my opinion the reverse (code still there, but is broken) is still worse. That happens in the python ecosystem all the time ... At least in this way you notice what is maintained and what not ... and can contribute a PR.

umbriquse:
Here are some of the time trials that i've done.

For rtol = 1e0 in `eval_psp_local_fourier` in `quadgk`
```
(#= /Users/jasonlehto/git/Dino.jl/dev/DFTK/src/terms/local.jl:86 =# @__LINE__(), iG, G, time() - time1) = (86, 1024, [-1, -1, 0], 8.518682956695557)
(#= /Users/jasonlehto/git/Dino.jl/dev/DFTK/src/terms/local.jl:86 =# @__LINE__(), iG, G, time() - time1) = (86, 2048, [-1, -1, 1], 9.731114864349365)
(#= /Users/jasonlehto/git/Dino.jl/dev/DFTK/src/terms/local.jl:86 =# @__LINE__(), iG, G, time() - time1) = (86, 3072, [-1, -1, 2], 7.254722833633423)
(#= /Users/jasonlehto/git/Dino.jl/dev/DFTK/src/terms/local.jl:86 =# @__LINE__(), iG, G, time() - time1) = (86, 4096, [-1, -1, 3], 7.212761878967285)
(#= /Users/jasonlehto/git/Dino.jl/dev/DFTK/src/terms/local.jl:86 =# @__LINE__(), iG, G, time() - time1) = (86, 5120, [-1, -1, 4], 7.632298946380615)
(#= /Users/jasonlehto/git/Dino.jl/dev/DFTK/src/terms/local.jl:86 =# @__LINE__(), iG, G, time() - time1) = (86, 6144, [-1, -1, 5], 7.296127080917358)
```

For rtol = 1e-1
```
(#= /Users/jasonlehto/git/Dino.jl/dev/DFTK/src/terms/local.jl:86 =# @__LINE__(), iG, G, time() - time1) = (86, 1024, [-1, -1, 0], 62.51997089385986)
(#= /Users/jasonlehto/git/Dino.jl/dev/DFTK/src/terms/local.jl:86 =# @__LINE__(), iG, G, time() - time1) = (86, 2048, [-1, -1, 1], 55.708820104599)
(#= /Users/jasonlehto/git/Dino.jl/dev/DFTK/src/terms/local.jl:86 =# @__LINE__(), iG, G, time() - time1) = (86, 3072, [-1, -1, 2], 60.34907603263855)
(#= /Users/jasonlehto/git/Dino.jl/dev/DFTK/src/terms/local.jl:86 =# @__LINE__(), iG, G, time() - time1) = (86, 4096, [-1, -1, 3], 61.12815594673157)
```
The G-Vectors for this grid ranges from 15 to -16. So using the data from above, with a relative tolerance of 0.1 it would take over 30 minutes to setup the grid alone.

I'll check if anybody else has tried to create a quick Hankel transform. In the case that nobody has, I'm a bit curious what the next step would be. Should I try to build a custom quick Hankel transform, or are there some other speedup methods that can be used to get a speedup in the factor of 10^6?

antoine-levitt:
Let's split correctness and performance. If you focus on getting something that works and is accurate (= reproduces results from abinit) even if it's relatively slow, we can merge that and we can always tackle the speed issue later. Regarding performance I think interpolation and fast hankel transform seems like a reasonable idea. Again I would look at how other people tackle this, since this must be pretty standard.

mfherbst:
I agree. Don't worry too much about speed for now and first get things to work.

Regarding your timings: Given that there is not a stupid bug (type instability etc) that's causing this, that's of course not long-term acceptable. But we can always see about Hankel.jl at a later point (when it maybe is more mature). Also as Antoine said, it's kind of a standard problem and numerical integration is certainly throwing away all the specific info there is, so I'm sure one can find something that is faster and still relatively easy to realise in Julia.

sethaxen:
> Given it's current status I'm pretty critical of Hankel.jl. The package is not well maintained (recent commits are GHA updates, which have been recommended by the community for a couple of months) and its dependency FunctionZeros.jl does not seem mature. So if we can avoid it, I'd not introduce that dependency.

Hankel.jl is indeed maintained; we have not been actively implementing new features. I echo the concern about FunctionZeros.jl as a dependency. All of its recent maintenance has been pushed by Hankel.jl contributors. I'm looking into alternatives.

> Regarding the compat I'll let @mfherbst answer, but it looks from https://github.com/chrisbrahms/Hankel.jl/blob/master/Project.toml#L14 that hankel.jl is restricted to a very old version of FunctionZeros, so probably they need to update their requirements.

Thanks for bringing this to my attention! I'll bump it.

mfherbst:
> Hankel.jl is indeed maintained;

Good to know and thanks for getting in touch @sethaxen and letting us know about the momentum to replace FunctionZeros.jl! With that in mind, I withdraw my concerns wrt. using Hankel.jl :smile:.

sethaxen:
> Good to know and thanks for getting in touch @sethaxen and letting us know about the momentum to replace FunctionZeros.jl! With that in mind, I withdraw my concerns wrt. using Hankel.jl 😄.

No problem! We've removed FunctionZeros as a dependency. Now we use GSL with Roots, but we'll probably switch to a pure Julia solution in the future. We still need to extensively test that the transform itself is stable for high orders.

I should note that there are a few different ways to compute the Hankel transform, which have tradeoffs in terms of speed, simplicity, and usefulness of the grid points for a specific application. QDHT, which Hankel.jl implements is one of the simpler approaches, but it is not the fastest. e.g. it is `O(N^2)` to transform with `N` being the number of grid points, but there are Hankel transforms that use the FFT and get something like `O(N log(N))` but with other tradeoffs like many grid points near the origin and few far away. If you end up going with a different Hankel transform approach, we'd be happy to accept an implementation in Hankel.jl.

mfherbst:
> QDHT, which Hankel.jl implements is one of the simpler approaches, but it is not the fastest.

I'd say that will suffice. Thanks for the advice @sethaxen! It's not going to be used in a big computational bottle neck as it's only needed once per calculation run and not in the inner loop.

antoine-levitt:
Yeah but if it's N^2 there's no real point in using this rather than a naive implementation, where we also control the grid... I have a call with abinit people thursday, I can ask them how they do it. I vaguely seem to remember having seen "spline" somewhere so I'm guessing they do spline interpolation both in q and r space combined with a fast FFT-based transform, which looks like a sensible thing to do.

antoine-levitt:
Asked Marc about it: apparently they just sample in q space (with 3k points by default: https://docs.abinit.org/variables/dev/#mqgrid) and for each of these q points they compute the integral. I think they do that by numerical quadrature: brute force O(N M), where N is the number of points in q space and M in r space. N and M are not that big. Evaluation of a bessel function is 100ns on my computer, so 3k times however many radial points there are (probably not that many) times 100ns should be fine. That's stored somewhere. Then whenever p(q) is needed, it's interpolated with splines or some such. Doing that precomputation is better than doint it on the fly, since there will generally be much more than 3k plane waves.

antoine-levitt:
@umbriquse Hi! Can I ask what was the status of this PR and whether you plan to continue it in the future? If not, I might take a stab at it. As noted in https://github.com/JuliaMolSim/DFTK.jl/issues/134 it looks to me like the XML format is simpler to parse.

umbriquse:
Hello @antoine-levitt! It seems I'm always 2 to 3 days from revisiting this PR and finishing it, so I'm a bit surprised to see that 14 months have passes since my last push. Regardless, if you would like to complete it then by all means feel free. You would definitely be able to finish this quicker than I would.  Otherwise I can revisit this PR sometime next week.

antoine-levitt:
It's probably better if I do the rebase/polishing and I'm happy to do that, but the hard part is testing it and making sure there's no issue with normalization/conventions/etc, so if you've already done that or have the workflow to do it ready it'd make my job a lot easier. There's no hurry so if you think you'll do it I'll leave it to you!

umbriquse:
I believe all of the testing for it has been pushed to this fork. So no, most of the tests still need to be written. I do have plans of finishing this PR, so the reminder will only speed this along. From memory it's still to slow to even test with, so after incorporating the change that was previously mentioned I'll flush out the tests. After a quick look through some tests for the HgH pseudo potential they seem exhaustive so would it be ok to use that file as an example for what's desired?

antoine-levitt:
Yes, we test (on master) the consistency between local and fourier representations of the projectors and the local potential so passing those would be a good start. Another good check is computing the GTH quantities on a grid, using that in the integration functions, and checking against the analytical results. If you have performance issues don't hesitate to post them here, optimizing julia code is somewhat of a hobby of mine.

antoine-levitt:
btw looking at the code I would imagine the reason why it's currently slow is that the integrand is not smooth, and quadgk isn't really built for that. If the integrand decays fast enough I would imagine a straightforward quadrature with uniform weights is actually super efficient here, as long as the grid size is << 2pi/qmax (see https://people.maths.ox.ac.uk/trefethen/sirev56-3_385.pdf for why). But the local potential includes a Coulomb part, right? If so, you can't just integrate it up to a max radius, that'll miss a large contribution to the integral. The simplest thing might be to subtract a known potential that has the same long-range behavior (eg Z exp(-alpha r^2)/r).

antoine-levitt:
Also, TM pseudopotentials seem to be slightly different from GTH pseudopotentials: they might not already be in Kleinman-Bylander separable form? The XML ONCVPSP format seems closer to GTH so easier to implement (and to parse!)

antoine-levitt:
I'm closing this since it's now superseded by @azadoks 's UPF parser, but feel free to reopen if there are useful bits of code there!

PR #404, opened 2021-01-12
Better precision in divided differences

antoine-levitt:
Trick shamelessly stolen from @sethaxen over at https://github.com/JuliaDiff/ChainRules.jl/pull/193. Shouldn't really matter too much in practice, but nice to have nonetheless.


mfherbst:
Nice! Always funny how in numerical error analysis sometimes a few characters of permuting stuff make a difference.

PR #410, opened 2021-01-28
Support saving results for MPI-parallelised runs

mfherbst:


src/PlaneWaveBasis.jl
```diff
@@ -194,34 +197,41 @@ build_kpoints(basis::PlaneWaveBasis, kcoords) =
...
+        fft_size, opFFT, ipFFT, opIFFT, ipIFFT,
         opFFT_unnormalized, ipFFT_unnormalized, opBFFT_unnormalized, ipBFFT_unnormalized,
         terms, symmetries)
     @assert length(kpoints) == length(kweights)
 
     # Instantiate the terms with the basis
-    for (it, t) in enumerate(model.term_types)
-        term_name = string(nameof(typeof(t)))
-        @timing "Instantiation $term_name" basis.terms[it] = t(basis)
+    if instantiate_terms
```
antoine-levitt, line 230: Hm that looks like a potential trap. What's the motivation here? To create dummy (non-functional) pwbasis? In which case can't you pass it a dummy model with no terms?

src/PlaneWaveBasis.jl
```diff
@@ -470,3 +480,65 @@ function PlaneWaveBasis(basis::PlaneWaveBasis; use_symmetry)
...
 end
+
+
+"""
+Gather the distributed k-Point data on the master process and return
+it as a `PlaneWaveBasis`. On the other (non-master) processes `nothing` is returned.
+The returned object should not be used for computations and only to extract data
+for post-processing and serialisation to disk.
+"""
+function mpi_kgather(basis::PlaneWaveBasis)
```
antoine-levitt, line 491: I'd rather we keep the mpi_* for actual MPI methods. `gather_kpts` (consistent with comm_kpts) or something?

src/PlaneWaveBasis.jl
```diff
@@ -470,3 +480,65 @@ function PlaneWaveBasis(basis::PlaneWaveBasis; use_symmetry)
...
+    # No need to allocate and setup a new basis object
+    mpi_nprocs(basis.comm_kpts) == 1 && return basis
+
+    # Gather k-Point info on master
+    kcoords = getproperty.(basis.kpoints, :coordinate)
+    kcoords = mpi_kgather(kcoords, basis)
+    ksymops = mpi_kgather(basis.ksymops, basis)
+
+    # Number of distinct k-Point coordinates is number of k-Points with spin 1
+    n_spinup_thisproc = count(isequal(1), getproperty.(basis.kpoints, :spin))
```
antoine-levitt, line 501: feels a bit convoluted with the count and the getproperty. Wouldn't a comprehension be clearer here?

src/PlaneWaveBasis.jl
antoine-levitt: This file is much too long already, would it make sense to move these methods elsewhere?

src/PlaneWaveBasis.jl
antoine-levitt: Note that in this substantial work is still being done to initialize the basis vectors

src/PlaneWaveBasis.jl
```diff
@@ -470,3 +480,65 @@ function PlaneWaveBasis(basis::PlaneWaveBasis; use_symmetry)
...
+    else
+        PlaneWaveBasis(basis.model,
+                       basis.Ecut,
+                       kcoords[1:n_kcoords],
+                       ksymops[1:n_kcoords],
+                       basis.symmetries;
+                       fft_size=basis.fft_size,
+                       kgrid=basis.kgrid,
+                       kshift=basis.kshift,
+                       variational=true,  # TODO hard-coded
```
antoine-levitt, line 515: assert the incoming pwbasis was variational

src/PlaneWaveBasis.jl
```diff
@@ -470,3 +480,65 @@ function PlaneWaveBasis(basis::PlaneWaveBasis; use_symmetry)
...
+and return it. On the other (non-master) processes `nothing` is returned.
+"""
+function mpi_kgather(data::AbstractArray, basis::PlaneWaveBasis)
+    master = tag = 0
+    n_kpts = sum(length, basis.krange_allprocs)
+
+    if MPI.Comm_rank(basis.comm_kpts) == master
+        allk_data = similar(data, n_kpts)
+        allk_data[basis.krange_allprocs[1]] = data
+        for rank in 1:mpi_nprocs(basis.comm_kpts) - 1  # Note: MPI ranks are 0-based
```
antoine-levitt, line 534: Don't you want to use MPI.gather here?

antoine-levitt: (individual sends don't scale very well compared to a gather)

src/external/jld2io.jl
```diff
@@ -22,22 +22,23 @@ function ScfSaveCheckpoints(filename="dftk_scf_checkpoint.jld2"; keep=false, ove
...
-        jld[string(symbol)] = getproperty(scfres, symbol)
-    end
-
-    jld
-end
-function save_scfres(file::AbstractString, scfres::NamedTuple, format::Val{:jld2})
+function save_scfres_master(file::AbstractString, scfres::NamedTuple, ::Val{:jld2})
+    # Notice: This function is only meaningful on an scfres which has gathered
+    # all k-Point information locally. So before using this in MPI-distributed calculations,
+    # ensure to call `mpi_kgather_scfres` on the scfres.
```
antoine-levitt, line 29: assert that this rank is master maybe?

src/external/jld2io.jl
```diff
@@ -93,16 +93,16 @@ JLD2.writeas(::Type{PlaneWaveBasis{T}}) where {T} = PlaneWaveBasisSerialisation{
 
 function Base.convert(::Type{PlaneWaveBasisSerialisation{T}},
                       basis::PlaneWaveBasis{T}) where {T}
-    if mpi_nprocs() > 1
-        error("JLD2 serialisation for PlaneWaveBasis only implemented for non-MPI calculations for now.")
-    end
-    n_kcoords = div(length(basis.kpoints), basis.model.n_spin_components)
+    # Notice: This function is only meaningful on a basis which has gathered
+    # all k-Point information locally. So before using this in MPI-distributed calculations,
+    # ensure to call `mpi_kgather` on the PlaneWaveBasis object.
```
antoine-levitt, line 98: same as above

src/external/jld2io.jl
```diff
@@ -116,5 +116,5 @@ function Base.convert(::Type{PlaneWaveBasis{T}},
     PlaneWaveBasis(serial.model, serial.Ecut, serial.kcoords,
                    serial.ksymops, serial.symmetries;
                    fft_size=serial.fft_size,
-                   kgrid=serial.kgrid, kshift=serial.kshift)
+                   kgrid=serial.kgrid, kshift=serial.kshift, variational=true)
```
antoine-levitt, line 119: also assert?

src/external/vtkio.jl
```diff
@@ -5,49 +5,55 @@ import WriteVTK: vtk_grid, vtk_save
 
 The function takes in the VTK filename and the scfres structure and stores into a VTK file.
 
-Grid Values:
+Parameters
+- `eigenvectors`: Store the eigenvectors or not. By default they are not stored.
```
antoine-levitt, line 9: we don't call them eigenvectors anywhere else do we?

src/external/vtkio.jl
```diff
@@ -5,49 +5,55 @@ import WriteVTK: vtk_grid, vtk_save
...
 
 MetaData:
-Energy , EigenValues, Fermi Level and occupation.
+- energies, eigenvalues, Fermi level and occupations.
 """
-function save_scfres(filename::AbstractString, scfres::NamedTuple, format::Val{:vts})
-    if mpi_nprocs() > 1
-        error("Saving VTK file with MPI is not supported yet.")
-    end
+function save_scfres_master(filename::AbstractString, scfres::NamedTuple, ::Val{:vts};
```
antoine-levitt, line 20: why vts and not vtk?

src/PlaneWaveBasis.jl
mfherbst: That's not easy to do, but I thought about adding a "variational" flag to pwbasis anyway.

src/PlaneWaveBasis.jl
mfherbst: I would love to, but it does not work, because you need an isbits type.

src/external/vtkio.jl
mfherbst: because that's the extension.

src/external/vtkio.jl
mfherbst: Yes whatever.

src/PlaneWaveBasis.jl
mfherbst: > can't you pass it a dummy model with no terms?

Not easily. I'd have to create a model from scratch, which I wanted to avoid to reduce coupling between the implementation details of the Model struct and this basis.

 > Note that in this substantial work is still being done to initialize the basis vectors

Yes maybe that optimisation is not worth it, I'm not sure. I wanted to avoid it because if you do checkpointing than this method will be run every SCF step. OTOH probably the communication to gather the k-Points is anyway dominating in this case and the initialisation does not really matter. You generally have a better intuition about this. If you think it's better to avoid this complication and fully initialise I don't mind.

src/PlaneWaveBasis.jl
mfherbst: gather_kpts is good. I'll think about the splitup.

src/PlaneWaveBasis.jl
antoine-levitt: Worth opening an issue at MPI, or converting to a canonical plain array? We probably need that at other places too.

src/PlaneWaveBasis.jl
antoine-levitt: I would say let's not optimize now and let's see if it matters afterwards. If you can't afford term creation, you possibly can't afford basis creation anyway

src/PlaneWaveBasis.jl
mfherbst: Does not work, because they are of different size for each k-Point.

Also I can contact the MPI folks, but I understand that they don't want to do it, because serialisation of arbitrary objects may lead to variable sizes on each process, that you need to communicate first before you can do a gatherv. So an MPI gather comes down to two gathers, which is perhaps surprising. Also it leads to copying quite a bit of data. So to me that is something that should be handled in the layer above MPI.jl.

antoine-levitt: OK. Note there is a generalized MPI gather (gatherv) where you take into account different sizes.

mfherbst: Yes but you still need isbits types. The problem is not just that you have different number of k-Points on each processor it's also that the \psi-Arrays (it's an array of arrays remember) themselves are of different size.

src/PlaneWaveBasis.jl
mfherbst: ok

PR #413, opened 2021-02-20
Remove RealFourierArray and refactor internal density datastructures

antoine-levitt:
Minimal working prototype of the API I have in mind. @mfherbst does that look like the right thing to do to you?


antoine-levitt:
After playing with it and looking at the SCF/mixing code it's clearly the right thing to do, so I'm going ahead with it.

antoine-levitt:
TODO list:

- [ ] interpolation
- [ ] external
- [ ] plotting
- [ ] examples
- [ ] tests
- [ ] postprocess/χ0
- [ ] postprocess/current
- [ ] terms/anyonic
- [ ] compute_kernels
- [ ] xc pass ρ directly
- [ ] kerker/chi0 mixing

antoine-levitt:
I'm adapting the code directly in chi0mixing, but I'm not quite sure the handling of the DC component is correct there in the case of non-RPA...

antoine-levitt:
Alright, this should be fully done now! The tests that still fail are tolerance problems. I don't think I changed anything in that respect (esp the convergence criteria could be affected by this changed, but I double checked), so let's just increase the tolerances (substantially, so that we don't need to change them every week)

antoine-levitt:
OK, this is done and ready for review. That was not so pleasant to do but I'm quite happy with how it turned out. In particular I deleted quite a bit of code, which is always a nice thing to do. We should have done this when we did the spin, but we had other things to do then.

Things to discuss:

- handling of DC component
- tolerances
- perf implications (should maybe do a quick benchmark to check I haven't done anything completely stupid)

antoine-levitt:
OK, this is done and ready for review. That was not so pleasant to do but I'm quite happy with how it turned out. In particular I deleted quite a bit of code, which is always a nice thing to do. We should have done this when we did the spin, but we had other things to do then.

Things to discuss:

- handling of DC component
- tolerances
- perf implications (should maybe do a quick benchmark to check I haven't done anything completely stupid)

mfherbst:
Ok, cool thanks for doing this. I'll try to take a look. Looks like I'm pretty tied in this week with setup etc, so no promises.

src/densities.jl
```diff
@@ -76,33 +79,42 @@ is not collinear the spin density is `nothing`.
...
         end
     end
 
     # Count the number of k-points modulo spin
     count = sum(length(basis.ksymops[ik]) for ik in 1:length(basis.kpoints)) ÷ n_spin
     count = mpi_sum(count, basis.comm_kpts)
-    ρs = [sum(getindex.(ρaccus, iσ)) / count for iσ in 1:n_spin]
-    mpi_sum!.(ρs, Ref(basis.comm_kpts))
+    ρ = sum(ρaccus) ./ count
+    mpi_sum!(ρ, basis.comm_kpts)
```
mfherbst, line 97: Why not put than in one line?

src/densities.jl
```diff
@@ -76,33 +79,42 @@ is not collinear the spin density is `nothing`.
...
-        ρtot  = from_fourier(basis, ρs[1] + ρs[2])  # up + down
-        ρspin = from_fourier(basis, ρs[1] - ρs[2])  # up - down
+total_density(ρ) = dropdims(sum(ρ; dims=4); dims=4)
+@views function spin_density(ρ)
+    if size(ρ, 4) == 2
+        ρ[:, :, :, 1] - ρ[:, :, :, 2]
     else
-        ρtot  = from_fourier(basis, ρs[1])
-        ρspin = nothing
+        0*ρ[:, :, :]
```
mfherbst, line 106: `false` or `falses(size)`?

src/external/etsf_nanoquanta.jl
```diff
@@ -164,14 +164,14 @@ function load_density(T, folder::EtsfFolder; magnetic_moments=[])
         ρtot_real = Array{T}(folder.den["density"][1, :, :, :, 1])
         @assert basis.fft_size == size(ρ_real)
 
-        return from_real(basis, ρtot_real), nothing
+        return ρ_from_total_and_spin(ρtot_real)
```
mfherbst, line 167: This is a bug: Second argument missing.

src/densities.jl
```diff
@@ -76,33 +79,42 @@ is not collinear the spin density is `nothing`.
...
     end
+end
 
-    ρtot, ρspin
+function ρ_from_total_and_spin(ρtot, ρspin)
+    n_spin = ρspin === nothing ? 1 : 2
+    ρ = similar(ρtot, size(ρtot)..., n_spin)
+    if n_spin == 1
+        ρ .= ρtot
```
mfherbst, line 114: Why not just reshape `ρtot`?

src/densities.jl
```diff
@@ -76,33 +79,42 @@ is not collinear the spin density is `nothing`.
...
+        ρ[:, :, :, 1] - ρ[:, :, :, 2]
     else
-        ρtot  = from_fourier(basis, ρs[1])
-        ρspin = nothing
+        0*ρ[:, :, :]
     end
+end
 
-    ρtot, ρspin
+function ρ_from_total_and_spin(ρtot, ρspin)
```
mfherbst, line 110: Why not just `from_total_and_spin`?

src/densities.jl
```diff
@@ -76,33 +79,42 @@ is not collinear the spin density is `nothing`.
...
+    mpi_sum!(ρ, basis.comm_kpts)
+    G_to_r(basis, ρ)
+end
 
-    @assert basis.model.spin_polarization in (:none, :spinless, :collinear)
-    if basis.model.spin_polarization == :collinear
-        ρtot  = from_fourier(basis, ρs[1] + ρs[2])  # up + down
-        ρspin = from_fourier(basis, ρs[1] - ρs[2])  # up - down
+total_density(ρ) = dropdims(sum(ρ; dims=4); dims=4)
+@views function spin_density(ρ)
```
mfherbst, line 102: These are not just useful for the density, but also for the potential, so it would be nice to see that reflected in the names, so how about `compute_total` and `compute_spin`? Also almost always you need both of them if you need one. Maybe we should just have a function `split_total_and_spin` or `compute_total_and_spin` to reflect that?

Also at least in theory these definitions change once we go beyond collinear spin ... I don't think we should worry about that now, but just to keep that in mind. That's amongst the reasons why I would go beyond a plain array, but instead a special type for spin-component arrays. Then you can just put a reference to a basis in the SpinArray and define total(::SpinArray) and spin(::SpinArray) and all is well. But of course we can still do that later when we need it.

src/interpolation.jl
```diff
@@ -5,10 +5,9 @@ Interpolate a function expressed in a basis `b_in` to a basis `b_out`
 This interpolation uses a very basic real-space algorithm, and makes
 a DWIM-y attempt to take into account the fact that b_out can be a supercell of b_in
 """
-function interpolate_density(ρ_in::RealFourierArray, b_out::PlaneWaveBasis)
-    ρ_out = interpolate_density(ρ_in.real, ρ_in.basis.fft_size, b_out.fft_size,
-                                ρ_in.basis.model.lattice, b_out.model.lattice)
-    from_real(b_out, ρ_out)
+function interpolate_density(ρ_in, b_in::PlaneWaveBasis, b_out::PlaneWaveBasis)
```
mfherbst, line 8: `basis_in` and `basis_out`?

src/guess_density.jl
```diff
@@ -1,32 +1,35 @@
...
+function guess_density(basis::PlaneWaveBasis, magnetic_moments=[])
+    guess_density(basis, basis.model.atoms, magnetic_moments)
 end
+@timing function guess_density(basis::PlaneWaveBasis{T}, atoms, magnetic_moments) where {T}
+    # build ρtot
+    gaussians_tot = [(T(n_elec_valence(spec)), T(atom_decay_length(spec)), pos)
+                     for (spec, positions) in atoms for pos in positions]
+    ρtot = gaussian_superposition(basis, gaussians_tot)
+
+    ρ = zeros(T, basis.fft_size..., basis.model.n_spin_components)
```
mfherbst, line 22: Unused

src/guess_density.jl
```diff
@@ -1,32 +1,35 @@
...
-guess_density(basis::PlaneWaveBasis) = guess_density(basis, basis.model.atoms)
-@timing function guess_density(basis::PlaneWaveBasis{T}, atoms) where {T}
-    # Return sum of atomic densities with all weights equal to 1
-    gaussians = [(T(n_elec_valence(spec)), T(atom_decay_length(spec)), pos)
-                 for (spec, positions) in atoms for pos in positions]
-    gaussian_superposition(basis, gaussians)
+function guess_density(basis::PlaneWaveBasis, magnetic_moments=[])
+    guess_density(basis, basis.model.atoms, magnetic_moments)
 end
+@timing function guess_density(basis::PlaneWaveBasis{T}, atoms, magnetic_moments) where {T}
```
mfherbst, line 16: I find this function a bit convoluted. I think what I would do is have a low-level internal function for the total density and one for the spin density and then `guess_density` just calls both and combines the result.

examples/scf_checkpoints.jl
```diff
@@ -81,15 +80,15 @@ loaded.energies
 # callback to [`self_consistent_field`](@ref), for example:
 
 callback = DFTK.ScfSaveCheckpoints()
-scfres = self_consistent_field(basis, tol=1e-2, ρspin=ρspin, callback=callback);
+scfres = self_consistent_field(basis; ρ=guess_density(basis, magnetic_moments), tol=1e-2, callback=callback);
```
mfherbst, line 83: Line too long.

examples/scf_checkpoints.jl
```diff
@@ -81,15 +80,15 @@ loaded.energies
...
 
 # Notice that using this callback makes the SCF go silent since the passed
 # callback parameter overwrites the default value (namely `DefaultScfCallback()`)
 # which exactly gives the familiar printing of the SCF convergence.
 # If you want to have both (printing and checkpointing) you need to chain
 # both callbacks:
 callback = DFTK.ScfDefaultCallback() ∘ DFTK.ScfSaveCheckpoints(keep=true)
-scfres = self_consistent_field(basis, tol=1e-2, ρspin=ρspin, callback=callback);
+scfres = self_consistent_field(basis; ρ=guess_density(basis, magnetic_moments), tol=1e-2, callback=callback);
```
mfherbst, line 91: Same here

src/postprocess/dos.jl
```diff
@@ -23,8 +23,8 @@ and temperature `T`. It increases with both `T` and better sampling of the BZ wi
 Fermi surface.
 """
 function compute_nos(ε, basis, eigenvalues; smearing=basis.model.smearing,
-                     temperature=basis.model.temperature, spins=1:basis.model.n_spin_components)
-    N = zero(ε)
+                     temperature=basis.model.temperature)
+    D = zeros(typeof(ε), basis.model.n_spin_components)
```
mfherbst, line 27: Typo

mfherbst (commented):
Here is my first set of comments. I did not manage to go through everything for now.

src/PlaneWaveBasis.jl
```diff
@@ -77,6 +77,8 @@ struct PlaneWaveBasis{T <: Real}
     # fft_size defines both the G basis on which densities and
     # potentials are expanded, and the real-space grid
     fft_size::Tuple{Int, Int, Int}
+    # factor for integrals in real space: sum(ρ) * integration_factor ~ ∫ρ
+    integration_factor::T  # = model.unit_cell_volume ./ prod(fft_size)
```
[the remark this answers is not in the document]
antoine-levitt: I'm not sure about dVol: it doesn't match our usual naming (should be d_vol?), and the name is not super clear. I left integration_factor for easy search/replace if we want to. I'm open to changing it. d_vol, dVol, dx, dr, d_omega?

src/PlaneWaveBasis.jl
```diff
@@ -13,9 +13,9 @@ More generally, a kpoint is a block of the Hamiltonian;
 eg collinear spin is treated by doubling the number of kpoints.
 """
 struct Kpoint{T <: Real}
-    model::Model{T}               # TODO Should be only lattice/atoms
+    model::Model{T}
     spin::Int                     # Spin component can be 1 or 2 as index into what is
-    #                             # returned by the `spin_components` function
+                                  # returned by the `spin_components` function
```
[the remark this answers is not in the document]
antoine-levitt: fix your editor? :-p I removed it for consistency with mapping and mapping_inv which don't have it

src/PlaneWaveBasis.jl
```diff
@@ -400,8 +404,15 @@ Perform an iFFT to obtain the quantity defined by `f_fourier` defined
 on the k-dependent spherical basis set (if `kpt` is given) or the
 k-independent cubic (if it is not) on the real-space grid.
 """
-function G_to_r(basis::PlaneWaveBasis, f_fourier::AbstractArray3)
-    G_to_r!(similar(f_fourier), basis, f_fourier)
+function G_to_r(basis::PlaneWaveBasis, f_fourier::AbstractArray; assume_real=true)
```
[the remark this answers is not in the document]
antoine-levitt: The problem is that this is used quite a bit, and if you forget it you get complex numbers everywhere. I'm not comfortable with the current solution either, but also at some point we might want to implement the optimization of symmetries in the Fourier transform of real functions and change this anyway... so not sure what to do here.

src/PlaneWaveBasis.jl
```diff
@@ -416,6 +427,9 @@ NOTE: If `kpt` is given, not only `f_fourier` but also `f_real` is overwritten.
 """
 @timing_seq function r_to_G!(f_fourier::AbstractArray3, basis::PlaneWaveBasis,
                              f_real::AbstractArray3)
+    if eltype(f_real) != complex(eltype(f_real))
```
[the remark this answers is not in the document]
antoine-levitt: cool, thanks!

src/densities.jl
```diff
@@ -44,7 +47,7 @@ grid `basis`, where the individual k-Points are occupied according to `occupatio
 `ψ` should be one coefficient matrix per k-Point. If the `Model` underlying the basis
 is not collinear the spin density is `nothing`.
 """
-@timing function compute_density(basis::PlaneWaveBasis, ψ, occupation)
+@views @timing function compute_density(basis::PlaneWaveBasis, ψ, occupation)
```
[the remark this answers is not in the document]
antoine-levitt: Actually I wondered if we could put a `@views` in front of *the whole of DFTK*: can we do that? We very very rarely (never?) use non-views slices, and when we do it's because we forgot to put a `@views`. The fact that julia's slices copy by default is a design mistake that was done because views used to be slow, and I hope they fix it for 2.0. So I'd rather put a `@views` in front of every function that uses slices and forget about it.

src/densities.jl
antoine-levitt: You mean merge with the next line? I like to keep MPI calls on separate lines so it's easy to ignore if you don't want to parse MPI mentally.

src/external/etsf_nanoquanta.jl
antoine-levitt: Indeed, well spotted. Should be fixed by switching to 
```
function ρ_from_total_and_spin(ρtot, ρspin=nothing)
```

src/densities.jl
antoine-levitt: To avoid aliasing and guarantee type stability

src/densities.jl
antoine-levitt: how do you know you're talking about densities and not wf/potentials?

src/densities.jl
antoine-levitt: Yes, but we manipulate densities in this 4D format a lot more than potentials, at least for now, so I kept it density specific for clarity. Yes I know non-collinear spin doesn't fit this API, but at least as a first non-optimized version we can do 5D (complex) densities [i,j,k,alpha,beta]: let's keep things simple for now and we'll see when we get to it.

src/guess_density.jl
antoine-levitt: good catch

src/PlaneWaveBasis.jl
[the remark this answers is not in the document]
mfherbst: I know this is inconsintent at the moment, but I can't be bothered to do a cleanup. Generally putting the hash earlier seems to break less stuff.

src/PlaneWaveBasis.jl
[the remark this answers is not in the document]
mfherbst: Yeah the thing about forgetting the postfix is true. The other option is to postfix the complex version, but that feels kind of weird. I suppose your solution is indeed the least problematic then.

src/densities.jl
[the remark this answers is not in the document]
mfherbst: I'm not sure about that. I had cases (could be in older Julia versions) where @views caused really subtle issues - both coding-wise and performance-wise, so I tend to use `@views` rather locally.
So definitely I'm against using `@views` for the full DFTK (can probably be done by putting it in front of the module definition) as that would be rather unexpected I'd say.

src/densities.jl
mfherbst: That's kind of the point. It should work for densities and potentials the same way, shouldn't it?

src/densities.jl
mfherbst: Ah you only refactored the densities and not the potentials as well. I have not read enough of the code to see that. But still ... perhaps go with the general names. What happens should be clear from context (we pretty much always use `ρ` or `V` as part of the variable names).

test/serialisation.jl
```diff
@@ -32,8 +32,7 @@ function test_scfres_agreement(tested, ref)
     @test tested.eigenvalues    == ref.eigenvalues
     @test tested.occupation     == ref.occupation
     @test tested.ψ              == ref.ψ
-    @test tested.ρ.real         ≈  ref.ρ.real     rtol=1e-14
-    @test tested.ρspin.real     ≈  ref.ρspin.real rtol=1e-14
+    @test tested.ρ         ≈  ref.ρ     rtol=1e-14
```
mfherbst, line 35: Approx not aligned.

test/silicon_lda.jl
```diff
@@ -18,13 +18,14 @@ function run_silicon_lda(T ;Ecut=5, grid_size=15, spin_polarization=:none, kwarg
 
     fft_size = grid_size * ones(3)
     Si = ElementPsp(silicon.atnum, psp=load_psp(silicon.atnum, functional="lda", family="hgh"))
-    magmoms = spin_polarization == :collinear ? [Si => zeros(2)] : []
     model = model_DFT(Array{T}(silicon.lattice), [Si => silicon.positions], [:lda_x, :lda_c_vwn],
-                      spin_polarization=spin_polarization, magnetic_moments=magmoms)
```
mfherbst, line 23: Are you sure that makes no difference?

test/silicon_pbe.jl
```diff
@@ -24,11 +24,13 @@ function run_silicon_pbe(T ;Ecut=5, grid_size=15, spin_polarization=:none, kwarg
     Si = ElementPsp(silicon.atnum, psp=load_psp(silicon.atnum, functional="pbe", family="hgh"))
     magmoms = spin_polarization == :collinear ? [Si => zeros(2)] : []
     model = model_DFT(Array{T}(silicon.lattice), [Si => silicon.positions], [:gga_x_pbe, :gga_c_pbe],
-                      spin_polarization=spin_polarization, magnetic_moments=magmoms)
```
mfherbst, line 27: Same here ... also if not needed remove the magmoms variable.

src/guess_density.jl
```diff
@@ -1,32 +1,37 @@
...
 
 Build a superposition of atomic densities (SAD) guess density.
 
 We take for the guess density a gaussian centered around the atom, of
 length specified by `atom_decay_length`, normalized to get the right number of electrons
 ```math
 \hat{ρ}(G) = Z \exp\left(-(2π \text{length} |G|)^2\right)
-```
+
+When magnetic moments are provided, construct a symmetry-broken density guess. The magnetic moments should be specified in units of ``μ_B``.
```
mfherbst, line 11: Line break.

test/silicon_redHF.jl
```diff
@@ -27,12 +27,13 @@ function run_silicon_redHF(T; Ecut=5, grid_size=15, spin_polarization=:none, kwa
     fft_size = grid_size * ones(3)
     Si = ElementPsp(silicon.atnum, psp=load_psp(silicon.atnum, functional="lda", family="hgh"))
     model = model_DFT(Array{T}(silicon.lattice), [Si => silicon.positions], [];
-                      temperature=0.05, spin_polarization=spin_polarization,
-                      magnetic_moments=[Si => zeros(2)])
```
mfherbst, line 31: magmoms thing.

test/bzmesh_symmetry.jl
```diff
@@ -26,6 +26,6 @@ include("testcases.jl")
         E2 = scfres.energies.total
 
         @test abs(E1 - E2) < 1e-10
-        @test norm(ρ1.fourier - ρ2.fourier) < 1e-8
+        @test norm(ρ1 - ρ2) .* sqrt(basis.integration_factor) < 1e-8
```
mfherbst, line 29: Why?

mfherbst (commented):
Sorry about the slow progress ... the next lot I got to.

src/PlaneWaveBasis.jl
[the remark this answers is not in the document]
antoine-levitt: I find it pretty ugly... can't we just harmonize everything to aligning the hashes?

src/PlaneWaveBasis.jl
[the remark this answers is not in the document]
antoine-levitt: OK let's do dvol

test/bzmesh_symmetry.jl
antoine-levitt: because that's what equivalent to the one above

test/silicon_lda.jl
antoine-levitt: Well it works, so it must be OK :-p

src/postprocess/dos.jl
```diff
@@ -23,8 +23,8 @@ and temperature `T`. It increases with both `T` and better sampling of the BZ wi
 Fermi surface.
 """
 function compute_nos(ε, basis, eigenvalues; smearing=basis.model.smearing,
-                     temperature=basis.model.temperature, spins=1:basis.model.n_spin_components)
-    N = zero(ε)
+                     temperature=basis.model.temperature)
+    N = zeros(typeof(ε), basis.model.n_spin_components)
```
mfherbst, line 27: If you change this you certainly also have to change the way things are summed into the N

src/scf/mixing.jl
```diff
@@ -85,16 +89,16 @@ from the current density of states at the Fermi level.
...
+        return mix(SimpleMixing(α=mixing.α), basis, δF)
     else
         n_spin = basis.model.n_spin_components
         Ω = basis.model.unit_cell_volume
-        dos  = [compute_dos(εF, basis, eigenvalues, spins=[σ]) / Ω for σ in 1:n_spin]
+        dos  = compute_dos(εF, basis, eigenvalues)
         kTF  = sqrt(4π * sum(dos))
-        ΔDOS = n_spin == 2 ? dos[1] - dos[2] : 0.0
-        mix(KerkerMixing(α=mixing.α, kTF=kTF, ΔDOS=ΔDOS), basis, δF, δFspin)
+        ΔDOS = n_spin == 2 ? dos[1]-dos[2] : 0.0
```
mfherbst, line 100: you really don't like spaces do you?

src/scf/mixing.jl
```diff
@@ -173,65 +174,39 @@ real space using a GMRES. Either the full kernel (`RPA=false`) or only the Hartr
...
+        δV = apply_kernel(basis, δρ; ρ=ρin, RPA=mixing.RPA)
+        δV .-= mean(δV)
+        Jδρ = copy(δρ)
         for apply_term! in χ0applies
-            apply_term!(JδF, δV, -1)  # JδF .-= χ0 * δV
+            apply_term!(Jδρ, δV, -1)  # JδF .-= χ0 * δV
         end
-        vec(JδF .-= mean(JδF))  # Zero DC component in total density response
+        Jδρ .-= mean(Jδρ)
+        vec(Jδρ)
```
mfherbst, line 201: why not keep it as one line?

src/scf/mixing.jl
```diff
@@ -173,65 +174,39 @@ real space using a GMRES. Either the full kernel (`RPA=false`) or only the Hartr
...
-    δρ .+= (sum(δF_tot.real) - sum(δρ)) / length(δF_tot)  # Set DC from δF
-
-    @views if n_spin == 1
-        from_real(basis, δρ[:, :, :, 1]), nothing
-    else
-        (from_real(basis, δρ[:, :, :, 1] .+ δρ[:, :, :, 2]),
-         from_real(basis, δρ[:, :, :, 1] .- δρ[:, :, :, 2]))
-    end
+    # Set DC from δF
+    δρ .+= mean(δF)
```
mfherbst, line 210: This won't work as you just removed the zero DC from δF. You need to store the DC and restore it here.

src/scf/scf_callbacks.jl
```diff
@@ -25,8 +24,10 @@ function ScfDefaultCallback()
             @printf "---   ---------------   ---------   --------%s   ----\n" magn[2]
         end
         E    = isnothing(info.energies) ? Inf : info.energies.total
-        Δρ   = norm(info.ρout.fourier - info.ρin.fourier)
-        magn = isnothing(info.ρ_spin_out) ? NaN : sum(info.ρ_spin_out.real) * dVol
+        Δρ   = norm(info.ρout - info.ρin) * sqrt(info.basis.dvol)
+        magn = size(info.ρout, 4) == 1 ?
+            NaN :
+            sum(spin_density(info.ρout)) * info.basis.dvol
```
mfherbst, line 30: Not a fan of this. In such case I prefer if ... else ... end

src/scf/scf_callbacks.jl
```diff
@@ -65,7 +68,10 @@ Flag convergence by using the L2Norm of the change between
 input density and unpreconditioned output density (ρout)
 """
 function ScfConvergenceDensity(tolerance)
-    info -> norm(info.ρout.fourier - info.ρin.fourier) < tolerance
+    function f(info)
+        norm(info.ρout - info.ρin) * sqrt(info.basis.dvol) < tolerance
+    end
+    f
```
mfherbst, line 74: Really? Come on that's four lines for something that is super clear in my opinion in one line.

src/terms/terms.jl
```diff
@@ -137,31 +135,25 @@ compute_kernel(::Term; kwargs...) = nothing  # By default no kernel
 
 
 """
-    apply_kernel(basis::PlaneWaveBasis, dρ, dρspin=nothing; kwargs...)
+    apply_kernel(basis::PlaneWaveBasis, dρ; kwargs...)
 
-Computes the potential response to a perturbation `(dρ, dρspin)` in real space.
-Returns the array `[dV_α, dV_β]` for collinear spin-polarized
-calculations, else the array [dV_{tot}].
+Computes the potential response to a perturbation dρ in real space.
```
mfherbst, line 140: Specify the structure of the returned dV?

src/terms/xc.jl
```diff
@@ -118,19 +117,14 @@ function compute_kernel(term::TermXc; ρ::RealFourierArray, ρspin=nothing, kwar
...
-        K_αtot  = Diagonal(vec(fac * (Kαα + Kαβ) / 2))
-        K_αspin = Diagonal(vec(fac * (Kαα - Kαβ) / 2))
-        K_βtot  = Diagonal(vec(fac * (Kβα + Kββ) / 2))
-        K_βspin = Diagonal(vec(fac * (Kβα - Kββ) / 2))
-
-        [K_αtot K_αspin;
-         K_βtot K_βspin]
+        K = fac .* [Diagonal(vec(Kαα)) Diagonal(vec(Kαβ));
+                    Diagonal(vec(Kβα)) Diagonal(vec(Kββ))]
+        reshape(K, 2*prod(term.basis.fft_size), 2*prod(term.basis.fft_size))
```
mfherbst, line 122: Why do you need that reshape?

src/terms/xc.jl
```diff
@@ -118,19 +117,14 @@ function compute_kernel(term::TermXc; ρ::RealFourierArray, ρspin=nothing, kwar
...
     end
 end
 
-function apply_kernel(term::TermXc, dρ::RealFourierArray, dρspin=nothing;
-                      ρ::RealFourierArray, ρspin=nothing, kwargs...)
+function apply_kernel(term::TermXc, dρ;
+                      ρ, kwargs...)
```
mfherbst, line 127: remove line break

src/terms/xc.jl
```diff
@@ -259,53 +251,50 @@ function libxc_symmetric_index(i, j)
...
-
-        if max_derivative > 0
-            ρ_fourier = similar(ρ.fourier, n_spin, basis.fft_size...)
-            ρ_fourier[1, :, :, :] .= @. (ρ.fourier + ρspin.fourier) / 2
-            ρ_fourier[2, :, :, :] .= @. (ρ.fourier - ρspin.fourier) / 2
+
+    # compute ρ_real and possibly ρ_fourier
+    ρ_real = similar(ρ, n_spin, basis.fft_size...)
+    for σ = 1:n_spin
+        ρ_real[σ, :, :, :] = ρ[:, :, :, σ]
```
mfherbst, line 278: permutedims?

src/terms/xc.jl
```diff
@@ -259,53 +251,50 @@ function libxc_symmetric_index(i, j)
...
+    end
+    if max_derivative > 0
+        ρf = r_to_G(basis, ρ)
+        ρ_fourier = similar(ρf, n_spin, basis.fft_size...)
+        for σ = 1:n_spin
+            ρ_fourier[σ, :, :, :] = ρf[:, :, :, σ]
```
mfherbst, line 284: also permutedims

src/terms/xc.jl
```diff
@@ -323,6 +312,7 @@ function LibxcDensity(basis, max_derivative::Integer, ρ::RealFourierArray, ρsp
     LibxcDensity(basis, max_derivative, ρ_real, ∇ρ_real, σ_real)
 end
 
+# See libxc documentation
```
mfherbst, line 315: Not really, rather see Libxc.jl documentation (which basically does not yet exist ;) ... I should write it at some point)

test/silicon_pbe.jl
```diff
@@ -22,13 +22,14 @@ function run_silicon_pbe(T ;Ecut=5, grid_size=15, spin_polarization=:none, kwarg
...
+                      spin_polarization=spin_polarization)
     basis = PlaneWaveBasis(model, Ecut, silicon.kcoords, silicon.ksymops; fft_size=fft_size)
 
     spin_polarization == :collinear && (ref_pbe = vcat(ref_pbe, ref_pbe))
-    run_scf_and_compare(T, basis, ref_pbe, ref_etot; kwargs...)
+    run_scf_and_compare(T, basis, ref_pbe, ref_etot;
+                        ρ=guess_density(basis),
+                        kwargs...)
```
mfherbst, line 32: could merge the lines?

src/terms/hartree.jl
```diff
@@ -51,17 +53,14 @@ function compute_kernel(term::TermHartree; kwargs...)
     # Note that `real` here: if omitted, will result in high-frequency noise of even FFT grids
     K = real(G_to_r_matrix(term.basis) * Diagonal(vec(vc_G)) * r_to_G_matrix(term.basis))
 
-    # Hartree kernel is independent of spin, so to apply it to (ρtot, ρspin)^T
-    # and obtain the same contribution to Vα and Vβ the operator has the block structure
-    #     ( K 0 )
-    #     ( K 0 )
     n_spin = term.basis.model.n_spin_components
-    n_spin == 1 ? K : [K 0I; K 0I]
+    n_spin == 1 ? K : reshape([K K; K K], n_spin*length(vc_G), n_spin*length(vc_G))
```
mfherbst, line 57: why reshape?

src/terms/hartree.jl
```diff
@@ -51,17 +53,14 @@ function compute_kernel(term::TermHartree; kwargs...)
...
-function apply_kernel(term::TermHartree, dρ::RealFourierArray, dρspin=nothing; kwargs...)
+function apply_kernel(term::TermHartree, dρ; kwargs...)
     @assert term.basis.model.spin_polarization in (:none, :spinless, :collinear)
-    kernel = from_fourier(dρ.basis, term.poisson_green_coeffs .* dρ.fourier)
-    n_spin = term.basis.model.n_spin_components
-    n_spin == 1 ? (kernel, ) : (kernel, kernel)  # Hartree term does not care about spin
+    dρtot = total_density(dρ)
+    dVσ = G_to_r(term.basis, term.poisson_green_coeffs .* r_to_G(term.basis, dρtot))
+    dV = similar(dρ)
+    dV .= dVσ
```
mfherbst, line 65: Why not just return the result of G_to_r ?

src/postprocess/chi0.jl
```diff
@@ -83,10 +77,11 @@ function compute_χ0(ham; droptol=0, temperature=ham.basis.model.temperature)
             ddiff = Smearing.occupation_divided_difference
             ratio = filled_occ * ddiff(model.smearing, E[m], E[n], εF, temperature)
             (n != m) && (abs(ratio) < droptol) && continue
-            # dVol because inner products have a dVol so that |f> becomes <dVol f|
```
mfherbst, line 86: can you keep that for stupid chemists like me?

src/postprocess/chi0.jl
```diff
@@ -97,22 +92,10 @@ function compute_χ0(ham; droptol=0, temperature=ham.basis.model.temperature)
...
-    end
-
-    if n_spin == 1
-        χ0  # Nothing to do here
-    else
-        # Add α and β rows to get response in ρtot, subtract them to get response in ρspin
-        [χ0[1:n_fft, :] + χ0[n_fft+1:2n_fft, :];
-         χ0[1:n_fft, :] - χ0[n_fft+1:2n_fft, :]]
+        ldos = compute_ldos(εF, basis, Es, Vs)
+        χ0 .+= vec(ldos) .* vec(ldos)' .* basis.dvol ./ sum(dos)
```
mfherbst, line 96: this is really nice ...

src/scf/chi0models.jl
```diff
@@ -20,23 +17,19 @@ For details see Herbst, Levitt 2020 arXiv:2009.01665
...
     function apply!(δρ, δV, α=1)
-        dotldosδV = sum(dot(ldos[σ], δV[σ].real) for σ = 1:n_spin)
-        for σ in 1:n_spin
-            δρ[:, :, :, σ] .+= α .* (-ldos[σ] .* δV[σ].real
-                                     .+ ldos[σ] .* dotldosδV .* dVol ./ dos)
-        end
-        δρ
+        δεF = dot(ldos, δV) .* basis.dvol
+        δρ .+= α .* (-ldos .* δV
+                     .+ ldos .* δεF ./ dos)
```
mfherbst, line 32: one line?

src/scf/chi0models.jl
```diff
@@ -20,23 +17,19 @@ For details see Herbst, Levitt 2020 arXiv:2009.01665
...
     iszero(basis.model.temperature) && return nothing
-    ldos = [compute_ldos(εF, basis, eigenvalues, ψ, spins=[σ]) for σ in 1:n_spin]
-    if maximum(maximum(abs, ldos[σ]) for σ in 1:n_spin) < eps(eltype(basis))
+    ldos = compute_ldos(εF, basis, eigenvalues, ψ)
+    if maximum(abs, ldos) < eps(eltype(basis))
         return nothing
     end
 
-    dos = sum(sum, ldos) * dVol  # Integrate LDOS to form total DOS
+    dos = sum(sum, ldos) * basis.dvol
```
mfherbst, line 28: one tiny inconsistency I see now is that dos / ldos sometimes mean the total dos / ldos and sometimes the spin-adapted. Perhaps we should use tdos / dos and tldos / ldos to distinguish that?

mfherbst (commented):
Ok that came out quite nicely!

examples/dielectric.jl
```diff
@@ -22,44 +22,9 @@ scfres = self_consistent_field(basis, tol=1e-14)
 
 # Apply ε† = 1 - χ0 (vc + fxc)
 function eps_fun(dρ)
-    dρ = reshape(dρ, size(scfres.ρ.real))
-    dρ = from_real(basis, dρ)
     dv = apply_kernel(basis, dρ; ρ=scfres.ρ)
-    χdv = apply_χ0(scfres.ham, scfres.ψ, scfres.εF, scfres.eigenvalues, dv...)[1]
-    vec((dρ - χdv).real)
+    χdv = apply_χ0(scfres.ham, scfres.ψ, scfres.εF, scfres.eigenvalues, dv)
```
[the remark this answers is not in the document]
antoine-levitt: woops...

src/densities.jl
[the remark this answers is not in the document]
antoine-levitt: Hm, I'm not sure. It'll be more confusing in any case no?

src/postprocess/dos.jl
antoine-levitt: woops... we should have a test for this.

src/scf/mixing.jl
antoine-levitt: wasn't on purpose, probably I deleted it and added back in again, forgetting about the spaces. However this is typically a case where no spaces makes the structure a lot clearer

src/scf/mixing.jl
antoine-levitt: why make it as one line? We literally never do this

src/scf/mixing.jl
antoine-levitt: duh, of course, thanks

src/scf/scf_callbacks.jl
antoine-levitt: I didn't find it very clear but OK, I'll just add parens

src/terms/xc.jl
antoine-levitt: would be a nice way to do it but less clear: this way the index order is clear for the rest of the code

src/terms/hartree.jl
antoine-levitt: a very fair question :-))

src/postprocess/chi0.jl
antoine-levitt: oh yes!

src/densities.jl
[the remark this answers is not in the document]
mfherbst: yeah ok, let's leave it.

src/scf/mixing.jl
mfherbst: depends what your eye is used to. I find the original easier to parse.

src/terms/xc.jl
mfherbst: yes I agree that's a fair point. On the other hand if you put permutedims and a comment it's too.

src/postprocess/chi0.jl
mfherbst: If you just add the |f> and the <dVol f| in your comment one does not need to know what a "dual" is.

src/postprocess/dos.jl
antoine-levitt: Yeah I did

antoine-levitt:
I fixed one issue with Hartree kernels. The rest are just tolerance errors, can we just bump them a lot / solve the equations more accurately in the tests, so we don't have to worry about this at every PR?

mfherbst:
I'll skip a "proper" cleanup of GGA for now. I'd say we first get this merged and then do it after.

PR #422, opened 2021-04-06
Add a nice introductory DFT reference article

mfherbst:
Came up in [JOSS review](https://github.com/JuliaCon/proceedings-review/issues/69).


antoine-levitt:
Good idea!

PR #423, opened 2021-04-08
Documentation for supported input and output formats

mfherbst:


docs/src/guide/input_output.jl
```diff
@@ -0,0 +1,83 @@
...
+# using a Quantum-Espresso input file, [Fe_afm.pwi](Fe_afm.pwi).
+# From this file the lattice, atomic positions and the initial magnetisation are read.
+# For more details about calculations on magnetic systems
+# using collinear spin, see [Collinear spin and magnetic systems](@ref).
+
+using DFTK
+
+## Parse the Quantum Espresso input file
+qe_input = "Fe_afm.pwi"
+atoms    = load_atoms(qe_input)
```
antoine-levitt, line 29: say somewhere that ASE does the magic of auto-detecting it's a QE input file (but it could have been some other format)

PR #437, opened 2021-06-03
Extra forces test for magnetic ground state

mfherbst:


antoine-levitt:
Nice. Might use a bit of refactoring, but better tests with redundancy than no tests!

mfherbst:
Nah can't be bothered to refactor right now, but I see your point.

Issue #441, opened 2021-06-06
Optimize linear algebra in LOBPCG


antoine-levitt:
https://github.com/JuliaMolSim/DFTK.jl/issues/434 shows that significant time is spent in block multiplications. We can probably save some time by enhancing our bmul to support bmul! (as well as the five-arg version). But in fact can't we just use https://github.com/JuliaArrays/BlockArrays.jl? I vaguely remember discarding this solution before, but I'm not sure why, it seems like it's working?

mfherbst:
I remember there was an issue with passing the structure on to some lower-lying BLAS / LAPACK / cholesky stuff. But of course if that's working now we might as well use it.

antoine-levitt:
Maybe, but looking at their issue list I don't see anything concerning. I do remember writing on an issue tracker, but probably that was another package?

antoine-levitt:
Yup, pretty much perfect:
```
using BlockArrays
a = (randn(2,2), randn(2,2), randn(2,2))
b = randn(6, 3)
c = randn(2, 3)
mortar(a) * b
mul!(c, a, b, 1, 1)
```

antoine-levitt:
OK so we should just rewrite LOBPCG with blockarrays instead of tuples. Essentially just replace things like `Ý = (X,R)` by `Y = mortar((X,R))`, and all custom linear algebra by plain old * and mul!. The only thing we'll be missing is the custom assume_hermitian thing but that can be added back later if we really want to

antoine-levitt:
Might also be a good opportunity to finally split off that code from DFTK...

PR #442, opened 2021-06-06
[WIP] Use BlockArrays in LOBPCG

antoine-levitt:
WIP because I figured out while reading the code that it's badly broken for generalized eigenproblem - which we don't care about in DFTK, but still.
Fix https://github.com/JuliaMolSim/DFTK.jl/issues/441

### DIFF
--- a/src/eigen/lobpcg_hyper_impl.jl
+++ b/src/eigen/lobpcg_hyper_impl.jl
@@ -6,57 +6,17 @@
 vprintln(args...) = nothing
 
 using LinearAlgebra
+using BlockArrays # used for the `mortar` command which makes block matrices
 
-# In the LOBPCG we deal with a subspace (X, R, P), where X, R and P are views
-# into bigger arrays. We don't store these subspace vectors as a dense matrix
-# or a special struct, but just as a tuple.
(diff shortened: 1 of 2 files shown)


antoine-levitt:
OK, good to go.

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -1,62 +1,36 @@
...
+# - Termination criteria for the orthogonalization are based on cheap estimates instead of costly explicit checks.
+# - The default tolerances are very tight, yielding a very stable algorithm. This can be tweaked (see keyword ortho_tol) to compromise on stability for less Cholesky factorizations.
+# - Implicit product updates (reuse of matvecs) are performed whenever it is safe to do so, ie only with orthogonal transformations. An exception is the B matrix reuse, which seems to be OK even with very badly conditioned B matrices. The code is easy to modify if this is not the case.
+# - The locking is performed carefully to ensure minimal impact on the other eigenvectors (which is not the case in many - all ? - other implementations)
```
mfherbst, line 13: Line length?

antoine-levitt: Frankly I don't see the point here, but OK

PR #446, opened 2021-06-10
Interpolate

gkemlin:
Adding the possibility in interpolation.jl to interpolate a blochwave from a basis to a smaller basis (and thus keeping only low frequencies from the small basis).


antoine-levitt:
Something has gone wrong with your git :-)

gkemlin:
Yep, I just send an email to both of you, sorry :sweat_smile:

PR #447, opened 2021-06-10
interpolation from a basis to a smaller basis

gkemlin:
Adding the possibility in interpolation.jl to interpolate a blochwave from a basis to a smaller basis (and thus keeping only low frequencies from the small basis). Should be good this time :stuck_out_tongue:

### DIFF
--- a/src/DFTK.jl
+++ b/src/DFTK.jl
@@ -89,7 +89,6 @@ export spin_density
 export ρ_from_total_and_spin
 include("densities.jl")
 include("interpolation.jl")
-export transfer_blochwave
 
 export PreconditionerTPA
 export PreconditionerNone
--- a/src/DFTK.jl
+++ b/src/DFTK.jl
@@ -89,6 +89,7 @@ export spin_density
 export ρ_from_total_and_spin
 include("densities.jl")
 include("interpolation.jl")
+export transfer_blochwave
 
 export PreconditionerTPA
 export PreconditionerNone
--- a/src/DFTK.jl
+++ b/src/DFTK.jl
@@ -88,7 +88,9 @@ export total_density
 export spin_density
 export ρ_from_total_and_spin
 include("densities.jl")
-include("interpolation.jl")
+include("interpolation_transfer.jl")
+export transfer_blochwave
+export transfer_blochwave_kpt
 
 export PreconditionerTPA
 export PreconditionerNone
--- a/src/interpolation.jl
+++ b/src/interpolation.jl
@@ -93,6 +93,7 @@ function transfer_blochwave_mapping(ik, basis_in::PlaneWaveBasis{T},
                                     basis_out::PlaneWaveBasis{T}) where T
     # Get indices of the G vectors of the old basis inside the new basis.
     idcsk_out = index_G_vectors.(Ref(basis_out), G_vectors(basis_in.kpoints[ik]))
+    filter!(e -> e != nothing, idcsk_out)
     # Linearise the indices
     idcsk_out = getindex.(Ref(LinearIndices(basis_out.fft_size)), idcsk_out)
 
--- a/src/interpolation.jl
+++ b/src/interpolation.jl
@@ -98,7 +98,6 @@ function transfer_blochwave_mapping(ik, basis_in::PlaneWaveBasis{T},
     # make sure that the indices linearization works. It is not an issue to
     # filter these vectors as this can only happen if Ecut_in > Ecut_out.
     filter!(!isnothing, idcsk_out)
-    # Linearize the indices
     idcsk_out = getindex.(Ref(LinearIndices(basis_out.fft_size)), idcsk_out)
 
     # Map to the indices of the corresponding G-vectors in G_vectors(kpt_out)
--- a/src/interpolation.jl
+++ b/src/interpolation.jl
@@ -86,9 +86,24 @@ function interpolate_kpoint(data_in::AbstractVecOrMat, kpoint_in::Kpoint, kpoint
 end
 
 """
-Compute indices from basis_in that are in basis_out. The output format is an array
-idcs_out of size length(basis_in.kpoint), where idcs_out[ik] contains the indices
-of the vectors in G_vectors(basis_in) that are also present in G_vectors(basis_out).
+Compute indices of the vectors in G_vectors(basis_in[ik]) that are also present
+in G_vectors(basis_out[ik]).
+"""
+function transfer_blochwave_mapping(ik, basis_in::PlaneWaveBasis{T},
+                                    basis_out::PlaneWaveBasis{T}) where T
+    # Get indices of the G vectors of the old basis inside the new basis.
+    idcsk_out = index_G_vectors.(Ref(basis_out), G_vectors(basis_in.kpoints[ik]))
+    # Linearise the indices
+    idcsk_out = getindex.(Ref(LinearIndices(basis_out.fft_size)), idcsk_out)
+
+    # Map to the indices of the corresponding G-vectors in G_vectors(kpt_out)
+    # this array might contains some nothings if basis_out has less G_vectors
+    # than basis_in at this kpoint
+    indexin(idcsk_out, basis_out.kpoints[ik].mapping)
+end
+
+"""
+Transfer Bloch wave between two basis sets. Limited feature set.
 
 If, for some kpt ik, basis_in has less vectors than basis_out, then idcs_out[ik] is
 the array of the indices of the G_vectors from basis_in in basis_out.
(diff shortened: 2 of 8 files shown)


src/interpolation.jl
```diff
@@ -85,10 +85,33 @@ function interpolate_kpoint(data_in::AbstractVecOrMat, kpoint_in::Kpoint, kpoint
     data_out
 end
 
+"""
+Compute indices from basis_in that are in basis_out, alongside the complementary
+set of indices of basis_out that are not in basis_in.
+"""
```
antoine-levitt, line 91: makes it sound like it returns two things but it just returns one?

antoine-levitt: Also extremely unclear what this function is supposed to do. Is that two bases for the same system? (in which case add a check for that). Are you talking about the k-dependent bases or the global one? The indices are in what array?

gkemlin: > makes it sound like it returns two things but it just returns one?

woups yes, my mistake

> Also extremely unclear what this function is supposed to do. Is that two bases for the same system? (in which case add a check for that). Are you talking about the k-dependent bases or the global one? The indices are in what array?

I'll try to make it clearer.

gkemlin:
I hope it is clearer now.

src/interpolation.jl
```diff
@@ -85,35 +85,84 @@ function interpolate_kpoint(data_in::AbstractVecOrMat, kpoint_in::Kpoint, kpoint
...
+
+Otherwise, if, for some kpt ik, basis_in has more vectors than basis_out, then
+idcs_out[ik] just keep the indices of the G_vectors from basis_in that are in basis_out.
+It is then of size G_vectors(basis_out.kpoints[ik]) and the interpolation can be done with
+ψ_out[ik] .= ψ_in[ik][idcs_in[ik], :]
+
+For the moment, only PlaneWaveBasis with same lattice and kgrid are supported.
 """
-function interpolate_blochwave(ψ_in, basis_in, basis_out)
+function grid_interpolation_indices(basis_in::PlaneWaveBasis{T},
```
antoine-levitt, line 106: Do this kpoint by kpoint

src/interpolation.jl
```diff
@@ -85,35 +85,84 @@ function interpolate_kpoint(data_in::AbstractVecOrMat, kpoint_in::Kpoint, kpoint
     data_out
 end
 
-
 """
-Interpolate Bloch wave between two basis sets. Limited feature set. Currently only
-interpolation to a bigger grid (larger Ecut) on the same lattice supported.
+Compute indices from basis_in that are in basis_out. The output format is an array
```
antoine-levitt, line 89: Move a lot of this doc into comments on interpolate_blochwaev

src/interpolation.jl
```diff
@@ -85,35 +85,84 @@ function interpolate_kpoint(data_in::AbstractVecOrMat, kpoint_in::Kpoint, kpoint
...
+    end
+
+    idcs_out
+end
+
+"""
+Interpolate Bloch wave between two basis sets. Limited feature set.
+Currently, only PlaneWaveBasis with same lattice and kgrid are supported.
+"""
+function interpolate_blochwave(ψ_in, basis_in::PlaneWaveBasis{T},
```
antoine-levitt, line 140: transfer_blochwave ?

src/interpolation.jl
antoine-levitt: transfer_blochwave_mapping ?

src/interpolation.jl
```diff
@@ -85,35 +85,66 @@ function interpolate_kpoint(data_in::AbstractVecOrMat, kpoint_in::Kpoint, kpoint
...
+"""
+Compute indices of the vectors in G_vectors(basis_in[ik]) that are also present
+in G_vectors(basis_out[ik]).
+"""
+function transfer_blochwave_mapping(ik, basis_in::PlaneWaveBasis{T},
+                                    basis_out::PlaneWaveBasis{T}) where T
+    # Get indices of the G vectors of the old basis inside the new basis.
+    idcsk_out = index_G_vectors.(Ref(basis_out), G_vectors(basis_in.kpoints[ik]))
+    filter!(e -> e != nothing, idcsk_out)
+    # Linearise the indices
```
antoine-levitt, line 97: ze

src/interpolation.jl
```diff
@@ -85,35 +85,66 @@ function interpolate_kpoint(data_in::AbstractVecOrMat, kpoint_in::Kpoint, kpoint
...
 
+"""
+Compute indices of the vectors in G_vectors(basis_in[ik]) that are also present
+in G_vectors(basis_out[ik]).
+"""
+function transfer_blochwave_mapping(ik, basis_in::PlaneWaveBasis{T},
+                                    basis_out::PlaneWaveBasis{T}) where T
+    # Get indices of the G vectors of the old basis inside the new basis.
+    idcsk_out = index_G_vectors.(Ref(basis_out), G_vectors(basis_in.kpoints[ik]))
+    filter!(e -> e != nothing, idcsk_out)
```
antoine-levitt, line 96: filter!(!isnothing)

src/interpolation.jl
```diff
@@ -85,35 +85,66 @@ function interpolate_kpoint(data_in::AbstractVecOrMat, kpoint_in::Kpoint, kpoint
...
 
         # Set values
         ψk_out = similar(ψ_in[ik], length(G_vectors(kpt_out)), n_bands)
         ψk_out .= 0
-        ψk_out[idcs_kpt_out, :] .= ψ_in[ik]
+        if !any(isnothing, idcsk_out)
+            # if true, then Ecut_out >= Ecut_in and we pad with zeros
+            ψk_out[idcsk_out, :] .= ψ_in[ik]
+        else
+            # else, then Ecut_in > Ecut_out and the mappig should be done the
```
antoine-levitt, line 143: pping

src/interpolation.jl
antoine-levitt: Add a comment saying why

src/interpolation.jl
```diff
@@ -85,35 +85,69 @@ function interpolate_kpoint(data_in::AbstractVecOrMat, kpoint_in::Kpoint, kpoint
...
+function transfer_blochwave_mapping(ik, basis_in::PlaneWaveBasis{T},
+                                    basis_out::PlaneWaveBasis{T}) where T
+    # Get indices of the G vectors of the old basis inside the new basis.
+    idcsk_out = index_G_vectors.(Ref(basis_out), G_vectors(basis_in.kpoints[ik]))
+    # In the case where G_vectors(basis_in.kpoints[ik]) are biggers than vectors
+    # in the fft_size box of basis_out, we need to filter out the "nothings" to
+    # make sure that the indices linearization works. It is not an issue to
+    # filter these vectors as this can only happen if Ecut_in > Ecut_out.
+    filter!(!isnothing, idcsk_out)
+    # Linearize the indices
```
mfherbst, line 101: That comment adds no info besides the next code line. Might as well remove it (I know I had it too, but it's not really needed).

src/interpolation.jl
```diff
@@ -85,35 +85,69 @@ function interpolate_kpoint(data_in::AbstractVecOrMat, kpoint_in::Kpoint, kpoint
...
+end
 
 """
-Interpolate Bloch wave between two basis sets. Limited feature set. Currently only
-interpolation to a bigger grid (larger Ecut) on the same lattice supported.
+Transfer Bloch wave between two basis sets. Limited feature set.
+For the moment, only PlaneWaveBasis with same lattice and kgrid are supported.
 """
-function interpolate_blochwave(ψ_in, basis_in, basis_out)
+function transfer_blochwave(ψ_in, basis_in::PlaneWaveBasis{T},
```
mfherbst, line 114: I called this function `interpolate_` to resemble the similar structure to the density and kpoint functions above. Any reason why you decided to rename?

test/interpolation.jl
```diff
@@ -25,3 +28,33 @@ using LinearAlgebra
...
+@testset "Interpolation of blochwave" begin
+    tol = 1e-7
+
+    Si = ElementPsp(silicon.atnum, psp=load_psp(silicon.psp))
+    model = model_LDA(silicon.lattice, [Si => silicon.positions])
+    kgrid = [2,2,2]
+    Ecut = 5
+    basis = PlaneWaveBasis(model, Ecut; kgrid=kgrid)
+
+    # Run nlsolve without guess
```
mfherbst, line 41: Well with zero guess really ...

src/DFTK.jl
```diff
@@ -89,6 +89,7 @@ export spin_density
 export ρ_from_total_and_spin
 include("densities.jl")
 include("interpolation.jl")
+export transfer_blochwave
```
mfherbst, line 92: If we make this part of the interface than we should consider doing the same for the density for consistency. Are we sure this is now settled enough to commit to this?

src/interpolation.jl
gkemlin: because when Ecut_out < Ecut_in it's not really interpolation anymore :thinking:

src/DFTK.jl
gkemlin: No, I did it to make it easier for me to use it, but we can remove it if you prefer !

test/interpolation.jl
gkemlin: I removed it, it was some copy-paste from other part of the code, but not very useful.

src/interpolation.jl
mfherbst: Ok that's true, but the same can be the case for `interpolate_kpoint` and we still call it `interpolate`. I personally don't mind which of the names we use, but it just feels strange to have one function called `interpolate_` and one `transfer_`. I guess transfer is more general and perhaps that should be used for all functions (and thus the file renamed to `basis_transfer.jl`). Thoughts?

Actually now that I think about that `transfer_blochwave` and `interpolate_kpoint` are actually quite related up to the point where one should probably merge these two ideas, no? In principle one could have a one-k-point-only version of `transfer_blochwave`, which could take two kpoint indices and two bases. We only use `interpolate_kpoint` once, where we can easily switch the interface from being `Kpoint`-based to being `PlaneWaveBasis`-based. OTOH perhaps a  `Kpoint`-based version of `transfer_blochwave` could also be possible and might be more general.

My point here is: I feel we are reinventing an existing concept, where we should generalise what we already have that it applies to both cases.

src/DFTK.jl
mfherbst: I don't feel we are settled enough on this stuff to make it part of the public interface.

gkemlin: Ok, I remove it then ! :)

src/interpolation.jl
gkemlin: I am not really sure that I see what you mean. You suggest to include `interpolate_kpoint` in `transfer_blochwave` rather than `transfer_blochwave_mapping` when it comes to transfer data from one k-point from the input basis to the output basis ?

gkemlin: Personnaly I have no opinion about what is best :sweat_smile: as long as in the end it does what I need :stuck_out_tongue:

mfherbst: Not quite ... sorry about the confusion. No what I mean is that `interpolate_kpoint` and `transfer_blochwave` are special cases of a more general function, which transfers data from one tuple (basis, kpoint) to another tuple (basis2, kpoint2) and I think with the infrastructure built in `interpolate_kpoint` and `transfer_blochwave` it should not be too hard and worth it to directly go for this general implementation.

mfherbst: Yes sure, it's just I think it might we worth to go the extra mile now that we have these two special cases and then we don't need to worry about this any more in the future. At the very least we should unify both concepts in a function of the same name and related interface (if you don't have the time to generalise fully right now).

gkemlin: Okay, it is clearer now, thanks ! 
Indeed it is worth it to do it now so that we don't have to worry about it in the future. I'll try to do this after dealing with Antoine reviews of newton PR ;)

mfherbst: sure.I'll do another pass on that as well after you guys are done. Thanks for your work in both these PRs. Well appreciated!

gkemlin:
I tried to answer your question and changed the interface by adding a function `transfer_blochwave_kpt` that takes as arguments tuples `(basis_in, kpt_in)` and `(basis_out, kpt_out)` to make the transfer. Then, we just have to call it in our function `transfer_blochwave`.

However, I do not really understand how `interpolate_kpoint` works, and I can't remove it (it is used for guesses, and should only be used for that according to the docstring). I chosed to let it for the moment (but if I use it as a replacement for `transfer_blochwave_kpt`, tests are failing.

mfherbst:
Fair enough. Last thing left to discuss on my end is the name `interpolate_` versus `transfer_`. Once we settle on that, I'm happy to take over and try to merge the functions as I suggested.

Yes it's true that this function is mentioned in the docs of the guess extrapolation, but that procedure is anyway super crude and far from perfect, so it's perfectly fine to swap this for something else.

gkemlin:
Hi! Sorry, I was on holiday. I would personally prefer to use `transfer`, and therefore change the name of the file to `interpolation_transfer.jl` or sthg like that.

mfherbst:
ok

gkemlin:
I don't know what is wrong with the documentation generation, but everything else seems to work and to be ready for merge.

PR #450, opened 2021-06-16
Potential mixing

mfherbst:
@antoine-levitt I made a big refactoring. The code is not yet tested to work, but I'm happy to get feedback.

### DIFF
--- a/src/eigen/diag.jl
+++ b/src/eigen/diag.jl
@@ -45,8 +45,7 @@ function diagonalize_all_kblocks(eigensolver, ham::Hamiltonian, nev_per_kpoint::
 
         prec = nothing
         prec_type !== nothing && (prec = prec_type(ham.basis, kpt))
-        results[ik] = eigensolver(ham.blocks[ik],
-                                  copy(guessk);  # Copy in case eigensolver modifies guess
+        results[ik] = eigensolver(ham.blocks[ik], guessk;
                                   prec=prec, tol=tol, miniter=miniter, maxiter=maxiter,
                                   n_conv_check=n_conv_check)
 
--- a/src/eigen/diag.jl
+++ b/src/eigen/diag.jl
@@ -28,7 +28,7 @@ function diagonalize_all_kblocks(eigensolver, ham::Hamiltonian, nev_per_kpoint::
         end
         # Get guessk
         @timing "QR orthonormalization" begin
-            if guess != nothing
+            if guess !== nothing
                 # guess provided
                 guessk = guess[ik]
             elseif interpolate_kpoints && ik > 1
@@ -45,7 +45,8 @@ function diagonalize_all_kblocks(eigensolver, ham::Hamiltonian, nev_per_kpoint::
 
         prec = nothing
         prec_type !== nothing && (prec = prec_type(ham.basis, kpt))
-        results[ik] = eigensolver(ham.blocks[ik], guessk;
+        results[ik] = eigensolver(ham.blocks[ik],
+                                  copy(guessk);  # Copy in case eigensolver modifies guess
                                   prec=prec, tol=tol, miniter=miniter, maxiter=maxiter,
                                   n_conv_check=n_conv_check)
 
--- a/src/eigen/lobpcg_hyper_impl.jl
+++ b/src/eigen/lobpcg_hyper_impl.jl
@@ -83,7 +83,7 @@ normest(M) = maximum(abs.(diag(M))) + norm(M - Diagonal(diag(M)))
 # Returns the new X, the number of Cholesky factorizations algorithm, and the
 # growth factor by which small perturbations of X can have been
 # magnified
-@timing function ortho(X; tol=2eps(real(eltype(X))))
+@timing function ortho!(X; tol=2eps(real(eltype(X))))
     local R
 
     # # Uncomment for "gold standard"
@@ -169,7 +169,7 @@ function drop!(X, tol=2eps(real(eltype(X))))
 end
 
 # Find X that is orthogonal, and B-orthogonal to Y, up to a tolerance tol.
-function ortho(X, Y, BY; tol=2eps(real(eltype(X))))
+function ortho!(X, Y, BY; tol=2eps(real(eltype(X))))
     T = real(eltype(X))
     # normalize to try to cheaply improve conditioning
     Threads.@threads for i=1:size(X,2)
@@ -185,7 +185,7 @@ function ortho(X, Y, BY; tol=2eps(real(eltype(X))))
         mul!(X, Y, BYX, -one(T), one(T)) # X -= Y*BY'X
         # If the orthogonalization has produced results below 2eps, we drop them
         # This is to be able to orthogonalize eg [1;0] against [e^iθ;0],
-        # as can happen in extreme cases in the ortho(cP, cX)
+        # as can happen in extreme cases in the ortho!(cP, cX)
         dropped = drop!(X)
         if dropped != []
             @views mul!(X[:, dropped], Y, BY' * (X[:, dropped]), -one(T), one(T)) # X -= Y*BY'X
@@ -194,7 +194,7 @@ function ortho(X, Y, BY; tol=2eps(real(eltype(X))))
             push!(ninners, 0)
             break
         end
-        X, ninner, growth_factor = ortho(X, tol=tol)
+        X, ninner, growth_factor = ortho!(X, tol=tol)
         push!(ninners, ninner)
 
         # norm(BY'X) < tol && break should be the proper check, but
@@ -249,7 +249,7 @@ end
     resid_history = zeros(real(eltype(X)), M, maxiter+1)
 
     # B-orthogonalize X
-    X = ortho(X, tol=ortho_tol)[1]
+    X = ortho!(copy(X), tol=ortho_tol)[1]
     if B != I
         BX = B*X
         B_ortho!(X, BX)
@@ -366,7 +366,7 @@ end
             cP = cX .- e
             cP = cP[:, Xn_indices]
             # orthogonalize against all Xn (including newly locked)
-            cP = ortho(cP, cX, cX, tol=ortho_tol)
+            cP = ortho!(cP, cX, cX, tol=ortho_tol)
 
             # Get new P
             new_P  = array_mul( Y, cP)
@@ -423,7 +423,7 @@ end
             Z  = full_X
             BZ = full_BX
         end
-        R .= ortho(R, Z, BZ; tol=ortho_tol)
+        R = ortho!(R, Z, BZ; tol=ortho_tol)
         if B != I
             mul!(BR, B, R)
             B_ortho!(R, BR)
--- a/src/eigen/lobpcg_hyper_impl.jl
+++ b/src/eigen/lobpcg_hyper_impl.jl
@@ -366,7 +366,7 @@ end
             cP = cX .- e
             cP = cP[:, Xn_indices]
             # orthogonalize against all Xn (including newly locked)
-            cP = ortho!(cP, cX, cX, tol=ortho_tol)
+            ortho!(cP, cX, cX, tol=ortho_tol)
 
             # Get new P
             new_P  = array_mul( Y, cP)
@@ -423,7 +423,7 @@ end
             Z  = full_X
             BZ = full_BX
         end
-        R = ortho!(R, Z, BZ; tol=ortho_tol)
+        ortho!(R, Z, BZ; tol=ortho_tol)
         if B != I
             mul!(BR, B, R)
             B_ortho!(R, BR)
--- a/src/scf/mixing.jl
+++ b/src/scf/mixing.jl
@@ -3,6 +3,15 @@ using IterativeSolvers
 using Statistics
 import Base: @kwdef
 
+abstract type Mixing end
+
+# Approximation to the dielectric matrix, which is diagonal
+# (such that both ρ-based and V-based SCFs can be preconditioned identically.)
+abstract type MixingDiagonal <: Mixing end
+function mix_potential(mixing::MixingDiagonal, basis::PlaneWaveBasis, δF; kwargs...)
+    mix(mixing, basis, δF; kwargs...)
+end
+
 # Mixing rules: (ρin, ρout) => ρnext, where ρout is produced by diagonalizing the
 # Hamiltonian at ρin These define the basic fix-point iteration, that are then combined with
 # acceleration methods (eg anderson). For the mixing interface we use `δF = ρout - ρin` and
@@ -14,10 +23,11 @@ import Base: @kwdef
 # process of formulating the fixed-point and solving it; we call "mixing" only the first part
 # The notation in this file follows Herbst, Levitt arXiv:2009.01665
 
+
 @doc raw"""
 Simple mixing: ``J^{-1} ≈ 1``
 """
-struct SimpleMixing; end
+struct SimpleMixing <: MixingDiagonal; end
 mix(::SimpleMixing, ::PlaneWaveBasis, δF; kwargs...) = δF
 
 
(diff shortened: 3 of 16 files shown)


src/eigen/diag.jl
```diff
@@ -45,7 +45,8 @@ function diagonalize_all_kblocks(eigensolver, ham::Hamiltonian, nev_per_kpoint::
 
         prec = nothing
         prec_type !== nothing && (prec = prec_type(ham.basis, kpt))
-        results[ik] = eigensolver(ham.blocks[ik], guessk;
+        results[ik] = eigensolver(ham.blocks[ik],
+                                  copy(guessk);  # Copy in case eigensolver modifies guess
```
antoine-levitt, line 49: Not super happy about this copy... Can't we just enforce that the eigensolver doesn't modify?

src/scf/mixing.jl
```diff
@@ -3,6 +3,15 @@ using IterativeSolvers
 using Statistics
 import Base: @kwdef
 
+abstract type Mixing end
+
+# Approximation to the dielectric matrix, which is diagonal
```
antoine-levitt, line 8: What does it mean diagonal in this sense?

antoine-levitt: Why not just get rid of this MixingDiagonal type, have functions mix_potential and mix_density, and define `mix_potential(mixing::Mixing, basis, deltaF) = mix_density(mixing::Mixing, basis, deltaF)`? That way by default it's the same, but mixing rules can overwrite it if needed.

src/scf/mixing.jl
```diff
@@ -158,13 +173,15 @@ real space using a GMRES. Either the full kernel (`RPA=false`) or only the Hartr
...
 """
-@kwdef struct χ0Mixing
+@kwdef struct χ0Mixing <: Mixing
     RPA::Bool = true       # Use RPA, i.e. only apply the Hartree and not the XC Kernel
     χ0terms   = χ0Model[Applyχ0Model()]  # The terms to use as the model for χ0
     verbose::Bool = false  # Run the GMRES verbosely
 end
 
 @views @timing "χ0Mixing" function mix(mixing::χ0Mixing, basis, δF; ρin, kwargs...)
+    # TODO This function is plain wrong if δV is a potential difference!
```
antoine-levitt, line 183: Will not be a problem with the above suggestion.

src/scf/potential_mixing.jl
```diff
@@ -0,0 +1,327 @@
...
+function scf_quadratic_model(info, info_next; modeltol=0.1)
+    T = eltype(ρ)
+
+    # Vout    = step(V), where step(V) = (Vext + Vhxc(ρ(Vin)))
+    # Vnext   = Vin + α0 * (Anderson(Vin, P⁻¹( Vout - Vin )) - Vin)
+    # ρin     = ρ(Vin)
+    # ρnext   = ρ(Vnext)
+    # α0 * δV = α0 * (Vnext - Vin) = α0 * (Anderson(Vin, P⁻¹( Vout - Vin )) - Vin)
+    #
+    Vin   = info.Vin
```
antoine-levitt, line 69: In julia 1.7 you can simplify this to (; Vin, Vout) = info (although that's not that interesting here)

src/terms/Hamiltonian.jl
```diff
@@ -163,24 +163,34 @@ end
...
+
+    pots = map(1:n_spin) do σ
+        # Get the first Hamiltonian block of this spin component
+        # (works since all local potentials are the same)
+        block = ham.blocks[first(krange_spin(ham.basis, σ))]
+        rs = [o for o in block.optimized_operators if o isa RealSpaceMultiplication]
+        only(rs).potential
+    end
+
+    # TODO This makes a copy ... might not always be what we want
```
antoine-levitt, line 180: why?

antoine-levitt (commented):
Did not look at the algorithm in too many details yet. Line searches are inherently messy but it doesn't look that bad, so good job!

src/terms/Hamiltonian.jl
mfherbst: Because you might want to get a view?

src/scf/potential_mixing.jl
mfherbst: Yes, I saw this.

src/scf/mixing.jl
mfherbst: I thought about that, but the fact that this works is really a special case for these types of dielectric models. Ok arguably "diagonal" is probably not the best term for it, but I thought it would be better to require an active opt-in into this equality rather than an opt-out. Don't you think?

src/eigen/diag.jl
mfherbst: We probably should. That was a quick fix I needed for something else. I'll try to see if I can fix it at another level.

src/terms/Hamiltonian.jl
mfherbst: ok

src/scf/mixing.jl
antoine-levitt: Most preconditioners are really very crude things that don't care so much whether it's potentials or densities, so I think it makes sense to have the simplest possible API and let more complicated algorithms overwrite when needed

mfherbst: Ok. If it shoots us in the back some day, I'll point at you ;)

antoine-levitt: deal

mfherbst:
Reverted #442 here to be able to test larger systems, see also #452.

mfherbst:
Pretty convinced now this works as it should. @antoine-levitt  I'm ready to merge this after a review.

src/scf/potential_mixing.jl
```diff
@@ -0,0 +1,345 @@
+using Statistics
+
+# Quick and dirty Anderson implementation ... lacks important things like
+# control of the condition number of the anderson matrix.
+# Not particularly optimised. Also should be moved to NLSolve ...
+function AndersonAcceleration(;m=Inf)
```
antoine-levitt, line 6: lower case for function names

antoine-levitt: or why not make the state explicit? anderson_acceleration!(Vs, PfVs, Vn, alphan, PfVn)?

antoine-levitt: I would also make the names fully generic and use the same conventions as in scf_solvers.jl: x for iterates, f for residuals

src/scf/potential_mixing.jl
```diff
@@ -0,0 +1,345 @@
...
+        reshape(Vₙ₊₁, size(Vₙ))
+    end
+end
+
+
+"""
+Accept a step if the energy is at most increasing by `max_energy` and the residual
+is at most `max_relative_residual` times the residual in the previous step.
+"""
+function ScfAcceptImprovingStep(;max_energy_change=1e-12, max_relative_residual=0.9)
```
antoine-levitt, line 52: This pattern of `function CallbackFunction(; params) info -> do_stuff(info, params)`, which we also use in scf_callbacks, is kind of solidifying, we should think about it a bit. Hiding params (or worse, state) in closures is not so great. Alternatives are
`struct CallbackFunction params end; do_stuff(::CallbackFunction, info)` or `struct CallbackFunction params end; (::CallbackFunction)(info)`. I tend to prefer the third, thoughts?

src/scf/potential_mixing.jl
```diff
@@ -0,0 +1,345 @@
...
+        energy_change = info_next.energies.total - info.energies.total
+        relative_residual = norm(info_next.Pinv_δV) / norm(info.Pinv_δV)
+
+        # Accept if energy goes down or residual decreases
+        accept = energy_change < max_energy_change || relative_residual < max_relative_residual
+        mpi_master() && @debug "Step $(accept ? "accepted" : "discarded")" energy_change relative_residual
+        accept
+    end
+end
+function ScfAcceptStepAll()
```
antoine-levitt, line 63: Put it before the others (for API example)

src/scf/potential_mixing.jl
```diff
@@ -0,0 +1,345 @@
...
+    accept_step(info, info_next) = true
+end
+
+
+function scf_quadratic_model(info, info_next; modeltol=0.1)
```
antoine-levitt, line 68: Move comments into docstring

antoine-levitt: and clarify the purpose and return value

src/scf/potential_mixing.jl
```diff
@@ -0,0 +1,345 @@
...
+end
+
+# Adaptive damping using a quadratic model
+@kwdef struct AdaptiveDamping
+    α_init = 0.5       # The initial damping value (used as first trial value in first SCF step)
+    α_min = 0.01       # Minimal damping
+    α_max = 1.0        # Maximal damping
+    α_trial_min = 0.1  # Minimal first trial damping used in a step
+    α_trial_enhancement = 1.1  # Enhancement factor to α_trial in case a step is immediately successful
+    modeltol = 0.1     # Maximum relative error for model to be considered trustworthy
```
antoine-levitt, line 124: on what?

src/scf/potential_mixing.jl
```diff
@@ -0,0 +1,345 @@
...
+        return info_next.α
+    end
+
+    α = scf_quadratic_model(info, info_next; modeltol=damping.modeltol)
+    isnothing(α) && (α = info_next.α / 2)  # Model failed ... use heuristics
+
+    # Adjust α to stay within desired range
+    α_sign = sign(α)
+    α = clamp(abs(α), damping.α_min, damping.α_max)
+    α = min(0.95abs(info_next.α), α)  # Avoid to get stuck
```
antoine-levitt, line 139: getting

src/scf/potential_mixing.jl
```diff
@@ -0,0 +1,345 @@
...
+end
+FixedDamping() = FixedDamping(0.8)
+propose_backtrack_damping(damping::FixedDamping) = damping.α
+initial_damping(damping::FixedDamping) = damping.α
+next_trial_damping(damping::FixedDamping, info, info_next, successful) = damping.α
+
+
+# Notice: For adaptive damping to run smoothly, multiple defaults need to be changed.
+#         See the function scf_potential_mixing_adaptive for that use case.
+@timing function scf_potential_mixing(
```
antoine-levitt, line 172: very annoying that we have to duplicate all these things. We should revisit the SCF & diagonalization parameter structs at some point

src/scf/potential_mixing.jl
```diff
@@ -0,0 +1,345 @@
...
+    max_backtracks=3,  # Maximal number of backtracking line searches
+)
+    # TODO Test other mixings and lift this
+    @assert (   mixing isa SimpleMixing
+             || mixing isa KerkerMixing
+             || mixing isa KerkerDosMixing)
+    damping isa Number && (damping = FixedDamping(damping))
+
+    # Initial guess for V and ψ (if none given)
+    if ψ !== nothing
```
antoine-levitt, line 202: is there a missing branch?

antoine-levitt (commented):
Not the prettiest code in DFTK, but that's kind of by nature... It's unfortunate to have separate code for potential and density mixing, it'd be nice to unify it someday.

src/scf/potential_mixing.jl
mfherbst: I agree. Thinking about the Anderson I also lean towards 2 or 3 as better Options. I think which is better depends a bit on the use case. What I don't like about the third option in general is that it suggests the struct is only meant to do one thing. For callbacks that is, however, the standard use case, so perhaps most can be converted to that pattern without much trouble.

src/scf/potential_mixing.jl
mfherbst: Agree. I think what we will need is also a generic "do scf" routine that dispatches to the various implementations we have (newton, potential mixing, density mixing etc). I was planning of writing an overview page contrasting the implementations as well. It starts to get a bit messy.

src/scf/potential_mixing.jl
mfherbst: More a misleading comment, actually.

src/scf/potential_mixing.jl
```diff
@@ -0,0 +1,345 @@
...
+
+            info_next    = EVρ(Vnext; ψ=guess, diagtol=diagtol)
+            Pinv_δV_next = mix_potential(mixing, basis, info_next.Vout - info_next.Vin; info_next...)
+            push!(diagonalization, info_next.diagonalization)
+            info_next = merge(info_next, (α=α, diagonalization=diagonalization,
+                                          ρin=info.ρout, n_iter=n_iter,
+                                          Pinv_δV=Pinv_δV_next))
+
+            successful = accept_step(info, info_next)
+            successful = MPI.bcast(successful, 0, MPI.COMM_WORLD)  # Ensure same successful
```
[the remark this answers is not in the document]
mfherbst: One would hope so, but it turned out that without the MPI broadcasts before the branching I ended up being stuck in MPI processes waiting for each other, so I figured the easy way out would be to synchronise explicitly, which fixed it. Certainly there are better ways to do this though ... I'll add a comment.

src/scf/potential_mixing.jl
```diff
@@ -0,0 +1,345 @@
...
+            α = α_next
+        end
+
+        # Switch off acceleration in case of very bad steps
+        ΔE = info_next.energies.total - info.energies.total
+        ΔE < 0 && (ΔEdown = -max(abs(ΔE), tol))
+        if !successful && n_acceleration_off == 0
+            if abs(ΔE) > abs(ratio_failure_accel_off * ΔEdown)
+                n_acceleration_off = 2
+                if mpi_master()
```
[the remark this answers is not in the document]
mfherbst: Yes, but otherwise you get it from each process.

src/scf/potential_mixing.jl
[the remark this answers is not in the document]
antoine-levitt: I don't want to go down the rabbit hole of having explicit MPI synchronize for every flag just for safety. Ideally we should figure out where the desynchronization comes from and fix it. Ie each process should have the exact same copy of the density/potential/etc

src/scf/potential_mixing.jl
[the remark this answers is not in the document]
antoine-levitt: same as the other thing, I'd rather avoid we get a lot of mpi statements everywhere. But I guess here it's OK because you use these debug thingies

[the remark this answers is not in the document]
antoine-levitt: If it turns out we use them a lot we should wrap them

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -249,7 +249,7 @@ end
     resid_history = zeros(real(eltype(X)), M, maxiter+1)
 
     # B-orthogonalize X
-    X = ortho(X, tol=ortho_tol)[1]
+    X = ortho!(copy(X), tol=ortho_tol)[1]
```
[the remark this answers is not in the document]
mfherbst: Because of what you said in the first review: Do the copy in the solver, not in the wrapper.

src/scf/potential_mixing.jl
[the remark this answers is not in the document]
mfherbst: Ok I agree, so where do you suggest to synchronise instead?

src/scf/potential_mixing.jl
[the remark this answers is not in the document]
mfherbst: yes that might be worth it. I found them quite useful here.

src/eigen/lobpcg_hyper_impl.jl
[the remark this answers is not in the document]
antoine-levitt: I was just thinking might as well overwrite it. But OK fair that we shouldn't mutate the input.

src/scf/potential_mixing.jl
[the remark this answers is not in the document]
mfherbst: but it's not so easy to do if you want to keep the source info correct, I'll not do it for now.

mfherbst:
@antoine-levitt Are you ok if I nuke the CROP solver or you want to keep it just in case we want to pick this up later?

antoine-levitt:
Keep it, no? It doesn't hurt anyone and I don't think there's many CROP implementations around

Issue #452, opened 2021-06-18
Timing regression due to #442


mfherbst:
There is a huge timing regression introduced by #442

The src/example/silicon.jl with Ecut=30 on my Laptop with 4 threads on current master:
```
─────────────────────────────────────────────────────────────────────────────────────────
                                                  Time                   Allocations      
                                          ──────────────────────   ───────────────────────
             Tot / % measured:                 14.1s / 96.4%           4.55GiB / 99.2%    

 Section                          ncalls     time   %tot     avg     alloc   %tot      avg
 ─────────────────────────────────────────────────────────────────────────────────────────
 self_consistent_field                 1    13.5s   100%   13.5s   4.51GiB  100%   4.51GiB
   LOBPCG                             60    11.6s  85.4%   193ms   3.70GiB  81.9%  63.1MiB
     rayleigh_ritz                   179    5.06s  37.3%  28.3ms   6.55MiB  0.14%  37.5KiB
     Hamiltonian multiplication      239    1.24s  9.18%  5.20ms    174MiB  3.77%   747KiB
       kinetic+local                 239    1.14s  8.42%  4.78ms   4.54MiB  0.10%  19.4KiB
       nonlocal                      239   64.3ms  0.47%   269μs   52.7MiB  1.14%   226KiB
     ortho                           683   63.2ms  0.47%  92.5μs   5.63MiB  0.12%  8.44KiB
   compute_density                     6    1.74s  12.9%   290ms    498MiB  10.8%  83.0MiB
   energy_hamiltonian                 13    204ms  1.51%  15.7ms    288MiB  6.24%  22.2MiB
     ene_ops                          13    158ms  1.16%  12.1ms    110MiB  2.38%  8.45MiB
       ene_ops: xc                    13    105ms  0.78%  8.10ms   32.0MiB  0.69%  2.46MiB
       ene_ops: hartree               13   23.9ms  0.18%  1.84ms   63.6MiB  1.38%  4.89MiB
       ene_ops: nonlocal              13   19.1ms  0.14%  1.47ms    755KiB  0.02%  58.1KiB
       ene_ops: kinetic               13   5.51ms  0.04%   424μs    408KiB  0.01%  31.4KiB
       ene_ops: local                 13   3.10ms  0.02%   239μs   12.9MiB  0.28%  0.99MiB
   QR orthonormalization              60   7.75ms  0.06%   129μs   12.4MiB  0.27%   211KiB
[...]

PR #453, opened 2021-06-18
Performance workaround for BlockArrays

antoine-levitt:


antoine-levitt:
Can you check it fixes the regression?

antoine-levitt:
Fixed in https://github.com/JuliaArrays/BlockArrays.jl/pull/173, so we can just revert this PR whenever that is in a release we have

mfherbst:
Nope, not fully cutting it, see #452.

PR #458, opened 2021-06-24
Remove BlockArrays workaround

antoine-levitt:
Also fix preconditioning while I'm at it


mfherbst:
And this needs to be a Float64? Would that not give trouble for other floating-point types?

antoine-levitt:
Fair

PR #463, opened 2021-06-29
Workaround reorganization

antoine-levitt:


src/workarounds/README.md
```diff
@@ -0,0 +1 @@
+Workarounds, fixes and hacks for various external libraries, should ideally be upstreamed.
```
mfherbst, line 1: add no warranty statement

mfherbst:
Could you grep for FourierTransforms.jl in the full project. I made all sorts of exceptions for this folder (e.g. coverage). Just change these as well

antoine-levitt:
I don't see anything?

mfherbst:
Ah yes, I changed that at some point, I remember now. Sorry ... but good that you checked again

PR #464, opened 2021-06-30
A few more tweaks for potential mixing

mfherbst:


mfherbst:
I would merge this, if you are ok with it @antoine-levitt

antoine-levitt:
LGTM!

antoine-levitt:
I just remembered we say
"As usual with MPI printing will be garbled. You can use
DFTK.mpi_master() || (redirect_stdout(); redirect_stderr())"
in the manual. So just do that and don't have these ugly `is_master() && @debug` statements ?

mfherbst:
Well no, because I want to be able to read them in my logs ;)

antoine-levitt:
You can do that with the redirect trick, no?

PR #465, opened 2021-07-01
Fermi-Dirac stability improvement

antoine-levitt:


src/Smearing.jl
```diff
@@ -75,14 +71,32 @@ function occupation_derivative(S::FermiDirac, x)
...
     - (xlogx(f) + xlogx(1 - f))
 end
+function occupation_divided_difference(S::FermiDirac, x, y, εF, temperature)
+    temperature == 0 && return occupation_divided_difference(None(), x, y, εF, temperature)
+    f(z) = occupation(S, (z-εF) / temperature)
+    fder(z) = occupation_derivative(S, (z-εF)/temperature) / temperature
+    # For a stable computation we use
+    # (fx - fy) = fx fy (exp(y) - exp(x)) = fx fy exp(x) expm1(y-x)
+    # which we symmetrize. This can overflow, in which case we fall back to the standard method
+    will_exp_overflow(z1, z2) = abs((z1-z2)/temperature) > log(floatmax(typeof(x))) / 4 # conservative
```
mfherbst, line 91: You know how I feel about long lines ...

src/Smearing.jl
```diff
@@ -75,14 +71,32 @@ function occupation_derivative(S::FermiDirac, x)
...
 # although that is not especially useful...
+function xlogx(x)
+    iszero(x) ? zero(x) : x * log(x)
+end
 function entropy(S::FermiDirac, x)
     f = occupation(S, x)
     - (xlogx(f) + xlogx(1 - f))
 end
+function occupation_divided_difference(S::FermiDirac, x, y, εF, temperature)
+    temperature == 0 && return occupation_divided_difference(None(), x, y, εF, temperature)
```
mfherbst, line 85: I think you should abbreviate temperature by `temp` or something like that. That word destroys the flow of reading in this function.

src/Smearing.jl
antoine-levitt: There's no logical place to break here

src/Smearing.jl
antoine-levitt: Good point, done

src/Smearing.jl
mfherbst: come on, just assign `floatmax(typeof(x))` to something, or `T = typeof(x)`

mfherbst:
No don't use `T`. We decided to reserve that for element type

src/Smearing.jl
antoine-levitt: less of a problem now that temperature is smaller

antoine-levitt:
zero risk of confusion here, and it's just for this module

mfherbst:
We discussed this and decided to drop using `T` for temperature for consistency. We deliberately did it also in this module, remember? I don't think it is a good idea to revert that decision.

antoine-levitt:
I think T is completely fine, but I also can't be bothered fighting this, so temp it is

mfherbst:
Me neither. That's why you fight once and than stick to it :smile:.

mfherbst:
CI takes unusually long ...

antoine-levitt:
Everything's fine locally so it's probably not the PR. Github misbehaving?

mfherbst:
Probably. I'll restart the CI.

antoine-levitt:
It works in the other PR. WTF.

PR #473, opened 2021-07-07
Overhaul pack/unpack routines to not assume that psi are fully occupied

antoine-levitt:
Will need this for finite temperature, and came up as a confusing error with @niklasschmitz

### DIFF
--- a/src/PlaneWaveBasis.jl
+++ b/src/PlaneWaveBasis.jl
@@ -596,16 +596,15 @@ function pack_ψ(ψ)
     vcat([vec(ψk) for ψk in ψ]...)
 end
 
-function unpack_ψ(x, ψ_for_shape) where T
-    n_bands = size(ψ_for_shape[1], 2)
-    lengths = length.(ψ_for_shape)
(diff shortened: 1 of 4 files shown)


src/PlaneWaveBasis.jl
```diff
@@ -591,25 +591,21 @@ end
...
 end
 
-function unpack_ψ(basis::PlaneWaveBasis{T}, x) where T
-    model = basis.model
-    filled_occ = filled_occupation(model)
-    n_spin = model.n_spin_components
-    n_bands = div(model.n_electrons, n_spin * filled_occ)
-
-    lengths = length.(G_vectors.(basis.kpoints)) .* n_bands
+function unpack_ψ(x, ψ_for_shape) where T
```
mfherbst, line 599: Remove where T

mfherbst, line 599: I think I would pass the shape explicitly or even better do an in-place version of this function. This interface is a bit unusual.

gkemlin (approved):
Good for me !

Issue #443, opened 2021-06-07
[WIP] Autodiff stresses


niklasschmitz:
Opening this to keep track of progress on obtaining stresses via autodiff.

# Goal

Calculate the stress as the total derivative of the total energy wrt lattice parameters via automatic differentiation. As this falls under scope of the Hellmann-Feynman theorem, we do not need to differentiate through the full SCF solve but rather only through a post-processing on the final solution `scfres`.
We start with the following [minimal example](https://github.com/niklasschmitz/DFTK.jl/blob/cafed13ddc5d3f95d370fafcdb01869f3a61a23b/test/autodiff-stress/stress-total.jl) of silicon with a single scalar lattice parameter `a`
```julia
using DFTK
using Test

function make_basis(a)
    lattice = a / 2 * [[0 1 1.];
                    [1 0 1.];
                    [1 1 0.]]
    Si = ElementPsp(:Si, psp=load_psp("hgh/lda/Si-q4"))
    atoms = [Si => [ones(3)/8, -ones(3)/8]]
    model = model_atomic(lattice, atoms, symmetries=false)
    kgrid = [1, 1, 1]  # k-point grid (Regular Monkhorst-Pack grid)
    Ecut = 15          # kinetic energy cutoff in Hartree
    PlaneWaveBasis(model, Ecut; kgrid=kgrid)
end

a = 10.26
scfres = self_consistent_field(make_basis(a), tol=1e-8)

function compute_energy(scfres_ref, a)
    basis = make_basis(a)
    energies, H = energy_hamiltonian(basis, scfres_ref.ψ, scfres_ref.occupation; ρ=scfres_ref.ρ)
    energies.total
end

function compute_stress(scfres_ref, a)
    Inf # TODO implement
end
@test compute_stress(scfres, a) ≈ FiniteDiff.finite_difference_derivative(a -> compute_energy(scfres, a), a) # -1.411
```

# Approach

We plan to try ForwardDiff.jl, ReverseDiff.jl and Zygote.jl.
For stresses only (#params < 10) we expect ForwardDiff to perform best. Going further the reverse modes of ReverseDiff and Zygote are also interesting as they could jointly evaluate stresses and other derivatives of the total energy (eg. forces) more efficiently.

Expected challenges:
- calls into non-Julia code such as FFTW (hurdle for ForwardDiff and ReverseDiff which don't support ChainRules, might need to manually overload)
[...]

PR #476, opened 2021-07-08
Hellmann-Feynman stresses via ForwardDiff and custom rules

niklasschmitz:
This PR implements experimental support for Hellmann-Feynman stresses via ForwardDiff. 
Coming from https://github.com/niklasschmitz/DFTK.jl/pull/2 and https://github.com/JuliaMolSim/DFTK.jl/issues/443

This does not include yet:
- external code like libxc

TODO list
- [x] re-enable lattice cond check
- [x] symmetries
- [x] safer type promotion in compute_density
- [x] solve DummyInplace code duplication with requires
- [x] use silicon numbers from `test/testcases.jl`


test/stresses.jl
```diff
@@ -0,0 +1,33 @@
...
+        atoms = [Si => [ones(3)/8, -ones(3)/8]]
+        model = model_DFT(lattice, atoms, [:lda_x, :lda_c_vwn]; symmetries=false)
+        kgrid = [4, 4, 4] # k-point grid (Regular Monkhorst-Pack grid)
+        Ecut = 7          # kinetic energy cutoff in Hartree
+        PlaneWaveBasis(model, Ecut; kgrid=kgrid, fft_size=[32, 32, 32])
+    end
+
+    function compute_energy(scfres_ref, a)
+        basis = make_basis(a)
+        energies, H = energy_hamiltonian(basis, scfres_ref.ψ, scfres_ref.occupation; ρ=scfres_ref.ρ)
```
antoine-levitt, line 23: I don't think you can do this because the normalization of the rho depends on the volume of the unit cell. The orbitals are fine because they are l^2 normalized to 1, independently of the basis. You have to recompute rho in the new basis with the old orbitals.

antoine-levitt: weird that the test doesn't pick that up. Can you try checking more carefully? Ie do the SCF to a very small precision and checking the result to more accuracy

antoine-levitt: Also we should probably check it for a metal. I believe Hellman-Feynmann is also valid in this case but we should be careful.

Project.toml
```diff
@@ -8,6 +8,7 @@ AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
 BlockArrays = "8e7c35d0-a365-5155-bbbb-fb81a777f24e"
 Conda = "8f4d0f93-b110-5947-807f-2305c1781a2d"
 FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
+FiniteDiff = "6a86dc24-6348-571c-b903-95158fe2bd41"
```
antoine-levitt, line 11: Can you add this to [extras] instead? That's where we put the "optional" deps used in tests

antoine-levitt (commented):
Wonderful! Can we just merge as is or is the type piracy wrt ForwardDiff a problem?

test/stresses.jl
niklasschmitz: That was indeed a bug! Fixed it now by recomputing rho and added a consistency check against non-Hellmann-Feynman finite diff of the whole solve

src/DFTK.jl
```diff
@@ -172,6 +172,8 @@ include("postprocess/chi0.jl")
 export compute_current
 include("postprocess/current.jl")
 
+include("workarounds/forwarddiff_rules.jl")
```
mfherbst, line 175: This needs to be conditionally included when ForwardDiff is available (using requires, I can do it if you want).

src/densities.jl
```diff
@@ -60,9 +60,10 @@ is not collinear the spin density is `nothing`.
     @assert n_k > 0
 
     # Allocate an accumulator for ρ in each thread for each spin component
-    ρaccus = [similar(view(ψ[1], :, 1), (basis.fft_size..., n_spin))
+    # TODO use something like T = promote_type(Array{eltype(basis)}, eltype(ψ), ...)
+    ρaccus = [similar(Array{complex(T)}, (basis.fft_size..., n_spin))
               for ithread in 1:Threads.nthreads()]
-
+
```
mfherbst, line 66: Whitespace commit

src/densities.jl
```diff
@@ -60,9 +60,10 @@ is not collinear the spin density is `nothing`.
     @assert n_k > 0
 
     # Allocate an accumulator for ρ in each thread for each spin component
-    ρaccus = [similar(view(ψ[1], :, 1), (basis.fft_size..., n_spin))
+    # TODO use something like T = promote_type(Array{eltype(basis)}, eltype(ψ), ...)
+    ρaccus = [similar(Array{complex(T)}, (basis.fft_size..., n_spin))
```
mfherbst, line 64: Not just that ... this also makes it more annoying when we think about GPUs, but ok we worry about that some other day.

src/densities.jl
```diff
@@ -60,9 +60,10 @@ is not collinear the spin density is `nothing`.
     @assert n_k > 0
 
     # Allocate an accumulator for ρ in each thread for each spin component
-    ρaccus = [similar(view(ψ[1], :, 1), (basis.fft_size..., n_spin))
+    # TODO use something like T = promote_type(Array{eltype(basis)}, eltype(ψ), ...)
```
mfherbst, line 63: I think the pattern would be `T = promote_type(T, eltype(\psi))` and then just use `T`

test/stresses.jl
```diff
@@ -0,0 +1,45 @@
+# Hellmann-Feynman stress
+# via ForwardDiff & custom FFTW overloads on ForwardDiff.Dual
```
mfherbst, line 2: We typically first do imports than comments

test/stresses.jl
```diff
@@ -0,0 +1,45 @@
...
+
+@testset "ForwardDiff stresses on silicon" begin
+    function make_basis(a)
+        lattice = a / 2 * [[0 1 1.];
+                        [1 0 1.];
+                        [1 1 0.]]
+        Si = ElementPsp(:Si, psp=load_psp(:Si, functional="lda"))
+        atoms = [Si => [ones(3)/8, -ones(3)/8]]
+        model = model_DFT(lattice, atoms, [:lda_x, :lda_c_vwn])
+        kgrid = [1, 1, 1] # k-point grid (Regular Monkhorst-Pack grid)
```
mfherbst, line 16: Two spaces before `#` ... also these comments are sort of not needed for the tests. It's a pretty standard setup. Also we have data for standard test cases centralised in `testcases.jl` So you better use that. See the other tests for examples (search for `silicon.lattice`).

src/workarounds/forwarddiff_rules.jl
```diff
@@ -0,0 +1,161 @@
+import ForwardDiff
+import AbstractFFTs
+
+# original PR by mcabbott: https://github.com/JuliaDiff/ForwardDiff.jl/pull/495
+# modified version: https://github.com/niklasschmitz/ForwardDiff.jl/blob/nfs/fft/src/fft.jl
```
mfherbst, line 5: Is this comment useful (in the sense: Will this not point to void when the branch is merged / deleted)

src/workarounds/forwarddiff_rules.jl
```diff
@@ -0,0 +1,161 @@
...
+
+ForwardDiff.value(x::Complex{<:ForwardDiff.Dual}) =
+    Complex(x.re.value, x.im.value)
```
mfherbst, line 8: This linebreak feels too much.

src/workarounds/forwarddiff_rules.jl
```diff
@@ -0,0 +1,161 @@
+import ForwardDiff
+import AbstractFFTs
+
```
mfherbst, line 3: Quite a bit of these workarounds are not covered by tests at the moment, which bothers me a bit. I know this will not last forever (hopefully), but if you can think of an easy way to increase coverage here @niklasschmitz (like another nice example or so) that would be good.

src/workarounds/forwarddiff_rules.jl
```diff
@@ -0,0 +1,161 @@
...
+# AbstractFFTs.complexfloat(x::AbstractArray{<:ForwardDiff.Dual}) = float.(x .+ 0im)
+AbstractFFTs.complexfloat(x::AbstractArray{<:ForwardDiff.Dual}) = AbstractFFTs.complexfloat.(x)
+AbstractFFTs.complexfloat(d::ForwardDiff.Dual{T,V,N}) where {T,V,N} = convert(ForwardDiff.Dual{T,float(V),N}, d) + 0im
+
+AbstractFFTs.realfloat(x::AbstractArray{<:ForwardDiff.Dual}) = AbstractFFTs.realfloat.(x)
+AbstractFFTs.realfloat(d::ForwardDiff.Dual{T,V,N}) where {T,V,N} = convert(ForwardDiff.Dual{T,float(V),N}, d)
+
+for plan in [:plan_fft, :plan_ifft, :plan_bfft]
+    @eval begin
+
```
mfherbst, line 28: I would not put an empty line here ...

src/workarounds/forwarddiff_rules.jl
```diff
@@ -0,0 +1,161 @@
...
+        AbstractFFTs.$plan(x::AbstractArray{<:ForwardDiff.Dual}, region=1:ndims(x); kwargs...) =
+            AbstractFFTs.$plan(ForwardDiff.value.(x) .+ 0im, region; kwargs...)
+
+        AbstractFFTs.$plan(x::AbstractArray{<:Complex{<:ForwardDiff.Dual}}, region=1:ndims(x); kwargs...) =
+            AbstractFFTs.$plan(ForwardDiff.value.(x), region; kwargs...)
+
```
mfherbst, line 34: ... or here.

src/workarounds/forwarddiff_rules.jl
```diff
@@ -0,0 +1,161 @@
...
+    end
+end
+
+function _apply_plan(p::AbstractFFTs.ScaledPlan{T,P,<:ForwardDiff.Dual}, x::AbstractArray) where {T,P}
+    _apply_plan(p.p, p.scale * x) # for when p.scale is Dual, need out-of-place
+end
+
+###
+### DFTK setup specific
+###
```
mfherbst, line 94: Just
```suggestion
#
# DFTK setup specific
#
```
or even just the comment surrounded by empty lines ... too much is just too much :smile:

src/workarounds/forwarddiff_rules.jl
```diff
@@ -0,0 +1,161 @@
...
+
+next_working_fft_size(::Type{<:ForwardDiff.Dual}, size) = size
+
+_fftw_flags(::Type{<:ForwardDiff.Dual}) = FFTW.MEASURE | FFTW.UNALIGNED
+
+# *** COPIED from fft_generic.jl *** TODO refactor
```
mfherbst, line 100: One way to do that is to have a file that is included with Requires for both types of dependencies, a bit like is done for the generic workarounds. I can do that if you want.

src/densities.jl
niklasschmitz: That's a good point.. Is there a good alternative here?  Another solution would be to go for non-mutating code and avoiding the typed pre-allocation, which we'd explore next with Zygote anyways.

src/workarounds/forwarddiff_rules.jl
niklasschmitz: agreed!

src/workarounds/forwarddiff_rules.jl
niklasschmitz: Agree, I removed the pointer to my branch

src/workarounds/forwarddiff_rules.jl
niklasschmitz: I factored this out into a new common `workarounds/dummy_inplace_fft.jl` with a flag `DUMMY_INPLACE_LOADED` checked at the requires.

src/densities.jl
```diff
@@ -79,7 +80,10 @@ is not collinear the spin density is `nothing`.
 
     Threads.@threads for (ikpts, ρaccu) in collect(zip(kpt_per_thread, ρaccus))
         ρaccu .= 0
-        ρ_k = similar(ψ[1][:, 1], basis.fft_size)
+
+        # TODO use something like T = promote_type(Array{eltype(basis)}, eltype(ψ), ...)
+        ρ_k = similar(Array{complex(T)}, basis.fft_size)
```
antoine-levitt, line 85: this allocates twice

test/stresses.jl
niklasschmitz: I now added the re-use of `atnum`, `psp` and `positions` from silicon in testcases.jl

src/densities.jl
mfherbst: Why twice? This just allocates the array and that's it.

antoine-levitt: oh right, sorry, not used to this pattern

src/densities.jl
niklasschmitz: I updated to use promote_type in combination with the three-arg version of `similar` specifying both the eltype and size https://github.com/JuliaMolSim/DFTK.jl/pull/476/commits/6a821bd79a920dafe7b2001db6609d8aba9644c3

src/densities.jl
niklasschmitz: https://github.com/JuliaMolSim/DFTK.jl/pull/476/commits/6a821bd79a920dafe7b2001db6609d8aba9644c3

src/Model.jl
```diff
@@ -103,8 +103,7 @@ function Model(lattice::AbstractMatrix{T};
         norm(lattice[:, i]) == norm(lattice[i, :]) == 0 || error(
             "For 1D and 2D systems, the non-empty dimensions must come first")
     end
-    cond(lattice[1:n_dim, 1:n_dim]) > 1e-5 || (
-        @warn "Your lattice is badly conditioned, the computation is likely to fail.")
+    _check_well_conditioned(lattice[1:d, 1:d]) || @warn "Your lattice is badly conditioned, the computation is likely to fail."
```
mfherbst, line 106: Line too long.

src/DFTK.jl
```diff
@@ -177,13 +177,19 @@ function __init__()
...
-    # included once.
+    # The global variables GENERIC_FFT_LOADED and DUMMY_INPLACE_LOADED
+    # make sure that things are only included once.
+    @require ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210" begin
+        !isdefined(DFTK, :DUMMY_INPLACE_LOADED) && include("workarounds/dummy_inplace_fft.jl")
+        include("workarounds/forwarddiff_rules.jl")
+    end
     @require IntervalArithmetic="d1acc4aa-44c8-5952-acd4-ba5d80a2a253" begin
         include("workarounds/intervals.jl")
+        !isdefined(DFTK, :DUMMY_INPLACE_LOADED) && include("workarounds/dummy_inplace_fft.jl")
```
antoine-levitt, line 188: I'm confused, why is that needed here?

mfherbst: It could happen someones uses ForwardDiff and DoubleFloats. So you don't want to include twice.

antoine-levitt: Sure but this loads the forwarddiff stuff when intervalarithmetic is loaded?

src/Model.jl
niklasschmitz: added a line break https://github.com/JuliaMolSim/DFTK.jl/pull/476/commits/30ada7f9967b715576a246d4a70d509b30a3f5d4

src/DFTK.jl
mfherbst: dummy inplace is also needed for generic ffts

src/Model.jl
```diff
@@ -103,8 +103,8 @@ function Model(lattice::AbstractMatrix{T};
         norm(lattice[:, i]) == norm(lattice[i, :]) == 0 || error(
             "For 1D and 2D systems, the non-empty dimensions must come first")
     end
-    cond(lattice[1:n_dim, 1:n_dim]) > 1e-5 || (
-        @warn "Your lattice is badly conditioned, the computation is likely to fail.")
+    _check_well_conditioned(lattice[1:d, 1:d]) || @warn (
```
mfherbst, line 106: This should be `1:n_dim` now.

src/DFTK.jl
antoine-levitt: then can we just have include("dummy_inplace_fft.jl") in fft_generic.jl?

niklasschmitz: dummy_inplace was previously part of generic_fft.jl (it does not include ForwardDiff stuff but is a common prerequisite for all three of ForwardDiff, DoubleFloats and IntervalArithmetic)

antoine-levitt: Or not... never mind anyway

src/Model.jl
niklasschmitz: ah yes!

src/DFTK.jl
mfherbst: we could, but I think that's only more confusing.

src/Model.jl
niklasschmitz: https://github.com/JuliaMolSim/DFTK.jl/pull/476/commits/cb466c1903f805af1008fb009472712cef51f8ea

src/DFTK.jl
antoine-levitt: Yup OK!

niklasschmitz:
Thanks @mfherbst @antoine-levitt !

antoine-levitt:
Thanks for the great work!

PR #477, opened 2021-07-08
Ω+K, split version

antoine-levitt:


src/scf/newton.jl
```diff
@@ -169,7 +169,8 @@ end
 # With χ04P = -Ω^-1,
 # (Ω+K)^-1 = Ω^-1 (1 - K(1+Ω^-1 K)^-1 Ω^-1)
 # (Ω+K)^-1 = -χ04P (1 + K(1 - χ04P K)^-1 χ04P)
-# (Ω+K)^-1 = -χ04P (1 + K(1 - χ0 K)^-1 χ04P)
+# (Ω+K)^-1 = -χ04P (1 + E K2P (1 - χ02P K2P)^-1 R χ04P)
+# where χ02P = R χ04P E and K2P = R K E
```
mfherbst, line 173: Ok, but this opens up the question what is R and E ... I feel we should document a few of these things at a central place and then just take it as granted knowledge

Talking about a hackaton, that's the sort of annoying tasks one could leave for that.

antoine-levitt: It is on top of the file. Agree the documentation is not great right now, I want to wait until things settle down to overhaul it

antoine-levitt: Also not sure how to communicate this best. E and R are nice but a bit on the formal side. The code is actually simpler than the formulas here.

PR #478, opened 2021-07-09
Psp sanity check for lower dimensionality

mfherbst:


antoine-levitt:
You 3g is better than mine... I don't like the fact that model has to know about psp. Can't you move that check to the psp themselves?

mfherbst:
Yes but the PsPs don't know anything about the model. It's the first place where the info is unified. I thought about moving it to the Ewald energy correction calculation for PsPs, but that sort of feels strange (it's the term where the issue is apparent, but it's not really the problem. The problem is that model and atoms definition do not fit).

antoine-levitt:
Yeah, I would move it to TermPspCorrection(basis::PlaneWaveBasis), so the pwbasis doesn't have to know about the terms. Agreed it's a bit late, but that doesn't really matter, does it?

mfherbst:
> so the pwbasis doesn't have to know about the terms

I suppose you mean model?

Yeah ok. At least it's the point where things effectively break.

By the way: I also noted our code is inconsistent for the case where there is a net charge in the unit cell. I know it does not make a lot of sense to do this, but people still do it in practice. Our treatment of the compensating background in Ewald / Psp / Hartree etc., however assumes that this is not the case.

antoine-levitt:
Yes, we could add a preventive assert there too. It does make sense but we need to think about it more carefully.

PR #482, opened 2021-07-12
Avoid mutation in ρ_from_total_and_spin

antoine-levitt:
Playing around with mutation. Debatable if it's clearer to avoid mutation...

### DIFF
--- a/src/densities.jl
+++ b/src/densities.jl
@@ -139,10 +139,9 @@ end
 
 function ρ_from_total_and_spin(ρtot, ρspin=nothing)
     if ρspin === nothing
-         # make a copy for consistency with the other case
-        cat(ρtot; dims=Val(4))
+        # Val used to ensure inferability
+        cat(ρtot; dims=Val(4))  # copy for consistency with other case
     else
-        # Val to ensure inferability
         cat((ρtot .+ ρspin) ./ 2,
             (ρtot .- ρspin) ./ 2; dims=Val(4))
     end
--- a/src/densities.jl
+++ b/src/densities.jl
@@ -138,13 +138,12 @@ total_density(ρ) = dropdims(sum(ρ; dims=4); dims=4)
 end
 
 function ρ_from_total_and_spin(ρtot, ρspin=nothing)
-    n_spin = ρspin === nothing ? 1 : 2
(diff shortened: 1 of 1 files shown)


mfherbst:
I don't think it's too bad.

src/densities.jl
```diff
@@ -138,13 +138,12 @@ total_density(ρ) = dropdims(sum(ρ; dims=4); dims=4)
...
-    ρ = similar(ρtot, size(ρtot)..., n_spin)
-    if n_spin == 1
-        ρ .= ρtot
+    if ρspin === nothing
+         # returns a copy for consistency with the other case
+        copy(reshape(ρtot, (size(ρtot)..., 1)))
     else
-        ρ[:, :, :, 1] .= (ρtot .+ ρspin) ./ 2
-        ρ[:, :, :, 2] .= (ρtot .- ρspin) ./ 2
+        # Val to ensure inferability
```
mfherbst, line 145: Can't one ensure that by providing the return type? I mean if we capture the element type of `ρtot` with a `where` than the return type is sort of clear or not?

mfherbst: Not sure if that's easier to parse though. Probably Val is still the better option.

antoine-levitt: not sure that's a good idea, because eg if rho is a GPU array it could be that cat still lives on the GPU so we don't want to assert it's an Array

src/densities.jl
```diff
@@ -138,13 +138,12 @@ total_density(ρ) = dropdims(sum(ρ; dims=4); dims=4)
 end
 
 function ρ_from_total_and_spin(ρtot, ρspin=nothing)
-    n_spin = ρspin === nothing ? 1 : 2
-    ρ = similar(ρtot, size(ρtot)..., n_spin)
-    if n_spin == 1
-        ρ .= ρtot
+    if ρspin === nothing
+         # returns a copy for consistency with the other case
+        copy(reshape(ρtot, (size(ρtot)..., 1)))
```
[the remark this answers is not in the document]
antoine-levitt: good idea

src/densities.jl
mfherbst: yes cat still lives on the GPU I checked that.

PR #486, opened 2021-07-16
Pwbasis refactor

antoine-levitt:
This makes more sense and is non-breaking. Based on that we can try to cut down the API further and remove a bunch of constructors...


antoine-levitt:
(it is technically breaking in the sense that it will not change fft_size for buggy generic FFTs behind your back and will just error, but I think that's fair)

src/PlaneWaveBasis.jl
```diff
@@ -140,24 +150,17 @@ end
...
-    # Validate kpoints and symmetries
-    @assert length(kcoords) == length(ksymops)
-    if symmetries === nothing
-        # TODO instead compute the group generated by with ksymops, or
-        # just retire this constructor. Not critical because this
-        # should not be used in this context anyway...
-        symmetries = vcat(ksymops...)
+                                kcoords::AbstractVector, ksymops,
+                                fft_size, variational, symmetries,
+                                kgrid, kshift, comm_kpts) where {T <: Real}
```
mfherbst, line 158: I wonder if we should keep the argument order. Perhaps it's better to first do the generic stuff (`model`, `Ecut`, `fft_size`, `variational`, `symmetries`) then the k-Point stuff (`kcoords`, `ksymops`, `kgrid`, `kshift`, `comm_kpts`).

src/PlaneWaveBasis.jl
```diff
@@ -140,24 +150,17 @@ end
...
+    if !(all(fft_size .== next_working_fft_size.(T, fft_size)))
+        error("Selected fft_size will not work for the buggy generic FFT routines")
```
mfherbst, line 160: This error message does not help a user that has never encountered this error and does not know about the internals.

src/PlaneWaveBasis.jl
```diff
@@ -140,24 +150,17 @@ end
...
     end
+    fft_size = Tuple{Int, Int, Int}(fft_size) # need explicit convert from
```
mfherbst, line 162: two spaces before `#` (is it that hard to write an emacs macro that fixes that?)

src/PlaneWaveBasis.jl
```diff
@@ -210,16 +210,17 @@ build_kpoints(basis::PlaneWaveBasis, kcoords) =
         "Non-variational calculations are experimental. " *
         "Not all features of DFTK may be supported or work as intended."
     )
-    kpoints = build_kpoints(model, fft_size, kcoords, Ecut; variational=variational)
+    kpoints = build_kpoints(model, fft_size, kcoords_thisproc, Ecut;
+                            variational)
```
mfherbst, line 214: here i would skip the linebreak.

src/PlaneWaveBasis.jl
```diff
@@ -231,7 +232,7 @@ build_kpoints(basis::PlaneWaveBasis, kcoords) =
 
     basis = PlaneWaveBasis{T}(
         model, Ecut, variational, kpoints,
-        kweights, ksymops, kgrid, kshift, comm_kpts, krange_thisproc, krange_allprocs,
+        kweights, ksymops_thisproc, kgrid, kshift, kcoords, ksymops, comm_kpts, krange_thisproc, krange_allprocs,
```
mfherbst, line 235: but this is definitely too long

src/PlaneWaveBasis.jl
```diff
@@ -245,13 +246,58 @@ build_kpoints(basis::PlaneWaveBasis, kcoords) =
...
 
+# This is the "internal" constructor; the higher-level one below should be preferred
+@timing function PlaneWaveBasis(model::Model{T}, Ecut::Number,
+                                kcoords::AbstractVector, ksymops, symmetries=nothing;
+                                fft_size=nothing, variational=true,
+                                optimize_fft_size=false, supersampling=2,
+                                kgrid=nothing, kshift=nothing,
+                                comm_kpts=MPI.COMM_WORLD,
+                               ) where {T <: Real}
+    mpi_ensure_initialized()
```
mfherbst, line 257: Why is this needed?

src/PlaneWaveBasis.jl
```diff
@@ -245,13 +246,58 @@ build_kpoints(basis::PlaneWaveBasis, kcoords) =
...
+Creates a new basis identical to `basis`, but with a custom set of kpoints
 """
 function PlaneWaveBasis(basis::PlaneWaveBasis, kcoords::AbstractVector,
-                        ksymops::AbstractVector, symmetries=nothing)
-    PlaneWaveBasis(basis.model, basis.Ecut, kcoords, ksymops, symmetries;
-                   fft_size=basis.fft_size, variational=basis.variational)
+                        ksymops::AbstractVector, symmetries=vcat(ksymops...))
+    PlaneWaveBasis(basis.model, basis.Ecut, kcoords, ksymops,
+                   basis.fft_size, basis.variational, basis.symmetries,
+                   nothing, nothing, # explicitly given kpoints
```
mfherbst, line 299: I think for clarity I would assign these to variables `kgrid` and `kshift` and pass them. That's one line more and we don't forget what `nothing` refers to. Also easier for later refactoring.

src/PlaneWaveBasis.jl
```diff
@@ -245,13 +246,58 @@ build_kpoints(basis::PlaneWaveBasis, kcoords) =
...
+    if symmetries === nothing
+        # TODO instead compute the group generated by ksymops (also in
+        # the other constructors). Not critical because this should
+        # not be used in this context anyway...
+        symmetries = vcat(ksymops...)
+    end
+    PlaneWaveBasis(model, Ecut,
+                   kcoords, ksymops,
+                   fft_size, variational, symmetries,
+                   kgrid, kshift, comm_kpts)
```
mfherbst, line 289: Strange formatting ...

src/fft.jl
```diff
@@ -101,44 +101,27 @@ end
...
-            "For a variational method, Ecut should be less than the maximal kinetic " *
-            "energy the grid supports ($max_E)"
-        )
-    else
-        # ensure no other options are set
-        @assert supersampling == 2
-        @assert !optimize_fft_size
+function compute_fft_size(model::Model{T}, Ecut, supersampling,
+                          variational, optimize_fft_size, kcoords) where {T}
+    @assert variational
```
mfherbst, line 106: Remove variational argument from this function and add the assertion to PWbasis constructor?

test/silicon_redHF.jl
```diff
@@ -24,7 +24,8 @@ function run_silicon_redHF(T; Ecut=5, grid_size=15, spin_polarization=:none, kwa
     ]
     ref_etot = -5.440593269861395
 
-    fft_size = grid_size * ones(3)
+    fft_size = fill(grid_size, 3)
+    fft_size = DFTK.next_working_fft_size.(T, fft_size) # ad-hoc fix for buggy generic FFTs
```
mfherbst, line 28: Actually I think asking the user to do this explicitly is better indeed. However in that case I would probably change the user-facing function to be just a call to a tuple and not require the extra broadcast.

src/PlaneWaveBasis.jl
antoine-levitt: Yeah I thought about that but probably if you want to start playing around with stuff like that you should be able to look at the code surrounding an error message; if not, you'd better turn back anyway

mfherbst:
I think I would already add a constructor `PlaneWaveBasis(model; Ecut=stuff, kwargs...)` if you agree to allow for a smooth transition in the future. When I record the JuliaCon workshop it would be nice to already support that pattern and don't show code examples that will break in a few weeks.

src/PlaneWaveBasis.jl
antoine-levitt: my avowed goal is that at some point you'll give up :-p I still don't understand why you care about that, I've never seen that rule before and it makes no sense to me

src/PlaneWaveBasis.jl
mfherbst: You could at least point out the name of the cryptic function to use for fixing this.

antoine-levitt: it's right there up one line from this error message. Looking up the stacktrace isn't particularly hard to do, and I don't think it's reasonable to babysit users who can't do that for this kind of advanced functionality

src/PlaneWaveBasis.jl
antoine-levitt: because we call MPI and it might be the first time we do

src/PlaneWaveBasis.jl
antoine-levitt: good point!

src/fft.jl
antoine-levitt: yup good

src/PlaneWaveBasis.jl
mfherbst: You really don't want to be nice, do you? I would consider my self an advanced user and I hate looking up what the function name was when I simply forgot to call the function. If you print it, however, I don't have to actively dig through the code next time I use this feature. I would appreciate that. I assume others, too. ;)

src/PlaneWaveBasis.jl
mfherbst: it makes comments easier to read. python uses the same convention (and so do the established julia styles if i remember correctly)

src/PlaneWaveBasis.jl
mfherbst: but the mpi call happens in the internal constructor, right? so there is another init call

src/PlaneWaveBasis.jl
mfherbst: well that's one of the reasons why I dislike "true", "false" or "nothing" as positional arguments ...

src/PlaneWaveBasis.jl
antoine-levitt: ah sorry I was not at the right line. Yup you're right

src/PlaneWaveBasis.jl
antoine-levitt: alright, alright. It's just so ugly that we even have this that I didn't want to enshrine it in an error message, but fair enough

mfherbst: Yeah, but I doubt there will be a generic pure-Julia FFT any time soon. Attention pops up periodically, but then it decays when people find out how much effort went into FFTW.

antoine-levitt: https://github.com/JuliaApproximation/FastTransforms.jl/blob/master/src/fftBigFloat.jl looks OK; doing a simple radix-2 FFT is not hard. It's probably not super fast, but we don't really care about that

mfherbst: I know about that code. We tried initially and then moved away from it. Don't remember why, though. Could be that the one we use now is faster, actually.

antoine-levitt:
OK let's do that. Thank god for the implicit X=X in kwargs in 1.6

mfherbst:
Oh yes! I love it.

PR #487, opened 2021-07-16
Stresses API

antoine-levitt:
Built on top of https://github.com/JuliaMolSim/DFTK.jl/pull/486

@niklasschmitz I'm getting an error in one of your fallbacks (looks like it's because I forward differentiate wrt several parameters at once), can you take a look?

- [x] Add tests
- [x] Add unfolding


niklasschmitz:
Indeed my norm rule hadn't covered the case of multiple partials in a Dual yet, I'll extend it shortly!

antoine-levitt:
Stresses are not correct with symmetries, as expected:
```
julia> ans
3×3 Matrix{Float64}:
 -0.00159939   7.98518e-5   7.98074e-5
  7.98518e-5  -0.00159939   7.98366e-5
  7.98074e-5   7.98366e-5  -0.00159939
julia> ans
3×3 Matrix{Float64}:
 -0.00181531   -0.000146066  -0.000119765
 -0.000146066  -0.00151338    0.000505329
 -0.000119765   0.000505329  -0.00146946
```
without and with symmetries, respectively, for a compressed silicon (a=9.26). The first is correct: negative stresses (and therefore positive pressure) that are only in the diagonal direction (because only a is a free parameter)

antoine-levitt:
So this needs an unfold_BZ(scfres), which is also needed for @LaurentVidal95 's Wannier PR anyway

src/postprocess/stresses.jl
```diff
@@ -0,0 +1,37 @@
...
+        new_basis = PlaneWaveBasis(new_model, basis.Ecut,
+                                   basis.kcoords_global, basis.ksymops_global,
+                                   basis.fft_size, basis.variational, basis.symmetries,
+                                   basis.kgrid, basis.kshift, basis.comm_kpts)
+        ρ = DFTK.compute_density(new_basis, ψ, occ)
+        energies, _ = energy_hamiltonian(new_basis, ψ, occ; ρ=ρ)
+        energies.total
+    end
+    ForwardDiff.gradient(M -> HF_energy((I+M) * basis.model.lattice),
+                         zeros(eltype(basis), 3, 3)) / det(basis.model.lattice)
```
mfherbst, line 33: It's so awesome that you can just write this ...

src/postprocess/stresses.jl
```diff
@@ -0,0 +1,37 @@
...
+end
+function compute_stresses(scfres)
+    compute_stresses(scfres.basis, scfres.ψ, scfres.occupation)
+end
```
mfherbst, line 37: I guess one should also have a `stresses_cart` for consistency?

antoine-levitt: I'm not super confortable with stress definitions so I'd rather have just one that matches what's found in the literature. Forces are more intuitive

src/postprocess/stresses.jl
antoine-levitt: well, with a bunch of @niklasschmitz 's magic behind it :-)

src/postprocess/stresses.jl
mfherbst: I see, but forces in the literature are usually cartesian as well. Perhaps we should only support cartesian forces? Or we wait until we understand more and decide then.

antoine-levitt: Well OK, cartesian stresses might make sense. Forces in reduced coordinates definitely makes sense, it's what you want to do for geometry optimization for instance. For stresses I'm not sure what to do about the division by volume, so let's leave it as is for now.

antoine-levitt:
TODO (reminder to self) add tests

antoine-levitt:
Good to go from my end

src/symmetry.jl
```diff
@@ -239,3 +239,81 @@ function check_symmetric(basis, ρin; tol=1e-10, symmetries=ρin.basis.model.sym
...
+    for ik_unfolded in 1:length(basis_unfolded.kpoints)
+        kpt_unfolded = basis_unfolded.kpoints[ik_unfolded]
+        ik_irred, symop = unfold_mapping(basis_irred, kpt_unfolded)
+        if !is_ψ
+            data_unfolded[ik_unfolded] = data[ik_irred]
+        else
+            @assert normalize_kpoint_coordinate(kpt_unfolded.coordinate) ≈ kpt_unfolded.coordinate
+            data_unfolded[ik_unfolded] = apply_ksymop(symop, basis_irred,
+                                                        basis_irred.kpoints[ik_irred],
+                                                        data[ik_irred])[2]
```
mfherbst, line 296: This is pretty cryptic and deserves a comment, also indention.

src/symmetry.jl
```diff
@@ -239,3 +239,81 @@ function check_symmetric(basis, ρin; tol=1e-10, symmetries=ρin.basis.model.sym
...
+        return data
+    end
+    if !(basis_irred.comm_kpts == basis_irred.comm_kpts == MPI.COMM_WORLD)
+        error("Brillouin zone symmetry unfolding not supported with MPI yet")
+    end
+    data_unfolded = similar(data, length(basis_unfolded.kpoints))
+    for ik_unfolded in 1:length(basis_unfolded.kpoints)
+        kpt_unfolded = basis_unfolded.kpoints[ik_unfolded]
+        ik_irred, symop = unfold_mapping(basis_irred, kpt_unfolded)
+        if !is_ψ
```
mfherbst, line 290: How about better calling this flag apply_symops or something like that.

src/symmetry.jl
```diff
@@ -239,3 +239,81 @@ function check_symmetric(basis, ρin; tol=1e-10, symmetries=ρin.basis.model.sym
...
+end
+unfold_array(basis_irred, basis_unfolded, data::Vector{Vector{T}}) where {T <: Number} =
+    unfold_array(basis_irred, basis_unfolded, data, false)
+unfold_ψ(basis_irred, basis_unfolded, ψ) = unfold_array(basis_irred, basis_unfolded, ψ, true)
+
+function unfold_BZ(scfres)
+    basis_unfolded = unfold_BZ(scfres.basis)
+    ψ = unfold_ψ(scfres.basis, basis_unfolded, scfres.ψ)
+    eigenvalues = unfold_array(scfres.basis, basis_unfolded, scfres.eigenvalues)
+    occupation = unfold_array(scfres.basis, basis_unfolded, scfres.occupation)
```
mfherbst, line 309: I think I would structure this differently. For me these three objects are inherently connected and I think it makes sense to fuse the loop, i.e. to handle them simultaneously. That way you also get rid of this strange `is_\psi` thingy. Also what other k-Point specific objects will we need to transfer in such an unfolding? I don't thing any more.

src/symmetry.jl
```diff
@@ -239,3 +239,81 @@ function check_symmetric(basis, ρin; tol=1e-10, symmetries=ρin.basis.model.sym
...
+    E, ham = energy_hamiltonian(basis_unfolded, ψ, occupation;
+                                scfres.ρ, eigenvalues, scfres.εF)
+    @assert E.total ≈ scfres.energies.total
+    new_scfres = (basis=basis_unfolded,
+                  ψ=ψ,
+                  ham=ham,
+                  eigenvalues=eigenvalues,
+                  occupation=occupation)
```
mfherbst, line 317: You can also do `(; ham, eigenvalues, ...)` in 1.6.

mfherbst (commented):
I'm not yet fully convinced about the name `unfold_BZ`, because nothing is really done to the BZ. It is only the symmetry that is undone, so IR-wedge -> full BZ.  Maybe `remove_symmetries_scfres` ? `disable_symmetries_scfres` ?

antoine-levitt:
> I'm not yet fully convinced about the name unfold_BZ, because nothing is really done to the BZ. It is only the symmetry that is undone, so IR-wedge -> full BZ. Maybe remove_symmetries_scfres ? disable_symmetries_scfres ?

Well the irreducible BZ gets unfolded into the reducible BZ, so the name works, no? Nothing is done to the symmetries field of scfres, so disable_symmetries would be a bit misleading

test/stresses.jl
```diff
@@ -8,38 +8,48 @@ include("testcases.jl")
...
-        lattice = a / 2 * [[0 1 1.];
-                           [1 0 1.];
-                           [1 1 0.]]
+    function make_basis(lattice, symmetry)
         Si = ElementPsp(silicon.atnum, psp=load_psp(silicon.psp))
         atoms = [Si => silicon.positions]
-        model = model_DFT(lattice, atoms, [:lda_x, :lda_c_vwn])
-        PlaneWaveBasis(model; Ecut=7, kgrid=[1, 1, 1])
+        model = model_DFT(lattice, atoms, [:lda_x, :lda_c_vwn]; symmetries=symmetry)
+        kgrid = [2, 2, 2]
```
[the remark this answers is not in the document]
antoine-levitt: I did test but then there were less symmetries, so this one is a nice enough test

src/symmetry.jl
```diff
@@ -239,3 +239,81 @@ function check_symmetric(basis, ρin; tol=1e-10, symmetries=ρin.basis.model.sym
...
+    for ik_irred = 1:length(basis_irred.kpoints)
+        kpt_irred = basis_irred.kpoints[ik_irred]
+        for symop in basis_irred.ksymops[ik_irred]
+            if normalize_kpoint_coordinate(symop[1] * kpt_irred.coordinate) ≈
+                normalize_kpoint_coordinate(kpt_unfolded.coordinate) && kpt_unfolded.spin == kpt_irred.spin
+                return ik_irred, symop
+            end
+        end
+    end
+    error("Invalid unfolding of BZ")
```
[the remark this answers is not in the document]
antoine-levitt: You mean like `@assert false`? Feels weird, I like error better

src/symmetry.jl
antoine-levitt: I don't want to give the impression this is a general thing : it's really specialized for psi

src/symmetry.jl
antoine-levitt: So you'd just merge the unfold_BZ and unfold_array functions? I kind of like the separation. You could have other things, like a partial density, a matrix element for a response computation, etc

test/stresses.jl
```diff
@@ -8,38 +8,48 @@ include("testcases.jl")
 # via ForwardDiff & custom FFTW overloads on ForwardDiff.Dual
 
 @testset "ForwardDiff stresses on silicon" begin
```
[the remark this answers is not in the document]
antoine-levitt: Yeah for that kind of things it would be useful to separate systems and features

src/postprocess/stresses.jl
```diff
@@ -0,0 +1,33 @@
...
+                                   basis.Ecut, basis.fft_size, basis.variational,
+                                   basis.kcoords_global, basis.ksymops_global,
+                                   basis.kgrid, basis.kshift, basis.symmetries,
+                                   basis.comm_kpts)
+        ρ = DFTK.compute_density(new_basis, scfres.ψ, scfres.occupation)
+        energies, _ = energy_hamiltonian(new_basis, scfres.ψ, scfres.occupation; ρ, scfres.eigenvalues, scfres.εF)
+        energies.total
+    end
+    ForwardDiff.gradient(M -> HF_energy((I+M) * scfres.basis.model.lattice),
+                         zeros(eltype(scfres.basis), 3, 3)) / scfres.basis.model.unit_cell_volume
```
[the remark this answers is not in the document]
antoine-levitt: can't use det(L) for 1D/2D systems

test/stresses.jl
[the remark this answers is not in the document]
antoine-levitt: ... actually I can't because we don't have generic LDA for collinear

antoine-levitt:
Review done, thanks!

test/stresses.jl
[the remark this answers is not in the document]
mfherbst: Ah yes ... that's a good point.

src/symmetry.jl
mfherbst: True, but in this setup you do a lot of duplicate work `unfold_mapping` is computed each `unfold_array` call. I can see us quickly overusing `unfold_BZ`, so I'm not too happy about this structure. If you have the unfolding as the outer loop it's still easy to add an extra line if you have an additional entry in your scfres.

src/symmetry.jl
mfherbst: Still a bit cryptic, especially the `[2]`. If you assign to a variable you can use `_, bla = ` and that makes clearer what happens.

mfherbst:
> Well the irreducible BZ gets unfolded into the reducible BZ, so the name works, no? Nothing is done to the symmetries field of scfres, so disable_symmetries would be a bit misleading

ok yes, it's only in the scfres where things are really happening that is true. So then `unfold_bz_scfres` (to match with our capitalisation convention in function calls).

src/symmetry.jl
antoine-levitt: Should we just define a SymOp struct with S and tau? Any reason we didn't do that before?

antoine-levitt: Ah sorry that's not a symop. So we redefine apply_ksymop to return a namedtuple instead of two things?

src/symmetry.jl
antoine-levitt: Unfold_mapping is cheap. I definitely would wait until we have a better idea of the use cases for this before doing any refactoring. Not convinced merging functions would help here

antoine-levitt:
Not a fan of unfold_bz_scfres(scfres). The long-term plan is for scfres to get a struct, so that this becomes a dispatch.

src/symmetry.jl
mfherbst: yes that's a good idea.

src/symmetry.jl
mfherbst: OTOH `is_psi` flag is really annoying and the wrapping it into two function names as well. It feels very hackish and I'm not a fan of this solution. If you swap inner and outer loop than you have that info and do not need that strangle function structure. I think that is the easiest solution and as you say, when we need more we refactor.

mfherbst:
ok then leave off the `scfres` if you want.

src/symmetry.jl
antoine-levitt: actually simply unpacking is simplest here

src/symmetry.jl
antoine-levitt: I've removed the extra functions which were more confusing than anything (trying to pretend this is a high-level API even if it's pretty hacky). But I really don't want to merge the functions : this is already a pain to write and read, having an extra function does help I feel

PR #488, opened 2021-07-17
Extend ForwardDiff fallback for SVector norm to handle multiple partials

niklasschmitz:
Needed for https://github.com/JuliaMolSim/DFTK.jl/pull/487


antoine-levitt:
Thanks!

PR #489, opened 2021-07-17
Allow passing Ecut as a keyword argument

mfherbst:
Also makes the compiler emit a deprecation warning (if explicitly turned on) about the old interface being deprecated now.


antoine-levitt:
Ah you beat me to it, I was just doing this now

antoine-levitt:
I had a
```
## Note the implicit passing of keyword arguments here:
## this is equivalent to PlaneWaveBasis(model; Ecut=Ecut, kgrid=kgrid)
```
in tutorial.md because that might be confusing to new julia users

mfherbst:
Sorry ... I thought it was clear in the comment I left that I would do it.

mfherbst:
good point

antoine-levitt:
There's a bunch of other places we could replace the pattern X=X (or X = X) by X, but it has been too long since I used regexpes

mfherbst:
Yeah, I know, but I suggest we just fix it when we find it.

antoine-levitt:
Yup, just might have been a good candidate while we were search&replacing stuff

PR #492, opened 2021-07-18
Support band structures for 1D systems

mfherbst:


antoine-levitt:
Cool! Maybe print from - 1/2 to 1/2?

mfherbst:
Ok, that actually works as well. Good idea.

PR #493, opened 2021-07-18
Search for more symmetries in larger systems

mfherbst:
Funny how many bugs you find when you start playing around ... turns out 384 is sort of the lower bound, not the upper. I should have read the C code.


antoine-levitt:
Gotta love these C APIs...

Issue #483, opened 2021-07-14
Replace pymatgen by Brillouin.jl


mfherbst:
As it says, use https://github.com/thchr/Brillouin.jl for the things we currently need pymatgen for.

mfherbst:
The way to do this is to use [`spg_get_dataset`](https://spglib.github.io/spglib/api.html#spg-get-dataset-and-spg-get-dataset-with-hall-number) from spglib, which returns the [`SpglibDataset`](https://spglib.github.io/spglib/dataset.html#spglib-dataset) in which there is the `spacegroup_number`. This can then be used as the `sgnum` argument to call `Brillouin.irrfbz_path` (using the conventional vectors from spglib as the lattice vectors).

harrisonlabollita:
Hey Michael, I can take care of this!

mfherbst:
Awesome. Then go right ahead!

mfherbst:
Thanks very much!

harrisonlabollita:
Hi, 

Just to confirm, you wanted to replace pymatgen in the [band_structure.jl ](https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/postprocess/band_structure.jl) file with features from Brillouin.jl? 

Where there any other places to touch other than there?

Thanks

mfherbst:
Yes, basically the blob to replace is https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/postprocess/band_structure.jl#L17--L34. But you will not only need Brillouin.jl. You also need to add another c-call to `spglib` which returns a struct, so that struct also needs to be wrapped in Julia code.

harrisonlabollita:
Okay, yes! This is all done. I just wanted to confirm before I made a pull request. Thanks

mfherbst:
I left a comment. Does this work without a segfault?

harrisonlabollita:
Okay, yes I see your comment. Yes, that struct worked without a segfault, but it might be better to incorporate the whole dataset. Comparing to the [python interface to spglib](https://github.com/spglib/spglib/blob/develop/python/spglib/spglib.py), there is a function there which returns that dataset as a python dictionary, as well as another function which returns only the space group number (calls the get dataset function then grabs the space group key from there). Would this be better?

mfherbst:
So far our approach to the spglib interface was to keep it as simple as possible. But of course having a way to get the full dataset would be nice. We could also use at other places in DFTK, actually.

[...]

PR #496, opened 2021-07-19
Replace pymatgen by Brillouin.jl #483

harrisonlabollita:
Completed #483 (replacing pymatgen with Brillouin.jl). 

Details:
- created an spglib dataset struct in [external/sglib.jl](https://github.com/harrisonlabollita/DFTK.jl/blob/master/src/external/spglib.jl)
- new ccall function to grab dataset with wrapper function to just extract the space group number
- [replaced the pymatgen function calls with Brillouin.jl function calls](https://github.com/harrisonlabollita/DFTK.jl/blob/master/src/postprocess/band_structure.jl) and reformat the outputs to match the expectations of the existing code.
- Brillouin.interpolate expects N (an integer number of datapoints) so I've repurposed kline_density accordingly.


antoine-levitt:
Awesome, thanks for doing this! Will review tomorrow. I think you need to add Brillouin to the deps and load it? It should be as simple as `Pkg.instantiate(DFTK); Pkg.add("Brillouin")` and commit the resulting Project.toml

src/external/spglib.jl
```diff
@@ -162,3 +162,51 @@ function spglib_get_stabilized_reciprocal_mesh(kgrid_size, rotations::Vector;
...
+	wyckoffs::Ptr{Cint}
+	site_symmetry_symbols::Ptr{NTuple{7, Cchar}}
+	equivalent_atoms::Ptr{Cint}
+	crystallographic_orbits::Ptr{Cint}
+	primitive_lattice::NTuple{9, Cdouble}
+	n_std_atoms::Cint
+	std_lattice::NTuple{9, Cdouble}
+	std_types::Ptr{Cint}
+        std_positions::NTuple{3, Cdouble}
+	std_rotation_matrix::NTuple{9, Cdouble}
```
mfherbst, line 188: The spacing is not uniform. We use 4 spaces indention.

src/external/spglib.jl
```diff
@@ -162,3 +162,51 @@ function spglib_get_stabilized_reciprocal_mesh(kgrid_size, rotations::Vector;
...
+	std_mapping_to_primitive::Ptr{Cint}
+	pointgroup_symbol::NTuple{6, Cchar}
+end
+
+function spglib_get_dataset(lattice, atoms; tol_symmetry=1e-5)
+
+    # Convert lattice and atoms to spglib and keep the mapping between our atoms
+    spg_lattice = copy(Matrix{Float64}(lattice)')
+    # and spglibs atoms
+    spg_positions, spg_numbers, spg_spins, atommapping = spglib_atoms(atoms)
```
mfherbst, line 198: If you don't need the last few args, just write
```suggestion
    spg_positions, spg_numbers, _ = spglib_atoms(atoms)
```

src/external/spglib.jl
mfherbst: Same at a couple of other places :smile:.

src/external/spglib.jl
```diff
@@ -162,3 +162,51 @@ function spglib_get_stabilized_reciprocal_mesh(kgrid_size, rotations::Vector;
...
+	std_rotation_matrix::NTuple{9, Cdouble}
+	std_mapping_to_primitive::Ptr{Cint}
+	pointgroup_symbol::NTuple{6, Cchar}
+end
+
+function spglib_get_dataset(lattice, atoms; tol_symmetry=1e-5)
+
+    # Convert lattice and atoms to spglib and keep the mapping between our atoms
+    spg_lattice = copy(Matrix{Float64}(lattice)')
+    # and spglibs atoms
```
mfherbst, line 197: I think if you don't need the mapping, then you can drop the comment.

mfherbst: Oh and no empty line after the function definition.

src/postprocess/band_structure.jl
```diff
@@ -1,4 +1,5 @@
 using PyCall
```
mfherbst, line 1: PyCall still needed here?

src/postprocess/band_structure.jl
```diff
@@ -162,6 +175,17 @@ function detexify_kpoint(string)
     string
 end
 
+function retexify_kpoint(string)
```
mfherbst, line 178: Not sure about this. Probably we can dump both `retexify_kpoint` and `detexify_kpoint` now that pymatgen is out of the loop.

src/external/spglib.jl
```diff
@@ -162,3 +162,51 @@ function spglib_get_stabilized_reciprocal_mesh(kgrid_size, rotations::Vector;
...
+			spg_lattice, spg_positions, spg_numbers, 
+			length(spg_numbers), tol_symmetry)
+   return unsafe_load(spg_dataset)
+end
+
+
+function spglib_get_spacegroup(lattice, atoms)
+    	spg_lattice = copy(Matrix{Float64}(lattice)')
+	spg_dataset = spglib_get_dataset(lattice, atoms)
+	return spg_dataset.spacegroup_number, spg_lattice
```
mfherbst, line 211: Why do you need to return the lattice here?

mfherbst (commented):
Thanks very much for this PR. I made a quick review round. I'm surprised how little code is needed ... as often in the Julia world.

TODO (mostly to myself):
Now that pymatgen is no longer a hard dependency there are also a few other cleanup things one can do (e.g. remove it from the build script, take care of it in the `load_lattice`-type functions).

harrisonlabollita:
Recommended and suggested have been changed

mfherbst:
Thanks! I'm a bit tied up with Juliacon at the moment. I'll do a full pass on Friday.

antoine-levitt:
The tests fail, but possibly that's expected? In that case can you fix the test?

src/external/spglib.jl
```diff
@@ -162,3 +162,48 @@ function spglib_get_stabilized_reciprocal_mesh(kgrid_size, rotations::Vector;
...
+function spglib_get_dataset(lattice, atoms; tol_symmetry=1e-5)
+    # Convert lattice and atoms to spglib
+    spg_lattice = copy(Matrix{Float64}(lattice)')
+    spg_positions, spg_numbers, _ = spglib_atoms(atoms)
+    # Get the spglib dataset 
+    spg_dataset = ccall((:spg_get_dataset, SPGLIB), Ptr{SpglibDataset}, 
+			(Ptr{Cdouble}, Ptr{Cdouble}, Ptr{Cint}, Cint, Cdouble),
+			spg_lattice, spg_positions, spg_numbers, 
+			length(spg_numbers), tol_symmetry)
+   return unsafe_load(spg_dataset)
```
antoine-levitt, line 202: fun that this just works! So there's no memory issue (eg leak or invalid access) with this?

src/postprocess/band_structure.jl
```diff
@@ -19,18 +19,30 @@ function high_symmetry_kpath(model; kline_density=20)
     # https://github.com/louisponet/DFControl.jl/blob/master/src/structure.jl
     # but for this the best way to go would be to refactor into a small "CrystalStructure"
     # julia module which deals with these sort of low-level details everyone can agree on.
-    pystructure = pymatgen_structure(model.lattice, model.atoms)
-    symm_kpath = pyimport("pymatgen.symmetry.bandstructure").HighSymmKpath(pystructure)
-    kcoords, labels = symm_kpath.get_kpoints(kline_density, coords_are_cartesian=false)
-
+    # pystructure = pymatgen_structure(model.lattice, model.atoms)
+    # symm_kpath = pyimport("pymatgen.symmetry.bandstructure").HighSymmKpath(pystructure)
+    # kcoords, labels = symm_kpath.get_kpoints(kline_density, coords_are_cartesian=false)
```
antoine-levitt, line 24: just remove this?

src/postprocess/band_structure.jl
```diff
@@ -19,18 +19,30 @@ function high_symmetry_kpath(model; kline_density=20)
...
+    
+    sgnum = spglib_get_spacegroup(model.lattice, model.atoms)
+    Rs = [model.lattice[i, :] for i in 1:size(model.lattice,1)]
+    kp        = irrfbz_path(sgnum, Rs)
+    N         = kline_density*length(vcat(kp.paths...))
+    kcoords   = collect(Brillouin.interpolate(kp, N))
```
antoine-levitt, line 30: Coverage says that this is untested, can you make sure that it is?

src/postprocess/band_structure.jl
```diff
@@ -19,18 +19,30 @@ function high_symmetry_kpath(model; kline_density=20)
...
-    symm_kpath = pyimport("pymatgen.symmetry.bandstructure").HighSymmKpath(pystructure)
-    kcoords, labels = symm_kpath.get_kpoints(kline_density, coords_are_cartesian=false)
-
+    # pystructure = pymatgen_structure(model.lattice, model.atoms)
+    # symm_kpath = pyimport("pymatgen.symmetry.bandstructure").HighSymmKpath(pystructure)
+    # kcoords, labels = symm_kpath.get_kpoints(kline_density, coords_are_cartesian=false)
+    
+    sgnum = spglib_get_spacegroup(model.lattice, model.atoms)
+    Rs = [model.lattice[i, :] for i in 1:size(model.lattice,1)]
+    kp        = irrfbz_path(sgnum, Rs)
```
antoine-levitt, line 28: can you prefix with Brillouin. so that it's clear where this comes from? For this kind of dependency it's maybe useful to not `using` but `import`, so that it's clearer where it's used

src/postprocess/band_structure.jl
```diff
@@ -19,18 +19,30 @@ function high_symmetry_kpath(model; kline_density=20)
...
+    for (key, val) in kp.points
+        labels_dict[string(key)] = val
+    end
+    kpath = []
+    for path in kp.paths
+        new_path = []
+        for el in path
+            push!(new_path, retexify_kpoint(string(el)))
+	end
+        push!(kpath, new_path)
```
antoine-levitt, line 42: indentation is a bit off. Also can't you just do `new_path = retexify_kpoint.(string.(path))`?

antoine-levitt: similarly for the kpath above, looks like that logic can be simplified. Generally we're trying to move away from the "push!" idiom

src/postprocess/band_structure.jl
antoine-levitt: Agree, can't we just remove all this?

src/external/spglib.jl
harrisonlabollita: It's really cool that this just works, however, I do encounter a segfault when I try to unsafe_load just one of the objects in the spglib struct. I'm not really sure why this is the case. Looking into it now. Other than that everything works!

src/postprocess/band_structure.jl
```diff
@@ -19,18 +19,21 @@ function high_symmetry_kpath(model; kline_density=20)
...
     # julia module which deals with these sort of low-level details everyone can agree on.
-    pystructure = pymatgen_structure(model.lattice, model.atoms)
-    symm_kpath = pyimport("pymatgen.symmetry.bandstructure").HighSymmKpath(pystructure)
-    kcoords, labels = symm_kpath.get_kpoints(kline_density, coords_are_cartesian=false)
-
+    
+    sgnum = spglib_get_spacegroup(model.lattice, model.atoms)
+    Rs = [model.lattice[i, :] for i in 1:size(model.lattice,1)]
+    kp        = Brillouin.irrfbz_path(sgnum, Rs)
+    N         = kline_density*length(vcat(kp.paths...))
```
thchr, line 26: I suspect this isn't jiving exactly with the differing notions of k-point density here: Brillouin's `interpolate` will try to distribute `N` points as equidistantly as possible across the entire path (with the distance metric evaluated in Cartesian space) - but `length(vcat(kp.paths))` will just give the number of unconnected paths overall. 
Maybe what you want a measure of the [overall path length](https://github.com/thchr/Brillouin.jl/blob/212eec65e97b3b9dcacfea9d02828b8c70ac6f55/src/interpolate-paths.jl#L102-L108).

harrisonlabollita: Yes, this would be much better, I think. Thanks!

src/external/spglib.jl
```diff
@@ -162,3 +162,82 @@ function spglib_get_stabilized_reciprocal_mesh(kgrid_size, rotations::Vector;
...
+function spglib_get_dataset(lattice, atoms; tol_symmetry=1e-5)
+    # Convert lattice and atoms to spglib
+    spg_lattice = copy(Matrix{Float64}(lattice)')
+    spg_positions, spg_numbers, _ = spglib_atoms(atoms)
+    # Get the spglib dataset 
+    spglib_dataset = ccall((:spg_get_dataset, SPGLIB), Ptr{SpglibDataset}, 
+	(Ptr{Cdouble}, Ptr{Cdouble}, Ptr{Cint}, Cint, Cdouble),
+	spg_lattice, spg_positions, spg_numbers, 
+	length(spg_numbers), tol_symmetry)
+    return unload_spglib_dataset(spglib_dataset)
```
mfherbst, line 236: You need to call `spg_free_dataset` to avoid a memory leak as `spg_get_dataset` returns a pointer to a dynamically allocated object on the C side. `unsafe_load` only converts this pointer to Julia, but does not take care to clean it up.

src/external/spglib.jl
```diff
@@ -162,3 +162,82 @@ function spglib_get_stabilized_reciprocal_mesh(kgrid_size, rotations::Vector;
...
+    n_std_atoms::Cint
+    std_lattice::NTuple{9, Cdouble}
+    std_types::Ptr{Cint}
+    std_positions::NTuple{3, Cdouble}
+    std_rotation_matrix::NTuple{9, Cdouble}
+    std_mapping_to_primitive::Ptr{Cint}
+    pointgroup_symbol::NTuple{6, Cchar}
+end
+
+function unload_spglib_dataset(dataset)
```
mfherbst, line 194: I get your idea to abstract the C interface completely, but I think
- this is a lot of boilerplate code we don't really need
- to truly abstract the interface you also need to take of unit conversions etc, which is quite a pain and not worth it in my opinion.

I would suggest to consider the dataset internal from the point of view of spglib.jl, i.e. inside spglib.jl the c-types and spglib conventions are ok to be used, so you don't need to convert to julia types. The things we need outside e.g. `spglib_get_spacegroup` for those we do the conversion to Julia types selectively, but only there.

src/postprocess/band_structure.jl
mfherbst: @harrisonlabollita Just to rule out another common pitfall: Have you taken unit conversions into account? Notice my comment about the `kline_density` in the 1D case for example: The kline density is in Angströms. I'm not sure what unit system is used for Brillouin. If I get it right from quickly looking at the code the carthesian coords in Brillouin are essential the units used by the lattice (right @thchr ?). So in that case you have a unit clash because our lattice is given in Bohrs. 

Also the kp is in fractional coords, so you need to make user to use cartesian lengths to compute the number of kpoints. 

OTOH it might make sense for Brillouin.jl to add a method, that takes a desired k-point density and does all the magic @thchr?

src/external/spglib.jl
mfherbst: Yes there is a memory issue. See my comment.

src/external/spglib.jl
harrisonlabollita: Okay I will work on this! So something like this probably?

`return ccall((:spg_free_dataset, SPGLIB), Ptr{SpglibDataset}, (Ptr{SpglibDataset} ,), spglib_dataset)
`

src/external/spglib.jl
harrisonlabollita: Okay! Once the memory issue is resolved I will fix this too

src/postprocess/band_structure.jl
thchr: Hmm, Brillouin doesn't make a concrete choice of any unit system: the impact of units should carry over through whatever you specify in the lattice vectors `Rs`: by default, `irrfbz_path` will return a **k**-path specified in reduced coordinates (i.e. in the basis of *primitive* reciprocal lattice). Similarly, if you pass that path to `interpolate`, the resulting interpolation points will be in reduced coordinates.
Internally, Brillouin's `interpolate` will first convert to cartesian coordinates to figure out the distances between segments in the **k**-path properly (just to be in an orthonormal basis) - but what it returns will be in reduced coordinates.

It might make sense for Brillouin to have an version of `interpolate` that takes a `density` kwarg - but I'm not completely sure about how to specify density meaningfully in the absence of a concrete unit system choice.

(The reason Brillouin doesn't have a default unit choice is that I'm in photonics where the units end up cancelling out due to a frequency-scale invariance)

thchr: While I remember it: one other possible convention pitfall is that Brillouin's `irrfbz_path` requires that you specify the direct (i.e. real-space) basis in a *conventional* setting (e.g., for, say, diamond this would be a cubic lattice) - but returns the reciprocal path in the basis of the *primitive* reciprocal basis.

This is just to say that it is important that `Rs` refers to the conventional direct basis (even though the unit cell you'd use in your calculation would be in the primitive direct basis). This was basically just adopted from the convention that SeeK-path went with. I don't know which setting spglib returns the basis in - so I just wanted to point this out clearly. 

(Right now, Brillouin doesn't have methods to transform back and forth between conventional/primitive settings for direct bases, only for reciprocal bases - but if you need that, I can add it)

harrisonlabollita:
In the spirit of keeping the spglib dataset within C, I've rewritten it such that I'm only accessing what is needed. In this case, the space group number and the standard lattice. I wasn't aware of this earlier, but there does exist a [Spglib Julia wrapper already](https://github.com/singularitti/Spglib.jl/tree/master/src), which has the full interface to the dataset implemented. Maybe it's worth using this, maybe not? The less dependencies the better probably?

I've written my own interpolate to generate the k points in fractional coordinates in the same style that pymatgen generates their k-points. Perhaps, this is something that could be added to Brillouin.jl directly? What do you think @thchr? Here is the commit cf39dd3

thchr:
> I've written my own interpolate to generate the k points in fractional coordinates in the same style that pymatgen generates their k-points. Perhaps, this is something that could be added to Brillouin.jl directly? What do you think @thchr? Here is the commit [cf39dd3](https://github.com/JuliaMolSim/DFTK.jl/commit/cf39dd3a0306c44fdff2b8389f49d11c0d044036)

Yeah, this seems nice to add; I'll add a variant of that (I think your version inserts the "end" points twice - but [the fix is minimal](https://gist.github.com/thchr/41f981382851b98268bcad9bf60bde59)). I think a signature like `Brillouin.interpolate(kp; density)` with `density` kwarg might make sense?

thchr:
@harrisonlabollita OK, I [added a method `interpolate(kp; density)`](https://github.com/thchr/Brillouin.jl/commit/afeecacfe0d7816328c0dc56e2b1f5f24179cb93) which should do what you want (I bumped the version of Brillouin to 0.4.1; it will take a few min before it's registered).
You probably still need to `collect` the result to get just the points themselves rather than the whole `KPathInterpolant` structure.

harrisonlabollita:
Awesome! Thanks so much

src/external/spglib.jl
```diff
@@ -162,3 +162,53 @@ function spglib_get_stabilized_reciprocal_mesh(kgrid_size, rotations::Vector;
...
+	(Ptr{Cdouble}, Ptr{Cdouble}, Ptr{Cint}, Cint, Cdouble),
+	spg_lattice, spg_positions, spg_numbers, 
+	length(spg_numbers), tol_symmetry)
+    return unsafe_load(spglib_dataset)
+end
+
+function spglib_get_std_lattice(lattice, atoms)
+    spglib_dataset = spglib_get_dataset(lattice, atoms)
+    return transpose(reshape(collect(spglib_dataset.std_lattice), 3, 3))
+end
```
mfherbst, line 209: Should be the same as `spglib_standardize_cell` I think  (to be checked), so we should only keep one of them.

src/external/spglib.jl
```diff
@@ -162,3 +162,53 @@ function spglib_get_stabilized_reciprocal_mesh(kgrid_size, rotations::Vector;
...
+
+function spglib_get_spacegroup(lattice, atoms)
+    spglib_dataset = spglib_get_dataset(lattice, atoms)
+    return spglib_dataset.spacegroup_number
```
mfherbst, line 213: No return needed. The return of the last statement is automatic in julia.

Also: You should convert from `Cint` to `Int` to avoid bleeding out c-types beyond `spglib.jl`.

src/postprocess/band_structure.jl
```diff
@@ -19,18 +19,20 @@ function high_symmetry_kpath(model; kline_density=20)
...
     # but for this the best way to go would be to refactor into a small "CrystalStructure"
     # julia module which deals with these sort of low-level details everyone can agree on.
-    pystructure = pymatgen_structure(model.lattice, model.atoms)
-    symm_kpath = pyimport("pymatgen.symmetry.bandstructure").HighSymmKpath(pystructure)
-    kcoords, labels = symm_kpath.get_kpoints(kline_density, coords_are_cartesian=false)
+    std_lattice = spglib_get_std_lattice(model.lattice, model.atoms)
+    sgnum = spglib_get_spacegroup(model.lattice, model.atoms)
+    Rs = [std_lattice[i, :] for i in 1:size(std_lattice,1)]
+    kp = Brillouin.irrfbz_path(sgnum, Rs)
+    kcoords = Brillouin.interpolate(kp, density=kline_density)
```
mfherbst, line 26: I think there is still a unit problem here (in lattice versus kline_denisty) and also there is the edge with the respective lattices to figure out (conventional versus primitive ...). I'll do some research on this.

test/compute_bands.jl
```diff
@@ -7,100 +7,88 @@ if mpi_nprocs() == 1  # not easy to distribute
...
-        [0.375000000000, 0.375000000000, 0.750000000000],
-        [0.625000000000, 0.250000000000, 0.625000000000],
-        [0.600000000000, 0.200000000000, 0.600000000000],
-        [0.575000000000, 0.150000000000, 0.575000000000],
-        [0.550000000000, 0.100000000000, 0.550000000000],
-        [0.525000000000, 0.050000000000, 0.525000000000],
-        [0.500000000000, 0.000000000000, 0.500000000000],
+        [0.025000000000, 0.000000000000, 0.025000000000],
+        [0.050000000000, 0.000000000000, 0.050000000000],
+	[0.075000000000, 0.000000000000, 0.075000000000],
```
mfherbst, line 15: I am dubious about this. If everything is done right you should get exactly the same k-points, no?

test/compute_bands.jl
```diff
@@ -125,9 +113,9 @@ if mpi_nprocs() == 1  # not easy to distribute
     for key in keys(ref_klabels)
         @test klabels[key] ≈ ref_klabels[key] atol=1e-15
     end
-
-    @test kpath[1] == [raw"\Gamma", "X", "W", "K", raw"\Gamma", "L", "U", "W", "L", "K"]
-    @test kpath[2] == ["U", "X"]
+    
+    @test kpath[1] = ["Γ", "X", "U"]
+    @test kpath[2] = ["K", "Γ", "L", "W", "X"]
```
mfherbst, line 118: Why is this different?

src/external/spglib.jl
mfherbst: I'm not sure about the return type, but otherwise that looks good. You might need to tell Julia to get the pointer, but it could also be that this is automatic, just try it.

src/postprocess/band_structure.jl
mfherbst: Yes this could be an issue @thchr Thanks for pointing that out. I'll do some research about the defails.

mfherbst:
> In the spirit of keeping the spglib dataset within C, I've rewritten it such that I'm only accessing what is needed. 

Perfect! I think that looks much better now.

> In this case, the space group number and the standard lattice. I wasn't aware of this earlier, but there does exist a Spglib Julia wrapper already, which has the full interface to the dataset implemented. Maybe it's worth using this, maybe not? The less dependencies the better probably?

I know of this wrapper, but it is basically unmaintained. Since we don't need all the features of spglib, we decided to run our own minimal interface. Of course if someone took over and maintained LibSymspg.jl than we could use that, too.

antoine-levitt:
Note this is a new wrapper : https://github.com/singularitti/Spglib.jl

mfherbst:
Ah great. Sure, we could use Spglib.jl. That sounds like a good idea.

mfherbst:
Ok, so `spglib_standardize_cell` (with `primitive=false`, the default) will return the conventional unit cell and not the primitive one, so exactly what `irrfbz_path` needs, but then the returned k-points are in fractional coordinates wrt. the lattice returned by `spglib_standardize_cell` (with `primitive=true`). Naturally that might not match the lattice actually used by DFTK. I know pymatgen throws a warning in such cases, which is also what we should do.

So we use `spglib_standardize_cell` (with `primitive=false`) to get the cell to be passed to Brillouin, check the lattice of DFTK against `spglib_standardize_cell` (with `primitive=true`) to throw an error and pass the fractional k-points from `Brillouin.interpolate(kp; density)` onwards to compute the bands.

test/compute_bands.jl
thchr: The [default path type choice for pymatgen seems to be the "SC" paths](https://github.com/materialsproject/pymatgen/blob/b789d74639aa851d7e5ee427a765d9fd5a8d1079/pymatgen/symmetry/bandstructure.py#L55) rather than SeeK-path/Hinuma (which is what Brillouin returns) - so that would cause a difference. The Hinuma paths are supposed to be more optimal, however.

mfherbst: Ok thanks @thchr, that clarifies it. In that case I will make a commit with the Hinuma testdata from pymatgen to allow proper. comparison.

harrisonlabollita:
Just to summarize the latest commits, I've switched to using Spglib.jl to get the space group number and the standard and primitive lattices. I've also added the quick check to make sure that the DFTK lattice matches the spglib primitive lattice, it will throw an error if they disagree.

mfherbst:
@harrisonlabollita Thanks. I'll have a closer look tmr.

> it will throw an error if they disagree.

In this case I would not be so harsh and just `@warn`, because sometimes you might want to do this on purpose.

harrisonlabollita:
Sure I can make it a warning with a nice message for the user!

src/postprocess/band_structure.jl
```diff
@@ -1,7 +1,6 @@
-using PyCall
-
+import Brillouin
+using Spglib
```
mfherbst, line 2: I would similarly vouch for an `import` instead of a `using` here.

src/postprocess/band_structure.jl
```diff
@@ -19,18 +18,32 @@ function high_symmetry_kpath(model; kline_density=20)
...
+    # get standardized cell and spacegroup number for Brillouin.jl 
+    spg_lattice = standardize_cell(structure, to_primitive=false).lattice
+    # comparge spglibe primitive to DFTK primitive
+    spg_primitive = standardize_cell(structure, to_primitive=true).lattice
+    spg_primitive ≈ transpose(model.lattice) || @warn "The DFTK lattice and spglib's primitive lattice do not agree."
+   
+    # get spacegroup num from spglib dataset
+    sgnum = get_spacegroup_number(structure)
+    
+    Rs = [spg_lattice[i, :] for i in 1:size(spg_lattice,1)]
```
mfherbst, line 35: Careful here: I think spglib uses lattice vectors as rows and not as columns (as we do in DFTK). Since all this conversion between different conventions gets pretty messy, I would suggest to do the wrapping of `Spglib.jl` to match DFTK conventions also in `spglib.jl`, where we actually already have a  `spglib_standardize_cell`  that you can adapt. 

Eventually we will replace all calls of DFTK to spglib_jll by calls to `Spglib.jl` so I think it would be good to keep all calls to spglib at a central place for now.

harrisonlabollita: I think Brillouin expects Rs to be a list of lattice vectors. This is why I create this from the spg_lattice, which is already formatted in rows. I will double check everything and move this into external/spglib.jl library so that everything that happens there and cleans up the band structure file.

mfherbst:
@harrisonlabollita I fixed a subtle remaining unit issue and made sure the function really replaces behaviour of pymatgen (which is a bit strange in the sense that it duplicates some k-points). I also reorganised the code a little.

I'm not fully happy with the structure as it is, but fixing that will require a more general reorganisation of the bandstructure code, which we might as well do in a follow-up PR. If the tests pass, I suggest we cleanup the PR, rebase against master and then just merge as is. Other thoughts?

harrisonlabollita:
@mfherbst I think that sounds like a plan! There is one issue that I haven't gotten a chancee to fix yet and that's the k-points generated by Brillouin.jl differ from the test_kpoints that you generated using pymatgen with the SeekPath option. I'm still trying to track down why the two arrays of k-coords are different, but once this is resolved I think we can proceed.

mfherbst:
I fixed the tests by committing to the output kpath produced by Brillouin.jl. By eye it looks similar and it seems the differences are only due to the slightly different ways the paths are constructed in pymatgen and Brillouin.jl. I have to admit I am not fully convinced this is the right thing to do, however. In particular when computing bands for supercells I have not yet verified the right thing happens.

So we should do a little more verification before merging this.

mfherbst:
I started to do some testing. For simple cases (e.g. primitive silicon unit cell) the obtained bands now look visually identical, save some differences in the precise kpoints that are selected along the paths. @harrisonlabollita good job!
Unfortunately for more complicated things there are some differences. My current test case is a silicon supercell.

- For reference I use the DFTK branch [kpathsc](https://github.com/JuliaMolSim/DFTK.jl/tree/kpathsc) in which I switch pymatgen to use Hinuma kpath definitions (which actually down the line dispatches to the `seekpath` python package). The [reference input file](https://github.com/JuliaMolSim/DFTK.jl/blob/kpathsc/examples/silicon_test_bands.jl) thus uses this kpath from `seekpath` 
and the [test input file](https://github.com/harrisonlabollita/DFTK.jl/blob/master/examples/silicon_test_bands.jl) uses the one from Brillouin.jl
- I compared the computed bands visually and they are noticably different on this example.
- I started comparing the fractional kpoints from both paths and they differ as well.
- I did some digging in `pymatgen` and found [this code block](https://github.com/materialsproject/pymatgen/blob/b789d74639aa851d7e5ee427a765d9fd5a8d1079/pymatgen/symmetry/kpath.py#L1021--L1060), which seems to do exactly the reordering of the fractional k-Point coordinates that explains the difference in the coordinates between both k-Paths.

As I have feared there seem to be a lot more subtleties that pymatgen does behind the back, but unfortunately it's not really explained why this is done.

@thchr Do you understand why pymatgen is doing this? Is it a reasonable thing to do?

thchr:
Hmm, yeah, I guess there are some subtleties which the "battle-hardened" nature of pymatgen guards against...

I think there's two main things that could cause differences in paths, one trivial, one not-so-trivial:
- **Interpolation:** It seems like pymatgen intentionally includes each "terminal" point of a segment twice, i.e. it is included at the start and end of each segment. Brillouin doesn't do that, since it seemed like wasted computational effort to me. @harrisonlabollita [had a PR that could make Brillouin do that](https://github.com/thchr/Brillouin.jl/pull/8) - we can revive that if you need it.
- **Setting differences:** This is probably the more fundamentally subtle source of potential problems (and the one you highlight above). In Brillouin, I basically just used the source code of SeeK-path and [their associated paper](https://doi.org/10.1016/j.commatsci.2016.10.015) as a guide - and from what I recall, the main point is that SeeK-path follows the International Table of Crystallography Vol A (ITA) conventions for setting choices. I.e. Brillouin does the same: cells are assumed to be in the primitive ITA setting (if there are counterexamples, this would be a bug in Brillouin then). I believe that this is *not* the choice made by SC - there's some differences (which are enumerated in the first several tables of the SeeK-path paper).
  I'm guessing those differences are what's being handled in the pymatgen code you linked @mfherbst . I think the origin of these differences could be understood better by (re)reading the SeeK-path paper's introductory sections, in principle.

harrisonlabollita:
@mfherbst I'm not entirely sure, what pymatgen's ``_trans_sc_to_Hin(sub_class)`` is actually doing. It seems the only call to this method in the ``KPathSeek`` class is to this [``_tmat``](https://github.com/materialsproject/pymatgen/blob/5ba1d857cfe1867913e1d7981d2285e79a9233e0/pymatgen/symmetry/kpath.py#L999) , but it doesn't look like this matrix is used anywhere else?

mfherbst:
Sorry about the delay. As @thchr recommended I read the paper by Hinuma, which was quite insightful. Fascinating to see that the question of how to plot band structures is still debated in literature. I was not aware of many of the subtleties.

### Interpolation
> It seems like pymatgen intentionally includes each "terminal" point of a segment twice, i.e. it is included at the start and end of each segment. 

Indeed, but that difference I already corrected for in my workaround (DFTK currently relies on this, but I agree it's wasted computation, so I will change that in the future). The "some differences in the precise kpoints that are selected along the path" I referred to are actually not this doubling, there are other differences. What I refer to is [this diff](https://github.com/JuliaMolSim/DFTK.jl/pull/496/commits/15da46af1e94223608448251d46876e83c5a79fe#diff-1984e98219ac13cd4a5d55184363c46aea594e7d1809b7a57735e466484518e4) where in some segments slightly more and in others slightly less k-points are used.

### Setting differences
> I'm guessing those differences are what's being handled in the pymatgen code you linked

From heaving looked at the paper I would also conclude that this is the case. The key point is that for certain classes of lattices the choice of conventional  lattice simply differs between spglib/Brillouin/Hinuma (all the same conventions for conventional lattice) and Setyawan-Curtarolo/pymatgen (again both the same, but different from the three others). Similarly the choice of primitive cell is not identical between these groups. The code section I link is essentially to account for that, i.e. to backtransform the fractional k-points in the Hinuma primitive cell to the fractional k-Points in the SC primitive cell (that is the second column of Table 5 of the Hinuma paper).

With that researched it is not surprising to obtain different bands, I'd say.

### Conclusion
I therefore think it's ok to fix the tests as I did (i.e. by switching to the output of Brillouin) and accept that some different band structures will now result. What is important is to check that DFTK is internally consistent, that is we fully commit to the Hinuma conventions.

I'll have to check the details again, but I'd say there is one aspect where we need something extra from Brillouin.jl @thchr, namely the primitive unit cell in the direct basis. The point is that at this point according to the Hinuma paper spglib and the Hinuma conventions differ. Therefore [this line](https://github.com/JuliaMolSim/DFTK.jl/blob/04836859edc88081619185ede1a12aa1e832095c/src/postprocess/band_structure.jl#L21) is actually not correct and should be replaced by the primitive cell used in Brillouin.jl. How could one get that information easily?

thchr:
Hi @mfherbst: Thanks for the nice summary. Regarding the last point: 
> [...] What is important is to check that DFTK is internally consistent, that is we fully commit to the Hinuma conventions.
> 
> I'll have to check the details again, but I'd say there is one aspect where we need something extra from Brillouin.jl @thchr, namely the primitive unit cell in the direct basis. The point is that in this aspect according to the Hinuma paper spglib and the Hinuma conventions differ. Therefore this line is actually not correct and should be replaced by the primitive cell used in Brillouin.jl. How could one get that information easily?

I'm not sure I completely understand: is spglib not following the International Table of Crystallography Vol. A (ITA) conventions? My impression was that both Hinuma and spglib followed ITA conventions. What I did was to *assume* that input is given consistently with ITA conventions, and then to do conventional <-> primitive transforms according to the ITA setting matrices (see [here](https://github.com/thchr/Brillouin.jl/blob/abf55d0f17e74a5e369eb7292c50020c344434c1/src/CrystallineBravaisVendor.jl#L71-L85)).

I basically copied over code from my other package, [Crystalline.jl](https://github.com/thchr/Crystalline.jl), to transform between conventional and primitive settings (to avoid depending on that package, which seemed overkill - ideally, I should factor out some stuff to some BravaisCore module or similar). 
I think you're right that Brillouin doesn't currently expose a function to transform the direct lattices, only the reciprocal lattices. I could add the latter (it's already in Crystalline) - but I want to make sure we agree about the Hinuma/spglib/ITA conventions first.

mfherbst:
> I'm not sure I completely understand: is spglib not following the International Table of Crystallography Vol. A (ITA) conventions? My impression was that both Hinuma and spglib followed ITA conventions.

For the conventional cell yes, the crystallographic conventional cell is used, but for the primitive a different convontion is used as the transformation matrices differ according to the Hinuma paper (see the remark below figures 7 and 9 on page 176 as well as the remark below Tables 65 and 66 on page 166).

> ideally, I should factor out some stuff to a BravaisCore module or similar).

I think that sounds like a good idea in the long run. 

> I could add the latter

That would be great and solve the remaning issue, thanks @thchr !

mfherbst:
@thchr @harrisonlabollita I was just wondering about the current status of this and what remains to be done before we can merge.

If I recall correctly it's only some way to get the primitive unit cell in the direct basis from Brillouin.jl which is missing. Am I right? If so, how much effort would it be to add this to Brillouin? Would that be something that could be quickly achieved in a PR or so? (I would not mind doing it if it is not too much effort).

harrisonlabollita:
I think @thchr mentioned above about having the code to do this written Crystalline.jl? Would this be a simple copy/paste @thchr?

thchr:
Sorry for dropping this bit on the floor here: I'll try to add a method to get the primitive unit cell basis from a conventional basis to Brillouin.jl next week. The transformation matrices are already implemented, and the [methods are already there for reciprocal bases](https://github.com/thchr/Brillouin.jl/blob/abf55d0f17e74a5e369eb7292c50020c344434c1/src/CrystallineBravaisVendor.jl#L102-L115) but just not for direct bases (they transform differently). There's already an [implementation in Crystalline.jl](https://github.com/thchr/Crystalline.jl/blob/5fee82840f95edd088591553e5bce3c86487ae3f/src/bravais.jl#L470-L474).

I think what stopped me from implementing this before was confusion about whether Hinuma follow the standard settings or not; it seems mostly so, but not entirely. I can see I had [noted something about that earlier in Crystalline.jl](https://github.com/thchr/Crystalline.jl/blob/5fee82840f95edd088591553e5bce3c86487ae3f/src/bravais.jl#L351-L362) but I need to revisit it.

thchr:
To avoid duplicating the same functionality across Brillouin.jl and Crystalline.jl, I've factored out the basis/setting-change functionality into a new small package Bravais.jl which lives in a [subdirectory of Crystalline.jl](https://github.com/thchr/Crystalline.jl/tree/master/Bravais).
It should be [registered in a few days](https://github.com/JuliaRegistries/General/pull/46168) and I'll then transition Brillouin.jl to depend on that.

With Bravais.jl, the syntax you'll probably find relevant here will be: 
- To transform a conventional direct basis `Rs` (which should ideally be an `SVector{D, SVector{D, Float64}}` but could also just be plain nested vectors) to a primitive setting, use `primitivize(DirectBasis(Rs))` (or `primitivize(DirectBasis{D}(Rs))` if the dimensionality `D` isn't inferable from `Rs`).
- To take a primitive direct basis to a conventional setting, use `conventionalize` instead of `primitivize`. 
- Similarly, to take transform a reciprocal basis instead, use `ReciprocalBasis` instead of `DirectBasis`.

Currently, Bravais.jl will assume the standard settings from the International Tables of Crystallography Volume A (equivalently, the setting choices followed on the Bilbao Crystallographic Server, see e.g. this [Table 2 of this paper](https://doi.org/10.1107/S205327331303091X)).
There's still the question of determining precisely when this doesn't coincide with Hinuma's setting choices (as far as I can tell, there's differences only in the 'A' and 'C' centering settings). I'll update here when I figure out the implications of that, if any.

mfherbst:
Thanks for the effort and update @thchr! Sounds good.

thchr:
> Awesome @thchr ! Once Brillouin.jl is updated, I can make the minor tweaks to the `band_structure.jl` file in DFTK

It's registered now (v0.5.0) 👍: https://github.com/JuliaRegistries/General/pull/46365

mfherbst:
> but I'm unsure why the latter results in a segfault when its called?

Me too, that's why I decided to leave as is for now, push through this PR and pick it up properly in a follow-up. But of course, feel free to investigate (might just be a very annoying time-sink).

> Maybe I missed a previous comment or discussion about this, but for this code that doubles the number of points, what needs to be done to remove this as the TODO indicates?

It's a bit annoying and quite specialised code I had to write to adhere to what pymatgen does. I'm happy to take care of that. It's to do with this [really annoying function](https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/postprocess/band_structure.jl#L59--L88) If you want to do it @harrisonlabollita that would be great, but I still suggest to do that in a follow-up PR.

PR #497, opened 2021-07-20
Small updates to the documentation

mfherbst:


docs/src/guide/parallelization.md
```diff
@@ -109,10 +109,12 @@ DFTK.mpi_master() || (redirect_stdout(); redirect_stderr())
 ```
 at the top of your script to disable printing on all processes but one.
 
-!!! warning "MPI-based parallelism is experimental"
-    Even though MPI-based parallelism shows the better scaling it is still
-    experimental and some routines (e.g. band structure and direct minimization)
-    are not compatible with it yet.
+!!! info "MPI-based parallelism not fully supported"
+    While standard procedures (such as the SCF or band structure calculations)
+    fully support MPI, not not all routines of DFTK are compatible with MPI yet.
```
antoine-levitt, line 114: not not

antoine-levitt: Add a reassurance that these routines error rather than give wrong results

PR #498, opened 2021-07-21
Simplify defining custom elements

mfherbst:


examples/custom_potential.jl
```diff
@@ -7,44 +7,37 @@
...
 
-# We need to extend two methods to access the real and Fourier forms of
-# the potential during the computations performed by DFTK
-function DFTK.local_potential_fourier(el::ElementCustomPotential, q::Real)
-    return el.pot_fourier(q)
+# We extend the two methods providing access to the real and Fourier
+# representation of the potential to DFTK.
+
+function DFTK.local_potential_fourier(el::ElementGaussian, q::Real)
+    el.α * exp(- (q * el.L)^2 / 4)
```
antoine-levitt, line 21: Normalizations look fishy here, there's no pi?

mfherbst: Yes I think they are, but I copied from Gaspard ...

mfherbst: Let me take a look.

examples/custom_potential.jl
```diff
@@ -7,44 +7,37 @@
...
-# We set the potential in its real and Fourier forms
-pot_real(x) = exp(-(x/L)^2)
-pot_fourier(q::T) where {T <: Real} = exp(- (q*L)^2 / 4);
-# And finally we build the elements and set their positions in the `atoms`
-# array. Note that in this example `pot_real` is not required as all applications
-# of local potentials are done in the Fourier space.
-nucleus = ElementCustomPotential(pot_real, pot_fourier)
-atoms = [nucleus => [x1*[1,0,0], x2*[1,0,0]]];
+x2 = 0.8
+nucleus = ElementGaussian(1.0, 0.5)
```
antoine-levitt, line 37: might as well give them a negative potential to get bound states

mfherbst: For GPE?

examples/custom_potential.jl
antoine-levitt: The convention we use is V(q) = int_{R^3} V(x) e^{-iqx} dx (documented in elements.jl) so very likely there's a factor missing

antoine-levitt: @gkemlin :-)

examples/custom_potential.jl
antoine-levitt: Yeah, sure, this GPE is not very different from the linear model. The current examples gets scattering states (which is fine in a finite box), but would be easier to see with bound states.

PR #499, opened 2021-07-22
Add some missing fallbacks for interval arithmetic

mfherbst:


mfherbst:
There is something wrong with the Windows CI that I don't think it's our fault. I'm just going to ignore that for now, see #501.

PR #500, opened 2021-07-22
Separate fft normalizations from plans

antoine-levitt:


src/PlaneWaveBasis.jl
```diff
@@ -198,20 +197,16 @@ end
     ksymops_thisproc = ksymops[krange_thisproc]
 
     # Setup fft_size and plans
-    (ipFFT_unnormalized,  opFFT_unnormalized,
-     ipBFFT_unnormalized, opBFFT_unnormalized) = build_fft_plans(T, fft_size)
+    (ipFFT,  opFFT, ipIFFT, opIFFT) = build_fft_plans(T, fft_size)
```
mfherbst, line 200: spacing

src/PlaneWaveBasis.jl
```diff
@@ -65,16 +65,15 @@ struct PlaneWaveBasis{T <: Real}
...
 
     ## Plans for forward and backward FFT
-    # These plans follow DFTK normalization conventions (see above)
+    # All these plans are *completely unnormalized* (eg FFT * IFFT != I)
+    # The normalizations are performed in G_to_r/r_to_G according to
+    # the DFTK conventions (see above)
     opFFT   # out-of-place FFT plan
     ipFFT   # in-place FFT plan
-    opIFFT  # inverse plans
+    opIFFT  # inverse plans (unnormalized; BFFT in FFTW terminology)
```
mfherbst, line 73: use `opBFFT` and `ipBFFT` ?

src/PlaneWaveBasis.jl
```diff
@@ -438,20 +433,24 @@ In-place version of `G_to_r`.
 @timing_seq function G_to_r!(f_real::AbstractArray3, basis::PlaneWaveBasis,
                              f_fourier::AbstractArray3)
     mul!(f_real, basis.opIFFT, f_fourier)
+    f_real .*= basis.G_to_r_normalization
```
mfherbst, line 436: Use 5-argument version of `mul!` ?

src/PlaneWaveBasis.jl
```diff
@@ -438,20 +433,24 @@ In-place version of `G_to_r`.
...
     fill!(f_real, 0)
     f_real[kpt.mapping] = f_fourier
 
-    # Perform an FFT
-    mul!(f_real, plan, f_real)
+    # Perform an IFFT
+    mul!(f_real, basis.ipIFFT, f_real)
+    if !skip_normalization
+        f_real .*= basis.G_to_r_normalization
+    end
```
mfherbst, line 452: 5-argument `mul!` ?

src/PlaneWaveBasis.jl
antoine-levitt: what's wrong with the spacing?

src/PlaneWaveBasis.jl
antoine-levitt: yup good point

src/PlaneWaveBasis.jl
antoine-levitt: Not sure FFTW supports it, better to be more explicit here

src/PlaneWaveBasis.jl
antoine-levitt: same as above

src/PlaneWaveBasis.jl
mfherbst: two spaces after ","

src/PlaneWaveBasis.jl
mfherbst: I just thought if they will eventually then we don't have to think about it, but I'm pretty sure right now they don't.

src/PlaneWaveBasis.jl
antoine-levitt: oof good eye

mfherbst:
I changed the `skip_normalisation=true` interface to `normalisation=false` because it removes the strange double negations.

antoine-levitt:
Normalization feels like a number. Rename to normalize?

mfherbst:
ok

PR #509, opened 2021-07-29
Make LDOS mixing the default

mfherbst:
WIP idea for now. Will polish.

### DIFF
--- a/src/scf/chi0models.jl
+++ b/src/scf/chi0models.jl
@@ -6,6 +6,32 @@ import Base: @kwdef
 
 abstract type χ0Model end
 
+function DefaultAdjustTemperature()
+    total_energy = NaN
+    lasttemp = Inf
+    function callback(temperature; energies, n_iter, kwargs...)
+        isnothing(energies) && return temperature
+        iszero(temperature) && return temperature
+        ΔE = abs(total_energy - energies.total)
+        total_energy = energies.total
+
+        # adjustment factor going smoothly from init to 1
+        # as x crosses from 1e-2 to 1e-5
+        init = 0.25
+        fac = log10(init / temperature)
+        f(x) = 10^(0.5fac - 0.5fac * erf(-1 * (log10(x) + 3.5)))
+
+        temperature > init && return temperature
+        if isnan(ΔE) || n_iter ≤ 1
+            return init
+        else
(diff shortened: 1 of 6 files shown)


mfherbst:
@antoine-levitt I would merge this now if you agree.

antoine-levitt:
I think the heuristic is reasonable, although temp 0.5 feels too large. I'm still not 100% sure we want to make it the default : on small systems this might account for a non insignificant portion of the computing time, and will not have a large effect. Can you test on eg a small aluminum system?

mfherbst:
Well that depends what one wants the default to do. Right now our default is SimpleMixing, which is equivalent to LdosMixing if you don't use temperature. So for insulators this PR changes nothing. For metals or inhomogeneous systems our current default will fail on larger systems. The question to ask is then (a) do we want a safe default to work out of the box on most cases, but potentially with a runtime overhead or (b) do we want our default to work fast on most systems and then we force the user to choose something else manually in case of failure.  
I would say our usual philosophy of choosing safe defaults is more in line with (a), thus `LdosMixing` by default. Stops no one from switching to Kerker for metals if they want to.

About the timings: For our [magnesium example](https://juliamolsim.github.io/DFTK.jl/stable/examples/metallic_systems/) on most cases I tried the runtime spent on the LDOS is ok for a default method. It sits around 10 - 15% of the time share of the SCF for a number of reasonable settings over different Ecuts and kgrids I tried. Only for strange setups (Ecut=100, only Gamma point) it starts to become a real issue (up to 25% runtime), but I would not consider that too relevant.

About the maximal temperature: That is just a conservative safe-guard. On all calculations I did the maximal temperature was 0.25. My idea was to have this cap for cases you want to go to really high smearing temperatures (i.e. beyond 1e-2) to aid convergence.

mfherbst:
Also we could still gain a bit of time by forwarding guesses to the next SCF step and by more carefully tuning the convergence of the gmres ... I have not really played with that so far.

antoine-levitt:
I'm not completely sure; since the LDOS at the initial guess might be completely stupid there's the risk that it blows up for no reason. OTOH we do want people to try this out, so why not!

mfherbst:
Yes indeed. Getting more people to try it would be one of my hopes as well.

PR #513, opened 2021-08-12
Fix stresses for very small occupations

antoine-levitt:


antoine-levitt:
@niklasschmitz does this seem sensible?

niklasschmitz:
I think specifying asymptotic behavior in the presence of overflow in exponentials is fair game, similar examples are discussed eg in the [jax docs](https://jax.readthedocs.io/en/latest/notebooks/Custom_derivative_rules_for_Python_code.html#numerical-stability).

src/workarounds/forwarddiff_rules.jl
```diff
@@ -141,3 +141,14 @@ function LinearAlgebra.norm(x::SVector{S,<:ForwardDiff.Dual{Tg,T,N}}) where {S,T
     dy = ntuple(j->real(dot(x_value, ForwardDiff.partials.(x,j))) * pinv(y), N)
     ForwardDiff.Dual{Tg}(y, dy)
 end
+
+# problem: the derivative of 1/(1+exp(x)) = -exp(x) / (1+exp(x))^2. When x is too large, exp(x) = Inf and this is a NaN.
```
mfherbst, line 145: Line length.

src/workarounds/forwarddiff_rules.jl
```diff
@@ -141,3 +141,14 @@ function LinearAlgebra.norm(x::SVector{S,<:ForwardDiff.Dual{Tg,T,N}}) where {S,T
...
+function Smearing.occupation(S::Smearing.FermiDirac, d::ForwardDiff.Dual{T}) where {T}
+    x = ForwardDiff.value(d)
+    if exp(x) > floatmax(typeof(x)) / 1e3
+        der = -zero(x)
+    else
+        der = -exp(x) / (1+exp(x))^2
```
mfherbst, line 151: This is one of the cases, where spaces around "+" make it easier to parse.

src/workarounds/forwarddiff_rules.jl
```diff
@@ -141,3 +141,14 @@ function LinearAlgebra.norm(x::SVector{S,<:ForwardDiff.Dual{Tg,T,N}}) where {S,T
     dy = ntuple(j->real(dot(x_value, ForwardDiff.partials.(x,j))) * pinv(y), N)
     ForwardDiff.Dual{Tg}(y, dy)
 end
+
+# problem: the derivative of 1/(1+exp(x)) = -exp(x) / (1+exp(x))^2. When x is too large, exp(x) = Inf and this is a NaN.
+function Smearing.occupation(S::Smearing.FermiDirac, d::ForwardDiff.Dual{T}) where {T}
+    x = ForwardDiff.value(d)
+    if exp(x) > floatmax(typeof(x)) / 1e3
+        der = -zero(x)
```
mfherbst, line 149: der -> `∂occ` ?

src/workarounds/forwarddiff_rules.jl
antoine-levitt: really? I find it easier to parse without the spacing. But I don't care so OK

antoine-levitt:
OK; in that case we have do something similar for the other smearing types. Actually the simplest solution (at least for forward) is probably to special-case the primal so it returns 0 whenever overflow occurs and then let autodiff figure it out.

PR #516, opened 2021-08-17
Fix and reword symmetries documentation section.

mfherbst:


antoine-levitt:
Thanks for doing this. I still want to think about it more, feel free to modify it and I'll make a final pass when I'm back from vacations

mfherbst:
So I merge and you make another pass later?

antoine-levitt:
Yup!

PR #519, opened 2021-08-20
[WIP] Hellmann-Feynman derivatives using ChainRules and Zygote

niklasschmitz:
This draft PR contains a rough sketch of enabling Hellmann-Feynman derivatives via reverse mode automatic differentiation, using ChainRules and Zygote. This is a complementary approach to the previously implemented HF stresses using ForwardDiff in https://github.com/JuliaMolSim/DFTK.jl/pull/476. It is technically more challenging, partly also due to current limitations of Zygote (no in-place array mutation, no try-catch, no MPI support) but has the promise of enabling cheap force computations for large systems and machine learning applications.

Currently, this still requires quite a few workarounds as can be seen in `workarounds/chainrules.jl`: 
- `G_to_r` and `r_to_G` rrules are expressed in terms of each other, hiding non-Julia FFTW and mutation
- complicated constructors (with sanity checks & control flow) have simplified counterparts that only compute the differentiable fields and return a NamedTuple instead of a full struct. An easier example is `_autodiff_Model_namedtuple`, whereas `_autodiff_PlaneWaveBasis_namedtuple` is arguably currently the most complicated due to necessary term instantiation handling.
- few functions are currently explicitly marked as non-differentiable such as `ElementPsp` and `G_vectors`
- a non-mutating functional (unoptimized) version of compute_density is included

A minimum working example of calculating a HF stress is included in `examples/zygote-stresses.jl`. Currently it is implemented only for Kinetic and AtomicLocal terms, but adding other terms should be similar in terms of the needed rewrite strategies.

Tests against FiniteDiff are included in test/chainrules.jl


mfherbst:
Is this superseeded by the zygote-scf branch?

niklasschmitz:
Indeed it is, this PR can thus be closed. For future reference, the new branch is at https://github.com/JuliaMolSim/DFTK.jl/tree/zygote-scf

PR #521, opened 2021-08-24
Fix more edge cases when computing the number of occupied bands

mfherbst:
This applies to cases where the number of electrons is odd.

### DIFF
--- a/src/PlaneWaveBasis.jl
+++ b/src/PlaneWaveBasis.jl
@@ -602,8 +602,9 @@ end
 
 # select the occupied orbitals assuming an insulator
 function select_occupied_orbitals(basis::PlaneWaveBasis, ψ)
-    model = basis.model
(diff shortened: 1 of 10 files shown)


src/PlaneWaveBasis.jl
```diff
@@ -604,7 +604,7 @@ end
 function select_occupied_orbitals(basis::PlaneWaveBasis, ψ)
     model = basis.model
     n_spin = model.n_spin_components
-    n_bands = div(model.n_electrons, n_spin * filled_occupation(model))
+    n_bands = div(model.n_electrons, n_spin * filled_occupation(model), RoundUp)
```
antoine-levitt, line 607: assert temp is zero?

mfherbst: Yes good point.

PR #522, opened 2021-08-24
Improve basis mapping

mfherbst:
This makes the basis transfer a bit faster (less allocations), more readable and allows to extract a sparse matrix mapping between the k-point bases.


src/DFTK.jl
```diff
@@ -88,6 +88,7 @@ export spin_density
 export ρ_from_total_and_spin
 include("densities.jl")
 include("interpolation_transfer.jl")
+export transfer_matrix
```
antoine-levitt, line 91: This function is not like the others... Build transfer matrix or something like that?

src/interpolation_transfer.jl
```diff
@@ -85,56 +86,83 @@ function interpolate_kpoint(data_in::AbstractVecOrMat, kpoint_in::Kpoint, kpoint
...
+        idcs_out = idcs_out[idcs_out .!= nothing]
+    end
+
+    idcs_in, idcs_out
 end
 
+
+"""
+Return a sparse matrix that maps quantities given in the
+`basis_in` basis and on `kpt_in` to quantities given in the `basis_out` basis
```
antoine-levitt, line 128: Just (basis_in, kpt_in)?

src/interpolation_transfer.jl
```diff
@@ -85,56 +86,83 @@ function interpolate_kpoint(data_in::AbstractVecOrMat, kpoint_in::Kpoint, kpoint
...
+    idcs_in, idcs_out = transfer_mapping(basis_in, kpt_in, basis_out, kpt_out)
+    sparse(idcs_out, idcs_in, true)
+end
+
+
+"""
+Return a list of sparse matrices (one per `k`-Point) that map quantities given in the
+`basis_in` basis to quantities given in the `basis_out` basis.
+"""
+function transfer_matrix(basis_in::PlaneWaveBasis{T}, basis_out::PlaneWaveBasis{T}) where T
```
antoine-levitt, line 142: I would remove that method. It doesn't mesh with its name (since it's matrices) and it's trivial

src/DFTK.jl
mfherbst: Good point. Though our prefix is `compute_` ...

src/interpolation_transfer.jl
mfherbst: Well it is a matrix (blocks of a a matrix) and most people will likely need that function and not the individual blocks (to transfer between bases). We can argue about names, but I would say this function and the basis/basis version of `transfer_blochwave` sit pretty much at the same level of "trivial". If we want one, we likely want the other.

PR #524, opened 2021-08-28
Improved printing with `Base.show`

mfherbst:
To avoid writing the same code over and over in user scripts ...


src/Model.jl
```diff
@@ -4,6 +4,9 @@
 # Contains the geometry information, but no discretization parameters.
 # The exact model used is defined by the list of terms.
 struct Model{T <: Real}
+    # Human-readable name for the model (like LDA, PBE, ...)
+    name::String
```
antoine-levitt, line 8: Hm, not so sure about this. A Model is both a mathematical model (functional form of the terms) as well as a specification of the geometry / atomic constitution. This field seems to specialize to just one of these things

src/Model.jl
```diff
@@ -240,3 +246,67 @@ function compute_unit_cell_volume(lattice)
     n_dim = count(!iszero, eachcol(lattice))
     abs(det(lattice[1:n_dim, 1:n_dim]))
 end
+
+
+function Base.show(io::IO, model::Model)
```
antoine-levitt, line 251: Move all these functions to printing.jl? Both Model and (especially) pwbasis are pretty crowded already

src/Model.jl
```diff
@@ -240,3 +246,67 @@ function compute_unit_cell_volume(lattice)
...
+        i < length(model.term_types) && println(io)
+    end
+end
+
+
+
+"""
+Returns the sum formula of the atoms list as a string.
+"""
+function sum_formula(atoms)
```
antoine-levitt, line 297: is "sum formula" a standard term? It doesn't seem to be. Chemical formula? (also add a dispatch on sum_formula(::Model))

src/energies.jl
```diff
@@ -6,14 +6,18 @@ struct Energies{T <: Number}
     # energies["TermName"]
     # parametrization on T acts as a nice check that all terms return correct type
     energies::OrderedDict{String, T}
+    total::T
```
antoine-levitt, line 9: is this really needed? I'd rather avoid precomputation as far as possible

src/Model.jl
```diff
@@ -95,6 +99,7 @@ function Model(lattice::AbstractMatrix{T};
     else
         @assert n_electrons isa Int
     end
+    isempty(terms) && error("Model without terms not supported.")
```
antoine-levitt, line 102: Can't you just leave this as a warning? As the rest of this PR shows it's sometimes useful for tests

antoine-levitt (commented):
Great, thanks for doing this!

src/Model.jl
mfherbst: It is in German :P But the English term is "molecular formula", so I'll use that.

src/Model.jl
mfherbst: ... which is why the default is now to make it a "Kinetic only" if you specify nothing. I have to say I find "model without terms" rather odd and I could imagine that we have code that implicitly assumes this (I have not checked). Certainly it is an easy assumption you might make without thinking about it, because to be fair I cannot imagine a realistic use case where you want a Model with truly no term in it.

But sure I can make it a warning if you insist.

src/Model.jl
mfherbst: Not sure. The point is that this ought to be changed whenever you change the details of the struct definition (e.g. add a field). So having both in the same file is an advantage. I think we better move other things out.

src/Model.jl
mfherbst: Yes true, I agree.

On the other hand this makes the short one-line printing of `Model` more meaningful without adding too much special casing depending on which terms are part of the model, that's why I did it. If you have a better idea how to easily achieve this, I'm happy to remove that field.

src/energies.jl
mfherbst: No, actually I destroyed some functionality with the changes I made in the class. I'll undo it.

src/Model.jl
antoine-levitt: I'm fine with whatever, it just looked simpler to allow it for tests but it's fine this way too!

src/Model.jl
antoine-levitt: I don't think this point of view (that everything that depends on details of the struct definition belongs to the file) is sustainable. We already have dependencies on "inner details" of pwbasis/model in a number of places, and I think it's completely fine. File splitting is done at a conceptual level of splitting a codebase along logical axes, not at a "java/python" level where it matters to the language

mfherbst: Yes, but that stuff breaks in case we change the struct and we catch it in the tests. If you rely on accurate printing in your science and that gets out of sync with the definition that can have pretty bad consequences. I've split out some other things into separate files, where I think it makes more sense.

antoine-levitt: I don't understand your point. Having it in the same file does not make it more or less likely that it stays in sync with the changes in the structure: that's the job of the tests (partially). We already have other places where the changes in X.jl need to be taken into account in Y.jl, that doesn't bother me at all. OTOH printing is kind of orthogonal to all other functionality and so having it all in printing.jl makes sense to me

src/G_to_r.jl
```diff
@@ -0,0 +1,129 @@
+#
+# Perform (i)FFTs.
```
antoine-levitt, line 2: rename to fft.jl and move the current fft.jl to fft_utils?

antoine-levitt: I do agree it's nice to move it out of pwbasis

PR #525, opened 2021-08-29
Some docs explaining periodic problems

mfherbst:


docs/src/guide/periodic_problems.jl
```diff
@@ -0,0 +1,321 @@
...
+# ```math
+# \begin{aligned}
+# -\Delta \left(e^{i k⋅x} ψ_{kj}(x)\right)
+#    &= k^2 e^{ik⋅x} ψ_{kj}(x) - 2ik e^{ik⋅x} (∇ψ_{kj}(x)) - e^{ik⋅x} (Δψ_{kj}(x)) \\
+#    &= e^{ik⋅x} (k^2 - 2ik ∇ - Δ) ) ψ_{kj}(x) \\
+#    &= e^{i k⋅x} (-i∇ + k)^2 ψ_{kj}(x) \\
+#    &= e^{i k⋅x} 2H_k ψ_{kj}(x),
+# \end{aligned}
+# ```
+# where we used
```
antoine-levitt, line 87: defined

docs/src/guide/periodic_problems.jl
```diff
@@ -0,0 +1,321 @@
...
+# The action of this operator on a function ``ψ_{kj}`` is given by
+# ```math
+#     H_k ψ_{kj} = e^{-i k⋅x} H e^{i k⋅x} ψ_{kj},
+# ```
+# which in particular implies that
+# ```math
+#    H_k ψ_{kj} = ε_{kj} ψ_{kj} \quad ⇔ \quad H (e^{i k⋅x} ψ_{kj}) = ε_{kj} (e^{i k⋅x} ψ_{kj}).
+# ```
+# To seek the eigenpairs of ``H`` we may thus equivalently
+# find the eigenpairs of *all* ``H_k``.
```
antoine-levitt, line 100: Say that the point of this is that the eigenvectors of H_k are periodic and can therefore be represented on a single unit cell (in contrast to those of H)

docs/src/guide/periodic_problems.jl
```diff
@@ -0,0 +1,321 @@
...
+#
+# A more detailed mathematical analysis shows that the transformation from ``H``
+# to the set of all ``H_k`` for a suitable set of values for ``k`` (detailes below)
```
antoine-levitt, line 103: detailed?

docs/src/guide/periodic_problems.jl
```diff
@@ -0,0 +1,321 @@
...
+# is actually a unitary transformation, the so-called **Bloch transform**.
```
antoine-levitt, line 104: not sure I would talk about the Bloch transform at that level

docs/src/guide/periodic_problems.jl
```diff
@@ -0,0 +1,321 @@
...
+# of all ``H_k`` blocks.
+#
+# [^1]: Notice that block-diagonal is a bit an abuse of terms here, since the Hamiltonian
+#       is not a matrix but an operator and the number of blocks is essentially infinite.
+#       The mathematically precise term is that the Bloch transform reveals the fibers
+#       of the Hamiltonian.
+#
+# ## The Brillouin zone
+#
+# The big mystery in the discussion so far is the parameter ``k`` of the Hamiltonian blocks.
```
antoine-levitt, line 125: Don't think I would call it a mystery

docs/src/guide/periodic_problems.jl
```diff
@@ -0,0 +1,321 @@
...
+#
+# - As discussed ``k`` is essentially a real number. It turns out, however, that some of
```
antoine-levitt, line 127: Why essentially? "An arbitrary real number"?

docs/src/guide/periodic_problems.jl
```diff
@@ -0,0 +1,321 @@
...
+#   these ``k∈\mathbb{R}`` give rise to operators related by unitary transformations
+#   (again due to translational symmetry).
```
antoine-levitt, line 129: Just explain it from the ansatz psi(x) = e^ikx u(x): k and k+2pi/a give the same class of functions

docs/src/guide/periodic_problems.jl
```diff
@@ -0,0 +1,321 @@
...
+# - The BZ in our example is thus ``B = [-π/a, π/a)``. The members of ``B``
+#   are typically called ``k``-points.
+#
+# ## Discretisation and plane-wave basis sets
+#
+# With what we discussed so far the strategy to find all eigenpairs of a periodic
+# Hamiltonian ``H`` thus reduces to finding the eigenpairs of all ``H_k`` with ``k ∈ B``.
+# This requires *two* discretisations:
+#
+#   - ``B`` is an overcountable set. To discretise we first only pick a finite number
```
antoine-levitt, line 151: what's an overcountable set?

docs/src/guide/periodic_problems.jl
```diff
@@ -0,0 +1,321 @@
...
+#     of ``k``-points. Usually this **``k``-point samping** is done by picking ``k``-points
+#     along a regular grid inside the BZ, the **``k``-grid**.A
```
antoine-levitt, line 153: Remove A

docs/src/guide/periodic_problems.jl
```diff
@@ -0,0 +1,321 @@
...
+n_bands = 6
+ρ0 = guess_density(basis)  # Just dummy, has no meaning in this model
+p  = plot_bandstructure(basis, ρ0, n_bands, kline_density=15, unit=u"hartree")
+
+# !!! note "Selection of k-point grids in `PlaneWaveBasis` construction"
+#     You might wonder why we only selected a single ``k``-point (clearly a very crude
+#     and inaccurate approximation). It turns out the `kgrid` parameter specified here
+#     is not actually used for plotting the bands. It is only used when solving more
+#     involved models like density-functional theory (DFT) where the Hamiltonian is
+#     non-linear and before plotting the bands therefore the self-consistent field
```
antoine-levitt, line 216: move therefore

docs/src/guide/periodic_problems.jl
```diff
@@ -0,0 +1,321 @@
...
+#   * density-functial theory models for treating electronic structures
+#     (see the [Tutorial](@ref)).
+#   * Gross-Pitaevskii models for bosonic systems
+#     (see [Gross-Pitaevskii equation in one dimension](@ref))
+#   * even some more unusual cases like anyonic models.
+#
+# In this tutorial we will go a little more low-level and directly provide
+# an analytic potential describing the interaction with the electrons to DFTK.
+#
+# First we define a custom Gaussian potential as a new "element" inside DFTK:
```
antoine-levitt, line 247: For this demo you don't want to just have a sin(x) thing?

docs/src/guide/periodic_problems.jl
```diff
@@ -0,0 +1,321 @@
...
+ham   = Hamiltonian(basis)
+
+## Extract the total potential term of the Hamiltonian and plot it
+potential = DFTK.total_local_potential(ham)[:, 1, 1]
+rvecs = collect(r_vectors_cart(basis))[:, 1, 1]  # slice along the x axis
+x = [r[1] for r in rvecs]                        # only keep the x coordinate
+plot(x, potential, label="", xlabel="x", ylabel="V(x)")
+
+# Notice how DFTK took care of the periodic wrapping of the potential values going
+# from ``0`` and ``100``.
```
antoine-levitt, line 301: don't understand the "going from 0 and 100"

docs/src/guide/periodic_problems.jl
```diff
@@ -0,0 +1,321 @@
...
+#
```
antoine-levitt, line 302: I would maybe make the point that the potential is the sum of local ("atomic") potentials, and that the atoms are implicitly periodized

docs/src/guide/periodic_problems.jl
```diff
@@ -0,0 +1,321 @@
...
+n_bands = 6
+ρ0 = zeros(eltype(basis), basis.fft_size..., 1)  # Just dummy, has no meaning in this model
+p = plot_bandstructure(basis, ρ0, n_bands, kline_density=15, unit=u"hartree")
+
+# The bands are noticably different.
+#  - The bands no longer overlap, meaning that the spectrum of $H$ is no longer continous
+#    but has gaps.
+#
+# - The two lowest bands are almost flat, which means that they represent
+#   two tightly bound and localised electrons inside the two Gaussians.
```
antoine-levitt, line 317: This is not obvious, so I would indicate that in writing (eg "this is because" instead of "which means" which suggests the reader is able to reproduce the derivation)

docs/src/guide/periodic_problems.jl
```diff
@@ -0,0 +1,321 @@
...
+#
+# - The higher the bands, the more curved they become. In other words the higher the
```
antoine-levitt, line 319: Rather say that it becomes free electron like? it's not obvious that the curvature is linked to the delocalization

antoine-levitt (commented):
Nice intro!

PR #528, opened 2021-09-30
Add test for self-adjointness of solve_ΩplusK

niklasschmitz:
cc @antoine-levitt

### DIFF
--- a/test/newton.jl
+++ b/test/newton.jl
@@ -1,0 +1,35 @@
+using Test
+using DFTK
+import DFTK: compute_projected_gradient, solve_ΩplusK, filled_occupation
+
+include("testcases.jl")
+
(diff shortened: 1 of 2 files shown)


antoine-levitt:
Good for me!

antoine-levitt:
Oh while you're at it can you also test apply_omega and apply_k (separately)? Same thing but just have apply_omega and apply_K instead of solve_omegaplusK

antoine-levitt:
Great! Can you just add a comment like "Omega is complex-linear and so self-adjoint as a complex operator, but K involves conjugates and is only a real-linear operator, hence we test using the real dot product"?

niklasschmitz:
I now added short comments and renamed the test file to `omegaplusk.jl`

PR #540, opened 2021-11-09
Fix spglib wrapper bug (wrong transpose)

### LINKED ISSUE #539: Wrong spglib spacegroup
Hi, I would like to report a possible bug in `DFTK.spglib_spacegroup_number`.

In the code below, `lattice1` is the convention used in DFTK examples, and `lattice2` is just the lattice vector with the sign of Cartesian x coordinates flipped. Hence, I expect that their spacegroup to be identical, but it is not the case.

```julia
using DFTK
a = 10.0

# Convention used in DFTK example
lattice1 = a / 2 * [[0 1 1.];
                    [1 0 1.];
                    [1 1 0.]]
atoms1 = [Si => [ones(3)/8, -ones(3)/8]]
model = model_LDA(lattice1, atoms1)
@show DFTK.spglib_spacegroup_number(model) # 227

# A different convention for the lattice vectors
lattice2 = a / 2 * [[0 -1 -1.]; # Sign of x coordinates flipped
                    [1 0 1.];
                    [1 1 0.]]
atoms2 = [Si => [ones(3)/8, -ones(3)/8]]
model2 = model_LDA(lattice2, atoms2)
@show DFTK.spglib_spacegroup_number(model2) # 139
```
```julia
(dfpt) pkg> status DFTK
      Status `~/dfpt/Project.toml`
  [acf6eb54] DFTK v0.4.0
  
(@v1.6) pkg> status Spglib
      Status `~/.julia/environments/v1.6/Project.toml`
  [f761d5c5] Spglib v0.5.1
```

jaemolihm:
Closes #539

It seems that this bug (wrong transpose) was added in #496 when moving from spglib_jll to Spglib.jl.

The tests are a bit verbose but I did not have a good idea to organize them.

### DIFF
--- a/src/external/spglib.jl
+++ b/src/external/spglib.jl
@@ -185,9 +185,9 @@ function get_spglib_lattice(model; to_primitive=false)
     #      Essentially this does not influence the standardisation,
     #      but it only influences the kpath.
     spg_positions, spg_numbers, _ = spglib_atoms(model.atoms)
-    structure = Spglib.Cell(transpose(model.lattice), spg_positions, spg_numbers)
+    structure = Spglib.Cell(model.lattice, spg_positions, spg_numbers)
     spglib_lattice = Spglib.standardize_cell(structure, to_primitive=to_primitive).lattice
-    Matrix(copy(spglib_lattice'))
+    Matrix(spglib_lattice)
 end
 
 
@@ -196,7 +196,7 @@ function spglib_spacegroup_number(model)
     # TODO Time-reversal symmetry disabled? (not yet available in DFTK)
     # TODO Are magnetic moments passed?
     spg_positions, spg_numbers, _ = spglib_atoms(model.atoms)
-    structure = Spglib.Cell(transpose(model.lattice), spg_positions, spg_numbers)
+    structure = Spglib.Cell(model.lattice, spg_positions, spg_numbers)
     spacegroup_number = Spglib.get_spacegroup_number(structure)
     spacegroup_number
 end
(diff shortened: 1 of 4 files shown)


mfherbst:
Thanks @jaemolihm  for the fix and the testcase and absolutely correct!

The annoying issue here is that the spglib python and C interface use *different* conventions for how the lattice vectors are organised (in rows versus in columns) and Spglib.jl seems to be following the python convention of organising it in columns, which is exactly the one we use in DFTK as well.

Could you add a comment about that in the appropriate places you modify? The relevant place in the Spglib.jl docs is 
https://singularitti.github.io/Spglib.jl/stable/api/.

Otherwise this is good to go!

src/external/spglib.jl
```diff
@@ -185,9 +185,9 @@ function get_spglib_lattice(model; to_primitive=false)
     #      Essentially this does not influence the standardisation,
     #      but it only influences the kpath.
     spg_positions, spg_numbers, _ = spglib_atoms(model.atoms)
-    structure = Spglib.Cell(transpose(model.lattice), spg_positions, spg_numbers)
+    structure = Spglib.Cell(model.lattice, spg_positions, spg_numbers)
     spglib_lattice = Spglib.standardize_cell(structure, to_primitive=to_primitive).lattice
-    Matrix(copy(spglib_lattice'))
+    Matrix(spglib_lattice)
```
mfherbst, line 190:
```suggestion
    Matrix(Spglib.standardize_cell(structure, to_primitive=to_primitive).lattice)
```

antoine-levitt:
Oh wow we should have caught that, thanks @jaemolihm! There's a comment at the top of the file so just clarify there that Spglib.jl used the column convention

src/external/spglib.jl
jaemolihm: Done!

jaemolihm:
I added the comment.
I also added "For example, `model.lattice[:, 1]` is the first real-space lattice vector." in `conventions.md` because it took me some time to understand what "in columns" mean 😂

PR #541, opened 2021-11-10
Fixed minor issues in periodic_problems tutorial

cw-tan:
I encountered some minor issues when going through the Periodic Systems [tutorial](https://juliamolsim.github.io/DFTK.jl/dev/guide/periodic_problems/), so I made this pull request. 

1. **"ElementGaussian has no field symbol"**

In the tutorial, upon running the line https://github.com/JuliaMolSim/DFTK.jl/blob/699cccd568daaef945fec66da51e0156ee4602bf/docs/src/guide/periodic_problems.jl#L318
I was met with the error message 
`Error showing value of type Model{Float64}:`
`ERROR: type ElementGaussian has no field symbol`

Though not an issue that would affect the later calculations, I avoided the error message by adding a dummy symbol so that the value of model would show in REPL. 

2. **Missing Bandstructure**

At the very bottom of the [tutorial](https://juliamolsim.github.io/DFTK.jl/dev/guide/periodic_problems/), the bandstructure plot seems to be empty. Changing the `kline_density` from `15` to `50` in https://github.com/JuliaMolSim/DFTK.jl/blob/699cccd568daaef945fec66da51e0156ee4602bf/docs/src/guide/periodic_problems.jl#L339
gave the expected bandstructure described at the end of the tutorial.


antoine-levitt:
> I was met with the error message
> Error showing value of type Model{Float64}:
> ERROR: type ElementGaussian has no field symbol

Hm, not sure your fix is the right one, can you post the backtrace for this? Maybe we expect Element to have a symbol field but shouldn't really?

For the band structure I don't know, @mfherbst what do you think?

cw-tan:
> > I was met with the error message
> > Error showing value of type Model{Float64}:
> > ERROR: type ElementGaussian has no field symbol
> 
> Hm, not sure your fix is the right one, can you post the backtrace for this? Maybe we expect Element to have a symbol field but shouldn't really?

Upon running the line, 
https://github.com/JuliaMolSim/DFTK.jl/blob/699cccd568daaef945fec66da51e0156ee4602bf/docs/src/guide/periodic_problems.jl#L318
I get
```
Model(custom, 1D):
    lattice (in Bohr)    : [100       , 0         , 0         ]
                           [0         , 0         , 0         ]
                           [0         , 0         , 0         ]
    unit cell volume     : 100 Bohr³

Error showing value of type Model{Float64}:
ERROR: type ElementGaussian has no field symbol
Stacktrace:
  [1] getproperty(x::ElementGaussian, f::Symbol)
    @ Base ./Base.jl:33
  [2] chemical_formula(atoms::Vector{Pair{Any, Vector{StaticArrays.SVector{3, Float64}}}})
    @ DFTK ~/.julia/packages/DFTK/ke1Dz/src/structure.jl:48
  [3] chemical_formula
    @ ~/.julia/packages/DFTK/ke1Dz/src/structure.jl:61 [inlined]
  [4] show(io::IOContext{Base.TTY}, #unused#::MIME{Symbol("text/plain")}, model::Model{Float64})
    @ DFTK ~/.julia/packages/DFTK/ke1Dz/src/show.jl:23
  [5] (::REPL.var"#38#39"{REPL.REPLDisplay{REPL.LineEditREPL}, MIME{Symbol("text/plain")}, Base.RefValue{Any}})(io::Any)
    @ REPL /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/REPL/src/REPL.jl:220
  [6] with_repl_linfo(f::Any, repl::REPL.LineEditREPL)
    @ REPL /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/REPL/src/REPL.jl:462
  [7] display(d::REPL.REPLDisplay, mime::MIME{Symbol("text/plain")}, x::Any)
    @ REPL /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/REPL/src/REPL.jl:213
  [8] display(d::REPL.REPLDisplay, x::Any)
    @ REPL /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/REPL/src/REPL.jl:225
  [9] display(x::Any)
    @ Base.Multimedia ./multimedia.jl:328
 [10] #invokelatest#2
    @ ./essentials.jl:708 [inlined]
 [11] invokelatest
    @ ./essentials.jl:706 [inlined]
 [12] print_response(errio::IO, response::Any, show_value::Bool, have_color::Bool, specialdisplay::Union{Nothing, AbstractDisplay})
    @ REPL /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/REPL/src/REPL.jl:247
 [13] (::REPL.var"#40#41"{REPL.LineEditREPL, Pair{Any, Bool}, Bool, Bool})(io::Any)
    @ REPL /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/REPL/src/REPL.jl:231
 [14] with_repl_linfo(f::Any, repl::REPL.LineEditREPL)
    @ REPL /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/REPL/src/REPL.jl:462
 [15] print_response(repl::REPL.AbstractREPL, response::Any, show_value::Bool, have_color::Bool)
    @ REPL /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/REPL/src/REPL.jl:229
 [16] (::REPL.var"#do_respond#61"{Bool, Bool, REPL.var"#72#82"{REPL.LineEditREPL, REPL.REPLHistoryProvider}, REPL.LineEditREPL, REPL.LineEdit.Prompt})(s::REPL.LineEdit.MIState, buf::Any, ok::Bool)
    @ REPL /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/REPL/src/REPL.jl:798
 [17] #invokelatest#2
    @ ./essentials.jl:708 [inlined]
 [18] invokelatest
    @ ./essentials.jl:706 [inlined]
 [19] run_interface(terminal::REPL.Terminals.TextTerminal, m::REPL.LineEdit.ModalInterface, s::REPL.LineEdit.MIState)
    @ REPL.LineEdit /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/REPL/src/LineEdit.jl:2441
 [20] run_frontend(repl::REPL.LineEditREPL, backend::REPL.REPLBackendRef)
    @ REPL /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/REPL/src/REPL.jl:1126
 [21] (::REPL.var"#44#49"{REPL.LineEditREPL, REPL.REPLBackendRef})()
    @ REPL ./task.jl:411 
```

I think you're right that ElementGaussian shouldn't even have the symbol field as it doesn't really make sense in this context (I assume `Model` just wants to label atomic pseudopotentials of specific atoms in the usual use case but that doesn't apply for this case of an arbitrary user-defined local potential.) I guess my changes were less of a "fix" but just a way to avoid the error messages so that the rest of `Model` is printed out for this tutorial example. Admittedly, the `atoms                : dummy_symbol₂` line below doesn't make much sense.

```
Model(custom, 1D):
    lattice (in Bohr)    : [100       , 0         , 0         ]
                           [0         , 0         , 0         ]
                           [0         , 0         , 0         ]
    unit cell volume     : 100 Bohr³

    atoms                : dummy_symbol₂
    atom potentials      : ElementGaussian(0.3, 10.0, :dummy_symbol)

    num. electrons       : 0
    spin polarization    : none
    temperature          : 0 Ha

    terms                : Kinetic(1)
                           AtomicLocal()
```

I'm not entirely sure, but I'm guessing that a possible, more fundamental fix would probably be to change something in the definition of `Model` so that it doesn't need a `symbol` to print its contents if it can't find any `symbol` fields.

antoine-levitt:
Yes: the responsible functions are Base.show(io::IO, ::MIME"text/plain", model::Model) in show.jl and chemical_formula in structure.jl. We could fix it eg by having `element_symbol = hasfield(typeof(element), :symbol) ? element.symbol : :Artificialum` or something nonsensical like that in `chemical_formula`, so it would still print something informative even on elements that don't have a symbol.

mfherbst:
I rather vote to use a function to extract the element symbol and have a fallback for the abstract type. Usually `:X` is used for a "dummy element".

The band structure thing is weird. One should not need as large a density as 50. Especially because it worked fine before we did the big restructuring PR (I checked), so maybe there is some unit conversion issue somewhere. I'll take a look later today.

antoine-levitt:
Thought of that as well but it's annoying to clog the namespace with an additional function that's pretty niche...

mfherbst:
I see, but it's pretty much the Julian way to deal with these types of problems. I think `chemical_symbol` is not such a bad name for such a function @antoine-levitt.

mfherbst:
> For the band structure I don't know, @mfherbst what do you think?

Indeed I made a mistake with the units. I think it's even fair to crank up to like `kline_density=100` (for both cases where `plot_bandstructure` is called).

antoine-levitt:
Fair for chemical_symbol!

cw-tan:
> Indeed I made a mistake with the units. I think it's even fair to crank up to like `kline_density=100` (for both cases where `plot_bandstructure` is called).

In that case, I changed the kline densities to 100 for both bandstructure calculations in the tutorial. I also reverted the changes to my earlier "fix" (commit d0f6174ef7231e761fde0898408fcc8e24b042f1). From what I gather, you want a solution that changes this https://github.com/JuliaMolSim/DFTK.jl/blob/4e5309d576af80729d7c2fbc84b9df38fe2d3742/src/structure.jl#L45
to look something more like this?
```
function chemical_formula(atoms)
    element_count = Dict{Symbol, Int}()
    for (element, positions) in atoms
        if chemical_symbol(element) in keys(element_count)                         # line changed
            element_count[chemical_symbol(element)] += length(positions)           # line changed
        else
            element_count[chemical_symbol(element)] = length(positions)            # line changed
        end
    end
    formula = join(string(elem) * string(element_count[elem])
                      for elem in sort(collect(keys(element_count))))
    for i in 0:9
        formula = replace(formula, ('0' + i) => ('₀' + i))
    end
    formula
end

function chemical_symbol(element)
    element_symbol = hasfield(typeof(element), :symbol) ? element.symbol : :X
    element_symbol
end
```

In `chemical_formula`, I only changed the instances of `element.symbol` to `chemical_symbol(element)`.

antoine-levitt:
The julia way of doing it is to use multiple dispatch: add to elements.jl `chemical_symbol(::Element) = :X`, which acts as a fallback. Then, whenever a new element is defined, eg ElementCoulomb, add a function `chemical_symbol(e::ElementCoulomb) = e.symbol`. That's the julia way of doing things and it has the advantage that there is a clear API for chemical_formula to query the symbol of an element (without relying on implementation details like the fact that some elements have a `symbol` field). Although to be honest in this case it feels like overengineering, and it could be simpler to just have what you have and say in the comment before struct Element that elements are expected to have a "symbol" field... @mfherbst what do you think?

Then chemical_formula is OK (although I would write symbol = chemical_symbol(element) and then reuse it for clarity)

mfherbst:
> Although to be honest in this case it feels like overengineering, and it could be simpler to just have what you have and say in the comment before struct Element that elements are expected to have a "symbol" field... @mfherbst what do you think?

I disagree, mainly because the multiple dispatch way is more idiomatic julia and what people would expect. Personally I also think this `hasfield(typeof(element), :symbol)` is really a nuisance and not exactly easy to read compared to implementing the fallback function. We also followed this pattern in other cases (e.g. `charge_nuclear`). Also looking into the future being based on functions gives us more flexibility to integrate with [AtomsBase.jl](https://github.com/JuliaMolSim/AtomsBase.jl).

@cw-tan With respect to the implementation, I would suggest to follow along the `charge_nuclear( )` function and its methods for the different elements, just grep for that in the `elements.jl` file and you should find some pointers.

mfherbst:
@cw-tan I was just wondering if you had a chance to work on this.

If not and you don't mind I would go ahead and fix the issues in the docs today or tomorrow to have it ready for the next patch release.

cw-tan:
@mfherbst Sorry, I've been busy the past few weeks and haven't followed up since my last comment. Feel free to fix the issue.

mfherbst:
@cw-tan  That's no problem. Thanks for letting me know. I'll take care of it :smile:.

mfherbst:
CI failure because of https://github.com/JuliaPlots/Plots.jl/issues/3964

Issue #538, opened 2021-11-07
Kinetic Terms


cw-tan:
Hi, I'm pretty new to Julia and DFTK but I was hoping to use DFTK for orbital-free benchmarking. I scanned through the documentation but wasn't sure if it had what I wanted. 

After getting the results from a self-consistent field calculation, is there a way to access the positive-definite kinetic energy density as defined [here](https://juliamolsim.github.io/DFTK.jl/dev/api/#DFTK.Kinetic), the kinetic energy contribution to the stress, and the Kohn-Sham potential? Thank you!

antoine-levitt:
Hi and thanks for your interest!

- The kinetic energy density you can compute it rather easily from the orbitals. See eg https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/postprocess/current.jl for a related computation. If you do code it up please contribute it to this package, it could be useful to other people (eg for meta-GGAs)!
- The kinetic contribution to the stress you can get by only keeping the Kinetic term in the new_model here: https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/postprocess/stresses.jl
- The Kohn-Sham potential we have, it's the function total_local_potential(ham::Hamiltonian) defined here: https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/terms/Hamiltonian.jl

All of this should be relatively easy, but you do have to look at the code a bit. The examples should help you get the structure. If you're stuck let us know and we can help.

May I ask what you want to do exactly? There's an open issue for Thomas-Fermi here: https://github.com/JuliaMolSim/DFTK.jl/issues/122

antoine-levitt:
Or for the stresses, even simpler: just change the energies.total line in compute_stresses to energies.kinetic

cw-tan:
Great, thanks! I literally installed Julia yesterday but I'm eager to have a go at it! I'm just hoping to inspect these quantities from Kohn-Sham DFT to possibly give some insight for the development of kinetic energy functional approximations.

antoine-levitt:
OK, happy discovery :-) I'm closing the issue for bookkeeping purposes, feel free to ask for help, here or on slack.

PR #543, opened 2021-11-12
Kinetic Energy Density function

cw-tan:
With reference to issue #538, it was suggested that a kinetic energy density function be contributed for other uses (such as meta-GGAs).

### DIFF
--- a/src/postprocess/ked.jl
+++ b/src/postprocess/ked.jl
@@ -1,0 +1,20 @@
+"""
+Computes the kinetic energy density (ked), 0.5 ∑ fn |∇ψn|^2.
+"""
+function compute_kinetic_energy_density(basis::PlaneWaveBasis, ψ, occupation)
+    @assert all(symop -> length(symop) == 1, basis.ksymops) == 1 
+    ked = zeros(eltype(basis), basis.fft_size)
+    for (ik, kpt) in enumerate(basis.kpoints)
+        for (n, ψnk) in enumerate(eachcol(ψ[ik]))
+            for α = 1:3
+                dαψnk = [im*(G+kpt.coordinate_cart)[α] for G in G_vectors_cart(kpt)] .* ψnk
+                dαψnk_real = G_to_r(basis, kpt, dαψnk)
+                ked .+= @. basis.kweights[ik] *
+                           occupation[ik][n] * 
+                           real( conj(dαψnk_real) * dαψnk_real)
+            end
+        end
+    end
+    DFTK.mpi_sum!(ked, basis.comm_kpts)
+    0.5 * ked
+end
--- a/src/postprocess/kinetic_energy_density.jl
+++ b/src/postprocess/kinetic_energy_density.jl
@@ -7,11 +7,11 @@ function compute_kinetic_energy_density(basis::PlaneWaveBasis, ψ, occupation)
     for (ik, kpt) in enumerate(basis.kpoints)
         for (n, ψnk) in enumerate(eachcol(ψ[ik]))
             for α = 1:3
-                dαψnk = [im*(G+kpt.coordinate_cart)[α] for G in G_vectors_cart(kpt)] .* ψnk
+                dαψnk = [im * (G + kpt.coordinate_cart)[α] for G in G_vectors_cart(kpt)] .* ψnk
                 dαψnk_real = G_to_r(basis, kpt, dαψnk)
                 ked .+= @. 0.5 * basis.kweights[ik] *
                            occupation[ik][n] * 
-                           real( conj(dαψnk_real) * dαψnk_real)
+                           real(conj(dαψnk_real) * dαψnk_real)
             end
         end
     end
--- a/src/postprocess/kinetic_energy_density.jl
+++ b/src/postprocess/kinetic_energy_density.jl
@@ -1,20 +1,20 @@
 """
-Computes the kinetic energy density (ked), 0.5 ∑ fn |∇ψn|^2.
+Computes the kinetic energy density (ked), 1/2 ∑ fn |∇ψn|^2.
 """
 function compute_kinetic_energy_density(basis::PlaneWaveBasis, ψ, occupation)
-    @assert all(symop -> length(symop) == 1, basis.ksymops) == 1 
+    @assert all(symop -> length(symop) == 1, basis.ksymops) == 1  # TODO lift this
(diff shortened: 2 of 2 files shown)


src/postprocess/ked.jl
```diff
@@ -0,0 +1,20 @@
...
+                dαψnk = [im*(G+kpt.coordinate_cart)[α] for G in G_vectors_cart(kpt)] .* ψnk
+                dαψnk_real = G_to_r(basis, kpt, dαψnk)
+                ked .+= @. basis.kweights[ik] *
+                           occupation[ik][n] * 
+                           real( conj(dαψnk_real) * dαψnk_real)
+            end
+        end
+    end
+    DFTK.mpi_sum!(ked, basis.comm_kpts)
+    0.5 * ked
```
antoine-levitt, line 19: absorb this factor into the ked .+= above (clearer and faster)

src/postprocess/ked.jl
```diff
@@ -0,0 +1,20 @@
...
+    @assert all(symop -> length(symop) == 1, basis.ksymops) == 1 
+    ked = zeros(eltype(basis), basis.fft_size)
+    for (ik, kpt) in enumerate(basis.kpoints)
+        for (n, ψnk) in enumerate(eachcol(ψ[ik]))
+            for α = 1:3
+                dαψnk = [im*(G+kpt.coordinate_cart)[α] for G in G_vectors_cart(kpt)] .* ψnk
+                dαψnk_real = G_to_r(basis, kpt, dαψnk)
+                ked .+= @. basis.kweights[ik] *
+                           occupation[ik][n] * 
+                           real( conj(dαψnk_real) * dαψnk_real)
```
antoine-levitt, line 14: Does this actually work? * does matrix multiplication, you want .* here

src/postprocess/ked.jl
```diff
@@ -0,0 +1,20 @@
+"""
+Computes the kinetic energy density (ked), 0.5 ∑ fn |∇ψn|^2.
```
antoine-levitt, line 2: we usually use 1/2 rather than 0.5 in the code (small detail, but might as well keep it for consistency)

src/postprocess/ked.jl
```diff
@@ -0,0 +1,20 @@
+"""
+Computes the kinetic energy density (ked), 0.5 ∑ fn |∇ψn|^2.
+"""
+function compute_kinetic_energy_density(basis::PlaneWaveBasis, ψ, occupation)
+    @assert all(symop -> length(symop) == 1, basis.ksymops) == 1
```
antoine-levitt, line 5: Keep the TODO in current that the restriction is to be lifted

antoine-levitt (commented):
Nice, thanks! Would be good to merge this with another function (we have a bunch of these "summers" already) but I can't figure out what so let's go with this.

- ked.jl is a bit obscure, call it kinetic_energy_density.jl
- Can you add a test? This will ensure that the code is regularly tested and so will not break eg whenever we decide to change APIs. Might as well test `compute_current`.

src/postprocess/ked.jl
cw-tan: It does work and I think it's because of the `@.` after `ked .+ =`.  That's how the `current.jl` function did it (with `@.` so it didn't use `.*` to broadcast the multiplication). I could remove the ` @.` and use `.*` instead, do you have a preference?

cw-tan:
Thanks for the review, I've made most of the changes. I'll work on a test for this (and for `compute_current` too if it's convenient).

src/postprocess/ked.jl
antoine-levitt: Oh of course you're right, sorry! No need to change anything then

src/postprocess/kinetic_energy_density.jl
```diff
@@ -0,0 +1,20 @@
+"""
+Computes the kinetic energy density (ked), 1/2 ∑ fn |∇ψn|^2.
+"""
+function compute_kinetic_energy_density(basis::PlaneWaveBasis, ψ, occupation)
+    @assert all(symop -> length(symop) == 1, basis.ksymops) == 1  # TODO lift this
+    ked = zeros(eltype(basis), basis.fft_size)
+    for (ik, kpt) in enumerate(basis.kpoints)
+        for (n, ψnk) in enumerate(eachcol(ψ[ik]))
+            for α = 1:3
+                dαψnk = [im*(G+kpt.coordinate_cart)[α] for G in G_vectors_cart(kpt)] .* ψnk
```
mfherbst, line 10: We try to keep one space around operators.

```suggestion
                dαψnk = [im * (G + kpt.coordinate_cart)[α] for G in G_vectors_cart(kpt)] .* ψnk
```

src/postprocess/kinetic_energy_density.jl
```diff
@@ -0,0 +1,20 @@
...
+    @assert all(symop -> length(symop) == 1, basis.ksymops) == 1  # TODO lift this
+    ked = zeros(eltype(basis), basis.fft_size)
+    for (ik, kpt) in enumerate(basis.kpoints)
+        for (n, ψnk) in enumerate(eachcol(ψ[ik]))
+            for α = 1:3
+                dαψnk = [im*(G+kpt.coordinate_cart)[α] for G in G_vectors_cart(kpt)] .* ψnk
+                dαψnk_real = G_to_r(basis, kpt, dαψnk)
+                ked .+= @. 0.5 * basis.kweights[ik] *
+                           occupation[ik][n] * 
+                           real( conj(dαψnk_real) * dαψnk_real)
```
mfherbst, line 14:
```suggestion
                           real(conj(dαψnk_real) * dαψnk_real)
```

mfherbst (commented):
LGTM. Just some nits about spacings :smile:

antoine-levitt:
Thanks @cw-tan !

antoine-levitt:
@mfherbst I think @cw-tan wanted to add a test so we could have merged this later, but that's OK. @cw-tan you'll have to make a new PR; for this you'll need to update your fork (see eg https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork)

mfherbst:
Ups sorry @cw-tan ... I was too quick then :smile:. Yes please, just open a new PR.

PR #549, opened 2021-12-01
Untested routine for wannier plotting

mfherbst:
@LaurentVidal95 @antoine-levitt In case any of you feels like it, you could check if this does the right thing. I do not have enough experience in Wannier functions to understand if it does or not (and I don't have time to check with another code).

If not I would just merge it as is. We don't give guarantees on this part of the code at the moment anyway.

### DIFF
--- a/examples/wannier90.jl
+++ b/examples/wannier90.jl
@@ -6,6 +6,10 @@
 # from an initial self consistent field calculation.
 # All processes are handled by calling the routine `run_wannier90`.
 #
+# !!! warning "No guarantees on Wannier90 interface"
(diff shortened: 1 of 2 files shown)


LaurentVidal95:
@mfherbst Testing it right now ! For now it seems to work great, I keep you informed !
By the way, it's very interesting to see the modifications and changes of syntax you made in the wannier90.jl file. Nice way to learn new ways to do Julia more efficiently (and elegantly).

mfherbst:
@LaurentVidal95 Thanks for the feedback (on the syntax and the plots) :smile:.

PR #550, opened 2021-12-03
Update to Wannier example and rename `atomic_symbol`

mfherbst:
- Wannier example is now the more interesting graphene case from @LaurentVidal95
- For the upcoming implementation of AtomsBase I renamed `chemical_symbol` to `atomic_symbol` (used in AtomsBase)


examples/wannier90.jl
```diff
@@ -11,21 +11,28 @@
 #     Bugs are likely and we welcome issues in case you find any!
 #
 # This example shows how to obtain the MLWFs corresponding
-# to the first eight bands of silicon. Since the bands 5 to 8 are entangled,
-# 12 bands are first computed to obtain 8 MLWFs by a disantanglement procedure.
+# to the first five bands of graphene. Since the bands 2 to 11 are entangled,
+# 15 bands are first computed to obtain 4 MLWFs by a disantanglement procedure.
```
antoine-levitt, line 15: Do you mean 5 here?

antoine-levitt: Hm, you have 4 below. @LaurentVidal95 I thought 5 was the correct number here?

antoine-levitt (commented):
OK for me if it matches what @LaurentVidal95 did! (there was some confusion yesterday but now I think is OK)

examples/wannier90.jl
mfherbst: Ah yes that's correct I think I forgot to account for a degeneracy in the bands.

antoine-levitt:
We rationalized this with Laurent as follows: there is a sigma Wannier on each bond plus a pz Wannier on each atom. On each unit cell there are two atoms, one bond in the unit cell, and four bonds shared with other unit cells, for a total of 2+1+4/2 = 5 Wannier. @LaurentVidal95 can you check the Wannier produced by this PR look like the ones you showed us yesterday?

examples/wannier90.jl
```diff
@@ -35,19 +42,26 @@ scfres = self_consistent_field(basis, tol=1e-12, n_bands=12);
 # "wannier90/wannier.wout", etc.). A different file prefix can be given with the
 # keyword argument `fileprefix` as shown below.
 #
-# We now solve for 8 MLWF using wannier90:
+# We now solve for 4 MLWF using wannier90:
```
jaemolihm, line 45: This should be 5 too

mfherbst: @jaemolihm good call!

LaurentVidal95:
> We rationalized this with Laurent as follows: there is a sigma Wannier on each bond plus a pz Wannier on each atom. On each unit cell there are two atoms, one bond in the unit cell, and four bonds shared with other unit cells, for a total of 2+1+4/2 = 5 Wannier. @LaurentVidal95 can you check the Wannier produced by this PR look like the ones you showed us yesterday?

Hi everyone, I am very sorry for the delay I wasn't recieving the git notifications... It is fixed now.
@antoine-levitt the code is the one I sent to Michael yesterday (within a few modifications of syntax) so it should provide the same results !

However I forgot yesterday to remove the kwargs `write_xyz=true` and `translate_home_cell=true`.
I used the xyz file for my own post-processing which has nothing to do with the wannierization in itself so I think we can remove them !

mfherbst:
@LaurentVidal95 I don't think it hurts. Let's just keep them.

PR #551, opened 2021-12-03
Refactor ene_ops and Terms interface

mfherbst:


src/PlaneWaveBasis.jl
```diff
@@ -224,13 +228,13 @@ end
     kweights = T.(model.n_spin_components .* kweights) ./ tot_weight
     @assert mpi_sum(sum(kweights), comm_kpts) ≈ model.n_spin_components
 
-    # Create dummy terms array for basis to handle
-    terms = Vector{Any}(undef, length(model.term_types))
-
+    # Volume element for integration
     dvol = model.unit_cell_volume ./ prod(fft_size)
```
antoine-levitt, line 232: isn't this already documented as a field in pwbasis?

src/terms/entropy.jl
```diff
@@ -5,26 +5,20 @@ This is in particular useful because the free energy,
 not the energy, is minimized at self-consistency.
 """
 struct Entropy end
-(::Entropy)(basis) = TermEntropy(basis)
+(::Entropy)(::AbstractBasis) = TermEntropy()
```
antoine-levitt, line 8: why?

src/terms/operators.jl
```diff
@@ -96,12 +96,12 @@ Nonlocal operator in Fourier space in Kleinman-Bylander format,
...
-    basis
-    kpoint
+struct NonlocalOperator{T <: Real, PT, DT} <: RealFourierOperator
+    basis::PlaneWaveBasis{T}
+    kpoint::Kpoint{T}
     # not typed, can be anything that supports PDP'ψ
-    P
-    D
+    P::PT
+    D::DT
```
antoine-levitt, line 104: On this and on others: why type the fields? This is not performance critical at all, and doesn't really help with readability (at least not these two)

src/PlaneWaveBasis.jl
mfherbst: I don't get what you mean by this comment, sorry. It's the place this quantity is computed.

src/terms/entropy.jl
mfherbst: You rather like `::Any`?

antoine-levitt: I just mean `basis` is more clear than `::AbstractBasis` to me

src/PlaneWaveBasis.jl
antoine-levitt: I mean the comment is unnecessary

src/terms/operators.jl
mfherbst: It does not really make it a lot worse in terms of readability in my opinion and at the same time downstream types (e.g. when computing P * D * P' * \psi can be fully inferred.

src/PlaneWaveBasis.jl
mfherbst: Ah yes ok that might be.

src/terms/entropy.jl
mfherbst: but that's an unused variable.

antoine-levitt: which is bad why?

mfherbst: Arguably no real difference here, but still flags up by my static code analysis.

src/terms/operators.jl
antoine-levitt: Inference does not matter at all here. My experience is that you can get into big messes because of eager typing of everything. Not particularly critical here, but not a good practice in my opinion

mfherbst: In this case I don't see the mess. It is a small struct with two fields to type-stably store some intermediate data. It's use is very local, the overhead in constructing and using it is zero due to this change. Has basically no downsides and only benefits (which might be small, but ok).

src/terms/entropy.jl
antoine-levitt: The answer to that is fix the tools, not fix the code... (which is why I'm very distrustful of analysis/formatting tools). Unused variables are completely fine, esp. with julia's somewhat informal way of doing interfaces. Anyway, doesn't really matter

src/terms/operators.jl
antoine-levitt: Oh it's just my general heuristics of more characters = bad :-p Just makes it a bit harder to parse, and generic structs are slightly nastier (eg to print)

mfherbst: Hmm. I have to say in the cases we have here I still lean to have the type annotations. It's not really make the code a lot longer or more complex, but it does give a bit of speed (ok arguably not in a part of the code it really matters, but energy_hamiltonian becomes notably faster).

antoine-levitt: My general rule is that type annotations (when done for performance and not for clarity or safety) are an optimization, and *any* optimization has a (sometimes hidden) cost (more things to parse when reading code, more potential bugs, less flexibility if we want to change something, more inference time, etc), so optimizations are only to be done if profiling demonstrates they are relevant. We can keep it if you really want but I just want to avoid it being a pattern

mfherbst: Ok, that's fair.

PR #552, opened 2021-12-03
Remove model backreference from `Kpoint` structure

mfherbst:


src/PlaneWaveBasis.jl
```diff
@@ -332,32 +319,77 @@ end
...
+    G_vectors_cart(basis::PlaneWaveBasis, kpt::Kpoint)
+
+The list of ``G`` vectors of a given `basis` or `kpt`, in cartesian coordinates.
+"""
 function G_vectors_cart(basis::PlaneWaveBasis)
     (basis.model.recip_lattice * G for G in G_vectors(basis.fft_size))
 end
+function G_vectors_cart(basis::PlaneWaveBasis, kpt::Kpoint)
+    (basis.model.recip_lattice * G for G in G_vectors(kpt))
+end
```
antoine-levitt, line 356: if we do that shouldn't we just have basis, kpt for G_vectors for consistency?

src/PlaneWaveBasis.jl
```diff
@@ -332,32 +319,77 @@ end
...
+q_vectors(kpt::Kpoint) = ((G + kpt.coordinate) for G in G_vectors(kpt))
+
+@doc raw"""
+    q_vectors_cart(basis::PlaneWaveBasis, kpt::Kpoint)
+
+The list of ``q = G + k`` vectors, in cartesian coordinates.
+"""
+function q_vectors_cart(basis::PlaneWaveBasis, kpt::Kpoint)
+    (basis.model.recip_lattice * (G + kpt.coordinate) for G in G_vectors(kpt))
+end
```
antoine-levitt, line 372: Hm, this is a great idea but maybe too clever? The trouble is that q_vectors could mean the set of all k+G, for all k and G

src/postprocess/current.jl
```diff
@@ -9,7 +9,7 @@ function compute_current(basis::PlaneWaveBasis, ψ, occupation)
         for (n, ψnk) in enumerate(eachcol(ψ[ik]))
             ψnk_real = G_to_r(basis, kpt, ψnk)
             for α = 1:3
-                dαψnk = [im*(G+kpt.coordinate_cart)[α] for G in G_vectors_cart(kpt)] .* ψnk
+                dαψnk = [im * q[α] for q in q_vectors_cart(basis, kpt)] .* ψnk
```
antoine-levitt, line 12: I have to admit that's pretty cool...

src/PlaneWaveBasis.jl
antoine-levitt: other issue is that q vectors is not such a standard thing. In fact, q is standard for the *phonon* Brillouin zone, which we'll want to have for phonons... So, kpG? Kinda ugly but descriptive...

mfherbst: Yes I agree. How about this fancy `k+G`? Might be too fancy?

src/PlaneWaveBasis.jl
mfherbst: I wondered about that myself. Given that we both wonder I do that for consistency.

src/PlaneWaveBasis.jl
antoine-levitt: Hm, yeah, probably too confusing, it's not clear a priori how to write this using tab completion. I think last time we landed on "unicode craziness OK for local variables in functions, but tricky in function names"... kpG, k_plus_G? k_plus_G_vectors is not too bad

mfherbst: I went with `kplusG_vectors` (like in `solve_ΩplusK`).

antoine-levitt: OK that makes sense!

src/PlaneWaveBasis.jl
```diff
@@ -12,30 +12,18 @@ More generally, a ``k``-point is a block of the Hamiltonian;
...
 """
 struct Kpoint{T <: Real}
-    model::Model{T}               # TODO Should be only lattice/atoms
     spin::Int                     # Spin component can be 1 or 2 as index into what is
                                   # returned by the `spin_components` function
     coordinate::Vec3{T}           # Fractional coordinate of k-point
     coordinate_cart::Vec3{T}      # Cartesian coordinate of k-point
     mapping::Vector{Int}          # Index of G_vectors[i] on the FFT grid:
-                                  # G_vectors(basis)[kpt.mapping[i]] == G_vectors(kpt)[i]
+                                  # G_vectors(bas)[kpt.mapping[i]] == G_vectors(bas, kpt)[i]
```
antoine-levitt, line 20: ?? We don't usually use bas

src/terms/kinetic.jl
```diff
@@ -12,8 +12,9 @@ struct TermKinetic <: Term
     kinetic_energies::Vector{Vector} # kinetic energy 1/2|G+k|^2 for every kpoint
 end
 function TermKinetic(basis::PlaneWaveBasis; scaling_factor=1)
-    kinetic_energies = [[scaling_factor * sum(abs2, G + kpt.coordinate_cart) / 2
-                         for G in G_vectors_cart(kpt)]
+    T = eltype(basis)
+    factor = scaling_factor / T(2)
```
antoine-levitt, line 16: I don't really like this, it doesn't make logical sense: the 1/2 and the scaling_factor belong to different parts of the formula, it doesn't make much sense to group them. Also is T(2) really necessary?

src/terms/nonlocal.jl
```diff
@@ -149,29 +148,28 @@ where pihat(q) = ∫_R^3 pi(r) e^{-iqr} dr
...
 
         # Combine with structure factors
         for r in positions
             # k+G in this formula can also be G, this only changes an unimportant phase factor
-            structure_factors = [cis(-2T(π)*dot(kpt.coordinate + G, r)) for G in G_vectors(kpt)]
-            for iproj = 1:count_n_proj(psp)
-                @views proj_vectors[:, offset+iproj] .= (structure_factors
-                                                        .* form_factors[:, iproj]
-                                                        ./ sqrt(model.unit_cell_volume))
+            structure_factors = map(q -> cis(-2T(π) * dot(q, r)), Gplusk_vectors(basis, kpt))
```
antoine-levitt, line 168: Do we "officially" want to use `map` instead of comprehensions? I quite like comprehensions, I find it's more readable than map, but I understand it makes more sense (eg for GPU)...

antoine-levitt (commented):
Cool, thanks! Should we also do it for G_to_r/r_to_G? Or let's wait until we actually need it...

src/terms/nonlocal.jl
mfherbst: Yeah I think we should at least slowly move towards it, even though I agree comprehensions are generally more readable.

src/terms/kinetic.jl
mfherbst: T(2) Yes, otherwise you get promotion to Float64 by default.

mfherbst:
What do you mean `Gplusk` or which aspect of this PR?

src/terms/nonlocal.jl
antoine-levitt: the main problem I have with the map is that it's less easy to see that q is in Gplusk

antoine-levitt: we could plausibly have a `@keep_types` macro that rewrites the type of the comprehension or something

mfherbst: Hmm. I don't think more macros will make it easier for AD suites etc. to process the code. Also generally comprehensions are type stable and can be type-annotated, so I don't see what such a macro would do differently.

PR #553, opened 2021-12-13
Include advertisement to DFTK school 2022

mfherbst:
Put some advertisement for next year's [DFTK school](https://school2022.dftk.org) into the docs.

### DIFF
--- a/docs/src/school.md
+++ b/docs/src/school.md
@@ -1,0 +1,2 @@
+# DFTK School 2022
+For more details and registration info see the [school's website](https://school2022.dftk.org).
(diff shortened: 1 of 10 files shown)


antoine-levitt:
Feels a bit overkill to have it in the docs, maybe just in the index.html?

mfherbst:
Ok, I made it more subtle, but also included something in the github repo.

docs/src/school.md
```diff
@@ -0,0 +1,2 @@
+# DFTK School 2022
+For more details and registration info see the [school's website](https://school2022.dftk.org).
```
antoine-levitt, line 2: so just remove it?

mfherbst: Ah that will just redirect. No-one will ever see this.

mfherbst: it's just because documenter does not allow you to put links into the sidebar.

antoine-levitt: Oh strange, ok

Issue #533, opened 2021-10-14
The kpoint symmetry and band plotting cleanup issue


mfherbst:
Some points to cleanup left after #496 is merged

### Quick n' easy
- [x] Drop enforcing `pymatgen` installation (also from the docs) (#534)
- [x] `high_symmetry_kpath` does not use atomic units (#536)
- [x] `plot_bandstructure` does not use atomic units consistently (#536)
- [x] `plot_bandstructure` interface is quirky for linear Hamiltonians (#536)

### Cleanup
- [x] Use `Spglib` and retire direct use of `spglib_jll` (#876)
- [x] Merge `spglib_standardize_cell` and `get_spglib_lattice` (#568)
- [x] Drop kpoint duplication (#555)
- [x] Band structures also for supercells (#555)
- [x] Use `Brillouin.jl` in `high_symmetry_kpath` also for 1D and 2D lattices (#555)
- [ ] Integrate more with `Brillouin.jl` for BZ plots or band structure plots (see also #503)

### Consistency
- [ ] Take care of time-reversal symmetry / absence thereof consistently for kpath (see [this discussion](https://github.com/JuliaMolSim/DFTK.jl/pull/496#discussion_r725203554))
- [ ] Take care of spin / magnetic moments in cell standardisation and kpath generation

jaemolihm:
Hi, I'm planning to make a PR to Brillouin.jl for a function that gives k path for arbitrary lattice vectors (https://github.com/thchr/Brillouin.jl/issues/13#issuecomment-968625017). After that, one can make `high_symmetry_kpath` in DFTK much simpler. Independently, I have a working branch that cleans up the kpath routines, e.g. deleting kpoint duplication. If nobody is working on these, I can pick it up and make a PR (after making a PR to Brillouin.jl).

mfherbst:
@jaemolihm  Yes that would be highly appreciated! Noone is working on these at the moment. Thanks for your work on this (truly annoying) aspects of solid-state modelling!

Issue #503, opened 2021-07-23
Refactor `plot_bandstructure` interface


mfherbst:
I've become a little unhappy about the `plot_bandstructure` interface, especially the fact that the kline_density has so strange units.
When Brillouin.jl is used this should be refactored.

thchr:
Yeah, I had just briefly seen it the other day: cool effort, definitely!

One thing that came to mind when I was browsing the code of AtomsBase.jl was that it might be nice for various crystallography-related packages to settle on some standard "back-bone" for how to specify and transform direct and reciprocal lattices. The motivation would be to reduce the setting-related uncertainty that we also discussed in #496.
The implementations in Bravais might be suitable in that regard (e.g., in place of `Box`).

mfherbst:
> The implementations in Bravais might be suitable in that regard (e.g., in place of Box).

Feel free to raise that aspect in `AtomsBase.jl`. Since AtomsBase is meant to be a bit more general (i.e. including molecular systems) I suppose it Bravais might not yet fit at that level, but I could imagine it to make sense to have a slightly more specialised implementation just for periodic problems ... and at that level Bravais could fit very well, I agree.

mfherbst:
See the code in https://github.com/thchr/Brillouin.jl/blob/master/src/requires/plotlyjs_dispersion.jl

PR #555, opened 2021-12-15
Band structure for non-standard lattice, cleanup

jaemolihm:
As I mentioned in #533, I make some improvements to the high-symmetry k-path and band plotting routines.

### Changes
* Use the new `Brillouin.irrfbz_path(cell::Spglib.Cell)` function (https://github.com/thchr/Brillouin.jl/pull/15) that gives a k-path for non-standard unit cell, added tests.
* Drop kpoint duplication (which was needed for interfacing with pymatgen) in high_symmetry_kpath
* Simplify band structure plotting. The concept of branches is removed.

### Things to be discussed
* [x] `Brillouin.irrfbz_path(cell::Spglib.Cell)` does not work for undistorted supercells. Currently, `high_symmetry_kpath` raises an error in such as case. Is there a better option?
  * Brillouin was updated to give k path for supercells, too.

### Example: band structure of silicons with a non-standard lattice parameters
```julia
using DFTK
using Plots

Ecut = 7
kgrid = [3, 3, 3]
psp = ElementPsp(14, psp=load_psp("hgh/lda/si-q4"))

# Non-standard lattice parameters that describe the same system as model_standard.
lattice = [0 1 1; 1 0 1; 1 1 0] .* 5.13
model = model_LDA(lattice, [psp => [ones(3)/8, -ones(3)/8]])
scfres = self_consistent_field(basis, tol=1e-12);
plot_bandstructure(scfres; kline_density=20)


lattice = [0 1 1; 1 0 1+3; 1 1 0+3] .* 5.13
model = model_LDA(lattice, [psp => [[-2,1,1]./8, -[-2,1,1]./8]], temperature=0.01)
scfres = self_consistent_field(basis, tol=1e-12);
plot_bandstructure(scfres; kline_density=20)
```
Standard:
image

Non-standard, old code:
image

Non-standard, new code:
image

### DIFF
--- a/src/postprocess/band_structure.jl
+++ b/src/postprocess/band_structure.jl
@@ -73,10 +73,11 @@ function kpath_get_branch(kpath::KPathInterpolant{D}, ibranch::Integer) where {D
     map(k -> vcat(k, zeros_like(k, 3 - D)), kpath.kpaths[ibranch])
 end
 
-@timing function compute_bands(basis::PlaneWaveBasis, kpath::KPathInterpolant;
+@timing function compute_bands(basis::PlaneWaveBasis, kcoords::AbstractVector;
                                n_bands=default_n_bands_bandstructure(basis.model),
                                ρ=nothing, eigensolver=lobpcg_hyper,
                                tol=1e-3, show_progress=true, kwargs...)
+    # kcoords are the kpoint coordinates in fractional coordinates
     if isnothing(ρ)
         if any(t isa TermNonlinear for t in basis.terms)
             error("If a non-linear term is present in the model the converged density is required " *
@@ -87,7 +88,6 @@ end
     end
 
     # Create basis with new kpoints, without any symmetry operations.
-    kcoords  = kpath_get_kcoords(kpath)
     kweights = ones(length(kcoords)) ./ length(kcoords)
     bs_basis = PlaneWaveBasis(basis, kcoords, kweights)
 
@@ -98,6 +98,9 @@ end
         @warn "Eigensolver not converged" iterations=band_data.iterations
     end
     merge((; basis=bs_basis), select_eigenpairs_all_kblocks(eigres, 1:n_bands))
+end
+function compute_bands(basis::PlaneWaveBasis, kpath::KPathInterpolant; kwargs...)
+    compute_bands(basis, kpath_get_kcoords(kpath))
 end
 
 
@@ -214,10 +217,9 @@ by default. Another standard choices is `unit=u"eV"` (electron volts).
 The `kline_density` is given in number of ``k``-points per inverse bohrs (i.e.
 overall in units of length).
 """
-function plot_bandstructure(basis::PlaneWaveBasis;
+function plot_bandstructure(basis::PlaneWaveBasis, kpath::KPath=irrfbz_path(basis.model);
                             εF=nothing, kline_density=40u"bohr",
                             unit=u"hartree", kwargs_plot=(; ),
-                            kpath::KPath=irrfbz_path(basis.model),
                             kwargs...)
     mpi_nprocs() > 1 && error("Band structures with MPI not supported yet")
     if !isdefined(DFTK, :PLOTS_LOADED)
@@ -232,7 +234,7 @@ function plot_bandstructure(basis::PlaneWaveBasis;
     band_data = compute_bands(basis, kinter; kwargs...)
     plot_band_data(kinter, band_data; εF, unit, kwargs_plot...)
 end
-function plot_bandstructure(scfres::NamedTuple;
+function plot_bandstructure(scfres::NamedTuple, kpath::KPath=irrfbz_path(scfres.basis.model);
                             n_bands=default_n_bands_bandstructure(scfres), kwargs...)
-    plot_bandstructure(scfres.basis; n_bands, ρ=scfres.ρ, εF=scfres.εF, kwargs...)
+    plot_bandstructure(scfres.basis, kpath; n_bands, ρ=scfres.ρ, εF=scfres.εF, kwargs...)
 end
--- a/src/postprocess/band_structure.jl
+++ b/src/postprocess/band_structure.jl
@@ -1,4 +1,4 @@
-import Brillouin
+using Brillouin
 import Brillouin.KPaths: Bravais
 
 @doc raw"""
@@ -14,54 +14,61 @@ overall in units of length).
 If the cell is a supercell of a smaller primitive cell, the standard ``k``-path of the
 associated primitive cell is returned. So, the high-symmetry ``k`` points are those of the
 primitive cell Brillouin zone, not those of the supercell Brillouin zone.
-"""
-function high_symmetry_kpath(model; kline_density=40, magnetic_moments=[])
-    kline_density = austrip(kline_density)
 
-    if model.n_dim == 1  # Return fast for 1D model
-        # TODO Is this special-casing of 1D is not needed for Brillouin.jl any more
-        #
-        # Just use irrfbz_path(1, DirectBasis{1}([1.0]))
-        # (see https://github.com/JuliaMolSim/DFTK.jl/pull/496/files#r725205860)
-        #
-        # Length of the kpath is recip_lattice[1, 1] in 1D
-        n_points = max(2, 1 + ceil(Int, kline_density * model.recip_lattice[1, 1]))
-        kcoords  = [@SVector[coord, 0, 0] for coord in range(-1//2, 1//2, length=n_points)]
-        klabels  = Dict("Γ" => zeros(3), "-½" => [-0.5, 0.0, 0.0], "½" => [0.5, 0, 0])
-        return (kcoords=kcoords, klabels=klabels,
-                kpath=[["-½", "½"]], kbranches=[1:length(kcoords)])
+Due to lacking support in `Spglib.jl` for two-dimensional lattices it is (a) assumed that
+`model.lattice` is a *conventional* lattice and (b) required to pass the space group
+number using the `sgnum` keyword argument.
+"""
+function high_symmetry_kpath(model; kline_density=40, sgnum=nothing, magnetic_moments=[])
+    if !isnothing(sgnum) && model.n_dim in (1, 3)
+        @warn("sgnum keyword argument unused in `high_symmetry_kpath` unused " *
+              "unless a 2-dimensional lattice is encountered.")
     end
 
-    # - Brillouin.jl expects the input direct lattice to be in the conventional lattice
-    #   in the convention of the International Table of Crystallography Vol A (ITA).
-    # - spglib uses this convention for the returned conventional lattice,
-    #   so it can be directly used as input to Brillouin.jl
-    # - The output k-Points and reciprocal lattices will be in the CDML convention.
-    kp     = Brillouin.irrfbz_path(spglib_cell(model, magnetic_moments))
-    kinter = Brillouin.interpolate(kp, density=kline_density)
+    # Brillouin.jl expects the input direct lattice to be in the conventional lattice
+    # in the convention of the International Table of Crystallography Vol A (ITA).
+    #
+    # The output of Brillouin.jl are k-Points and reciprocal lattice vectors
+    # in the CDML convention.
+    if model.n_dim == 1
+        # Only one space group; avoid spglib here
+        kp = Brillouin.irrfbz_path(1, [[model.lattice[1, 1]]], Val(1))
+    elseif model.n_dim == 2
+        if isnothing(sgnum)
+            error("sgnum keyword argument (specifying the ITA space ""group number) " *
+                  "is required for band structure plots in 2D lattices.")
+        end
+        # TODO We assume to have the conventional lattice here.
+        direct_basis = Bravais.DirectBasis(collect(eachcol(model.lattice)))
+        kp = Brillouin.irrfbz_path(sgnum, direct_basis)
+    elseif model.n_dim == 3
+        # Brillouin.jl has an interface to Spglib.jl to directly reduce the passed
+        # lattice to the ITA conventional lattice and so the Spglib cell can be
+        # directly used as an input.
+        cell, _ = spglib_cell(model, magnetic_moments)
+        kp      = Brillouin.irrfbz_path(cell)
+    end
+    kinter  = Brillouin.interpolate(kp, density=austrip(kline_density))
 
-    # TODO Need to take care of time-reversal symmetry here!
+    # TODO In case of absence of time-reversal symmetry we need to explicitly
+    #      add the inverted kpath here!
     #      See https://github.com/JuliaMolSim/DFTK.jl/pull/496/files#r725203554
 
-    kcoords = vcat(kinter.kpaths...)
-    klabels = kp.points
-    kpath   = [[string(el) for el in path] for path in kp.paths]
-    kbranches = [1:length(kinter.kpaths[1])]
-    for n in length.(kinter.kpaths)[2:end]
-        push!(kbranches, kbranches[end].stop+1:kbranches[end].stop+n)
-    end
-    (; kcoords, klabels, kpath, kbranches)
+    kinter
 end
 
+"""Return kpoint coordinates in reduced coordinates"""
+function kpath_get_kcoords(kpath::KPathInterpolant{D}) where {D}
+    map(k -> vcat(k, zeros_like(k, 3 - D)), kpath)
+end
+function kpath_get_branch(kpath::KPathInterpolant{D}, ibranch::Integer) where {D}
+    map(k -> vcat(k, zeros_like(k, 3 - D)), kpath.kpaths[ibranch])
+end
 
-@timing function compute_bands(basis, kcoords;
+@timing function compute_bands(basis::PlaneWaveBasis, kpath::KPathInterpolant;
                                n_bands=default_n_bands_bandstructure(basis.model),
                                ρ=nothing, eigensolver=lobpcg_hyper,
                                tol=1e-3, show_progress=true, kwargs...)
-    # Create basis with new kpoints, without any symmetry operations.
-    kweights = ones(length(kcoords)) ./ length(kcoords)
-    bs_basis = PlaneWaveBasis(basis, kcoords, kweights)
-
     if isnothing(ρ)
         if any(t isa TermNonlinear for t in basis.terms)
             error("If a non-linear term is present in the model the converged density is required " *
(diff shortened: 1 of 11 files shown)


src/postprocess/band_structure.jl
```diff
@@ -99,35 +83,37 @@ end
...
-
-    map(branches[2:end]) do branch
-        branch_data = Dict(key => data[key][branch.kindices, :, :] for key in keys(data))
-        ret = (klabels=(getlabel(kcoords[branch.kindices[1]]),
-                        getlabel(kcoords[branch.kindices[end]])),
-               kdistances=branch.kdistances,
-               kindices=branch.kindices)
-        merge(ret, (; branch_data...))
-    end
+    kdistances, tick_distances, tick_labels
```
mfherbst, line 116: As a general comment: In such cases I would better return a named tuple compared to a tuple. You can still use it as a tuple, plus it has the additional benefit that the included "name" of the object carries more meaning.

src/postprocess/band_structure.jl
```diff
@@ -154,26 +140,10 @@ function prepare_band_data(band_data; datakeys=[:λ, :λerror],
...
-        end
-        push!(tick_labels, label)
-        push!(tick_distances, distance)
-    end
 
-    (branches=branches, ticks=(distances=tick_distances, labels=tick_labels),
-     n_bands=n_bands, n_kcoord=n_kcoord, n_spin=n_spin, basis=basis)
+    kdistances, tick_distances, tick_labels = kdistances_and_ticks(kcoords_cart, klabels_cart)
+
+    (;ticks=(distances=tick_distances, labels=tick_labels), kdistances, n_bands, n_kcoord, n_spin, basis, data...)
```
mfherbst, line 146: We try to keep lines shorter than around 90 characters.

mfherbst (commented):
Thanks very much @jaemolihm! Indeed a great improvement to usability. I left some comments in the code on the points to be discussed.

src/postprocess/band_structure.jl
```diff
@@ -99,35 +83,37 @@ end
 end
 
 
-function split_into_branches(kcoords, data::Dict, klabels::Dict)
+function kdistances_and_ticks(kcoords, klabels::Dict)
```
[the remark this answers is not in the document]
jaemolihm: Indeed, I hadn't thought of the problem. Thanks for pointing it out.

If the band discontibuity is the only problem, what about storing only the indices of the co tinuous k points as a vector of ranges, e.g. [1:5, 6:8], and using them for plotting?

src/postprocess/band_structure.jl
```diff
@@ -30,39 +30,23 @@ function high_symmetry_kpath(model; kline_density=40)
...
-    kp     = Brillouin.irrfbz_path(sgnum, direct_basis)
+    # TODO: Wrap this into a function? (Repeated pattern in spglib.jl)
+    spg_positions, spg_numbers, _ = spglib_atoms(model.atoms)
+    cell = Spglib.Cell(model.lattice, spg_positions, spg_numbers)
+    dset = Spglib.get_dataset(cell)
+
+    # If the input cell is a supercell (without any distortion), Brillouin.irrfbz_path does not work.
+    if !isapprox(det(cell.lattice), det(dset.primitive_lattice)) # check volumes agree
+        error(DomainError(cell, "`cell` is a supercell and untreatable by `Brillouin.irrfbz_path`. "))
+    end
```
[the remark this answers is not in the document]
jaemolihm: The issue is that irrfbz_path in Brillouin.jl will error, not warn. I agree that giving some k path is better than giving nothing.

If the model is not treatable because it is a supercell, I think there are two ways to give a k path: use a the cell same lattice but (1) with maximal symmetry (i.e. one atom per unit cell) or (2) with minimal symmetry.

src/postprocess/band_structure.jl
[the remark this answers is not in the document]
mfherbst: That would probably be a simple and sufficient solution, I agree.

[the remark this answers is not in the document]
mfherbst: On that note: I also find the `klabels` dict a bit clunky. How about changing it to a list of pairs, similar what Brillouin's `kp.points` contains. What to use there depends of course what make the overall implementation and handling of the data generation and plotting easiest.

src/postprocess/band_structure.jl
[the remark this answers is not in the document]
mfherbst: My first rationale would be that if you have a supercell, you sort of look into that cell as one entity, (e.g. because you later want to perturb it and break the symmetry), so I would argue one would want minimal symmetry.

But now that I think about this some more I'm not sure if the best way to go about this is to be automatic. It would be better in such "unusual" cases to force the user to specify what he wants. What I'm thinking is to ask for an explicit list of symmetries to be taken into account when doing the plotting. In practice we already have `model.symmetries` in `model`, which specifies the symmetries the user has set to be used. Ideally one would just use that as a default and allow the user of `high_symmetry_kpath` to overwrite them.

If given those symmetries the algorithm works, it should just go through. Else I think its fine to error and ask the user to manually tinker with the symmetries.

What do you think @jaemolihm?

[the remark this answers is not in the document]
jaemolihm: I think that the idea of using the given symmetry is great. But, I cannot think of how to implement it.

To run `irrfbz_path`, one needs 1) the space group, and 2) the standard primitive cell. The space group for given symmetry can be easily found using spglib. But, AFAIK spglib does not have a function to standardize the cell using only given symmetries.

Also, now I am not sure how to implement the minimal symmetry option either because of the same reason. One could construct an artificial cell with the same lattice but many atoms randomly placed to break all symmetries, but this seems a bit too dirty.

Another option is to use take the k path of the primitive unit cell (which will have a large BZ than the supercell) and fold it back to the supercell BZ. A downside for this option is that the high symmetry k point labels lose their meaning.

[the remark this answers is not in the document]
mfherbst: > But, I cannot think of how to implement it.

No worries! I did not expect this to be easily doable ... I was just hypothesising about the "best case". The implementation we had before we went to Bravais has been way less correct / useful in any case, so we are certainly moving in the right direction with this PR :smile:.

> But, AFAIK spglib does not have a function to standardize the cell using only given symmetries.

I think you can do that, but it probably involves doing some nasty calls deep into the library, which will not be fun ... or easy to do. Probably best to put this on the long list of #533.

> Another option for solution, not discusseded in my previous comment, is to use take the k path of the primitive unit cell (which will have a large BZ than the supercell) and fold it back to the supercell BZ. A downside for this option is that the high symmetry k point labels lose their meaning.

Actually all things considered, I think this is short-term the best solution. Great suggestion!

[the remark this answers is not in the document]
jaemolihm: Hi, I made a PR to Brillouin.jl https://github.com/thchr/Brillouin.jl/pull/16 that implements the alternative I mentioned (taking k path of the primitive cell). If that gets merged, I will update this PR.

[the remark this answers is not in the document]
jaemolihm: Now, `Brillouin.irrfbz_kpath`, as well as `DFTK.high_symmetry_kpath` gives the primitive cell k path for undistorted supercells.

src/postprocess/band_structure.jl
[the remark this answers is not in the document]
jaemolihm: I re-introduced branches as a list of ranges, so the points are not connected at discontinuous k points (see the figure below).

> On that note: I also find the klabels dict a bit clunky. How about changing it to a list of pairs, similar what Brillouin's kp.points contains. What to use there depends of course what make the overall implementation and handling of the data generation and plotting easiest.

I do not understand this point, could you clarify? `klabels` is a mapping from k point labels to the vector. Since this mapping is not ordered, I think a dict is more adequate than a list of pairs. Also, Brillouin's `kp.points` is also a dict (only difference is that the keys are symbols, not strings): https://github.com/thchr/Brillouin.jl/blob/master/src/KPaths.jl#L55-L61

jaemolihm:
### Updated example
```julia
using DFTK
using Plots

Ecut = 7
kgrid = [3, 3, 3]
psp = ElementPsp(14, psp=load_psp("hgh/lda/si-q4"))

# Non-standard lattice parameters that describe the same system as model_standard.
lattice = [0 1 1; 1 0 1; 1 1 0] .* 5.13
model = model_LDA(lattice, [psp => [ones(3)/8, -ones(3)/8]])
scfres = self_consistent_field(basis, tol=1e-12);
p_standard = plot_bandstructure(scfres; kline_density=20)


lattice = [0 1 1; 1 0 1+3; 1 1 0+3] .* 5.13
model = model_LDA(lattice, [psp => [[-2,1,1]./8, -[-2,1,1]./8]], temperature=0.01)
scfres = self_consistent_field(basis, tol=1e-12);
p_nonstandard = plot_bandstructure(scfres; kline_density=20)

lattice = [0 1 2; 1 0 2; 1 1 0] .* 5.13
model = model_LDA(lattice, [psp => [[1,1,1/2]./8, -[1,1,1/2]./8, [1,1,9/2]./8, -[1,1,9/2]./8]]);
basis = PlaneWaveBasis(model; Ecut, kgrid, kshift);
scfres = self_consistent_field(basis, tol=1e-12);
p_super = plot_bandstructure(scfres; kline_density=20)

Plots.plot(p_standard, p_nonstandard, p_super, layout=(1, 3), title=["standard" "non-standard" "supercell"])
plot!(size=(1000, 400))
```

image

src/postprocess/band_structure.jl
```diff
@@ -30,39 +32,17 @@ function high_symmetry_kpath(model; kline_density=40)
...
-    kcoords = empty(first(kinter.kpaths))
-    for kbranch in kinter.kpaths
-        idcs = findall(k -> any(sum(abs2, k - kcomp) < 1e-5
-                                for kcomp in values(kp.points)), kbranch)
-        @assert length(idcs) ≥ 2
-        idcs = idcs[2:end-1]  # Don't duplicate first and last
-        idcs = sort(append!(idcs, 1:length(kbranch)))
-        append!(kcoords, kbranch[idcs])
-    end
+    kcoords = vcat(kinter.kpaths...)
```
mfherbst, line 45: I wonder if it is wise to "throw away" the branches here. Would it not simplify things to just keep the list of list of kpoints here and only issue a `vcat( )` when constructing the PlaneWaveBasis? That way you might be able to get rid of the `kdistances_and_ticks` function completely without preventing any of the branch-based tricks in the plotting.

src/postprocess/band_structure.jl
```diff
@@ -30,39 +32,17 @@ function high_symmetry_kpath(model; kline_density=40)
...
 
     T = eltype(kcoords[1])
     klabels = Dict{String,Vector{T}}(string(key) => val for (key, val) in kp.points)
```
mfherbst, line 48: Similarly here, I don't see the reason to convert this to string. We needed that for pymatgen compatibility, but we might as well keep the symbol now.

src/postprocess/band_structure.jl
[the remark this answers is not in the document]
mfherbst: Ah sorry, then that's an error on my side. In that case I am all for keeping the Dict (with the additional suggestion to not convert the keys to string).

test/compute_bands.jl
```diff
@@ -168,46 +162,27 @@ end
...
         end
     end
-    for ibr in 1:3
-        @test ret.branches[ibr].kdistances == ref_kdist[:, ibr]
-    end
-
+    @test ret.kdistances == ref_kdist
     @test ret.ticks.labels == ["Γ", "X", "W | U", "X"]
-    @test ret.ticks.distances == [0.0, ref_kdist[end, 1], ref_kdist[end, 2], ref_kdist[end, 3]]
+    @test ret.ticks.distances == ref_kdist[[1,3,5,8]]
```
mfherbst, line 185: Spacing after commas:
```suggestion
    @test ret.ticks.distances == ref_kdist[[1, 3, 5, 8]]
```

test/compute_bands.jl
```diff
@@ -168,46 +162,27 @@ end
...
+    @test ret.n_kcoord == 8
     @test ret.n_bands  == 4
 
-    @test ret.branches[1].kindices == [1, 2, 3]
-    @test ret.branches[2].kindices == [4, 5, 6]
-    @test ret.branches[3].kindices == [7, 8, 9]
-
-    @test ret.branches[1].klabels == ("Γ", "X")
-    @test ret.branches[2].klabels == ("X", "W")
-    @test ret.branches[3].klabels == ("U", "X")
```
mfherbst, line 180: In case we decide to keep the `kdistances_and_ticks` function, I'd still test the branch splitting explicitly by checking the branch index ranges in the `branches` field.

test/compute_bands.jl
```diff
@@ -222,3 +197,25 @@ end
...
+@testset "High-symmetry kpath for nonstandard lattice" begin
+    testcase = silicon
+    spec = ElementPsp(testcase.atnum, psp=load_psp(testcase.psp))
+    lattice_standard = [0 1 1; 1 0 1; 1 1 0] .* 5.13
+    model_standard = model_LDA(lattice_standard, [spec => [ones(3)/8, -ones(3)/8]])
+
+    # Non-standard lattice parameters that describe the same system as model_standard.
+    lattice_nonstandard = copy(lattice_standard)
+    lattice_nonstandard[:, 3] .+= lattice_nonstandard[:, 1] .* 3
+    model_nonstandard = model_LDA(lattice_nonstandard, [spec => [[-2,1,1]./8, -[-2,1,1]./8]])
```
mfherbst, line 210:
```suggestion
    model_nonstandard = model_LDA(lattice_nonstandard,
                                  [spec => [[-2, 1, 1]/8, -[-2, 1, 1]/8]])
```

test/compute_bands.jl
```diff
@@ -222,3 +197,25 @@ end
...
+
+    data_standard = high_symmetry_kpath(model_standard)
+    data_nonstandard = high_symmetry_kpath(model_nonstandard)
+
+    # Check the k points are the same in Cartesian coordinates.
+    for (k_standard, k_nonstandard) in zip(data_standard.kcoords, data_nonstandard.kcoords)
+        @test model_standard.recip_lattice * k_standard ≈ model_nonstandard.recip_lattice * k_nonstandard
```
mfherbst, line 217:
```suggestion
        @test(  model_standard.recip_lattice    * k_standard
              ≈ model_nonstandard.recip_lattice * k_nonstandard)
```

mfherbst (commented):
Thanks very much for the update. The output is pretty much exactly what I had in mind.

The only major thing left to discuss is whether the rather annoying `kdistances_and_ticks` function is needed at all, see my comments. If you think it's not possible to easily get rid of it, than we can also merge it as is.

Also: Could you please lower-bound the version of `Brillouin.jl` in `Project.toml` to ensure people use a version that is compatible with this code.

thchr:
I don't know to which extent this is relevant - but reading along here, I saw that there's some tools for band structure plots. I just wanted to point out that there's some [related code](https://github.com/thchr/Brillouin.jl/blob/master/src/requires/plotlyjs_dispersion.jl) for this in Brillouin (but it is using PlotlyJS.jl rather than Plots.jl). There's some examples of its use in the [documentation](https://thchr.github.io/Brillouin.jl/dev/kpaths/#Band-structure).

mfherbst:
@thchr Thanks for the pointer. I made a comment about this in #503 and #533. If we could somehow merge our routines (and generalise them to Plots rather than just Plotly), I'd be happy to scrape our code on this.

src/postprocess/band_structure.jl
jaemolihm: Sorry for the delay. I'm afraid getting rid of the kdistances_and_ticks function completely is not possible, because one still needs to compute the kdistances and tick labels.

Also, since `prepare_band_data` accepts `band_data`, it knows the k points only through `basis`, i.e. it cannot directly access `kcoords` from `high_symmetry_kpath`. But I agree that completely discarding branch information is not good, so I added `kbranches` to the output of `high_symmetry_kpath` (which was already the case for 1D).

src/postprocess/band_structure.jl
jaemolihm: Done. Now the Symbol to String conversion is done in `kdistances_and_ticks`.

test/compute_bands.jl
jaemolihm: Done

test/compute_bands.jl
jaemolihm: Now `branches` is an input of `kdistances_and_ticks`.

test/compute_bands.jl
jaemolihm: Done

test/compute_bands.jl
jaemolihm: Done

jaemolihm:
I added compat bounds for brillouin.jl: 

> I think it could be worth advertising this feature in some of the examples in the docs. I guess [the supercell example](https://docs.dftk.org/stable/examples/pymatgen/) or the [slab example](https://docs.dftk.org/stable/examples/ase/) are suitable for that purpose. What do you think @jaemolihm?

I agree that it will be good to advertise this feature. I will add one when we are settled with the inner workings.

jaemolihm:
> I made a comment about this in #503 and #533. If we could somehow merge our routines (and generalise them to Plots rather than just Plotly), I'd be happy to scrape our code on this.

Out of curiosity (I have little knowledge of plotting with Julia - I have only used PyPlot.jl): What is the benefit of using Plots? Does it enable use of any backend the user wants?

src/postprocess/band_structure.jl
jaemolihm: Now I am not sure this was a good idea. `branches` can be easily reconstructed from `basis.kcoords_global`. Keeping `branches` generated by Brillouin makes the code a little simpler, but when it comes to user-defined k paths, it is required for the user to provide `branches` by themselves.

If we just throw away `branches` by Brillouin, then for both Brillouin k path and user-defined k path, `branches` can be automatically calculated in `kdistances_and_ticks`. So, the user does not need to provide `branches`.

mfherbst:
> Out of curiosity (I have little knowledge of plotting with Julia - I have only used PyPlot.jl): What is the benefit of using Plots? Does it enable use of any backend the user wants?

Indeed. That is why we try to build our infrastructure on top of that.

src/postprocess/band_structure.jl
mfherbst: That's a good point. I see two ways out of this:

- If the user does not supply the branches, everything is on one branch (i.e. what you do now)
- We return Brillouin's `kbranches`, but don't actually use them, i.e. this argument in the returned named tuple is just for information and we re-compute the kbranches as has been done before.

I am tempted to go for the first option, because "custom kpaths" is sort of an advanced feature and then people probably know what they want wrt. the branches. Also to me `prepare_band_data` has always been an internal function (it's not exported for that reason).

But of course @jaemolihm ... if you have a use case which gets harder with this refactoring, than please do speak up ... after all you are doing the work :smile:.

antoine-levitt:
Yeah, I've never really bought that argument. In theory it's great and all, but in practice it just adds a layer of complexity on top of something that is already very messy. I've never been able to get into it, and in particular I've never managed to do something as basic as having multiple plots (figure() in pyplot)

mfherbst:
Hmm. I disagree. I think `Plots` really made it more intuitive to plot things, but ok, different people have different opinions.

Re subplots, just concatenate: `p = plot(...); q = plot(...); side_by_side = plot(p, q)`

antoine-levitt:
Yeah but it's super annoying, I can never remember how it works, and as I recall it imposes a structure that makes it very hard to use in scripts (run several analyses one after another). Anyway, I have nothing against using it in DFTK, esp because not having to have a working python environment to use DFTK is great.

mfherbst:
Partially closes #678.

PR #556, opened 2021-12-21
Turn generators into arrays

antoine-levitt:


mfherbst:
I made some cleanup. It turns out that not using generators really kills us in the `compute_densities` due to the symmetrisation stuff. We loose about 10% on my experiments with silicon.

The Hartree thing we can argue about, but with the explicit lattice multiplication as I do it it cuts down cost by factor 100 (ok still neglibile).

src/PlaneWaveBasis.jl
```diff
@@ -325,6 +325,12 @@ The wave vectors `G` in reduced (integer) coordinates for a cubic basis set
...
 function G_vectors(fft_size::Union{Tuple,AbstractVector})
+    start = .- cld.(fft_size .- 1, 2)
+    stop  = fld.(fft_size .- 1, 2)
+    axes  = [[collect(0:stop[i]); collect(start[i]:-1)] for i in 1:3]
+    [Vec3{Int}(i, j, k) for i in axes[1], j in axes[2], k in axes[3]]
+end
+function G_vectors_generator(fft_size::Union{Tuple,AbstractVector})
     start = .- cld.(fft_size .- 1, 2)
     stop  = fld.(fft_size .- 1, 2)
     axes = [[collect(0:stop[i]); collect(start[i]:-1)] for i in 1:3]
```
antoine-levitt, line 336: have one `collect` the other?

antoine-levitt: and a comment saying what this is for

antoine-levitt:
OK that makes sense. I'd leave out the hartree term (if that is a bottleneck we're dead in lots of other places anyway)

src/PlaneWaveBasis.jl
mfherbst: With the collect it's 100-fold slower than with the code duplication.

antoine-levitt: wow really? It must hit the captured variables in closure thing, or something similar. OK then just add a comment saying that?

PR #557, opened 2021-12-21
Implement Marzari-Vanderbilt-DeVita-Payne (cold) smearing

azadoks:
As reference:
https://gitlab.com/QEF/q-e/-/blob/develop/Modules/wgauss.f90
https://gitlab.com/QEF/q-e/-/blob/develop/Modules/w0gauss.f90

I've also locally implemented arbitrary-order Methfessel-Paxton smearing, but I wasn't sure if that was something you're interested in. For that, I used [this](http://www.hector.ac.uk/cse/distributedcse/reports/conquest/conquest/node6.html) PDF plus some inspection of the QE implementation in the files above.
I saw two options for implementing the order as a parameter: adding an argument to `occupation` and `entropy`, or adding a parameter to the struct. In the end, I used a parameterized struct `MethfesselPaxton{N} <: SmearingFunction` where `N` is the order of the Hermite delta expansion. I'm not sure how this would interact with `ForwardDiff`, but I have the occupation derivative implemented by hand in any case.

Let me know if this would be interesting; I can make another PR or put it in with this one.


src/Smearing.jl
```diff
@@ -130,10 +130,12 @@ function MethfesselPaxton(order::Integer)
     end
 end
 
-# TODO: Marzari-Vanderbilt "cold smearing"
+struct Cold <: SmearingFunction end
```
antoine-levitt, line 133: Hm, ColdSmearing maybe? Or MarzariVanderbilt? Cold is a bit short...

src/Smearing.jl
```diff
@@ -130,10 +130,12 @@ function MethfesselPaxton(order::Integer)
...
+occupation(S::Cold, x) = 2 - erfc(x - 1/√2) + sqrt(2 / typeof(x)(π)) * exp(-x^2 + √2x - 1/2)
```
antoine-levitt, line 134: You have to be a bit more careful here: x - 1/sqrt(2) will compute 1/sqrt(2) in Float64, for a loss of precision if you're computing in higher precision. Do typeof(2). Same for 1/2. Yeah it's annoying: there's a macro to do it automatically (ChangePrecision.jl) but here it's just simpler to do it.

Actually for this function and the others, it'd be nicer to type the argument (`occupation(S::Cold, x::T)`) and then do T(2), not sure why we didn't do it before.

antoine-levitt: Also I worry reading this whether √2x is √2 x or √(2x), can you add parentheses on this?

antoine-levitt (commented):
Thanks a lot for the PR, super nice to have this! That file dates back and could use a bit of cleanup, I'll do that after this PR.

Higher order MP would be nice to have (not that anybody would use them probably, but still). Re API adding an argument would be awkward here: better to have a struct. Having a parametrized struct here is more trouble than it's worth. Better just to have the order as a field in the MethfesselPaxton struct: simply
```
struct MethfesselPaxton
    order::Int
end
```
This way it's even compatible with the current API. It won't compile to branchless code, but this function is not performance critical at all, so simplicity > micro-optimization.

antoine-levitt:
Re ForwardDiff, it should just work. In any case it gets covered by the tests if you add it to the `smearing_methods` list, so if it doesn't work it'll tell you.

src/Smearing.jl
mfherbst: I think MarzariVanderbilt is good. That's also used by QE for input.

mfherbst: Also could you add the reference of the equations to the top of the page (see the other comments).

src/Smearing.jl
mfherbst: > Actually for this function and the others, it'd be nicer to type the argument (occupation(S::Cold, x::T)) and then do T(2), not sure why we didn't do it before.

agree.


> Also I worry reading this whether √2x is √2 x or √(2x), can you add parentheses on this?


Or just use the sqrt function :smile:

mfherbst (commented):
Thanks very much for the contribution!

Yes we would be interested in the arbitrary Mathfessel-Paxton as well in case you have it. After all this is a code to allow experimentation as well as realistic calculations, so both is appreciated.

Regarding forward-diff I'm also not too worried. It works quite ok with the MP smearing. Of course if you want to, feel free to implement the analytical derivative as well. The main reason why we use forwarddiff is because we were lazy.

mfherbst:
Also @azadoks a small suggestion for the docs, which we would really appreciate if you got around doing it:

It would be super cool now that we have quite a selection of smearings to maybe have a documentation page or example that just showcases them. Does not need to be something fancy, maybe just an aluminium cell with a small k-grid or something where it gives (slightly) different numbers to show the point.

Of course no worries if you don't manage :smile:.

src/Smearing.jl
azadoks: I chose to name it cold because that's most often what Prof. Marzari uses in our lab.
I agree though that it's more commonly called Marzari-Vanderbilt, so I'll change it to that for now.

src/Smearing.jl
azadoks: I can go through all of the functions so that where necessary the signatures are `occupation(S::SmearingFunction, x::T) where T` and `entropy(S::Smearingfunction, x::T) where T`.
This way, I can try to clean up all of the `typeof(x)(pi)` etc.

I'll also just use the `sqrt` function 😅 .

azadoks:
> Thanks very much for the contribution!

Glad to help!

> Yes we would be interested in the arbitrary Mathfessel-Paxton as well in case you have it. After all this is a code to allow experimentation as well as realistic calculations, so both is appreciated.

I'll add this in the next commit.

> Regarding forward-diff I'm also not too worried. It works quite ok with the MP smearing. Of course if you want to, feel free to implement the analytical derivative as well. The main reason why we use forwarddiff is because we were lazy.

I have implementations for the occupation derivatives of all of the smearing functions (Fermi-Dirac, Gaussian, Cold/Marzari-Vanderbilt, and arbitrary order Methfessel-Paxton). I think it may make the code more insightful for research purposes to have these so users/developers/scientists can reference them. I would still leave the default dispatch as a call to `ForwardDiff` in case anyone wants to develop some new smearing without having to write out the occupation derivative and also just to send the message that these are just the same derivatives you would get from `ForwardDiff`
Thoughts?

mfherbst:
Sounds perfect to me, thanks @azadoks!

antoine-levitt:
One thing I definitely would like to avoid is having derivatives that are not used by default, because then they would not be tested. More generally, ideally I'd like to live in a world where nobody ever writes derivatives by hand. We're not there yet for complicated multivariable functions, but for this kind of one variable functions forward diff does a wonderful job, so why not just use it? More code = more bugs.

azadoks:
I've replaced the `MethfesselPaxton1` and `MethfesselPaxton2` structs with `MethfesselPaxton`, renamed `Cold` to `MarzariVanderbilt`, added some sources to the top of the file, and worked a bit on the type stability issues that came up in the discussion.

I didn't include the by-hand derivatives because this seems still to be up for discussion.
I either directly translated all the derivatives from Quantum-Espresso's source or, in the case of MethfesselPaxton, made them into (in my opinion) _much_ clearer code.
In both cases, I tested a bit against the Q-E implementations, and aside from some precision differences and the truncation that they use (I guess for speed-up or to avoid over / underflows), they seem to be correct.

I could do some work to thoroughly test against the Q-E implementations and against `ForwardDiff` if you'd like to include the derivative functions. If not, then I'm happy to leave them out.

antoine-levitt:
Thanks! The tests are failing, maybe some typo somewhere? Re derivatives I'd be of the opinion to just not include them, unless there's a specific need for them?

Note we also have https://github.com/JuliaMolSim/DFTK.jl/pull/513#event-5159480887 still open that we should fix at some point, so if that interests you it's up for grabs!

src/Smearing.jl
```diff
@@ -94,46 +97,44 @@ end
...
-    x == Inf && return zero(x)
-    x == -Inf && return one(x)
-    occupation(Gaussian(), x) + A(1, typeof(x))*H1(x)*exp(-x^2)
+entropy(S::Gaussian, x::T) where T = 1 / (2 * sqrt(T(pi))) * exp(-x^2)
+
+struct MarzariVanderbilt <: SmearingFunction end
+occupation(S::MarzariVanderbilt, x::T) where T = 2 - erfc(x - 1/sqrt(T(2))) + sqrt(2 / T(π)) * exp(-x^2 + sqrt(T(2)) * x - 1/T(2))
+entropy(S::MarzariVanderbilt, x::T) where T = 1 / sqrt(2 * T(π)) * (x - 1/sqrt(T(2))) * exp(-x^2)
+
+A(T::Type, n) = (-1)^n / (factorial(n) * 4^n * sqrt(T(π)))
```
antoine-levitt, line 106: no need to type T here

src/Smearing.jl
```diff
@@ -94,46 +97,44 @@ end
...
+function H(x, n)
+    if iszero(n)
```
antoine-levitt, line 108: you can just have n == 0 here. Also add a comment saying this is the standard Hermite function with what wikipedia calls the "physicists convention"

src/Smearing.jl
```diff
@@ -94,46 +97,44 @@ end
...
+        return one(x)
+    elseif isone(n)
```
antoine-levitt, line 110: same, n==1

src/Smearing.jl
```diff
@@ -94,46 +97,44 @@ end
...
-    if order == 0
-        Gaussian()
-    elseif order == 1
-        MethfesselPaxton1()
-    elseif order == 2
-        MethfesselPaxton2()
-    else
-        error("Not implemented")
+function occupation(S::MethfesselPaxton, x::T) where T
+    S₀ = 1/T(2) * erfc(-x)
```
antoine-levitt, line 121: you can just do erfc(-x)/2

src/Smearing.jl
```diff
@@ -94,46 +97,44 @@ end
...
+    ΣSₙ = zero(x)
+    for i in 1:S.order
```
antoine-levitt, line 123: I usually like to use `=` for simple numerical loops, and `in` for more collection-type things, eg `for iband = 1:nband` but `for G in G_vectors(basis)`. But we don't really do this consistently in DFTK anyway (I'm not sure Michael agrees with me on this)

src/Smearing.jl
```diff
@@ -94,46 +97,44 @@ end
...
+        ΣSₙ -= A(T, i) * H(x, 2i + 1)
+    end
+    return ΣSₙ * exp(-x^2)
 end
-
-# TODO: Marzari-Vanderbilt "cold smearing"
 
 # List of available smearing functions
-smearing_methods = (None, FermiDirac, Gaussian, MethfesselPaxton1, MethfesselPaxton2)
+smearing_methods = (None, FermiDirac, Gaussian, MarzariVanderbilt, MethfesselPaxton)
```
antoine-levitt, line 137: that won't work (see test/occupation.jl which calls `method()`). The simplest is probably to change this to
```
smearing_methods = (None(), FermiDirac(), Gaussian(), MarzariVanderbilt(), MethfesselPaxton(1), MethfesselPaxton(2))
```
and make corresponding changes to test/occupation.jl. And maybe just move that list to test/occupation.jl anyway...

src/Smearing.jl
```diff
@@ -94,46 +97,44 @@ end
...
-function MethfesselPaxton(order::Integer)
-    if order == 0
-        Gaussian()
-    elseif order == 1
-        MethfesselPaxton1()
-    elseif order == 2
-        MethfesselPaxton2()
-    else
-        error("Not implemented")
+function occupation(S::MethfesselPaxton, x::T) where T
```
antoine-levitt, line 120: much nicer than what we had before, thanks!

src/Smearing.jl
azadoks: Yes, this was the first issue with the tests, I've (hopefully) fixed it in the latest commit. Thanks!

azadoks:
Thanks @antoine-levitt for all the helpful comments!

I went through the tests in `occupations.jl` and moved the `smearing_methods` tuple there.
After I got the tests working, I noticed that the Marzari-Vanderbilt and Methfessel-Paxton occupations and smearing were mirrored around the y-axis compared to Gaussian and Fermi-Dirac (I guess a convention difference w.r.t. QuantumEspresso).

I also squashed some bugs that the testing caught, fingers crossed that they pass now!

src/Smearing.jl
```diff
@@ -93,47 +96,50 @@ function occupation_divided_difference(S::FermiDirac, x, y, εF, temp)
...
-        Gaussian()
-    elseif order == 1
-        MethfesselPaxton1()
-    elseif order == 2
-        MethfesselPaxton2()
-    else
-        error("Not implemented")
+function entropy(S::MethfesselPaxton, x::T) where T
+    exp_term = exp(-x^2)
+    S₀ = A(T, 0) / 2 * exp_term
```
antoine-levitt, line 135: just write it out explicitly. Hopefully the compiler is suffiicently smart (or will become sufficiently smart) to recognize `exp` is pure, but even if not this is not performance critical

test/occupation.jl
```diff
@@ -4,18 +4,29 @@ using SpecialFunctions
 
 include("testcases.jl")
 
+smearing_methods = (
+        DFTK.Smearing.None(),
+        DFTK.Smearing.FermiDirac(),
+        DFTK.Smearing.Gaussian(),
+        DFTK.Smearing.MarzariVanderbilt(),
+        DFTK.Smearing.MethfesselPaxton(1),
```
antoine-levitt, line 12: if you want to be super fancy you can do `(DFTK.Smearing.MethfesselPaxton(n) for n=1:4)...` but this is fine too

antoine-levitt (approved):
Great! Good to go on my end

azadoks:
Thanks again for the review!

I cleaned up the MP code like you suggested and added some comments.
MP with `order > 16` fails a bunch of the tests, so I added an inner constructor restricting the order to 14, which is still way overkill, but I thought it would be good to save the knowledge.

I also added some notes that for MV/Cold and MP smearing, the Fermi level is not unique.
A colleague of mine, Flaviano Jose dos Santos, has implemented a mixed zero-finding algorithm in QE for ensuring that the most reasonable Fermi level is chosen in these cases.
He found with another colleague that choosing the wrong Fermi level significantly impacted the structural relaxation of _roughly_ 30% of our high-throughput database, so this could be another good issue to look into now that MV smearing is implemented.

antoine-levitt:
> MP with order > 16 fails a bunch of the tests, so I added an inner constructor restricting the order to 14, which is still way overkill, but I thought it would be good to save the knowledge.

That's kind of arbitrary: it depends on the particular values used for the test, the precision used, etc. I'd rather either have a principled test based on well-understood principles or have no check at all. The simple implementation of hermite polynomials by the recurrence relations is possibly unstable for high order. I'd say let's not bother: the user is free to choose crazy values if they want, but of course there's no guarantee on what happens then. 

> He found with another colleague that choosing the wrong Fermi level significantly impacted the structural relaxation of roughly 30% of our high-throughput database, so this could be another good issue to look into now that MV smearing is implemented.

Oof that's scary... In theory for convergence it doesn't really matter which level you choose for convergence (see last sentence p11 of https://arxiv.org/pdf/1805.07144.pdf), although of course in practice it does.

azadoks:
Ok, I've removed the check.

On the Fermi level issue, you can see the [relevant implementation by Flaviano here](https://gitlab.com/QEF/q-e/-/blob/develop/PW/src/efermig.f90).

It looks like the general workflow is to:
1. If Gaussian/FD, use bisection and return
2. If MP/MV, use bisection with Gaussian smearing to get an initial guess
3. If the initial guess gives the right number of electrons, return
4. If not, use Newton's method with the initial guess to refine the answer
5. If Newton's method gives the right number of electrons, return
6. If not, use bisection with MP/MV smearing and return (with fingers crossed that the fermi level is correct; the error message says that this is could be caused by smearing over a gap.)

Taking a glance at `occupation.jl`, it looks like you're already checking the number of electrons condition, so at least you get an error if the wrong zero is found (much better than what QE was doing).
It could still be nice to try to implement the Newton's method second step or similar to try to recover from more conditions that would error.

antoine-levitt:
Makes sense to initialize with a "safer" method, although that can't be completely foolproof. How do you define a good vs a spurious Fermi level? Do you somehow check that some of the occupations are in the "lobes" where the occupation is negative / larger than 1? Or is a bad Fermi level one where the derivative of the number of electrons vs energy (ie the DOS) is somehow too large?

azadoks:
These are all great questions, but unfortunately I don't know. I'll have to wait until after the holidays to ask how Flaviano thought about this.

The best we can do at the moment is to look at the QE source code where Flaviano has implemented his method.
You're right that it seems that the "safer" initialization is not foolproof, but the only implemented fallback is to return to QE's old behavior of running bisection and hoping that it finds a reasonable zero.

From a brief discussion I had with Flaviano a few weeks ago, it does seem that his conceptual definition of a spurious Fermi level was one that is in one of the "lobes" where the occupation is negative (MP) or greater than 1 (MP/MV), though it doesn't look like the method does anything to check this (i.e. it only checks that the number of electrons is correct).

I guess what I can suggest for the moment is to open a separate issue for this; I can work on a draft PR with the secondary Newton's method for MP/MV, and we can discuss if there is any further improvement to be made.

src/Smearing.jl
```diff
@@ -93,47 +96,57 @@ function occupation_divided_difference(S::FermiDirac, x, y, εF, temp)
...
-    if order == 0
-        Gaussian()
-    elseif order == 1
-        MethfesselPaxton1()
-    elseif order == 2
-        MethfesselPaxton2()
-    else
-        error("Not implemented")
+function entropy(S::MethfesselPaxton, x::T) where T
+    # Equivalent to A(T, i) * (H(x, 2i) / 2 + 2i * H(x, 2i - 2)), but H(x, i<0) is not defined and gives an error.
```
antoine-levitt, line 140: just define it equal to zero? Also means you don't have to have the n==1 branch

antoine-levitt:
Sure, let's just merge this when @mfherbst agrees. (but see my last comment for a slight simplification). Again in theory (when the temperature goes to zero) everything is fine if the code picks the wrong Fermi level. I'm generally not super concerned about being very smart in picking the Fermi level and I'd rather keep the code simple and naive (so that at least when it fails it fails in simple and predictable ways).

antoine-levitt:
Btw, the reason why I'm not super motivated about improving this is that I think MP methods are a bit of a hack anyway and should be replaced by better integration methods (like tetrahedra methods). There's a plan for playing around with that in DFTK, so we'll see what comes out of it.

src/Smearing.jl
mfherbst: We are not consistent, but I agree with using `=` for plain numbers.

src/Smearing.jl
```diff
@@ -93,47 +96,57 @@ function occupation_divided_difference(S::FermiDirac, x, y, εF, temp)
...
+end
+function occupation(S::MethfesselPaxton, x::T) where T
     x == Inf && return zero(x)
     x == -Inf && return one(x)
-    occupation(MethfesselPaxton1(), x) + A(2, typeof(x))*H3(x)*exp(-x^2)
+    f₀ = erfc(x) / 2  # 0-order Methfessel-Paxton smearing is Gaussian smearing
+    Σfₙ = zero(x)
+    for i in 1:S.order
+        Σfₙ += A(T, i) * H(x, 2i - 1)
+    end
```
mfherbst, line 136: Actually I would write this as
```suggestion
    Σfₙ = sum(i -> A(T, i) * H(x, 2i - 1), 1:S.order)
```

src/Smearing.jl
mfherbst: Agree. I think it's pretty common to do that.

src/Smearing.jl
```diff
@@ -93,47 +96,57 @@ function occupation_divided_difference(S::FermiDirac, x, y, εF, temp)
...
-        MethfesselPaxton2()
-    else
-        error("Not implemented")
+function entropy(S::MethfesselPaxton, x::T) where T
+    # Equivalent to A(T, i) * (H(x, 2i) / 2 + 2i * H(x, 2i - 2)), but H(x, i<0) is not defined and gives an error.
+    # Also equivalent to Gaussian entropy (A(0) == 1/√π)
+    s₀ = A(T, 0) / 2 * exp(-x^2)
+    Σsₙ = zero(x)
+    for i in 1:S.order
+        Σsₙ += A(T, i) * (H(x, 2i) / 2 + 2i * H(x, 2i - 2))
```
mfherbst, line 145: Can use the same trick as before to do the sums.

src/Smearing.jl
```diff
@@ -3,12 +3,18 @@
 # Ref for the equations:
 #    - M. Methfessel, A. T. Paxton 1989
 #       "High-precision sampling for Brillouin-zone integration in metals"
+#     - http://www.hector.ac.uk/cse/distributedcse/reports/conquest/conquest/node6.html
+#       "Methfessel-Paxton Approximation to Step Function"
+#     - https://gitlab.com/QEF/q-e/-/blob/bcbb9787c58f17a467653707917172ffc4053dd9/Modules/wgauss.f90
+#       Smearing function implementation in Quantum-Espresso
+#     - https://gitlab.com/QEF/q-e/-/blob/bcbb9787c58f17a467653707917172ffc4053dd9/Modules/w0gauss.f90
+#       Occupation derivative and entropy implementations in Quantum-Espresso
```
mfherbst, line 11: We tend to also put original references, if we have them handy. Could you add the Cold Smearing paper?

test/occupation.jl
```diff
@@ -4,18 +4,26 @@ using SpecialFunctions
 
 include("testcases.jl")
 
+smearing_methods = (
+        DFTK.Smearing.None(),
+        DFTK.Smearing.FermiDirac(),
+        DFTK.Smearing.Gaussian(),
+        DFTK.Smearing.MarzariVanderbilt(),
+        (DFTK.Smearing.MethfesselPaxton(i) for i in 1:4)...
```
mfherbst, line 12: Might be a bit too fancy ...

mfherbst: I just saw Antoine asked you to do this, nevermind. It does not matter so much.

mfherbst (commented):
Thanks for the implementation and the interesting discussion and sorry I was a bit quiet due to holidays.

mfherbst:
> Once we settle in to a final version of the code, how should I proceed with adding some documentation?

I would add a nice example explaining and contrasting different smearing methods for maybe an aluminium or iron test case? Just add a new file to the `example` folder. Since the file is interpreted by Documenter, you have to be a little careful with the formatting (basically all non-commented stuff is interpreted as code blocks, which are executed). Best follow along a different example, e.g. maybe https://github.com/JuliaMolSim/DFTK.jl/blob/master/examples/custom_solvers.jl. To get your example included in the docs, add it to https://github.com/JuliaMolSim/DFTK.jl/blob/master/docs/src/index.md. You should then be able to build it locally by running the `make.jl` script in the `docs` subfolder.

antoine-levitt:
Are you sure this warrants a full new example? We already have too many of those. Maybe just add to https://docs.dftk.org/dev/examples/metallic_systems/?

mfherbst:
My main point for this is to present users an overview of what smearing / BZ integration methods we have available. Doesn't need to be a new example for that purpose, I guess.

> We already have too many of those.

I don't think there can be too many examples as long as they are not bloaty and redundant :smile:.

antoine-levitt:
> I don't think there can be too many examples as long as they are not bloaty and redundant smile.

Examples can serve two purposes: either teach users how to do in DFTK something they already know, or to teach them what is the thing in question. For the first one, it's a bit trivial here: replace FermiDirac() by whatever(). For the second one, it's much trickier - something like http://theossrv1.epfl.ch/Main/ElectronicTemperature would be relevant, but it's hard to do in an example - especially given our constraints that it has to run on CI. Maybe we could relax that constraint and pre-generate figures, and so do something like in the above link?

mfherbst:
> For the first one, it's a bit trivial here: replace FermiDirac() by whatever()

Indeed. That one could just say by adding a sentence to another metal example.

> Maybe we could relax that constraint and pre-generate figures, and so do something like in the above link?

Yes, I've been thinking about that as well a few times - along with the question to add a few more expensive tests for tracking timings, for example. I only fear if we don't run example code at least from time to time it will stop working. Ideally one would run the example code using crude setups, but show the plots from more realistic numerical parameters. I suppose that is doable, but probably out of scope for this PR.

azadoks:
Hi guys, happy new year!
I just returned from holiday, so I've gone over your suggestions and fiddled with the code.

It seems like the documentation issue might be a bit out of scope for this PR, but I do like @antoine-levitt 's idea w.r.t. the plots of force/energy vs. smearing+temperature/k-points. It would be nice to have a consistent set of plots and some test cases both for learning and development purposes.

mfherbst:
Hi @azadoks and happy new year to you as well!

> It seems like the documentation issue might be a bit out of scope for this PR
> It would be nice to have a consistent set of plots and some test cases both for learning and development purposes.

Agree: https://github.com/JuliaMolSim/DFTK.jl/issues/567

> Indeed. That one could just say by adding a sentence to another metal example. (https://github.com/JuliaMolSim/DFTK.jl/pull/557#issuecomment-1008272052)

Would you be ok to quickly do that

mfherbst:
Thanks!

mfherbst:
Thanks @azadoks !

PR #558, opened 2021-12-28
Integrate with AtomsBase

mfherbst:
Integration PR with [https://github.com/JuliaMolSim/AtomsBase.jl](AtomsBase.jl). Perspective is shown in [examples/atomsbase.jl](https://github.com/JuliaMolSim/DFTK.jl/blob/atomsbase/examples/atomsbase.jl). For now mainly a showcase to see if two packages (AtomIO, DFTK) can communicate sensibly via the AtomsBase interface.

I removed the nested list in the atoms datastructure to keep it easier in sync with the system (simplifies interfaces for force calculation for  example).

**Update:** I rebased heavily to be in sync with the changes we made recently.

Todo:
- [x] Rewrite internal atoms datastructure and get rid of all compatibility stuff 
- [x] Get rid of obsessive type annotations
- [x] Get PR in AtomsBase merged
- [x] Adjust compats
- [x] Write tests wrt. AtomsBase features.
- [x] Get tests to pass
- [x] Documentation


mfherbst:
@antoine-levitt This is ready, if you want to have a look.

docs/src/guide/input_output.jl
```diff
@@ -40,7 +40,13 @@ magnetic_moments = load_magnetic_moments(qe_input)
...
 # See [Creating slabs with ASE](@ref) for more details.
 
-atoms = [ElementPsp(el.symbol, psp=load_psp(el.symbol, functional="pbe")) for el in atoms]
+atoms = map(atoms) do el
+    if el.symbol == :Fe
+        ElementPsp(:Fe, psp=load_psp("hgh/pbe/fe-q16.hgh"))
+    else
+        error("Unsupported element: $el")
+    end
+end;
```
antoine-levitt, line 49: oof that's a lot. Just have the thing before and add a `@assert all(el -> el.symbol== :Fe, atoms)`?

src/standard_models.jl
```diff
@@ -50,6 +57,7 @@ DOI:10.1103/PhysRevLett.77.3865
 function model_PBE(lattice::AbstractMatrix, atoms::Vector, positions::Vector; kwargs...)
     model_DFT(lattice, atoms, positions, [:gga_x_pbe, :gga_c_pbe]; kwargs...)
 end
+model_PBE(system::AbstractSystem; kwargs...) = call_with_system(model_PBE, system; kwargs...)
```
antoine-levitt, line 60: You can do all of these with an eval loop: `for fun in (:a, :b) @ eval $fun() ...`

src/Model.jl
```diff
@@ -98,6 +103,7 @@ function Model(lattice::AbstractMatrix{T}, atoms=Element[], positions=Vec3{T}[];
                spin_polarization=default_spin_polarization(magnetic_moments),
                symmetries=default_symmetries(lattice, atoms, positions, magnetic_moments,
                                              spin_polarization, terms),
+               system=construct_system(lattice, atoms, positions, magnetic_moments),
```
antoine-levitt, line 106: Hm, this allows passing both a lattice/atoms/positions and a system without enforcing consistency. Do you really need it as an input arg?

src/Model.jl
```diff
@@ -168,9 +175,19 @@ function Model(lattice::AbstractMatrix{<:Quantity}, args...; kwargs...)
...
-normalize_magnetic_moment(::Nothing)  = Vec3{Float64}(zeros(3))
-normalize_magnetic_moment(mm::Number) = Vec3{Float64}(0, 0, mm)
-normalize_magnetic_moment(mm::AbstractVector) = Vec3{Float64}(mm)
+"""
+    Model(system::AbstractSystem; kwargs...)
+
+AtomsBase-compatible Model constructor. Sets structural information (`atoms`, `positions`,
+`lattice`, `n_electrons` etc.) from the passed `system`.
+"""
+Model(system::AbstractSystem; kwargs...) = call_with_system(Model, system; kwargs...)
```
antoine-levitt, line 184: How about putting this and the other from standard_model into atomsbase? Generally it'd be nice to separate them to the max

antoine-levitt (commented):
Still not 100% sure the added functionality is worth the integration effort. I would prefer a looser integration, in particular not having Model depend on System.

src/atomsbase.jl
```diff
@@ -0,0 +1,84 @@
+# Key functionality to integrate DFTK and AtomsBase
```
antoine-levitt, line 1: doesn't really need to be in the src/ main dir, move it to external?

mfherbst:
Thanks for the review. To comment on the main point

> I would prefer a looser integration, in particular not having Model depend on System.

Hmm. From my point of view a crucial aspect is to have an integration both ways, i.e. that you can't just use AtomsBase to setup a calculation in DFTK, but that anyone using DFTK can also get an AtomsBase-compatible representation of the exact problem that DFTK has been dealing with.

Along these lines what would be nice is to have have the exposed AtomsBase-compatible representation of DFTK's state be exactly the identical object that was passed to DFTK in case the `model_DFT(system)`-like constructors have been used. But maybe that is a bit overengineered (certainly leads to tight coupling). Because if one gives up on that, one could get rid of the `system` kwarg in the model for example and the associated potential contradicting information. I'm ok with that if you prefer.

Yet another option is to not store the system in the Model at all, but to only offer a public function to make an `AbstractSystem` from a `Model`. That sounds tempting at first, but it involves potentially discarding information (e.g. the initial magnetic moments), which could be confusing people, so I did not do that.

mfherbst:
> The magnetic moment thing is unsatsfying anyway and really does not belong to the model (maybe we should move it to the basis, since we discard symmetries there

Yes but it is frequently stored in input files and atomic structure files, that's the original rationale for having it in the AtomsBase world. Ok, let's go with the slim integration first without storing the system. Adding it later can always be done if needed.

Issue #560, opened 2022-01-11
Make perturbation-based error bounds available in master


mfherbst:
As briefly discussed offline with @gkemlin, we should get the implementation discussed in https://arxiv.org/abs/2111.01470 up to speed and readily available for use in DFTK. Of course let's not worry too much about the details such as performance initially.

@gkemlin Where is the code currently located? How close is it to being mergeable? If I recall correcty most aspects (grid transfer, newton solver etc.) are anyway already available, right?

antoine-levitt:
We discussed with Gaspard, a good way to do this at least now is as an example

antoine-levitt:
I'm kind of reluctant to have it as a black-box thing without more extensive testing. But if we have plenty of caveats then we can possibly do that yeah

gkemlin:
At the moment, the files are located on my computer. I had a branch called "estimators" that I cleaned and I think can be removed cause I don't really remember what I modified on it. And then can open a new branch to merge it with master. 

I think an example file following what we do in the paper can be easily generated at the moment (which would be great also to include a link in the paper after its review before publishing it).

To include it as a ready-to-use function, we should think about what we want : use it as an error estimator ? a way to compute more precise forces ?

gkemlin:
(An example file for silicon would be no more than few dozens of lines)

mfherbst:
> I'm kind of reluctant to have it as a black-box thing without more extensive testing.

Ok, that makes sense.

> I think can be removed  

done

>  use it as an error estimator ? a way to compute more precise forces ?

Well ideally both. One sort of includes the other, no?

But ok, I'm convinced. Let's have it as an example with a link to the paper first. @gkemlin When you make the PR can you also take a quick pass at https://docs.dftk.org/stable/publications/ to see if it's up to date.

mfherbst:
Silicon is a bit boring. If it's not too expensive, can you go for the TiO2?

gkemlin:
Yes of course ! Do we do that for one couple (Ecut, Ecut_ref) or do we seek to plot a figure like Figure 9 in https://arxiv.org/pdf/2111.01470.pdf ? Generating the whole figure takes time, I would rather aim to a well chosen couple to see the effectiveness of the bound.

I'll try to propose you something in the coming week then :-)

> Well ideally both. One sort of includes the other, no?
Yes, true, the question is more about the output we want but it's not a big issue.

[...]

PR #561, opened 2022-01-12
new example using AD to compute error bounds for forces

gkemlin:
Here is a first shot about the example mentioned [here](https://github.com/JuliaMolSim/DFTK.jl/issues/560). Some remarks go with this PR:
- I had to modify `local.jl` and `nonlocal.jl` in the parts where the forces are computed to avoid type errors that were being raised when using AD to compute the differential of the forces.
- The core of the example is the inversion of the Schur system and the computation of dF, but we still need to define all the operators needed to strictly follow our [paper](https://hal.inria.fr/hal-03408321), which results in a quite long example.
- Docs should compile :)


src/terms/local.jl
```diff
@@ -90,8 +90,9 @@ function (::AtomicLocal)(basis::PlaneWaveBasis{T}) where {T}
 end
 
 @timing "forces: local" function compute_forces(::TermAtomicLocal,
-                                                basis::PlaneWaveBasis{T},
-                                                ψ, occ; ρ, kwargs...) where {T}
+                                                basis::PlaneWaveBasis,
+                                                ψ, occ; ρ, kwargs...)
+    T = real(eltype(ψ[1]))
     atoms = basis.model.atoms
```
antoine-levitt, line 96: we've already had this issue I think, and I'm not sure where we landed. Should we promote T from the eltype of basis and that of T? CC @niklasschmitz

gkemlin: Yup, I remember we talked together about that in Roscoff, but then I kept the modification in my branch :sweat_smile:

examples/error_estimates_forces.jl
```diff
@@ -0,0 +1,184 @@
...
+# We then try to improve ``F(P)`` using the first order linearization:
+# ```math
+# F(P) = F(P_*) + {\rm d}F(P)\cdot(P-P_*).
+# ```
+# To this end, we use the `ForwardDiff.jl` package to compute ``{\rm d}F(P)``
+# using automatic differentiation.
+function df(basis, occupation, ψ, δψ)
+    δρ = DFTK.compute_δρ(basis, ψ, δψ, occupation)
+    function f(ε)
+        compute_forces(basis, ψ .+ ε.*δψ, occupation; ρ=ε.*δρ)
```
antoine-levitt, line 167: I'm confused by the rho here. Can't you just do rho = compute_rho(basis, psi+eps dpsi)?

antoine-levitt (commented):
Haven't checked carefully but it looks OK!

examples/error_estimates_forces.jl
```diff
@@ -0,0 +1,184 @@
...
+ψr = DFTK.transfer_blochwave(scfres.ψ, basis, basis_ref)
+ρr = compute_density(basis_ref, ψr, scfres.occupation)
+_, ham = energy_hamiltonian(basis_ref, ψr, scfres.occupation; ρ=ρr);
+# We then compute several quantities that we need to evaluate the error bounds.
+# - Compute the residual ``R(P)``, and remove the virtual orbitals, as required in [`src/scf/newton.jl`](https://github.com/JuliaMolSim/DFTK.jl/blob/fedc720dab2d194b30d468501acd0f04bd4dd3d6/src/scf/newton.jl#L121).
+res = DFTK.compute_projected_gradient(basis_ref, ψr, scfres.occupation)
+res = DFTK.select_occupied_orbitals(basis_ref, res)
+ψr = DFTK.select_occupied_orbitals(basis_ref, ψr);
+# - Compute the error ``P-P_*`` on the orbitals ``ψ`` and ``ψ_*`` by aligning the eigenvectors: this is done by solving ``\min |ψ - ψ_*U|`` for ``U`` unitary matrix of size ``N\times N`` (``N`` being the number of electrons) whose solution is ``U = S(S^*S)^{-1/2}`` where ``S`` is the overlap matrix ``ψ_*^*ψ``.
+function compute_error(basis, ϕ, ψ)
```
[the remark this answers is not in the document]
gkemlin: nope, because this function computes the error phi - psi, after aligning them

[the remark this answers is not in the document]
antoine-levitt: Ah right, sorry. Then the comment needs to use the same notation

examples/error_estimates_forces.jl
gkemlin: Hum, I'm not convinced. With that there is no need to recompute rho every time f is being called, why do you find it confusing ?

antoine-levitt: OK on reflection I agree it's always true for any reasonable case. Still, computing drho by compute_drho is basically doing the job of the AD, no reason to do it

gkemlin: So, when I do that with
```
function df(basis, occupation, ψ, δψ)                                                                            
    function f(ε)                                                               
        compute_forces(basis, ψ .+ ε.*δψ, occupation;  ρ=compute_density(basis, ψ .+ ε.*δψ, occupation))        
    end                                                                         
    ForwardDiff.derivative(f, 0)                                                
end; 
```
then I have again this type error for ForwardDiff mentioned above.

antoine-levitt: Hm interesting. Because T now has to be eltype(rho)? So we have to promote T from rho, psi and basis?!

gkemlin: Well, I don't know... In https://github.com/JuliaMolSim/DFTK.jl/blob/master/examples/forwarddiff.jl, it's `compute_δρ` which is being used.

antoine-levitt: I see. I thought we had a ForwardDiff frule for compute_rho anyway, but apparently we don't. In this case as a workaround you can just do what you do now but with  ρ=ρ+ε.*δρ

examples/error_estimates_forces.jl
```diff
@@ -0,0 +1,184 @@
...
+kgrid = [1, 1, 1]
+Ecut_ref = 50
+basis_ref = PlaneWaveBasis(model; Ecut=Ecut_ref, kgrid=kgrid)
+tol = 1e-6;
+
+# ## Computations
+# We compute the reference solution ``P_*`` from which we will compute the
+# references forces.
+scfres_ref = self_consistent_field(basis_ref, tol=tol, callback=info->nothing)
+ψ_ref = DFTK.select_occupied_orbitals(basis_ref, scfres_ref.ψ)
```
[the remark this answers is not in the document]
gkemlin: returning the occupations without the virtual states would suit the best in my opinion (as it is computed just below that)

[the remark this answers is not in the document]
gkemlin: then we have to modify a few files to adapt the output but that's not a big deal

src/terms/local.jl
mfherbst: `promote_type` sounds fair to me. The only potential issue I see is that it can introduce very subtle type instabilities if we are not careful (e.g. `promote_type(ForwardDiff.Dual{Float64}, Float64) == Real` only `promote_type(ForwardDiff.Dual{Float64, 1, 1}, Float64) <: Dual`).

examples/error_estimates_forces.jl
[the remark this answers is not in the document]
mfherbst: can you also allow a tolerance parameter to specify when an orbital is considered "empty" (than this function can also be useful if we have temperature).

examples/error_estimates_forces.jl
mfherbst: > I see. I thought we had a ForwardDiff frule for compute_rho anyway, but apparently we don't. 
not yet

> as a workaround you can just do what you do now

can you also add a comment for that. We should really have the frule.

src/terms/local.jl
antoine-levitt: ok but with forwarddiff all types should be the latter, not the former, no?

mfherbst: sure. I just wanted to point out that in case the input is garbage the `promote_type` approach will make the function still work and in this way effectively hide the problem. That could make debugging harder in the future.

examples/error_estimates_forces.jl
antoine-levitt: or just make compute_rho compatible with forwarddiff anyway?

mfherbst: Don't we have the symmetry issue to work around in that case. I don't remember or was this reverse diff?

examples/error_estimates_forces.jl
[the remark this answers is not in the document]
gkemlin: yup I can do that

examples/error_estimates_forces.jl
antoine-levitt: reverse diff is definitely tricky because we mutate a lot, but forward diff should be OK

src/terms/local.jl
gkemlin: So I should use `promote_type(eltype(basis), real(eltype(ψ[1])))` ?

mfherbst: Exactly.

examples/error_estimates_forces.jl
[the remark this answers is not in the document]
gkemlin: Just a question about this tolerance stuff for temperature: will it be an issue if the chosen tolerance is such that the returned occupation array has for instance 3 orbitals considered as occupied on one k-point and 4 on another ?

I'm thinking of the example in the doc https://docs.dftk.org/stable/examples/metallic_systems/ where the occupation array is
```
 [1.9999999999077942, 1.9999975862813721, 0.004016889271919304, 3.0004096949872443e-15, 1.1115968307815484e-18, 1.1115046659398737e-18, 7.978744566129872e-19, 7.978122691422918e-19, 3.295359361101644e-22]
 [1.9999997482868157, 1.9949327490941309, 0.0017168724035947042, 1.892360555738378e-6, 8.115209558888162e-8, 9.359437682919286e-13, 6.491292093256281e-14, 1.2325425611911432e-16, 1.0641105741057316e-16]
 [1.3387789742542426, 1.3344369226783412, 1.3344251909990323, 0.0002626279766479971, 6.750936578704564e-5, 6.750709969200984e-5, 8.293218866624051e-13, 1.505956414918229e-13, 1.5053101610349944e-13]
```
and thus taking the tolerance to differentiate between virtual and occupied equal to 1e-4 would select 3 orbitals in the two first k-points and 4 in the third k-point.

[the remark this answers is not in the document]
antoine-levitt: yeah this doesn't really make much sense in the metallic case (or at least it's pretty tricky). Or are you concerned about unconverged bands @mfherbst ?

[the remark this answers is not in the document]
gkemlin: Yep, we had included this test in `select_occupied_orbitals` 
https://github.com/JuliaMolSim/DFTK.jl/blob/fedc720dab2d194b30d468501acd0f04bd4dd3d6/src/orbitals.jl#L5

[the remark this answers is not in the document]
mfherbst: Not really. I only want a way to distinguish occupied from unoccupied bands also for metals, that's all. So indeed I would also vote for  just returning the max (so 4 bands everywhere).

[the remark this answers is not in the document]
gkemlin: Yes, that's a solution. It won't change anything for insulators, maybe we rise a warning with metals ?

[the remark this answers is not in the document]
antoine-levitt: but it doesn't really make sense to talk about unoccupied bands for metals. Ideally we should just change self_consistent_field to accept as input a max occupancy and adjust the number of bands dynamically, which would make this irrelevant. We still do have the ambiguity (both in the insulator and metallic case) that it is sometimes useful (and sometimes not) to have partially converged bands. Maybe we should have psi_extra, occupation_extra fields in scfres, so that psi and occupation refer only to converged, occupied (according to the user-given criterion) bands and the extra bands are partially converged bands included to make it easier to converge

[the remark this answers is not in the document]
mfherbst: ok than don't call it unoccupied bands. The point is just to strip off those bands with small occupation, that can be ignored.

I agree that the best solution would be to have a dynamic adjustment of the bands during the SCF (https://github.com/JuliaMolSim/DFTK.jl/issues/562). Actually the "extra" thing you mention sort of already exists, as `n_ep_extra` is always returned, so you can just drop the last few off, both for metals and insulators.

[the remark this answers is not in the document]
gkemlin: From what I understand, the idea is just to have this tool to remove bands we don't want to use (it's not even an exported function), so for the moment maybe we can just use something like that
```
function select_occupied_orbitals(ψ, occupation; threshold=0.0)
    N = maximum(length.([filter(x->x>threshold, occk) for occk in occupation])) 
   [ψk[:, 1:N] for ψk in ψ], [occk[1:N] for occk in occupation]
end
```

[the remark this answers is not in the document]
mfherbst: Better use `findfirst` instead of `filter` and `length`. You can assume occupations are partially ordered.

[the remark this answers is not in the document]
mfherbst: Or if you want to be super fancy, then `findlast`, because very likely there will be fewer orbitals stripped off than kept.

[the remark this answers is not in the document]
gkemlin: @antoine-levitt would this suits you, at least before having dynamic adjustment in the SCF ?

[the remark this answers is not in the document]
antoine-levitt: I don't get the problem with having different number of bands for different kpoints?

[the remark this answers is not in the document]
mfherbst: I think that could break stuff in DFTK as we assume the number to be identical at places, that's my only concern.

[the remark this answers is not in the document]
gkemlin: Yes, that was my concern too.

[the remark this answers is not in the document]
antoine-levitt: Do we? We shouldn't: in the metallic case we want to be able to have a different number of bands per kpoint. Psi already has a varying number of rows, so having a varying number of columns is OK. `unpack_psi` looks like it does assume but in fact doesn't: we should remove `n_bands` there (@gkemlin can you do it while you're at it?)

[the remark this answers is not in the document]
mfherbst: I'm all for removing that constraint. That will also help with https://github.com/JuliaMolSim/DFTK.jl/issues/562.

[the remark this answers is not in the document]
antoine-levitt: I'm not sure we even have that constraint currently, but yes a more systematic review would be nice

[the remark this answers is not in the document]
gkemlin: Ok, so I changed the function to
```
function select_occupied_orbitals(basis, ψ, occupation; threshold=0.0)
    N = [findlast(x -> x > threshold, occk) for occk in occupation]
    selected_ψ = [ψk[:, 1:N[ik]] for (ik,ψk) in enumerate(ψ)]
    selected_occ = [occk[1:N[ik]] for (ik,occk) in enumerate(occupation)]

    # if we have an insulator, sanity check that the orbitals we kept are the
    # occupied ones
    if threshold == 0.0
        model = basis.model
        filled_occ = filled_occupation(model)
        n_spin = model.n_spin_components
        n_bands = div(model.n_electrons, n_spin * filled_occ, RoundUp)
        @assert n_bands == size(selected_ψ[1], 2)
    end

    selected_ψ, selected_occ
end
```
and changed the output in the place where it was used before.

src/orbitals.jl
```diff
@@ -1,10 +1,23 @@
...
-    n_spin = model.n_spin_components
-    @assert iszero(basis.model.temperature)
-    n_bands = div(model.n_electrons, n_spin * filled_occupation(model), RoundUp)
-    [ψk[:, 1:n_bands] for ψk in ψ]
+# Returns the occupied orbitals and the occupation array without virtual states
+# (or states with small occupation level for metals).
+# threshold is a parameter to distinguish between states we want to keep and the
+# others when using temperature. It is set to 0.0 by default, to treat with insulators.
+function select_occupied_orbitals(basis, ψ, occupation; threshold=0.0)
+    N = [findlast(x -> x > threshold, occk) for occk in occupation]
```
mfherbst, line 6: A bit nit-picking, but maybe `>=` instead.

src/terms/local.jl
```diff
@@ -90,8 +90,9 @@ function (::AtomicLocal)(basis::PlaneWaveBasis{T}) where {T}
 end
 
 @timing "forces: local" function compute_forces(::TermAtomicLocal,
-                                                basis::PlaneWaveBasis{T},
-                                                ψ, occ; ρ, kwargs...) where {T}
+                                                basis::PlaneWaveBasis,
+                                                ψ, occ; ρ, kwargs...)
+    T = promote_type(eltype(basis), real(eltype(ψ[1])))
```
mfherbst, line 95: I wonder if we should keep the `where {T}` and then just use the paradigm
```
T = promote_type(T, real(eltype(ψ[1])))
```

examples/error_estimates_forces.jl
```diff
@@ -0,0 +1,182 @@
+# # Practical error bounds for the forces
+#
+# This is a simple example showing how to compute error estimates for the forces on a
+# ``{\rm TiO}_2`` molecule, from which we can either compute asymptotically valid error
+# bounds or increase the precision on the computation of the forces.
+#
+# The strategy we follow is described with more details in[^CDKL2021] and we
```
mfherbst, line 7: Space after "in" ?

examples/error_estimates_forces.jl
```diff
@@ -0,0 +1,182 @@
...
+kgrid = [1, 1, 1]
+Ecut_ref = 35
+basis_ref = PlaneWaveBasis(model; Ecut=Ecut_ref, kgrid)
+tol = 1e-8;
+
+# ## Computations
+# We compute the reference solution ``P_*`` from which we will compute the
+# references forces.
+scfres_ref = self_consistent_field(basis_ref, tol=tol, callback=info->nothing)
+ψ_ref, occupation = DFTK.select_occupied_orbitals(basis_ref, scfres_ref.ψ, scfres_ref.occupation)
```
mfherbst, line 53: Line a bit long.

examples/error_estimates_forces.jl
```diff
@@ -0,0 +1,182 @@
...
+# We compute a variational approximation of the reference solution with
+# smaller `Ecut`. Be careful to choose `Ecut` not too close to `Ecut_ref`.
+# /!\ The current choice `Ecut_ref=35` is such that the reference solution is
```
mfherbst, line 56: Maybe use one of the fancy boxes like in https://github.com/JuliaMolSim/DFTK.jl/blob/master/docs/src/guide/tutorial.jl#L20

examples/error_estimates_forces.jl
mfherbst, line 53: occupation_ref for consistency

examples/error_estimates_forces.jl
```diff
@@ -0,0 +1,182 @@
...
+# not converged and `Ecut=15` is such that the asymptotic regime
+# (crucial to validate the approach) is barely established.
+Ecut = 15
+basis = PlaneWaveBasis(model; Ecut=Ecut, kgrid)
+scfres = self_consistent_field(basis, tol=tol, callback=info->nothing)
+ψr = DFTK.transfer_blochwave(scfres.ψ, basis, basis_ref)
```
mfherbst, line 62: Maybe add that the ψr, ρr and Er quantities are basically the quantities computed by the crude SCF extended to the reference basis.

examples/error_estimates_forces.jl
```diff
@@ -0,0 +1,182 @@
+# # Practical error bounds for the forces
```
mfherbst, line 1: Generally a few more empty lines between comments and code would make the .jl file easier to read.

examples/error_estimates_forces.jl
```diff
@@ -0,0 +1,182 @@
...
+    Nk = length(basis.kpoints)
+    err = similar(ϕ)
+    for ik = 1:Nk
+        ϕk = ϕ[ik]
+        ψk = ψ[ik]
+        S = ψk'ϕk
+        U = S*(S'S)^(-1/2)
+        err[ik] = ϕk - ψk*U
+    end
+    err
```
mfherbst, line 81:
```suggestion
    map(zip(ϕ, ψ)) do (ϕk, ψk)
        S = ψk'ϕk
        U = S*(S'S)^(-1/2)
        ϕk - ψk*U
    end
```

examples/error_estimates_forces.jl
```diff
@@ -0,0 +1,182 @@
...
+end
+err = compute_error(basis_ref, ψr, ψ_ref);
+# - Compute ``{\bm M}^{-1}R(P)`` with ``{\bm M}^{-1}`` defined in[^CDKL2021]:
+P = [PreconditionerTPA(basis_ref, kpt) for kpt in basis_ref.kpoints]
+for (ik, ψk) in enumerate(ψr)
```
mfherbst, line 86: could use zip pattern again

examples/error_estimates_forces.jl
```diff
@@ -0,0 +1,182 @@
...
+    δφnk = sqrt.(Pk.mean_kin[n] .+ Pk.kin) .* δφnk
+    δφnk = DFTK.proj_tangent_kpt(δφnk, φk)
+    δφnk = sqrt.(Pk.mean_kin[n] .+ Pk.kin) .* δφnk
+    δφnk = DFTK.proj_tangent_kpt(δφnk, φk)
+end
+function apply_inv_M(φk, Pk, δφnk, n)
+    δφnk = DFTK.proj_tangent_kpt(δφnk, φk)
+    function op(x)
+        apply_M(φk, Pk, x, n)
+    end
```
mfherbst, line 100:
```suggestion
    op(x) = apply_M(φk, Pk, x, n)
````

File: examples/error_estimates_forces.jl (hunk 11 above)
Comment (mfherbst) [reply examples/error_estimates_forces.jl:1]: Could you also line-break the long comment lines.

File: examples/error_estimates_forces.jl [hunk 15]
```diff
@@ -0,0 +1,182 @@
...
+ψ, _ = DFTK.select_occupied_orbitals(basis, scfres.ψ, scfres.occupation)
+e1 = DFTK.solve_ΩplusK(basis, ψ, rhs, occupation; tol_cg=tol)
+e1 = DFTK.transfer_blochwave(e1, basis, basis_ref)
+res_schur = e1 + Mres;
+
+# ## Error estimates
+# We start with different estimations of the forces on the first ``\rm Ti`` atom, in direction 1.
+# - Forces computed with the reference solution:
+f_ref = compute_forces(scfres_ref)
+println("F(P_*) = $(f_ref[1][1][1])")
```
mfherbst, line 152: Add a comment explaining the `[1][1][1]`

examples/error_estimates_forces.jl
```diff
@@ -0,0 +1,182 @@
...
+# ```math
+# F(P) = F(P_*) + {\rm d}F(P)\cdot(P-P_*).
+# ```
+# To this end, we use the `ForwardDiff.jl` package to compute ``{\rm d}F(P)``
+# using automatic differentiation.
+function df(basis, occupation, ψ, δψ, ρ)
+    δρ = DFTK.compute_δρ(basis, ψ, δψ, occupation)
+    function f(ε)
+        compute_forces(basis, ψ .+ ε.*δψ, occupation; ρ=ρ+ε.*δρ)
+    end
```
mfherbst, line 166: I put this as an anonymous function (`ε -> ...`) directly to forward-diff.

examples/error_estimates_forces.jl
```diff
@@ -0,0 +1,182 @@
...
+    ForwardDiff.derivative(f, 0)
+end;
+# - Computation of the forces by a linearization argument if we have access to the actual error ``P-P_*``:
+df_err = df(basis_ref, occupation, ψr, DFTK.proj_tangent(err, ψr), ρr)
+println("F(P) - df(P).(P-P_*) = $(f[1][1][1]-df_err[1][1][1])")
```
mfherbst, line 171: `\cdot<TAB>` inserts a nice middle-centred dot

mfherbst (commented):
Thanks Gaspard and sorry I only got round to leave proper comments now.

examples/error_estimates_forces.jl
```diff
@@ -0,0 +1,182 @@
...
+
+# We apply a small displacement to one of the ``\rm Ti`` atoms to get nonzero forces.
+el, pos = atoms[1]
+atoms[1] = Pair(el, pos .+ [[-0.22, 0.28, 0.35] / 10, [0, 0, 0]]);
+# We build a model with one k-point only, not too high `Ecut_ref` and small tolerance to
+# limit computational time. These parameters can be increased for more precise
+# results.
+model = model_LDA(lattice, atoms)
+kgrid = [1, 1, 1]
+Ecut_ref = 35
```
mfherbst, line 45: Is there any chance you can make that a bit smaller still? Just because it adds about 3 minutes to our CI runtime. It is a nice showcase, so I'd keep it as is if needed.

antoine-levitt: I think we need to figure this out more generally. As we go towards more complicated systems / postprocessings, we can't really expect all examples to run in reasonable time on CI. How about we just give up running the examples on CI and just pre-generate them? Of course it means that they won't be run as CI, so we'd have to move tests to the testsuite and not rely on examples (eg some tests can just be a reduced version of the examples).

src/orbitals.jl
gkemlin: If we use `>=` then `threshold=0.0" would select all the states for insulators, so I'm not sure that's a good option. However, I agree that `>` is not very good either, but I'm not sure there is a better solution ? Maybe `findlast(x -> x > threshold, occk) - 1` ?

src/terms/local.jl
gkemlin: Well, if we choose to do that, it raises the error
```
LoadError: syntax: local variable name "T" conflicts with a static parameter
```
so maybe rather something like
```
@timing "forces: local" function compute_forces(::TermAtomicLocal, basis::PlaneWaveBasis{TT}, ψ, occ; ρ, kwargs...) where TT  
T = promote_type(TT, real(eltype(ψ[1])))                             

```

examples/error_estimates_forces.jl
gkemlin: Yes, I know some lines are long... I tried to did better in the next commit. However, for the lines with the markdown syntax for list `-` I don't know if there is another way : if I break the line in the comment then the output in the doc is weird...

examples/error_estimates_forces.jl
gkemlin: I tried to use smaller `Ecut_ref` but then the results are not really convincing : the reference solution is not converged (it is not clear if it is converged), and with smaller `Ecut` then the linear regime is not established.

mfherbst: I see. Ok, let's leave it as is for now and we try to figure out something more clever to do with this issue in general.

examples/error_estimates_forces.jl
mfherbst: I think if you line-break than you have to indent it appropriately, e.g.
```julia
# - abc def
#   ghi
```
Did you try that? 

Other option is of course to drop the leading `-` ...

gkemlin: yep, corrected

src/orbitals.jl
mfherbst: Yeah good point. Let's leave it as is.

src/terms/local.jl
mfherbst: ok.

PR #564, opened 2022-01-13
Generalize PowerNonlinearity to LocalNonlinearity

antoine-levitt:
This is breaking, do we want to be fancy and have it be working with a deprecation?


src/terms/local_nonlinearity.jl
```diff
@@ -0,0 +1,35 @@
...
+
+    # In the case of collinear spin, the potential is spin-dependent
+    ops = [RealSpaceMultiplication(basis, kpt, potential[:, :, :, kpt.spin])
+           for kpt in basis.kpoints]
+    (E=E, ops=ops)
+end
+
+
+function compute_kernel(term::TermLocalNonlinearity, ::AbstractBasis; ρ, kwargs...)
+    fp(ρ) = ForwardDiff.derivative(term.f, ρ)
```
mfherbst, line 26: As I said my only remark here would be why to use ForwardDiff instead of Zygote (which generates the differentiated code). I have not tested this, but I could imagine that using Zygote vectorizes better and thus is overall faster. Not that it matters here, I just think we should discuss it once and reach a conclusion.

antoine-levitt: Both have the theoretical potential to just unroll everything and get optimal code. I trust forward more than reverse (because it's an easier task for the ad) for such simple functions, but we can benchmark and see who's faster if you want. Or, when diffractor's forward mode works, we can use that.

mfherbst: Exactly I just wanted to benchmark this once and see what comes out. I agree with you that ForwardDiff is the simplest option and thus preferable in case the difference in timings is negligible.

antoine-levitt:
```julia
function f(ρ)
    # Ugly automatically generated code. TODO code from the equations
    x0  = sqrt(5913592351)
    x1  = 2^(2/3)*3^(1/6)
    x2  = pi^(1/6)
    x3  = ρ^(1/6)
    x4  = x2*x3
    x5  = 468052052839577158149*x0/928058053502134909615000000
    x6  = 6^(1/3)
    x7  = 12500*x4*x6
    x8  = pi^(1/3)
    x9  = ρ^(1/3)
    x10 = x8*x9
    x11 = 1 / (46593*x1.*x10 + x7 + 323380*sqrt(pi)*sqrt(ρ))

    x5*atan(x0*x4/(12500*x1 + 46593*x4)) + 310907*log(x11*x7)/10000000 + 76037485627899*log(x11*x4*(25000*x1 + 5249*x4)^2/100000)/78468213432500000
end

using ForwardDiff
@btime ForwardDiff.derivative(f, 1.2)

using Zygote
@btime Zygote.gradient(f, 1.2)

using Diffractor
@btime Diffractor.gradient(f, 1.2)
```
```
  288.088 ns (0 allocations: 0 bytes)
  356.692 ns (1 allocation: 16 bytes)
  6.314 μs (225 allocations: 6.56 KiB)
```
So let's stick with ForwardDiff

mfherbst: ok :+1:

mfherbst: but wait you did not benchmark the vectorizing aspect

mfherbst:
> This is breaking, do we want to be fancy and have it be working with a deprecation?

We could. I mean if we refactor the atoms structure (which will be soon) we will have to make a breaking release anyway. Then we can remove the deprecation.

src/terms/local_nonlinearity.jl
antoine-levitt: I very much doubt the primal vectorizes, let alone the ad

mfherbst: Ok, indeed no real changes if one goes to larger input array sizes.

PR #565, opened 2022-01-13
Simplify xc fallbacks, allow second derivatives, remove codegen

antoine-levitt:


antoine-levitt:
Hm actually it's hidden from the tests because the libxc version is preferred... I don't have a good solution to test the kernel then

antoine-levitt:
Can we get rid of the `Val` business? I think it's an unecessary optimization (to be checked but I'm pretty sure that the cost, if any, of dynamic dispatch is O(1) in broadcast)

antoine-levitt:
Ah nevermind you actually use it for dispatch. OK that's good! I'm still not as big a fan of dispatch are you are, but here it makes sense.

PR #569, opened 2022-01-22
More XC fallback implementations

mfherbst:


src/xc/xc_evaluate.jl
```diff
@@ -7,28 +7,54 @@ end
...
 
 # Function that always dispatches to the DFTK fallback implementations
 function xc_fallback!(func::Functional, ::Val{:lda}, ρ::AbstractArray;
                       zk=nothing, vrho=nothing, v2rho2=nothing)
     func.n_spin == 1  || error("Fallback functionals only for $(func.n_spin) == 1")
     zk = reshape(zk, size(ρ))
 
-    fE = ρ -> energy_per_particle(Val(func.identifier), ρ)
-    if zk !== nothing
+    fE(ρ) = energy_per_particle(Val(func.identifier), ρ)
```
antoine-levitt, line 20: I was wary of doing this because of the super nasty https://github.com/JuliaLang/julia/issues/15602, but OK it's not an issue here if we don't redefine it.

antoine-levitt (commented):
My my you have fun weekends :-) but great! How did you check those?

mfherbst:
> How did you check those?

Against libxc within double precision accuracy using BigFloat for our fallbacks (see unit tests).

antoine-levitt:
Ah right, duh. Great then!

PR #571, opened 2022-01-22
Implement meta-GGA functionals

mfherbst:
I start to really like our code :smile:. One evening and 110 lines of code and I got to do the first SCAN calculation of silicon. Not sure yet if the results are correct, but hey!


src/terms/xc.jl
```diff
@@ -83,36 +100,58 @@ end
...
         potential .*= term.scaling_factor
     end
-    ops = [RealSpaceMultiplication(basis, kpt, potential[:, :, :, kpt.spin])
-           for kpt in basis.kpoints]
-    (E=E, ops=ops)
+    ops = map(basis.kpoints) do kpt
+        if haskey(terms, :vtau) && any(x -> abs(x) ≥ term.potential_threshold, terms.vtau)
+            # Need meta-GGA non-local operator
+            # Note: Minus in front of scaling coefficient comes from partial integration
+            Vτ = -term.scaling_factor .* permutedims(terms.vtau, (2, 3, 4, 1))
```
antoine-levitt, line 107: pretty confusing to have Vτ be -vtau... are these names standard?

antoine-levitt: Actually it's probably better to put the minus in the operator: `-div(Agrad)`. That way it's a positive operator when A is positive, and the standard kinetic energy when A=1/2.

src/terms/xc.jl
```diff
@@ -83,36 +100,58 @@ end
...
-Etot = ∫ ρ E(ρ,σ), where σ = |∇ρ|^2 and E(ρ,σ) is the energy per unit particle.
-libxc provides the scalars
+    Etot = ∫ ρ ε(ρ,σ,τ,Δρ)
+where ε(ρ,σ,τ,Δρ) is the energy per unit particle, σ = |∇ρ|², τ = ∑ᵢ |∇ϕᵢ|²
+is the kinetic energy density and Δρ is the Laplacian of the density.
+
+Libxc provides the scalars
     Vρ = ∂(ρ E)/∂ρ
     Vσ = ∂(ρ E)/∂σ
+    Vτ = ∂(ρ E)/∂σ
```
antoine-levitt, line 129: typo

src/terms/xc.jl
```diff
@@ -182,17 +221,21 @@ end
...
 # stores the input to libxc in a format it likes
 struct LibxcDensity
-    basis
+    basis::PlaneWaveBasis
     max_derivative::Int
     ρ_real    # density ρ[iσ, ix, iy, iz]
     ∇ρ_real   # for GGA, density gradient ∇ρ[iσ, ix, iy, iz, iα]
     σ_real    # for GGA, contracted density gradient σ[iσ, ix, iy, iz]
+    Δρ_real   # for (some) mGGA, Laplacian of the density Δρ[iσ, ix, iy, iz]
 end
```
antoine-levitt, line 239: why isn't tau in this list?

src/terms/xc.jl
```diff
@@ -381,11 +430,14 @@ function add_kernel_gradient_correction!(δV, terms, density, perturbation, cros
...
 
 
-function Libxc.evaluate(xc::Functional, density::LibxcDensity; kwargs...)
+function Libxc.evaluate(xc::Functional, density::LibxcDensity; τ=nothing, kwargs...)
     if xc.family == :lda
         evaluate(xc; rho=density.ρ_real, kwargs...)
     elseif xc.family == :gga
         evaluate(xc; rho=density.ρ_real, sigma=density.σ_real, kwargs...)
+    elseif xc.family == :mgga
+        lapl = needs_laplacian(xc) ? (; lapl=density.Δρ_real) : (; )
```
antoine-levitt, line 439: isn't this more complicated than a plain if?

antoine-levitt (commented):
Great! A minima you can test it with hamiltonian consistency

src/terms/xc.jl
```diff
@@ -12,10 +12,13 @@ struct Xc
...
     density_threshold::Union{Nothing,Float64}
+
+    # Threshold for potential terms: Below this value a potential term is counted as zero.
+    potential_threshold::Float64
 end
 Xc(symbols::Symbol...; kwargs...) = Xc([symbols...]; kwargs...)
-function Xc(symbols::Vector; scaling_factor=1, density_threshold=nothing)
-    Xc(convert.(Symbol, symbols), scaling_factor, density_threshold)
+function Xc(symbols::Vector; scaling_factor=1, density_threshold=nothing, potential_threshold=1e-14)
+    Xc(convert.(Symbol, symbols), scaling_factor, density_threshold, potential_threshold)
```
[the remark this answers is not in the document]
mfherbst: yes.

src/terms/xc.jl
mfherbst: Because it's not a density derivative, but generated from the orbitals. Physically that sort of lives in a different compartment.

src/terms/xc.jl
mfherbst: oh nothing is really standard in this regime. There is at least two philosophies in naming DFT things in the literature. `vtau` is what Libxc uses. I agree about the naming and the scaling, I'll put the constants into the operator. Then if V=1 it is just the kinetic energy (in my original implementation I forgot an 1/2).

src/terms/operators.jl
```diff
@@ -125,6 +125,29 @@ end
...
+
+    # TODO This allocates *twice* on every apply!
+    for α = 1:3
+        # Note: The two 1im from both derivatives give rise to the .-=
+        ∂αψ_real = G_to_r(op.basis, op.kpoint, G_plus_k[α] .* ψ.fourier)
+        Vτ∇ψ = r_to_G(op.basis, op.kpoint, ∂αψ_real .* op.Vτ)  # allocates!
+        Hψ.fourier .-= G_plus_k[α] .* Vτ∇ψ  # collapses both imaginary units from the derivatives
+    end
+end
+# TODO Implement  Matrix(op::NonlocalMetaGGA)
```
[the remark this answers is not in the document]
mfherbst: ähm ... nothing, same thing it's not implemented.

src/terms/xc.jl
antoine-levitt: meh, code organization trumps physical relevance here I would say, but as you wish. Another option would be to rename the struct to LibxcInput

mfherbst: Yeah, ok, that's a good point. I'll think about it.

mfherbst:
One can probably still improve a bit in the mGGA performance, but I'll leave it here for now. The allocs are only on the small k-Point specific grid.

src/terms/Hamiltonian.jl
```diff
@@ -2,121 +2,149 @@
...
     # The operators are vectors of RealFourierOperator,
     # not typed because of type invariance issues.
     operators::Vector  # the original list of RealFourierOperator
                        # (as many as there are terms), kept for easier exploration
-    optimized_operators::Vector  # the optimized list of RealFourierOperator, to be used for applying
+    optimized_operators::Vector  # Optimized list of RealFourierOperator, for application
+end
+
+# More optimized HamiltonianBlock for the important case of a DFT Hamiltonian
+struct DftHamiltonianBlock <: HamiltonianBlock
```
antoine-levitt, line 20: caps for DFT?

src/terms/Hamiltonian.jl
```diff
@@ -2,121 +2,149 @@
...
+        fourier_ops  = filter(o -> o isa FourierMultiplication,   optimized_operators)
+        real_ops     = filter(o -> o isa RealSpaceMultiplication, optimized_operators)
+        nonlocal_ops = filter(o -> o isa NonlocalOperator,        optimized_operators)
+        divAgrid_ops = filter(o -> o isa DivAgradOperator,        optimized_operators)
+        @assert length(divAgrid_ops) < 2
+
+        if (length(fourier_ops) == length(real_ops) == length(nonlocal_ops) == 1)
+            return DftHamiltonianBlock(basis, kpoint, operators, optimized_operators,
+                                       only(fourier_ops), only(real_ops), only(nonlocal_ops),
+                                       something(divAgrid_ops..., Some(nothing)),
```
antoine-levitt, line 47: that pattern doesn't exactly roll off the tongue

src/terms/Hamiltonian.jl
```diff
@@ -2,121 +2,149 @@
...
+struct GenericHamiltonianBlock <: HamiltonianBlock
     basis::PlaneWaveBasis
     kpoint::Kpoint
 
     # The operators are vectors of RealFourierOperator,
     # not typed because of type invariance issues.
     operators::Vector  # the original list of RealFourierOperator
                        # (as many as there are terms), kept for easier exploration
-    optimized_operators::Vector  # the optimized list of RealFourierOperator, to be used for applying
+    optimized_operators::Vector  # Optimized list of RealFourierOperator, for application
```
antoine-levitt, line 16: just don't do the optimization in the generic hamiltonian code? Makes it even clearer

src/terms/Hamiltonian.jl
```diff
@@ -2,121 +2,149 @@
...
-        Threads.@threads for iband = 1:nband
+# Fast version, specialized on DFT models. Minimizes the number of allocations
+@views @timing "DftHamiltonian multiplication" function LinearAlgebra.mul!(Hψ::AbstractArray,
+                                                                           H::DftHamiltonianBlock,
+                                                                           ψ::AbstractArray)
+    n_bands = size(ψ, 2)
+    have_divAgrad = !isnothing(H.divAgrad_op)
+
+    potential = H.real_op.potential
+    potential /= prod(H.basis.fft_size)  # because we use unnormalized plans
```
antoine-levitt, line 120: for additional speed

src/terms/Hamiltonian.jl
mfherbst: nah. I know we are not consistent, but I prefer camel-case also in abbreviations.

src/terms/Hamiltonian.jl
mfherbst: I hope it's better now.

PR #574, opened 2022-01-28
Replace `@timing_seq` by thread-local timers

mfherbst:


antoine-levitt:
Very cool! But you removed a bunch (sternheimer and ffts) that seem useful?

mfherbst:
well yes, but the _seq stuff was disabled by default and has always been a bit tricky to enable (clean precompile cache, set environment variable, disable threadying, etc.). So I honestly never used it. Did you? I think it's better to track these things by appropriate thread-local timers and timers in surrounding sections now.

Regarding the Sternheimer, yes that we could annotate better, but as it stands the timer has been disabled the whole time (it was as `_seq` timer) and no-one noticed :smile:.

antoine-levitt:
OK !

PR #575, opened 2022-01-29
Optimization to density building

antoine-levitt:


src/densities.jl
```diff
@@ -5,6 +5,17 @@
 #        compute_density, compute_δρ, compute_kinetic_energy_density
 #      - Use symmetrization instead of explicit use of symmetry operators
 
+function _check_positive(ρ)
+    minimum(ρ) < 0 && @warn("Negative ρ detected",
+                            min_ρ=minimum(ρ))
```
mfherbst, line 10: Formatting

src/densities.jl
```diff
@@ -5,6 +5,17 @@
...
+end
+function _check_total_charge(basis, ρ, N)
+    n_electrons = sum(ρ) * basis.dvol
+    if abs(n_electrons - N) > sqrt(eps(eltype(basis)))
```
mfherbst, line 14: T could be inferred from ρ and instead of passing basis, pass dvol (makes function more flexible)

src/densities.jl
```diff
@@ -118,18 +112,15 @@ end
...
     mpi_sum!(δρ_fourier, basis.comm_kpts)
     count = sum(length(basis.ksymops[ik]) for ik in 1:length(basis.kpoints)) ÷ n_spin
     count = mpi_sum(count, basis.comm_kpts)
+    lowpass_for_symmetry!(δρ_fourier, basis)
     δρ = G_to_r(basis, δρ_fourier) ./ count
-    # Check sanity in the density, we should have ∫δρ = 0
-    if abs(sum(δρ)) > sqrt(eps(T))
-        @warn "Non-neutral δρ" sum_δρ=sum(δρ)
-    end
+    _check_total_charge(basis, δρ, 0)
```
mfherbst, line 123: The error message in the warning is very misleading in this case.

src/densities.jl
```diff
@@ -54,15 +57,6 @@ is not collinear the spin density is `nothing`.
...
     n_spin = basis.model.n_spin_components
 
-    # Sanity checks
-    @assert n_k == length(ψ)
-    @assert n_k == length(occupation)
-    for ik in 1:n_k
-        @assert length(G_vectors(basis, basis.kpoints[ik])) == size(ψ[ik], 1)
-        @assert length(occupation[ik]) == size(ψ[ik], 2)
-    end
-    @assert n_k > 0
```
mfherbst, line 64: Not convinced. Quite an elementary piece of the code. Having this helps to spot errors (like wrongful setup of the \psi array).

antoine-levitt: not really though, it's pretty deep inside scf & stuff, so by that point you're expected to have setup this stuff correctly. If these aren't consistent, you'll run into a lot of pretty clear trouble. It's leftover for when we were figuring out this stuff, but not really needed anymore. Otherwise why would we have these checks here and not everywhere else?

src/densities.jl
antoine-levitt: Hm true.

antoine-levitt: although, really? looks quite OK to me actually

src/PlaneWaveBasis.jl
```diff
@@ -236,7 +239,8 @@ end
 
 # This is the "internal" constructor; the higher-level one below should be preferred
 @timing function PlaneWaveBasis(model::Model{T}, Ecut::Number,
-                                kcoords::AbstractVector, ksymops, symmetries=nothing;
+                                kcoords::AbstractVector, ksymops,
+                                symmetries = symmetries_preserving_kcoords_ksymops(model.symmetries, kcoords, ksymops);
```
mfherbst, line 243: As you know long lines make me uneasy. How about setting this to nothing and inferring the symmetries inside as we did before?

src/densities.jl
mfherbst: ok

src/densities.jl
```diff
@@ -118,18 +111,15 @@ end
...
     mpi_sum!(δρ_fourier, basis.comm_kpts)
     count = sum(length(basis.ksymops[ik]) for ik in 1:length(basis.kpoints)) ÷ n_spin
     count = mpi_sum(count, basis.comm_kpts)
+    lowpass_for_symmetry!(δρ_fourier, basis)
     δρ = G_to_r(basis, δρ_fourier) ./ count
-    # Check sanity in the density, we should have ∫δρ = 0
-    if abs(sum(δρ)) > sqrt(eps(T))
-        @warn "Non-neutral δρ" sum_δρ=sum(δρ)
-    end
+    _check_total_charge(basis.dvol, δρ, 0)
```
mfherbst, line 122: I still think this will yield to a confusing error message. People will run response or AD when this pops up and then reading something about the number of electrons being 0 is clearly not what they expect.

src/postprocess/chi0.jl
```diff
@@ -254,19 +254,18 @@ function apply_χ0(ham, ψ, εF, eigenvalues, δV; kwargs_sternheimer...)
            Smearing.occupation.(model.smearing, (eigenvalues[ik] .- εF) ./ model.temperature)
            for ik = 1:length(basis.kpoints)]
 
+
```
mfherbst, line 257: kill one empty line

test/hamiltonian_consistency.jl
```diff
@@ -30,17 +30,21 @@ function test_consistency_term(term; rtol=1e-4, atol=1e-8, ε=1e-8, kgrid=[1, 2,
...
+            end
+            E, _ = energy_hamiltonian(basis, ψ_trial, occupation; ρ=ρ_trial)
+            E.total
         end
 
-        @assert length(basis.terms) == 1
-        E0, ham = energy_hamiltonian(basis, ψ, occupation; ρ=ρ)
-        E1, _ = energy_hamiltonian(basis, ψ_trial, occupation; ρ=ρ_trial)
-        diff = (E1.total - E0.total)/ε
+        diff = (compute_E(ε)-compute_E(-ε))/(2ε)
```
mfherbst, line 47: spacing

src/symmetry.jl
```diff
@@ -66,13 +66,32 @@ end
 Filter out the symmetry operations that respect the symmetries of the discrete BZ grid
 """
 function symmetries_preserving_kgrid(symmetries, kcoords)
+    kcoords_normalized = normalize_kpoint_coordinate.(kcoords)
+    T = eltype(kcoords[1])
+    tol = T <: Rational ? 0 : sqrt(eps(T))
+    is_approx_in(x, X, tol) = any(y -> norm(x-y) ≤ tol, X)
```
mfherbst, line 72: This is exactly what isapprox is doing when you have a vector. Might as well write this as:
```
is_approx_in(x, X; atol) = any(isapprox(x; atol), X)
```
or directly inline that below?

src/symmetry.jl
```diff
@@ -66,13 +66,32 @@ end
...
     end
     filter(symop -> preserves_grid(symop[1]), symmetries)
 end
 
+
+"""
+Find the subset of symmetries compatible with the grid induced by the given kcoords and ksymops
+"""
+function symmetries_preserving_kcoords_ksymops(symmetries, kcoords, ksymops)
+    all_kcoords = []
```
mfherbst, line 85: typing

src/symmetry.jl
```diff
@@ -66,13 +66,32 @@ end
...
+    for ik=1:length(kcoords)
+        for symop in ksymops[ik]
+            push!(all_kcoords, symop[1]*kcoords[ik])
```
mfherbst, line 88: spacing

src/symmetry.jl
```diff
@@ -66,13 +66,32 @@ end
...
+            for k in kcoords_normalized)
     end
     filter(symop -> preserves_grid(symop[1]), symmetries)
 end
 
+
+"""
+Find the subset of symmetries compatible with the grid induced by the given kcoords and ksymops
+"""
+function symmetries_preserving_kcoords_ksymops(symmetries, kcoords, ksymops)
```
mfherbst, line 84: I wonder if it would be nicer/easier to read if we expose `preserves_grid` as a function as well as a function to build all `kcords` (we might already have this) and then just perform the action of this function beast by just using `filter` and the latter two functions.

src/PlaneWaveBasis.jl
antoine-levitt: That's super ugly, we're not going to not use default argument values just because it makes for long lines. I've broken the line, but of course the issue is that symmetries_preserving_kcoords_ksymops is a ridiculous name. How about symmetries_preserving_BZ(symmetries, kcoords) and symmetries_preserving_BZ(symmetries, kcoords, ksymops)?

src/densities.jl
antoine-levitt: It's not really feasible long-term to have different error messages for the AD and not AD version. Simplest is just to turn off the check?

src/symmetry.jl
antoine-levitt: good catch, I thought I already did it but no

src/symmetry.jl
antoine-levitt: do you know of a good way to do this without mutation?

src/symmetry.jl
antoine-levitt: I'd rather not anticipate possible future API needs that we don't know yet, so let's just do that if we find that we need preserves_grid. Hopefully we will never have to touch this code ever again (where have I heard that before?...)

src/symmetry.jl
```diff
@@ -68,8 +68,8 @@ Filter out the symmetry operations that respect the symmetries of the discrete B
 function symmetries_preserving_kgrid(symmetries, kcoords)
     kcoords_normalized = normalize_kpoint_coordinate.(kcoords)
     T = eltype(kcoords[1])
-    tol = T <: Rational ? 0 : sqrt(eps(T))
-    is_approx_in(x, X, tol) = any(y -> norm(x-y) ≤ tol, X)
+    atol = T <: Rational ? 0 : sqrt(eps(T))
+    is_approx_in(x, X, tol) = any(y -> isapprox(x, y; atol), X)
```
mfherbst: No need to specify `atol`. What you are doing here is the default behaviour.

src/symmetry.jl
mfherbst: Hmm. I understand you don't want to touch this again, but in my experience it's better to build complex functionality like this by chaining small functions (such as preserve_kgrids predicate and a generator for all kcoords). A generator for all kcoords sounds like a pattern we might actually have somewhere in the code without realising it for other reasons (I have not checked though).

src/symmetry.jl
antoine-levitt: hm indeed!

antoine-levitt: although, no: default is to use rtol, which will not be approximate when comparing to zero

src/densities.jl
mfherbst: Actually I agree. This is probably not very useful to have here.

src/symmetry.jl
mfherbst: ok, let's keep it as is then.

src/symmetry.jl
mfherbst: Yes. Apply a map to Iterators.product.

src/PlaneWaveBasis.jl
```diff
@@ -277,7 +276,10 @@ end
 Creates a new basis identical to `basis`, but with a custom set of kpoints
 """
 @timing function PlaneWaveBasis(basis::PlaneWaveBasis, kcoords::AbstractVector,
-                                ksymops::AbstractVector, symmetries=vcat(ksymops...))
+                                ksymops::AbstractVector,
+                                symmetries=symmetries_preserving_kcoords_ksymops(basis.model.symmetries,
+                                                                                 kcoords,
+                                                                                 ksymops))
```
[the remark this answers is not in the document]
antoine-levitt: yes, why? it's not a kwarg

src/symmetry.jl
mfherbst: ok then, just keep it with the mutation.

src/PlaneWaveBasis.jl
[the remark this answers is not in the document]
antoine-levitt: the tests use it as I recall, and the next constructor won't compute it. Of course we should just remove this mess, but let's wait until we get a clear idea of what symmetrization do we want exactly.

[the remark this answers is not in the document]
mfherbst: I just checked and I only found the band structure stuff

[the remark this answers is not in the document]
antoine-levitt: let's just wait until I complete this round of PRs (maybe this kind of arguments will change) and then we can try to simplify the constructor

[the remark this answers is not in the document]
mfherbst: ok

src/PlaneWaveBasis.jl
mfherbst: I guess we wait with the renaming until you are done with the other refactoring. But I'm ok with `symmetries_preserving_bz`

antoine-levitt: I went with preserving_kgrid (because the symmetries always preserve the bz, it's just the discrete one, ie the grid, that is not preserved)

mfherbst: yeah, ok. That's not a bad name indeed.

PR #576, opened 2022-01-29
Cleanup Chi0Mixing docstrings and timings

mfherbst:
In my tests employing LdosMixing for potential-mixing SCF iterations actually works way worse than with density-mixing, which surprises me a bit. My suspicion it's got to do with the fact that we actually apply ε rather than ε^\dagger during the linear solve, such that we cannot benefit from the implict low-pass filtering done by the Hartree kernel. Any thoughts @antoine-levitt ?

**Edit:** I'll split the LdosMixing changes to another PR (#587) to get the the other misc. changes into master.

### DIFF
--- a/src/eigen/lobpcg_hyper_impl.jl
+++ b/src/eigen/lobpcg_hyper_impl.jl
@@ -169,7 +169,7 @@ function drop!(X, tol=2eps(real(eltype(X))))
 end
 
 # Find X that is orthogonal, and B-orthogonal to Y, up to a tolerance tol.
-function ortho!(X, Y, BY; tol=2eps(real(eltype(X))))
+@timing "ortho! X vs Y" function ortho!(X, Y, BY; tol=2eps(real(eltype(X))))
     T = real(eltype(X))
     # normalize to try to cheaply improve conditioning
     Threads.@threads for i=1:size(X,2)
@@ -322,8 +322,10 @@ end
         end
         vprintln(niter, "   ", resid_history[:, niter+1])
         if precon !== I
-            precondprep!(precon, X) # update preconditioner if needed; defaults to noop
-            ldiv!(precon, new_R)
+            @timing "preconditioning" begin
+                precondprep!(precon, X) # update preconditioner if needed; defaults to noop
+                ldiv!(precon, new_R)
+            end
         end
 
         ### Compute number of locked vectors
--- a/src/scf/mixing.jl
+++ b/src/scf/mixing.jl
@@ -202,7 +202,7 @@ real space using a GMRES. Either the full kernel (`RPA=false`) or only the Hartr
     RPA::Bool = true       # Use RPA, i.e. only apply the Hartree and not the XC Kernel
     χ0terms   = χ0Model[Applyχ0Model()]  # The terms to use as the model for χ0
     verbose::Bool = false   # Run the GMRES verbosely
-    reltol::Float32 = 0.01  # Relative tolerance for GMRES
+    reltol::Float64 = 0.01  # Relative tolerance for GMRES
 end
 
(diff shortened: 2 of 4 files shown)


src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -213,9 +213,6 @@ function ortho!(X, Y, BY; tol=2eps(real(eltype(X))))
     end
     vprintln("ortho choleskys: ", ninners) # get how many Choleskys are performed
 
-    # @assert (norm(BY'X)) < tol
```
antoine-levitt, line 216: These are useful postconditions, put a comment instead if you want to remove them

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -313,7 +310,9 @@ end
         end
 
         ### Compute new residuals
-        new_R = new_AX .- new_BX .* λs'
+        @timing "residual" begin
+            new_R = new_AX .- new_BX .* λs'
```
antoine-levitt, line 314: I'd rather not time O(N) ops, if we time this we have to time every op

src/scf/chi0models.jl
```diff
@@ -16,6 +16,8 @@ For details see Herbst, Levitt 2020 arXiv:2009.01665
 """
 @kwdef struct LdosModel <: χ0Model
     adjust_temperature = IncreaseMixingTemperature()
+    characteristic_length = 0  # in Bohr (0 means low-pass filtering is disabled)
+                               # Good values for filtering are between 2 and 0.1 Bohr
```
antoine-levitt, line 20: Clarify the effect of this : it will smooth out LDOS below this length, right?

src/scf/chi0models.jl
```diff
@@ -28,6 +30,16 @@ function (χ0::LdosModel)(basis; eigenvalues, ψ, εF, kwargs...)
         return nothing
     end
 
+    if χ0.characteristic_length > 0
+        length_fourier = 2π / χ0.characteristic_length
```
antoine-levitt, line 34: Aka wavelength

src/scf/chi0models.jl
```diff
@@ -28,6 +30,16 @@ function (χ0::LdosModel)(basis; eigenvalues, ψ, εF, kwargs...)
...
+        G₀ = 2length_fourier / 3
+        Gslope = length_fourier / 6
+        f_lowpass(G) = (1 - erf((G - G₀) / Gslope))/ 2
```
antoine-levitt, line 37: Erfc? I would have used gaussian filtering but why not

src/scf/chi0models.jl
```diff
@@ -72,8 +84,7 @@ function (χ0::DielectricModel)(basis; kwargs...)
 end
 
 """
-Full χ0 application, optionally dropping terms or disabling Sternheimer.
-All keyword arguments passed to [`apply_χ0`](@ref).
+Full χ0 application. All keyword arguments passed to [`apply_χ0`](@ref).
```
antoine-levitt, line 87: Reason why I hate docstrings :p

src/scf/mixing.jl
```diff
@@ -182,45 +199,79 @@ real space using a GMRES. Either the full kernel (`RPA=false`) or only the Hartr
 @kwdef struct χ0Mixing <: Mixing
     RPA::Bool = true       # Use RPA, i.e. only apply the Hartree and not the XC Kernel
     χ0terms   = χ0Model[Applyχ0Model()]  # The terms to use as the model for χ0
-    verbose::Bool = false  # Run the GMRES verbosely
+    verbose::Bool = false   # Run the GMRES verbosely
+    reltol::Float32 = 0.01  # Relative tolerance for GMRES
 end
```
antoine-levitt, line 204: 32?

PR #580, opened 2022-02-03
Fix type in periodic problems guide

epolack:
The order seems wrong line 88, and ambiguous line 95+.

### DIFF
--- a/docs/src/guide/periodic_problems.jl
+++ b/docs/src/guide/periodic_problems.jl
@@ -85,15 +85,15 @@
 # ```math
 #    -i∇ \left( e^{i k⋅x} f \right) 
 #    = -i\frac{d}{dx} \left( e^{i k⋅x} f \right)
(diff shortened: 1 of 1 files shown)


docs/src/guide/periodic_problems.jl
```diff
@@ -85,15 +85,15 @@
 # ```math
 #    -i∇ \left( e^{i k⋅x} f \right) 
 #    = -i\frac{d}{dx} \left( e^{i k⋅x} f \right)
-#    = k e^{i k⋅x} f -i∇ e^{i k⋅x} f = (-i∇ + k) e^{i k⋅x} f.
+#    = k e^{i k⋅x} f -i e^{i k⋅x} ∇f = e^{i k⋅x} (-i∇ + k) f.
```
antoine-levitt, line 88: I usually prefer to keep (-i nabla) together (because it's the p operator and keeping it together avoids some sign mistakes)

antoine-levitt:
Great, thanks!!

PR #581, opened 2022-02-04
AtomicNonlocal Zygote compatibility

mtowara:
I added my changes for non-local, and updated for the new Hamiltonian structure.
Replaced some map(reduce()) with mapreduce.

I'll do a seperate PR against master for nonlocal.jl once I'm happy with the changes there


niklasschmitz (approved):
LGTM

PR #583, opened 2022-02-10
Proper integration of SCF with ForwardDiff

mfherbst:


src/scf/newton.jl
```diff
@@ -158,10 +158,10 @@ function solve_ΩplusK(basis::PlaneWaveBasis{T}, ψ, rhs, occupation;
     J = LinearMap{T}(ΩpK, size(rhs_pack, 1))
 
     # solve (Ω+K) δψ = rhs on the tangent space with CG
-    δψ = cg(J, rhs_pack, Pl=FunctionPreconditioner(f_ldiv!),
-            reltol=0, abstol=tol_cg, verbose=verbose)
+    δψ, history = cg(J, rhs_pack, Pl=FunctionPreconditioner(f_ldiv!),
+                  reltol=0, abstol=tol_cg, verbose=verbose, log=true)
 
-    unpack(δψ)
+    (; δψ=unpack(δψ), history)
```
antoine-levitt, line 164: I quite like this pattern of returning named tuples for non-trivial functions: it's sufficiently flexible that we can return whatever extra information we want, and the caller is free to ignore

src/terms/local_nonlinearity.jl
```diff
@@ -11,10 +11,10 @@ struct TermLocalNonlinearity{TF} <: TermNonlinear
 end
 (L::LocalNonlinearity)(::AbstractBasis) = TermLocalNonlinearity(L.f)
 
-function ene_ops(term::TermLocalNonlinearity, basis::PlaneWaveBasis, ψ, occ; ρ, kwargs...)
+function ene_ops(term::TermLocalNonlinearity, basis::PlaneWaveBasis{T}, ψ, occ; ρ, kwargs...) where {T}
     fp(ρ) = ForwardDiff.derivative(term.f, ρ)
     E = sum(term.f.(ρ)) * basis.dvol
-    potential = fp.(ρ)
+    potential = convert_dual.(T, fp.(ρ))
```
antoine-levitt, line 17: it's not clear to me why that is needed?

src/workarounds/forwarddiff_rules.jl
```diff
@@ -127,6 +127,75 @@ function _is_well_conditioned(A::AbstractArray{<:ForwardDiff.Dual}; kwargs...)
     _is_well_conditioned(ForwardDiff.value.(A); kwargs...)
 end
 
+# Convert and strip off duals if that's the only way
```
antoine-levitt, line 130: put this kind of generic helper functions to the top of the file

src/terms/local_nonlinearity.jl
mfherbst: In case you have a dual in the function returning a potential. (like in the forwarddiff example)

antoine-levitt: this is scary to me, how do you know you're not discarding important information? Also will we need to pepper the code with this? Let's discuss on a call

mfherbst: The way out is to properly refactor the floating-point type handling as we do it now. I agree that this should be done. This is only the "quick way out".

You are not discarding important information, since this `convert_dual` is only triggered in case the type of the PlaneWaveBasis (which in our current setup overrules everything) is non-dual and thus it is sort of requested to do a non-dual computation. Pushing the dual through in that case you will anyway get inconsistent results as not all dual rules will be triggered.

examples/forwarddiff.jl
```diff
@@ -43,43 +43,7 @@ polarizability_fd = let
...
-    T = ForwardDiff.tagtype(ε)
-    basis = make_basis(ForwardDiff.value(ε); kwargs...)
-    basis_dual = make_basis(ε; kwargs...)
-    ψ, ρ, δψ, δρ = self_consistent_field_dual(basis, basis_dual; tol)
-    ρ_dual = ForwardDiff.Dual{T}.(ρ, δρ)
-    dipole(basis_dual, ρ_dual)
-end;
-
-# This setup allows to compute the polarizability via automatic differentiation:
+# We do the same thing using automatic differentiation:
```
niklasschmitz, line 46: Should we hint to the user here that this is doing implicit differentiation based on our own rules?

niklasschmitz (commented):
This looks good to me from a quick overview; The conversion of `construct_value(...)` seems a reasonable choice, and the forwarddiff polarizability example looks a lot better!

PR #584, opened 2022-02-10
Schur complement in Sternheimer solver

gkemlin:
We changed the way the Sternheimer equations are solved to compute response in chi0.jl as well as some new features:
- Schur trick for insulators that uses the information from nonconverged extra bands (if there is temperature, the old version is used)
- all `apply_χ0` functions now take as kwargs extra wavefunctions and eigenvalues (nothing by default)
- new argument for `apply_χ0` that takes `scfres` as input (which simplifies the `dieletric.jl` example) and uses the extra bands associated to `scfres`
- new tests for `test/chi0.jl` with `scfres` argument, both for silicon and magnesium to anticipate future work on Sternheimer for metals ==> shall we keep the first one ?


src/postprocess/chi0.jl
```diff
@@ -105,31 +105,105 @@ precondprep!(P::FunctionPreconditioner, ::Any) = P
...
+        precon = PreconditionerTPA(basis, kpoint)
+        precondprep!(precon, ψnk)
+        function R_ldiv!(x, y)
+            x .= R(precon \ R(y))
+        end
+        J = LinearMap{eltype(ψk)}(RAR, size(Hk, 1))
+        # tol_cg should not be too tight, and in particular not be
+        # too far below the error in the ψ. Otherwise Q and H
+        # don't commute enough, and an oversolving of the linear
+        # system can lead to spurious solutions
```
antoine-levitt, line 178: Would be interesting to explore this a bit theoretically. R should help with the non-commutation, right? So the condition is something like tol >> ||R [Q,H] R||? Can we prove that something like this holds? Doe the gap play a role? If so which one? Generally speaking I haven't seen any error analysis for response-type linear problems, it'd be interesting to make even a rough analysis.

src/postprocess/chi0.jl
```diff
@@ -105,31 +105,105 @@ precondprep!(P::FunctionPreconditioner, ::Any) = P
...
+        Z = cg(J, bb, Pl=FunctionPreconditioner(R_ldiv!),
+               reltol=0, abstol=tol_cg, verbose=verbose)
+
+        # 2) solve for αn now that we know Z
+        αn = YHYi * Y' * (b - H(Z))
```
antoine-levitt, line 183: either both alpha and Z should be index by n, or none of the two

src/postprocess/chi0.jl
```diff
@@ -105,31 +105,105 @@ precondprep!(P::FunctionPreconditioner, ::Any) = P
...
+
+        # 3) put things together
```
antoine-levitt, line 185: kind of self-explanatory, you've got this formula above

antoine-levitt (commented):
Cool! I left some comments. There's some things I think can be simplified or improved, but maybe it's best to wait until the metals are done and refactor then? We can discuss next week if you want.

src/postprocess/chi0.jl
```diff
@@ -105,31 +105,105 @@ precondprep!(P::FunctionPreconditioner, ::Any) = P
...
+
+    if iszero(temperature) && !isnothing(ψk_extra)
+        # We use a Schur decomposition of the orthogonal of the occupied states
+        # where we still know some information from the partially converged
+        # nonoccupied states (in particular, they are Rayleigh-Ritz wrt to H).
+        # /!\ this is only implemented for insulators at the moment, WIP for
+        # /!\ metals (the old sternheimer is still used, see the "else" case)
+
+        # computed but nonconverged bands
+        Y = ψk_extra
```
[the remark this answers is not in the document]
gkemlin: Ok, I have made some modifications in this spirit but indeed, we should wait to see how it looks like for metals before taking a final decision.

src/postprocess/chi0.jl
```diff
@@ -105,31 +105,105 @@ precondprep!(P::FunctionPreconditioner, ::Any) = P
...
+
+        # Y are not converged but have been Rayleigh-Ritzed
+        # so YHY should be a real diagonal matrix
+        H(ϕ) = Hk * ϕ - εnk * ϕ
+        YHY = real.(Diagonal(εk_extra .- εnk))
+        YHYi = real.(Diagonal(1 ./ (εk_extra .- εnk)))
+
+        # computing products once and for all
+        HY = H(Y)
+        HYIY = HY * YHYi * Y'
```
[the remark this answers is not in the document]
gkemlin: That is true, I removed it.

src/postprocess/chi0.jl
gkemlin: Yes, that is exactly what we intend to explore with @mfherbst after I implement your review ! From what we already saw, R does help indeed.

src/postprocess/chi0.jl
```diff
@@ -180,7 +254,9 @@ function compute_αmn(fm, fn, ratio)
     ratio * fn / (fn^2 + fm^2)
 end
 
-@views @timing function apply_χ0_4P(ham, ψ, occ, εF, eigenvalues, δHψ; kwargs_sternheimer...)
+@views @timing function apply_χ0_4P(ham, ψ, occ, εF, eigenvalues, δHψ;
+                                    ψ_extra=nothing, eigenvalues_extra=nothing,
```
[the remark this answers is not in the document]
gkemlin: well that is weird because it should be called, let's see what happens after next CI test

src/postprocess/chi0.jl
```diff
@@ -245,9 +329,17 @@ Get the density variation δρ corresponding to a total potential variation δV.
 Note: This function assumes that all bands contained in `ψ` and `eigenvalues` are
 sufficiently converged. By default the `self_consistent_field` routine of `DFTK`
 returns `3` extra bands, which are not converged by the eigensolver
-(see `n_ep_extra` parameter). These should be discarded before using this function.
+(see `n_ep_extra` parameter). These should be discarded from `ψ` before using this
+function. However, one can still use additional information from nonconverged bands
+stored in `ψ_extra` (typically, the `3` extra bands returned by default in SCF
+routine).
 """
```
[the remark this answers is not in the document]
gkemlin: see function below which takes scfres as an argument

test/chi0.jl
```diff
@@ -93,3 +94,105 @@ for temperature in (0, 0.03), spin_polarization in (:none, :collinear)
                   temperature=temperature, spin_polarization=spin_polarization)
     end
 end
+
+# ==================== Test χ0 with SCF extra bands ===========================
+
+function test_chi0_scf(testcase; symmetry=false, use_symmetry=false, temperature=0,
```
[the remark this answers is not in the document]
gkemlin: The test above use fully converged states (cf eigen(Hk)) so we thought it would be interesting to turn that into a scfres-type interface. I agree that then the first test above can be removed, if you think that now it is useless.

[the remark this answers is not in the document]
gkemlin: We also includeed chi0 test for metals to anticipate the future :)

src/postprocess/chi0.jl
```diff
@@ -245,9 +317,16 @@ Get the density variation δρ corresponding to a total potential variation δV.
...
-function apply_χ0(ham, ψ, εF, eigenvalues, δV; kwargs_sternheimer...)
+function apply_χ0(ham, ψ, εF, eigenvalues, δV;
+                  ψ_extra=[zeros(size(ψk,1), 0) for ψk in ψ],
+                  ε_extra=[zeros(0) for ψk in ψ],
```
mfherbst, line 327:
```suggestion
                  ψ_extra=[zeros(size(ψk, 1), 0) for ψk in ψ],
                  eigenvalues_extra=[zeros(0) for _ in eigenvalues],
```

src/postprocess/chi0.jl
```diff
@@ -180,7 +248,10 @@ function compute_αmn(fm, fn, ratio)
     ratio * fn / (fn^2 + fm^2)
 end
 
-@views @timing function apply_χ0_4P(ham, ψ, occ, εF, eigenvalues, δHψ; kwargs_sternheimer...)
+@views @timing function apply_χ0_4P(ham, ψ, occ, εF, eigenvalues, δHψ;
+                                    ψ_extra=[zeros(size(ψk,1), 0) for ψk in ψ],
+                                    ε_extra=[zeros(0) for ψk in ψ],
```
mfherbst, line 253: See suggestions below.

src/postprocess/chi0.jl
```diff
@@ -103,33 +103,101 @@ LinearAlgebra.ldiv!(y::T, P::FunctionPreconditioner, x) where {T} = P.preconditi
...
+    # nonoccupied states (in particular, they are Rayleigh-Ritz wrt to Hk).
+    #
+    # /!\ This is only implemented for insulators at the moment, WIP for
+    # metals, in which case we use empty arrays and all computations are
+    # equivalent to invert the Sternheimer system in Ran(1-P) without the Schur
+    # trick.
+
+    # TODO
+    # finite temperature not supported yet, so we remove extra bands and perform
+    # all the computations with empty arrays
```
mfherbst, line 128: I'd just say:
```suggestion
    # finite temperature not supported yet => remove extra bands
```

src/postprocess/chi0.jl
```diff
@@ -103,33 +103,101 @@ LinearAlgebra.ldiv!(y::T, P::FunctionPreconditioner, x) where {T} = P.preconditi
...
+                            εk_extra=zeros(0), schur=true,
+                            tol_cg=1e-12, verbose=false)
     basis = Hk.basis
     kpoint = Hk.kpoint
+    model = basis.model
+    temperature = model.temperature
+
+    # We use a Schur decomposition of the orthogonal of the occupied states
+    # where we still know some information from the partially converged
+    # nonoccupied states (in particular, they are Rayleigh-Ritz wrt to Hk).
```
mfherbst, line 119: maybe:
```suggestion
    # into a part where we have the partially converged, non-occupied bands
    # (which are Rayleigh-Ritz wrt to Hk) and the rest.
```

src/postprocess/chi0.jl
```diff
@@ -103,33 +103,101 @@ LinearAlgebra.ldiv!(y::T, P::FunctionPreconditioner, x) where {T} = P.preconditi
...
 precondprep!(P::FunctionPreconditioner, ::Any) = P
 
-# Solves Q (H-εn) Q δψn = -Q rhs
-# where Q is the projector on the orthogonal of ψk
-function sternheimer_solver(Hk, ψk, ψnk, εnk, rhs; tol_cg=1e-6, verbose=false)
+# Solves (1-P) (H-εn) (1-P) δψn = - (1-P) rhs
+# where 1-P is the projector on the orthogonal of ψk
+function sternheimer_solver(Hk, ψk, ψnk, εnk, rhs;
+                            ψk_extra=zeros(size(ψk,1), 0),
+                            εk_extra=zeros(0), schur=true,
```
mfherbst, line 110: Why would you ever turn the Schur off explicitly? You can do so just by passing no extra bands. I would remove that kwarg.

src/postprocess/chi0.jl
```diff
@@ -103,33 +103,101 @@ LinearAlgebra.ldiv!(y::T, P::FunctionPreconditioner, x) where {T} = P.preconditi
...
 
-    # we err on the side of caution here by applying Q *a lot*
-    # there are optimizations to be made here
-    Q(ϕ) = ϕ - ψk * (ψk' * ϕ)
-    function QHQ(ϕ)
-        Qϕ = Q(ϕ)
-        Q(Hk * Qϕ - εnk * Qϕ)
+    # Projectors:
+    # projector onto the computed and converged states, which are the occupied
+    # states in the case of insulators
```
mfherbst, line 137: They are also "occupied" for metals (in the sense of non-zero occupation). I would remove that half-sentence.

src/postprocess/chi0.jl
```diff
@@ -103,33 +103,101 @@ LinearAlgebra.ldiv!(y::T, P::FunctionPreconditioner, x) where {T} = P.preconditi
...
+    P(ϕ) = ψk * (ψk' * ϕ)
+    # projector onto the computed but nonconverged states
+    P_free(ϕ) = ψk_extra * (ψk_extra' * ϕ)
```
mfherbst, line 140: Why `P_free` that carries no meaning for me. `P_extra` ?

src/postprocess/chi0.jl
```diff
@@ -103,33 +103,101 @@ LinearAlgebra.ldiv!(y::T, P::FunctionPreconditioner, x) where {T} = P.preconditi
...
+    # projector onto the computed (converged and unconverged) states
+    P_computed(ϕ) = P(ϕ) + P_free(ϕ)
+    # Q = 1-P is the projector onto the orthogonal of converged states
+    Q(ϕ) = ϕ - P(ϕ)
```
mfherbst, line 144: Yeah I agree with @antoine-levitt  ... we should find better names for these guys once we have the metal case.

src/postprocess/chi0.jl
```diff
@@ -103,33 +103,101 @@ LinearAlgebra.ldiv!(y::T, P::FunctionPreconditioner, x) where {T} = P.preconditi
...
+    # R = 1-P_computed is the projector onto the orthogonal of computed states
+    R(ϕ) = ϕ - P_computed(ϕ)
+
+    # We put things into the form
+    # δψkn = ψk_extra * αkn + δψkn_uncomputed ∈ Ran(Q)
```
mfherbst, line 149: Could use unicode subscripts

src/postprocess/chi0.jl
```diff
@@ -103,33 +103,101 @@ LinearAlgebra.ldiv!(y::T, P::FunctionPreconditioner, x) where {T} = P.preconditi
...
+    #
+    # This can be summarized as the following:
+    #
+    # <---- P ----><------------ Q = 1-P -----------------
+    #              <-- P_free --->
+    # <--------P_computed -------><-- R = 1-P_computed ---
+    # |-----------|--------------|------------------------
+    # 1     N_occupied  N_occupied + N_extra
+
+
```
mfherbst, line 163: remove one empty line

src/postprocess/chi0.jl
```diff
@@ -103,33 +103,101 @@ LinearAlgebra.ldiv!(y::T, P::FunctionPreconditioner, x) where {T} = P.preconditi
...
+    # ψk_extra are not converged but have been Rayleigh-Ritzed (they are NOT
+    # eigenvectors of H) so H(ψk_extra) = ψk_extra'(Hk-εn)ψk_extra should be a
```
mfherbst, line 165: Spaces:
```suggestion
    # eigenvectors of H) so H(ψk_extra) = ψk_extra' (Hk-εn) ψk_extra should be a
```

src/postprocess/chi0.jl
```diff
@@ -103,33 +103,101 @@ LinearAlgebra.ldiv!(y::T, P::FunctionPreconditioner, x) where {T} = P.preconditi
...
+    # real diagonal matrix.
+    H(ϕ) = Hk * ϕ - εnk * ϕ
+    ψk_exHψk_ex = real.(Diagonal(εk_extra .- εnk))
```
mfherbst, line 168: Aargh these names are not so great. But ok, let's bikeshed that later.

mfherbst: Also:
```suggestion
    ψk_exHψk_ex = Diagonal(real.(εk_extra .- εnk))
```

src/postprocess/chi0.jl
```diff
@@ -103,33 +103,101 @@ LinearAlgebra.ldiv!(y::T, P::FunctionPreconditioner, x) where {T} = P.preconditi
...
+    # R * (H - εn) * (1 - M * (H - εn)) * R * δψkn_uncomputed = R * (1 - M) * b
+    #
+    # where M = ψk_extra * (ψk_extra'(H-εn)ψk_extra)^{-1} * ψk_extra'
+    # is defined above and b is the projection of -rhs onto Ran(Q).
+    #
+    b = -Q(rhs)
+    bb = R(b -  H(ψk_extra * (ψk_exHψk_ex \ ψk_extra'b)))
+    function RAR(ϕ)
+        Rϕ = R(ϕ)
+        ARϕ = Rϕ - ψk_extra * (ψk_exHψk_ex \ ψk_extra'H(Rϕ))
```
mfherbst, line 184: What is A?

mfherbst (commented):
Thanks. I also agree that we should bikeshed the names after we know the metals.

Regarding the public interface of `apply_χ0`, I don't like the `ε_extra` kwargs as it breaks with the name for the `eigenvalues`, I'd call it `eigenvalues_extra` for consistency.

src/postprocess/chi0.jl
gkemlin: Isn't `eigenvalues_extra` too long ?

src/postprocess/chi0.jl
gkemlin: > Aargh these names are not so great. But ok, let's bikeshed that later.
Yes, I am not very imaginative when it comes to this...

test/chi0.jl
```diff
@@ -4,92 +4,103 @@ using LinearAlgebra: norm
...
+            D = compute_dos(εF, basis, res.eigenvalues)
+            LDOS = compute_ldos(εF, basis, res.eigenvalues, res.ψ)
         end
 
         if !symmetry
-            # Test compute_χ0 against finite differences
-            χ0 = compute_χ0(ham0)
-            diff_computed_χ0 = reshape(χ0 * vec(δV), basis.fft_size..., n_spin)
-            @test norm(diff_findiff - diff_computed_χ0) < testtol
-
```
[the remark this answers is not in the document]
gkemlin: Yep, I agree. But then we have to find a compromise between small Ecut and small fft_size so that the compuation is not too long and high Ecut so that LOBPCG doesn't crash. 

For the moment, I included this test for small Ecut and fft_size, but replace the LOBPCG by full diagonlization, but then the extra eigenvectors are fully converged.

src/postprocess/chi0.jl
mfherbst: yes, but it's consistent. You have tab completion and all.

src/postprocess/chi0.jl
```diff
@@ -181,23 +176,23 @@ function sternheimer_solver(Hk, ψk, ψnk, εnk, rhs;
...
     end
     precon = PreconditionerTPA(basis, kpoint)
     precondprep!(precon, ψnk)
     function R_ldiv!(x, y)
         x .= R(precon \ R(y))
     end
     J = LinearMap{eltype(ψk)}(RAR, size(Hk, 1))
-    δψkn_uncomputed = cg(J, bb, Pl=FunctionPreconditioner(R_ldiv!),
+    δψknᴿ = cg(J, bb, Pl=FunctionPreconditioner(R_ldiv!),
                          reltol=0, abstol=tol_cg, verbose=verbose)
```
mfherbst, line 189: alignment

src/postprocess/chi0.jl
```diff
@@ -181,23 +176,23 @@ function sternheimer_solver(Hk, ψk, ψnk, εnk, rhs;
     bb = R(b -  H(ψk_extra * (ψk_exHψk_ex \ ψk_extra'b)))
     function RAR(ϕ)
         Rϕ = R(ϕ)
-        ARϕ = Rϕ - ψk_extra * (ψk_exHψk_ex \ ψk_extra'H(Rϕ))
-        R(H(ARϕ))
+        Rϕ .-= ψk_extra * (ψk_exHψk_ex \ ψk_extra'H(Rϕ))
+        R(H(Rϕ))
```
mfherbst, line 180: Isn't this a little misleading? `Rϕ = R(ϕ)` feels right, but with the in-place update that is no longer true. I thought by `A` you mean the Schur complement because then `ARϕ` not a bad variable name, but then you should explain what `A` is.

test/chi0.jl
```diff
@@ -4,92 +4,103 @@ using LinearAlgebra: norm
...
         end
     end
 end
 
-for temperature in (0, 0.03), spin_polarization in (:none, :collinear)
-    for use_symmetry in (false, true), symmetry in (false, true)
-        test_chi0(;symmetry=symmetry, use_symmetry=use_symmetry,
-                  temperature=temperature, spin_polarization=spin_polarization)
+for testcase in (silicon, magnesium)
+    temp = isnothing(testcase.temperature) ? (0, 0.03) : (testcase.temperature)
```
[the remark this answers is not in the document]
mfherbst: Ah yes, of course, stupid me. In that case I would just make it more explicit:
```julia
for (case, temperatures) in [(silicon, (0, 0.03)), (magnesium, (0, ))]
```

test/chi0.jl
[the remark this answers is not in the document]
mfherbst: You can just make a flag to run the `compute_χ0` only in some of the cases, for which you lower `Ecut` and `fft_size` explictly and switch the solver to full (just do this via kwargs to the test function).

The function is pretty simple and not as important as the `apply_χ0`, so you don't need to test it so exhaustively wrt. metals/insulators etc. Just cover a reasonable subset.

test/chi0.jl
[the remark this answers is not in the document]
mfherbst: or maybe skip some of the loops and write out the cases explicitly.

test/chi0.jl
```diff
@@ -90,12 +95,16 @@ function test_chi0(testcase; symmetry=false, use_symmetry=false, temperature=0,
...
         for use_symmetry in (false, true), symmetry in (false, true)
-            test_chi0(testcase; symmetry, use_symmetry,
-                      temperature, spin_polarization)
+            test_chi0(case; symmetry, use_symmetry, temperature, spin_polarization)
+            # additional test for compute_χ0 in some specific cases
+            if iszero(temperature) && !symmetry
+                test_chi0(case; symmetry, use_symmetry, temperature, spin_polarization,
+                          eigensolver=diag_full, Ecut=3, fft_size=[10, 1, 10],
+                          compute_full_χ0=true)
+            end
```
mfherbst, line 107: Just make another for loop block for this, I think that is less confusing.

mfherbst:
Also keep an eye on the failing MPI test.

mfherbst:
(Btw the electron mismatch warnings are artificial and will be fixed in https://github.com/JuliaMolSim/DFTK.jl/pull/596/files)

gkemlin:
> Also keep an eye on the failing MPI test.

Yes, I'm not sure what is failing here, I'll be happy to talk about that.

test/chi0.jl
gkemlin: what do you mean by "another loop block" ? on the flag being true or false ?

mfherbst: what i mean is instead of "filtering out" the cases with `iszero(temperature) && !symmetry`, just have another loop only for the silicon test case where you run the modified `test_chi0` with the `compute_full_χ0=true`.

PR #585, opened 2022-02-11
iFFT on the supercell and plotting wannier functions with paraview

LaurentVidal95:
I added some routines in src/fft.jl and src/external/wannier90.jl to perfom ifft on the supercell and used WriteVTK to plot wannier functions with paraview. This is a first naive implementation but since we talk about it several times I thought the ifft might be usefull for someone else !

### DIFF
--- a/src/external/wannier90.jl
+++ b/src/external/wannier90.jl
@@ -1,4 +1,6 @@
 using Dates
+using DelimitedFiles
+using WriteVTK
 
 """
 Write a win file at the indicated prefix.
@@ -287,3 +289,79 @@ default_wannier_centres(n_wannier) = [rand(1, 3) for _ in 1:n_wannier]
     @timing "Wannierization" wannier90_jll.wannier90(exe -> run(Cmd(`$exe $prefix`; dir)))
     fileprefix
 end
+
+@doc raw"""
+    For a given set of bands ``\{Ψ_{nk}\}_{1\leq n \leq N_in}``
+    and a Wannier90 file containing unitary matrices (for disentanglement or
+    Wannierization) returns a set of band ``{Ψ_out_{nk}}_{1\leq n \leq N_out}``
+    with:
+    ``Ψ_out_{nk} = \sum\limits_{m=1}^{N_in} U_{mn}Ψ_{nk}``
+    ``N_in > N_out`` for disentanglement.
+    ``N_in = N_out`` for wannierization.
+"""
+function apply_U_matrices(file, ψ)
+    # Extract all Uk
+    @assert(isfile(file))
+    Uks = readdlm(file); num_kpts, num_ψout, num_ψin = Uks[2,1:3]
+    len_Uk = num_ψin*num_ψout
+
+    ψ_out = copy(ψ) # Better ideas than copy to inherit the type of ψ in ψ_out ?
+    for k in 1:num_kpts
+        # Locate the beginning of the matrix Uk written as a vector
+        i_Uk = 4 + (k-1)*(len_Uk+1)
+        Uk = reshape([x+im*y for (x,y) in eachrow(Uks[i_Uk:i_Uk+(len_Uk-1), 1:2])],
+                     num_ψin, num_ψout)
+        # Check that Uk is unitary
+        (norm(Uk'Uk - diagm(ones(num_ψout))) > 1e-8) && (@error("U matrix is not unitary"))
+        # Compute all ψ_out[k][:,n]
+        ψ_out_k = zero(ψ[k][:,1:num_ψout])
+        for n in 1:num_ψout
+            ψ_out_k[:,n] = ψ[k][:,1:num_ψin]*Uk[:,n]
+        end
+        ψ_out[k] = ψ_out_k
+    end
+    ψ_out
+end
+
+@doc raw"""
+    Wraps disentanglement (if needed) and wannierization to return Wanniers function
+    as a table ``W[k][G,n]``
+    The prefix in argument is the one returned by the `run_wannier90` routine above.
+    `ψ_unfold` is the array of Bloch waves from an "unfolded" self-consistent-field result.
+"""
+function extract_wannier_functions(prefix, ψ_unfold)
+    # Apply disentanglement if needed
+    if isfile("$(prefix)_u_dis.mat")
+        ψ_unfold = apply_U_matrices("$(prefix)_u_dis.mat", ψ_unfold)
+    end
+    # Construct Wannier functions
+    apply_U_matrices("$(prefix)_u.mat", ψ_unfold)
+end
+
+function extract_wannier_functions(prefix, scfres::NamedTuple)
+    scfres_unfold = unfold_bz(scfres);
+    extract_wannier_functions(prefix, scfres_unfold.ψ)
+end
+
+"""
+   Produce vts files that are readable with Paraview to plot Wannier functions.
+   The wannier functions are given in Bloch decomposition format.
+"""
+function plot_wannier_function(basis::PlaneWaveBasis, w_fourier, n_band)
+    basis_supercell = cell_to_supercell(basis)
+    # Extract real supercell grid
+    r_vec = collect(r_vectors_cart(basis_supercell))
+    x = [r[1] for r in r_vec]; y = [r[2] for r in r_vec]; z = [r[3] for r in r_vec];
+    # Perfom ifft
+    wn_real = G_to_r_supercell(basis, [wk[:, n_band] for wk in w_fourier])
+    vtk_grid("wannier_$(n_band)", x, y, z) do vtk
+        vtk["value"]=real.(wn_real)
+    end
+    return nothing
+end
+
+"""
+    Same as above but only the prefix of wannierization is given
+"""
+plot_wannier_function(scfres::NamedTuple, prefix::String, n_band) =
+    plot_wannier_function(unfold_bz(scfres.basis), extract_wannier_functions(prefix, scfres), n_band)
--- a/src/fft.jl
+++ b/src/fft.jl
@@ -61,6 +61,73 @@ end
 
 
 @doc raw"""
+    Replace 0 by 1 in k_shift to avoid NaN in supercell fft.
+"""
+function clean_shift(basis::PlaneWaveBasis)
+    shift = Vector(basis.kshift)
+    for (i,x) in enumerate(shift)
+        iszero(x) && (shift[i] = one(Float64))
+    end
+    shift
+end
+
+@doc raw"""
+    For a given k point `k` return all `k+G`
+    for `G` in `G_vectors(k)` as a G vector of the 
+    supercell in reduced coordinates
+"""
+function kpG_reduced_supercell(basis::PlaneWaveBasis, kpt::Kpoint)
+    weigths =  basis.kgrid ./ clean_shift(basis)
+    map(Gpk -> round.(Int64, Gpk .* weigths), [G .+ kpt.coordinate
+                                         for G in G_vectors(basis, kpt)])
+end
+
+@doc raw"""
+    Construct a PlaneWaveBasis adapted to fft and ifft in the supercell.
+    Its fft_grid contains all the `k+G` vector of the initial given basis.
+    This amouts to take a single `k` point and multiply each fft_size
+    component by the number of unit cell in the supercell, which is
+    equal to `k_grid` with a correction due to an eventual `k_shift`.
+"""
+function cell_to_supercell(basis::PlaneWaveBasis)
+    # Number of unit cell in the supercell on each axis
+    supercell_size = basis.kgrid ./ clean_shift(basis)
+
+    # New basis and model parameters
+    model = basis.model
+    supercell = ( model.lattice' .* supercell_size )'
+    fft_size_supercell = Int64.(basis.fft_size .* supercell_size)
+
+    # Assemble new model and new basis
+    model_supercell = Model(supercell, atoms=model.atoms)
+    PlaneWaveBasis(model_supercell, basis.Ecut, fft_size_supercell,
+                   basis.variational, [zeros(Int64, 3)],
+                   [[one(SymOp)]], # only works with "unfolded" basis
+                   [1,1,1], # k_grid = Γ point only
+                   zeros(Int64, 3), basis.symmetries, basis.comm_kpts)
+end
+
+"""
+    Convert an array of format `[ψ(1), ψ(2), ... ψ(n_kpts)]` into a 3 dimensional
+    tensor adapted to ifft in the supercell.
+"""
+function cell_to_supercell(basis::PlaneWaveBasis, basis_supercell::PlaneWaveBasis,
+                           ψ_fourier)
+    ψ_fourier_supercell = zeros(ComplexF64, Tuple(basis_supercell.fft_size))
+    for (ik, kpt) in enumerate(basis.kpoints)
+        ikpG = DFTK.index_G_vectors.(basis_supercell, kpG_reduced_supercell(basis, kpt))
+        ψ_fourier_supercell[ikpG] .= ψ_fourier[ik]
+    end
+    ψ_fourier_supercell
+end
+
+function G_to_r_supercell(basis::PlaneWaveBasis, ψ_fourier)
+    basis_supercell = cell_to_supercell(basis)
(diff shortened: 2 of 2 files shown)


antoine-levitt:
Thanks for the PR! I'll take a look. For plots, does https://makie.juliaplots.org/stable/examples/plotting_functions/volume/#example_18412623800959278449 not do what you want?

src/external/wannier90.jl
```diff
@@ -287,3 +289,79 @@ default_wannier_centres(n_wannier) = [rand(1, 3) for _ in 1:n_wannier]
     @timing "Wannierization" wannier90_jll.wannier90(exe -> run(Cmd(`$exe $prefix`; dir)))
     fileprefix
 end
+
+@doc raw"""
+    For a given set of bands ``\{Ψ_{nk}\}_{1\leq n \leq N_in}``
+    and a Wannier90 file containing unitary matrices (for disentanglement or
+    Wannierization) returns a set of band ``{Ψ_out_{nk}}_{1\leq n \leq N_out}``
```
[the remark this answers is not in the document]
antoine-levitt: can you call this function in the tests, just so that it doesn't bitrot?

src/external/wannier90.jl
```diff
@@ -287,3 +289,79 @@ default_wannier_centres(n_wannier) = [rand(1, 3) for _ in 1:n_wannier]
...
+    ``N_in > N_out`` for disentanglement.
+    ``N_in = N_out`` for wannierization.
+"""
+function apply_U_matrices(file, ψ)
+    # Extract all Uk
+    @assert(isfile(file))
+    Uks = readdlm(file); num_kpts, num_ψout, num_ψin = Uks[2,1:3]
+    len_Uk = num_ψin*num_ψout
+
+    ψ_out = copy(ψ) # Better ideas than copy to inherit the type of ψ in ψ_out ?
```
antoine-levitt, line 308: copy is fine

src/external/wannier90.jl
```diff
@@ -287,3 +289,79 @@ default_wannier_centres(n_wannier) = [rand(1, 3) for _ in 1:n_wannier]
...
+    for k in 1:num_kpts
+        # Locate the beginning of the matrix Uk written as a vector
+        i_Uk = 4 + (k-1)*(len_Uk+1)
+        Uk = reshape([x+im*y for (x,y) in eachrow(Uks[i_Uk:i_Uk+(len_Uk-1), 1:2])],
+                     num_ψin, num_ψout)
+        # Check that Uk is unitary
+        (norm(Uk'Uk - diagm(ones(num_ψout))) > 1e-8) && (@error("U matrix is not unitary"))
```
antoine-levitt, line 315: Uk'Uk - I

antoine-levitt: I never remember the difference between error and `@error`, but I think we use `error`? @mfherbst

src/external/wannier90.jl
```diff
@@ -287,3 +289,79 @@ default_wannier_centres(n_wannier) = [rand(1, 3) for _ in 1:n_wannier]
...
+    Uks = readdlm(file); num_kpts, num_ψout, num_ψin = Uks[2,1:3]
+    len_Uk = num_ψin*num_ψout
+
+    ψ_out = copy(ψ) # Better ideas than copy to inherit the type of ψ in ψ_out ?
+    for k in 1:num_kpts
+        # Locate the beginning of the matrix Uk written as a vector
+        i_Uk = 4 + (k-1)*(len_Uk+1)
+        Uk = reshape([x+im*y for (x,y) in eachrow(Uks[i_Uk:i_Uk+(len_Uk-1), 1:2])],
+                     num_ψin, num_ψout)
+        # Check that Uk is unitary
```
antoine-levitt, line 314: unitary only applies to square matrices. People usually write "has orthogonal columns"

antoine-levitt: also comment is redundant with the following

src/external/wannier90.jl
```diff
@@ -287,3 +289,79 @@ default_wannier_centres(n_wannier) = [rand(1, 3) for _ in 1:n_wannier]
...
+        (norm(Uk'Uk - diagm(ones(num_ψout))) > 1e-8) && (@error("U matrix is not unitary"))
+        # Compute all ψ_out[k][:,n]
```
antoine-levitt, line 316: would be cleaner to split into two functions: read_amn and apply_U or something. Also is this not just `psi_out = [psi[ik] * U[ik] for ik=1:Nk]`? If so you maybe don't even need a function for that

src/external/wannier90.jl
```diff
@@ -287,3 +289,79 @@ default_wannier_centres(n_wannier) = [rand(1, 3) for _ in 1:n_wannier]
...
+        for n in 1:num_ψout
+            ψ_out_k[:,n] = ψ[k][:,1:num_ψin]*Uk[:,n]
+        end
+        ψ_out[k] = ψ_out_k
+    end
+    ψ_out
+end
+
+@doc raw"""
+    Wraps disentanglement (if needed) and wannierization to return Wanniers function
```
antoine-levitt, line 327: without the s

src/external/wannier90.jl
```diff
@@ -287,3 +289,79 @@ default_wannier_centres(n_wannier) = [rand(1, 3) for _ in 1:n_wannier]
...
+    as a table ``W[k][G,n]``
+    The prefix in argument is the one returned by the `run_wannier90` routine above.
+    `ψ_unfold` is the array of Bloch waves from an "unfolded" self-consistent-field result.
```
antoine-levitt, line 330: careful that we use unfolded in a (related but) different sense (unfold the symmetries of a BZ). Let's discuss this on slack

src/external/wannier90.jl
```diff
@@ -287,3 +289,79 @@ default_wannier_centres(n_wannier) = [rand(1, 3) for _ in 1:n_wannier]
...
+"""
+function extract_wannier_functions(prefix, ψ_unfold)
```
antoine-levitt, line 332: feels like this is more of an internal thing than a proper function

src/external/wannier90.jl
```diff
@@ -287,3 +289,79 @@ default_wannier_centres(n_wannier) = [rand(1, 3) for _ in 1:n_wannier]
...
+    # Apply disentanglement if needed
+    if isfile("$(prefix)_u_dis.mat")
```
antoine-levitt, line 334: confused by these .mat files, are they wannier90's?

src/external/wannier90.jl
```diff
@@ -287,3 +289,79 @@ default_wannier_centres(n_wannier) = [rand(1, 3) for _ in 1:n_wannier]
...
+    apply_U_matrices("$(prefix)_u.mat", ψ_unfold)
+end
+
+function extract_wannier_functions(prefix, scfres::NamedTuple)
+    scfres_unfold = unfold_bz(scfres);
+    extract_wannier_functions(prefix, scfres_unfold.ψ)
+end
+
+"""
+   Produce vts files that are readable with Paraview to plot Wannier functions.
```
antoine-levitt, line 347: vtk?

src/fft.jl
```diff
@@ -60,6 +60,73 @@ function G_to_r(basis::PlaneWaveBasis, kpt::Kpoint, f_fourier::AbstractVector; k
 end
 
 
+@doc raw"""
+    Replace 0 by 1 in k_shift to avoid NaN in supercell fft.
```
antoine-levitt, line 64: ?

src/fft.jl
```diff
@@ -60,6 +60,73 @@ function G_to_r(basis::PlaneWaveBasis, kpt::Kpoint, f_fourier::AbstractVector; k
...
+function kpG_reduced_supercell(basis::PlaneWaveBasis, kpt::Kpoint)
+    weigths =  basis.kgrid ./ clean_shift(basis)
+    map(Gpk -> round.(Int64, Gpk .* weigths), [G .+ kpt.coordinate
+                                         for G in G_vectors(basis, kpt)])
+end
+
+@doc raw"""
+    Construct a PlaneWaveBasis adapted to fft and ifft in the supercell.
+    Its fft_grid contains all the `k+G` vector of the initial given basis.
+    This amouts to take a single `k` point and multiply each fft_size
```
antoine-levitt, line 88: amounts

src/fft.jl
```diff
@@ -60,6 +60,73 @@ function G_to_r(basis::PlaneWaveBasis, kpt::Kpoint, f_fourier::AbstractVector; k
...
+    component by the number of unit cell in the supercell, which is
+    equal to `k_grid` with a correction due to an eventual `k_shift`.
+"""
+function cell_to_supercell(basis::PlaneWaveBasis)
```
antoine-levitt, line 92: those are generally useful functions, maybe move them to a supercell.jl

src/fft.jl
```diff
@@ -60,6 +60,73 @@ function G_to_r(basis::PlaneWaveBasis, kpt::Kpoint, f_fourier::AbstractVector; k
...
+    # Number of unit cell in the supercell on each axis
+    supercell_size = basis.kgrid ./ clean_shift(basis)
+
+    # New basis and model parameters
+    model = basis.model
+    supercell = ( model.lattice' .* supercell_size )'
+    fft_size_supercell = Int64.(basis.fft_size .* supercell_size)
+
+    # Assemble new model and new basis
+    model_supercell = Model(supercell, atoms=model.atoms)
```
antoine-levitt, line 102: those atoms are now wrong

src/fft.jl
```diff
@@ -60,6 +60,73 @@ function G_to_r(basis::PlaneWaveBasis, kpt::Kpoint, f_fourier::AbstractVector; k
...
+                   [[one(SymOp)]], # only works with "unfolded" basis
+                   [1,1,1], # k_grid = Γ point only
+                   zeros(Int64, 3), basis.symmetries, basis.comm_kpts)
+end
+
+"""
+    Convert an array of format `[ψ(1), ψ(2), ... ψ(n_kpts)]` into a 3 dimensional
+    tensor adapted to ifft in the supercell.
+"""
+function cell_to_supercell(basis::PlaneWaveBasis, basis_supercell::PlaneWaveBasis,
```
antoine-levitt, line 114: why not a cell_to_supercell(scfres)?

antoine-levitt: @mfherbst do we not have routines like this somewhere else?

antoine-levitt (commented):
Very cool! I think the cell_to_supercell thing is a generally super useful thing to have, would be good to have it generally and tested (with a bunch of checks, eg that the energy of scf(basis_supercell) is prod(kgrid) * energy of scf(basis), that the scfres of the supercell matches that of the scf(basis_supercell), etc). Not everything has to work in this PR but it'd be good to have the general structure in place and add to it later.

antoine-levitt:
I could do a full review yet because the coverage thing keeps bugging me and I can't find a way to turn it off during review. I'll do a new pass after this round!

src/external/wannier90.jl
antoine-levitt: ah no I see that you use it in the sense that we do in the symmetry.

src/external/wannier90.jl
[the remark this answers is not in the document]
LaurentVidal95: OK I wasn't so clear ! 
Wannier90 only the U matrices that allow to construct the Wannier functions from the Bloch waves. But it gives two files:
- `prefix_u_dis.mat` to construct a new set of Bloch wave corresponding to disentanglement
- `prefix_u.mat` which construct the MLWFs from the "disentangled" Bloch waves.
This is why I use the routine "apply_U_matrices" two times, one in order to construct the "disentangled" Bloch waves, and the other to construct the Wannier. So I kind of chose to call the first step "disentanglement" and the second step "wannierization".

But I agree that the wannierization is both these steps rather that the second one only. 

We should find a more suitable name.

src/fft.jl
LaurentVidal95: When there is a k_shift 1/2 on one axis, the number of unit cells on this axis in the supercell is multiplied by 2.
So the dimensions of the unit cell are in that direction: `kgrid / k_shift`.

To keep the same formula when there is no k_shift I chose to replace zeros by ones in the shift.
That way I can write: `supercell_dimensions = kgrid ./ k_shift`

I wanted to avoid to treat each axis independantly, but I agree that this trick affects the readability of the code. 
If you have some ideas on how to do that without the `clean_shift` function I will be glad to change that !

LaurentVidal95: Ok I think I will just do a `supercell_size` function which will be more clear !

src/external/wannier90.jl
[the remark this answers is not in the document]
antoine-levitt: is the amn not exactly the Uk? Let's talk tomorrow.

src/external/wannier90.jl
LaurentVidal95: Its a bit tricky because if the bands given to wannier90 were not the full set of bands in psi_in, the U matrices do not have the proper dimension to do psi[ik]*U[ik]
However it can be done and provide the same result by completing U by zeros in the first dimension I think !

src/external/wannier90.jl
LaurentVidal95: Yes I had trouble to find a good name for it. It is unfolded in the sense of symmetries but since we do not have any mention of the "folded" basis in this routine, maybe we can call it `\psi_nosym` or something equivalent ?

src/external/wannier90.jl
mfherbst: Yes, because `@error` does not abort.

src/fft.jl
mfherbst: Yes, we have https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/interpolation_transfer.jl

mfherbst:
Indeed very nice to have the supercell fuctions. And `cell_to_supercell` is not a bad name, I think.

I got a few usual issues with formatting and then also conditional loading of packages. We already have a function to dump VTK stuff, which is conditionally loaded. Could be good to integrate in there.

Also I feel some of what is done here should perhaps not go to DFTK, but to a separate Wannier90.jl package? Of course that can be done at a later point, but just as an idea.

src/fft.jl
LaurentVidal95: Nice ! I'm definitely using those routines !

LaurentVidal95:
Thanks for the review !

> I got a few usual issues with formatting and then also conditional loading of packages. We already have a function to dump VTK stuff, which is conditionally loaded. Could be good to integrate in there.

I've chanced upon these routine during the week end ! I will indeed use those rather than WriteVTK. For that I just have to code a `cell_to_supercell(scfres)` and apply `save_scfres` to the new `scfres`. I'm almost done but I think I have a problem with the Ecut. Shouldn't I take a greater Ecut in the supercell ? I'm afraid to lose some information on the way otherwise. 

> 
> Also I feel some of what is done here should perhaps not go to DFTK, but to a separate Wannier90.jl package? Of course that can be done at a later point, but just as an idea.

I can also do a quick Wannier90.jl package if you fell that's better !

antoine-levitt:
> Shouldn't I take a greater Ecut in the supercell ? I'm afraid to lose some information on the way otherwise.

No particular reason to. The transfer should be lossless (in the sense that it should be theoretically invertible). We can discuss it when you want

> I can also do a quick Wannier90.jl package if you fell that's better !

Might be a good idea in the future. Let's see how things develop; right now it doesn't really hurt to have these routines in DFTK, but if they start to become too big we can split them off to a separate package.

LaurentVidal95:
Ok I pushed a first clean version using more routines that are already implemented in DFTK, tests to cover all functions and (as much as I could) clear documentation. 
EDIT: Sorry I added a correction to the test on ``cell_to_supercell`` where I accidentally broke the ``unfold_bz`` routine by imposing some symmetries.

mfherbst:
Hi @LaurentVidal95. No worries. I've been pretty busy in the past days as well.

It's a bit tricky to review your PR because for some reason github shows not only your changes, but also a lot of other things that happened on master. I don't exactly understand what has been happening here, but in case you remember which files you edited, I think the easiest is to make a fresh branch and put your changes there by moving the files over. Then you can either force-push to your `ifft_supercell` branch or make a new PR.

Let me know, if you need some help on this.

PR #589, opened 2022-02-14
Use symmetrization in density construction

antoine-levitt:
No threads for now

### DIFF
--- a/src/densities.jl
+++ b/src/densities.jl
@@ -1,5 +1,6 @@
 ## Densities (and potentials) are represented by arrays
 ## ρ[ix,iy,iz,iσ] in real space, where iσ ∈ [1:n_spin_components]
+using ForwardDiff
 
 function _check_positive(ρ)
     minimum(ρ) < 0 && @warn("Negative ρ detected", min_ρ=minimum(ρ))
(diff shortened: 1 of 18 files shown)


src/densities.jl
```diff
@@ -24,42 +25,42 @@ grid `basis`, where the individual k-points are occupied according to `occupatio
     ρ = similar(ψ[1], T, (basis.fft_size..., basis.model.n_spin_components))
     ρ .= 0
     ψnk_real = zeros(complex(T), basis.fft_size)
-    for ik = 1:length(basis.kpoints)
-        kpt = basis.kpoints[ik]
+    for (ik, (weight, kpt)) in enumerate(zip(basis.kweights, basis.kpoints))
```
antoine-levitt, line 28: not too sure about that, enumerate(zip) is really the place where I feel it's simpler to just iterate over indices...

src/densities.jl
```diff
@@ -24,42 +25,42 @@ grid `basis`, where the individual k-points are occupied according to `occupatio
...
 # Variation in density corresponding to a variation in the orbitals and occupations.
-@views @timing function compute_δρ(basis::PlaneWaveBasis, ψ, δψ, occupation, δoccupation=zero.(occupation))
-    ForwardDiff.derivative(ε -> compute_density(basis,
-                                                [ψ[ik] .+ ε .* δψ[ik] for ik = 1:length(ψ)],
-                                                [occupation[ik] .+ ε .* δoccupation[ik] for ik = 1:length(ψ)]),
-                           zero(eltype(basis)))
+@views @timing function compute_δρ(basis::PlaneWaveBasis{T}, ψ, δψ,
+                                   occupation, δoccupation=zero.(occupation)) where T
+    ForwardDiff.derivative(zero(T)) do ε
+        ψ_ε   = [ψk   .+ ε .* δψk   for (ψk,   δψk)   in zip(ψ, δψ)]
```
antoine-levitt, line 48: +1 on the `do` thing. The zip... is that really clearer than the shorter version above?

mfherbst: To me yes, I find zips (and enumerate(zip)s) way better than explicit loops over indices. Certainly a personal thing regarding the readability, but the functional style generally makes it also easier for GPU, AD and parallelisation packages, so it's not bad to get used to it.

antoine-levitt: I agree for "large" indices, such as G. For small indices, such as k or n, there's no real difference so it comes to readability. People in scientific computing, for better or worse, are very used to thinking in terms of indices; fancy iterators can sometimes make things more obscure. For this one I don't particularly care but the compute_density above for instance with the enumerate(zip) really feels a bit too clever

mfherbst: hmm ok, fair enough. I'll remove the zip.

Issue #515, opened 2021-08-15
Use symmetrization for forces


antoine-levitt:
Instead of unfolding manually in nonlocal.jl. See https://github.com/JuliaMolSim/DFTK.jl/pull/511

PR #596, opened 2022-02-17
Symmetrize forces

antoine-levitt:
Fix https://github.com/JuliaMolSim/DFTK.jl/issues/515


src/symmetry.jl
```diff
@@ -251,6 +251,26 @@ function symmetrize_stresses(lattice, symmetries, stresses)
...
+    for (iel, (element, positions)) in enumerate(atoms)
+        for symop in symmetries
+            (W, w) = get_Ww(symop)
+            for (iat, at) in enumerate(positions)
+                # see (A.27) of https://arxiv.org/pdf/0906.2569.pdf
+                # (but careful that our symmetries are r -> Wr+w, not R(r+f))
+                other_at = W\(at-w)
+                tol = SYMMETRY_TOLERANCE
+                i_other_at = findfirst(a -> all(isinteger.(rationalize.(a - other_at; tol))),
+                                       positions)
```
mfherbst, line 266:
```suggestion
                i_other_at = findfirst(positions) do a
                    all(isinteger.(rationalize.(a - other_at; tol=SYMMETRY_TOLERANCE)))
                end
```

src/terms/nonlocal.jl
```diff
@@ -54,42 +54,38 @@ end
...
-            mpi_sum!(fr, basis.comm_kpts)  # TODO take that out to gain latency
-            forces[iel][ir] += fr
+        for (ik, kpt) in enumerate(basis.kpoints)
+            # we compute the forces from the irreductible BZ; they are symmetrized later
+            qs_cart = Gplusk_vectors_cart(basis, kpt)
+            qs = Gplusk_vectors(basis, kpt)
+            form_factors = build_form_factors(el.psp, qs_cart)
+            for (ir, r) in enumerate(positions)
+                fr = zeros(T, 3)
+                for α = 1:3
```
mfherbst, line 64:
```suggestion
                fr = map(1:3) do α
```

src/SymOp.jl
```diff
@@ -17,10 +17,9 @@ const SYMMETRY_TOLERANCE = 1e-5
 # Represents a symmetry (S,τ)
 struct SymOp
     S::Mat3{Int}
-    τ::Vec3{Float64}
+    τ::Vec3{Float64}  # floating-point type fixed by spglib
```
mfherbst, line 20: Are you sure we want to do it like this. We have to explicitly convert to the PlaneWaveBasis type very time we apply the symmetries then (think of a Float32 run), which one can easily forget. Either we ensure to convert to the floating-point type of the model in the spglib routines (and use that T here) or we need to explicitly convert every time we use the symmetry operation.

src/SymOp.jl
```diff
@@ -17,10 +17,9 @@ const SYMMETRY_TOLERANCE = 1e-5
...
     function SymOp(S, τ)
-        τ = τ .- floor.(τ)
```
mfherbst, line 22: Also is this not important to "normalise" the τ to [0, 1) (or at least roughtly). I think we need this at places.

src/SymOp.jl
mfherbst: Ok, I check this is actually only at very few places as most routines don't even need `τ`. Maybe the explicit type conversion after using `τ` is in fact the easiest.

src/SymOp.jl
mfherbst: stupid me. That's why there is the mod.

src/SymOp.jl
antoine-levitt: OK done

PR #597, opened 2022-02-17
Simplify the erfc cutoff in ewald

antoine-levitt:


mfherbst (approved):
Nice. I wonder why we did not do this straight away.

antoine-levitt:
It was a long time ago :-)

PR #599, opened 2022-02-21
Support for multiple partials in ForwardDiff SCF

niklasschmitz:
This PR adds support for multiple partials in the ForwardDiff version of `self_consistent_field`, enabling usage of `ForwardDiff.gradient` for vector parameters. For the implicit differentiation, each tangent direction is solved with `solve_ΩplusK` independently.

### DIFF
--- a/src/workarounds/forwarddiff_rules.jl
+++ b/src/workarounds/forwarddiff_rules.jl
@@ -199,9 +199,7 @@ function self_consistent_field(basis_dual::PlaneWaveBasis{T};
                     DT(imag(ψi), imag.(δψi)))
         end
     end
-    ρ_out = map(scfres.ρ, δρ...) do ρi, δρi...
-       DT(ρi, δρi)
-    end
+    ρ_out = map((ρi, δρi...) -> DT(ρi, δρi), scfres.ρ, δρ...)
 
     merge(scfres, (; ham=ham_dual, basis=basis_dual, energies=energies_dual, ψ=ψ_out,
                      occupation=occupation_dual, ρ=ρ_out, eigenvalues=eigenvalues_dual,
--- a/src/workarounds/forwarddiff_rules.jl
+++ b/src/workarounds/forwarddiff_rules.jl
@@ -180,25 +180,24 @@ function self_consistent_field(basis_dual::PlaneWaveBasis{T};
 
     ## Implicit differentiation
     hamψ_dual = ham_dual * ψ_dual
-    N = ForwardDiff.npartials(T)
-    δresults = ntuple(N) do j
-        δHψj = [ForwardDiff.partials.(δHψₖ, j) for δHψₖ in hamψ_dual]
-        δψj, responsej = solve_ΩplusK(basis, ψ, -δHψj, occupation;
+    δresults = ntuple(ForwardDiff.npartials(T)) do α
+        δHψα = [ForwardDiff.partials.(δHψk, α) for δHψk in hamψ_dual]
+        δψα, responseα = solve_ΩplusK(basis, ψ, -δHψα, occupation;
                                       tol_cg=scfres.norm_Δρ, verbose=response.verbose)
-        δρj = compute_δρ(basis, ψ, δψj, occupation)
-        δψj, δρj, responsej
+        δρα = compute_δρ(basis, ψ, δψα, occupation)
+        δψα, δρα, responseα
     end
-    δψ = [δψj for (δψj, δρj, responsej) in δresults]
-    δρ = [δρj for (δψj, δρj, responsej) in δresults]
-    response = [responsej for (δψj, δρj, responsej) in δresults]
+    δψ = [δψα for (δψα, δρα, responseα) in δresults]
+    δρ = [δρα for (δψα, δρα, responseα) in δresults]
+    response = [responseα for (δψα, δρα, responseα) in δresults]
 
     ## Convert, combine and return
     DT = ForwardDiff.Dual{ForwardDiff.tagtype(T)}
-    ψ_out = map(ψ, δψ...) do ψₖ, δψₖ...
-        map(ψₖ, δψₖ...) do ψi, δψi...
+    ψ_out = map(ψ, δψ...) do ψk, δψk...
+        map(ψk, δψk...) do ψi, δψi...
             Complex(
-                DT(real(ψi), map(real, δψi)),
-                DT(imag(ψi), map(imag, δψi)),
+                DT(real(ψi), real.(δψi)),
+                DT(imag(ψi), imag.(δψi)),
             )
         end
     end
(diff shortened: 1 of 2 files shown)


src/workarounds/forwarddiff_rules.jl
```diff
@@ -180,15 +180,31 @@ function self_consistent_field(basis_dual::PlaneWaveBasis{T};
...
+    end
+    δψ = [δψj for (δψj, δρj, responsej) in δresults]
+    δρ = [δρj for (δψj, δρj, responsej) in δresults]
+    response = [responsej for (δψj, δρj, responsej) in δresults]
 
     ## Convert, combine and return
     DT = ForwardDiff.Dual{ForwardDiff.tagtype(T)}
-    ψ_out = [Complex.(DT.(real(ψₖ), real(δψₖ)), DT.(imag(ψₖ), imag(δψₖ))) for (ψₖ, δψₖ) in zip(ψ, δψ)]
-    ρ_out = DT.(scfres.ρ, δρ)
+    ψ_out = map(ψ, δψ...) do ψₖ, δψₖ...
```
antoine-levitt, line 197: I'd rathe we typed the k in plain ASCII (that's what we do in several other places)

src/workarounds/forwarddiff_rules.jl
```diff
@@ -180,15 +180,31 @@ function self_consistent_field(basis_dual::PlaneWaveBasis{T};
 
     ## Implicit differentiation
     hamψ_dual = ham_dual * ψ_dual
-    δHψ = [ForwardDiff.partials.(δHψₖ, 1) for δHψₖ in hamψ_dual]
-    δψ, response = solve_ΩplusK(basis, ψ, -δHψ, occupation;
-                                tol_cg=scfres.norm_Δρ, verbose=response.verbose)
-    δρ  = compute_δρ(basis, ψ, δψ, occupation)
+    N = ForwardDiff.npartials(T)
+    δresults = ntuple(N) do j
```
antoine-levitt, line 184: would be good to use more distinctive indices, since this can get confusing. How about Npartials and \alpha?

src/workarounds/forwarddiff_rules.jl
```diff
@@ -180,15 +180,31 @@ function self_consistent_field(basis_dual::PlaneWaveBasis{T};
...
+    response = [responsej for (δψj, δρj, responsej) in δresults]
 
     ## Convert, combine and return
     DT = ForwardDiff.Dual{ForwardDiff.tagtype(T)}
-    ψ_out = [Complex.(DT.(real(ψₖ), real(δψₖ)), DT.(imag(ψₖ), imag(δψₖ))) for (ψₖ, δψₖ) in zip(ψ, δψ)]
-    ρ_out = DT.(scfres.ρ, δρ)
+    ψ_out = map(ψ, δψ...) do ψₖ, δψₖ...
+        map(ψₖ, δψₖ...) do ψi, δψi...
+            Complex(
+                DT(real(ψi), map(real, δψi)),
```
antoine-levitt, line 200: for this we tend to use `real.()`

antoine-levitt, line 200: for this we tend to use `real.()`

src/workarounds/forwarddiff_rules.jl
mfherbst: Acutally ... I'm not sure that's such a bad idea. I find the subscript k rather readable.

src/workarounds/forwarddiff_rules.jl
```diff
@@ -180,15 +180,30 @@ function self_consistent_field(basis_dual::PlaneWaveBasis{T};
...
-    δHψ = [ForwardDiff.partials.(δHψₖ, 1) for δHψₖ in hamψ_dual]
-    δψ, response = solve_ΩplusK(basis, ψ, -δHψ, occupation;
-                                tol_cg=scfres.norm_Δρ, verbose=response.verbose)
-    δρ  = compute_δρ(basis, ψ, δψ, occupation)
+    δresults = ntuple(ForwardDiff.npartials(T)) do α
+        δHψα = [ForwardDiff.partials.(δHψk, α) for δHψk in hamψ_dual]
+        δψα, responseα = solve_ΩplusK(basis, ψ, -δHψα, occupation;
+                                      tol_cg=scfres.norm_Δρ, verbose=response.verbose)
+        δρα = compute_δρ(basis, ψ, δψα, occupation)
+        δψα, δρα, responseα
```
mfherbst, line 188: Maybe add an underscore before the α to make it  more readable?

src/workarounds/forwarddiff_rules.jl
```diff
@@ -180,15 +180,30 @@ function self_consistent_field(basis_dual::PlaneWaveBasis{T};
...
+    end
+    δψ = [δψα for (δψα, δρα, responseα) in δresults]
+    δρ = [δρα for (δψα, δρα, responseα) in δresults]
+    response = [responseα for (δψα, δρα, responseα) in δresults]
```
mfherbst, line 192: = / for / in alignment helps readability here

src/workarounds/forwarddiff_rules.jl
```diff
@@ -180,15 +180,30 @@ function self_consistent_field(basis_dual::PlaneWaveBasis{T};
...
     ## Convert, combine and return
     DT = ForwardDiff.Dual{ForwardDiff.tagtype(T)}
-    ψ_out = [Complex.(DT.(real(ψₖ), real(δψₖ)), DT.(imag(ψₖ), imag(δψₖ))) for (ψₖ, δψₖ) in zip(ψ, δψ)]
-    ρ_out = DT.(scfres.ρ, δρ)
+    ψ_out = map(ψ, δψ...) do ψk, δψk...
+        map(ψk, δψk...) do ψi, δψi...
+            Complex(
+                DT(real(ψi), real.(δψi)),
+                DT(imag(ψi), imag.(δψi)),
+            )
```
mfherbst, line 201: A bit less spaceous:
```suggestion
            Complex(DT(real(ψi), real.(δψi)),
                    DT(imag(ψi), imag.(δψi)))
```

src/workarounds/forwarddiff_rules.jl
antoine-levitt: I don't really like unicode subscripts. It's rather gimmicky, and some fonts just don't support it. The place where I agree it makes sense is for longer subscripts, like x_{n+1}. Here ψk is not really worse than ψₖ... Also if we want to do it we should do it everywhere

src/workarounds/forwarddiff_rules.jl
```diff
@@ -180,15 +180,30 @@ function self_consistent_field(basis_dual::PlaneWaveBasis{T};
...
+        end
+    end
+    ρ_out = map(scfres.ρ, δρ...) do ρi, δρi...
+       DT(ρi, δρi)
+    end
```
mfherbst, line 206: Maybe in one line ?
```suggestion
    ρ_out = map((ρi, δρi...) -> DT(ρi, δρi), scfres.ρ, δρ...)
```

src/workarounds/forwarddiff_rules.jl
mfherbst: yes sure, consistency is clear.

mfherbst: ok let's leave it for now

PR #601, opened 2022-02-21
Using pseudo-inverse for one- and two-dimensional cells

epolack:
If the lattice has rank smaller than three, then the routine fails.

### DIFF
--- a/src/postprocess/forces.jl
+++ b/src/postprocess/forces.jl
@@ -28,7 +28,7 @@ which has the same structure as the `atoms` object passed to the underlying [`Mo
 function compute_forces_cart(basis::PlaneWaveBasis, ψ, occ; kwargs...)
     lattice = basis.model.lattice
     forces = compute_forces(basis::PlaneWaveBasis, ψ, occ; kwargs...)
-    [[lattice \ f for f in forces_for_element] for forces_for_element in forces]
+    [[pinv(lattice)*f for f in forces_for_element] for forces_for_element in forces]
 end
 
 function compute_forces(scfres)
--- a/src/postprocess/forces.jl
+++ b/src/postprocess/forces.jl
@@ -28,7 +28,8 @@ which has the same structure as the `atoms` object passed to the underlying [`Mo
 function compute_forces_cart(basis::PlaneWaveBasis, ψ, occ; kwargs...)
     lattice = basis.model.lattice
     forces = compute_forces(basis::PlaneWaveBasis, ψ, occ; kwargs...)
-    [[pinv(lattice)*f for f in forces_for_element] for forces_for_element in forces]
+    pinv_lattice = pinv(lattice)
+    [[pinv_lattice * f for f in forces_for_element] for forces_for_element in forces]
 end
 
 function compute_forces(scfres)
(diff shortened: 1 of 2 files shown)


src/postprocess/forces.jl
```diff
@@ -28,7 +28,7 @@ which has the same structure as the `atoms` object passed to the underlying [`Mo
 function compute_forces_cart(basis::PlaneWaveBasis, ψ, occ; kwargs...)
     lattice = basis.model.lattice
     forces = compute_forces(basis::PlaneWaveBasis, ψ, occ; kwargs...)
-    [[lattice \ f for f in forces_for_element] for forces_for_element in forces]
+    [[pinv(lattice)*f for f in forces_for_element] for forces_for_element in forces]
```
antoine-levitt, line 31: michael will say spaces :-p Also precompute the pinv

mfherbst, line 31: Does this work with forward diff? See also the comment in 
https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/structure.jl#L6--L19

antoine-levitt (commented):
A bit of a hack but let's do it like this. We don't really use it at other places yet... A better solution could be to have a function to_cart or to_reduced that would do this, but it's a bit ambiguous (because vectors can be covariant or contravariant)

src/postprocess/forces.jl
antoine-levitt: OK good point; then just refactor to have a compute_inverse_lattice function and call it

src/structure.jl
```diff
@@ -1,24 +1,32 @@
...
 
 
+"""
+Compute the reciprocal lattice.
+We use the convention that the reciprocal lattice is the set of G vectors such
+that G ⋅ R ∈ 2π ℤ for all R in the lattice.
+"""
+function compute_recip_lattice(lattice::AbstractMatrix{T}) where {T}
+    2T(π) * compute_inverse_lattice(lattice)
+end
```
mfherbst, line 27: T not needed, could make this a one-liner.

epolack: I am not sure how you write one-liners.
Btw, with the use of pseudo-inverse I do not seem to obtain the correct forces anymore. I will invistigate a bit more.

mfherbst: Ah no sorry, what I had in mind does not work, because you need the `T`

src/structure.jl
```diff
@@ -1,24 +1,32 @@
...
 """
-function compute_recip_lattice(lattice::AbstractMatrix{T}) where {T}
+function compute_inverse_lattice(lattice::AbstractMatrix{T}) where {T}
     # Note: pinv pretty much does the same, but the implied SVD causes trouble
     #       with interval arithmetic and dual numbers, so we go for this version.
     n_dim = count(!iszero, eachcol(lattice))
     @assert 1 ≤ n_dim ≤ 3
     if n_dim == 3
-        2T(π) * inv(lattice')
+        inv(lattice')
```
antoine-levitt, line 10: this is not the inverse lattice, it's the inverse transpose lattice. Don't transpose here, and transpose in compute_recip_lattice

PR #603, opened 2022-02-23
Fix stress symmetrization

antoine-levitt:
Based on @jaemolihm 's PR. Also introduce a couple of helper functions to hopefully get less confused between the different conversions


src/Model.jl
```diff
@@ -216,4 +226,37 @@ end
...
+
+
+#=
+There are two types of quantities, depending on how they transform under change of coordinates.
+
+Positions transform with the lattice: r_cart = lattice * r_red. We term them vectors.
+
+Linear forms on vectors (anything that appears in an expression f⋅r)
+transform with the inverse lattice transpose: f_cart = lattice' \ f_red. We term them covectors.
+Examples of covectors are forces.
```
mfherbst, line 241: I think it's a good idea to add the one line of arguments I've put into the forces function here.

src/Model.jl
```diff
@@ -216,4 +226,37 @@ end
...
+
+Reciprocal vectors are a special case: they are covectors, but conventionally have an
+additional factor of 2π in their definition, so they transform rather with 2π times the
+inverse lattice transpose: q_cart = 2π lattice' \ q_red = recip_lattice * q_red.
+=#
+vector_red_to_cart(model::Model, rred) = model.lattice * rred
+vector_cart_to_red(model::Model, rcart) = model.inv_lattice * rcart
+covector_red_to_cart(model::Model, fred) = model.inv_lattice' * fred
+covector_cart_to_red(model::Model, fcart) = model.lattice' * fcart
```
mfherbst, line 250: Equals alignment.

src/Model.jl
```diff
@@ -216,4 +226,37 @@ end
...
+Linear forms on vectors (anything that appears in an expression f⋅r)
+transform with the inverse lattice transpose: f_cart = lattice' \ f_red. We term them covectors.
+Examples of covectors are forces.
+
+Reciprocal vectors are a special case: they are covectors, but conventionally have an
+additional factor of 2π in their definition, so they transform rather with 2π times the
+inverse lattice transpose: q_cart = 2π lattice' \ q_red = recip_lattice * q_red.
+=#
+vector_red_to_cart(model::Model, rred) = model.lattice * rred
+vector_cart_to_red(model::Model, rcart) = model.inv_lattice * rcart
```
mfherbst, line 248: Should add quick tests, just to ensure forward and backward gives an identity.

src/Model.jl
```diff
@@ -216,4 +226,37 @@ end
...
+
+#=
+Transformations on vectors and covectors are matrices and comatrices. 
+Examples of matrices are the symmetries in real space (W)
+Examples of comatrices are the symmetries in reciprocal space (S)
+=#
+matrix_red_to_cart(model::Model, Ared) = model.lattice * Ared * model.inv_lattice
+matrix_cart_to_red(model::Model, Acart) = model.inv_lattice * Acart * model.lattice
+comatrix_red_to_cart(model::Model, Bred) = model.inv_lattice' * Bred * model.lattice'
+comatrix_cart_to_red(model::Model, Bcart) = model.lattice' * Bcart * model.inv_lattice'
```
mfherbst, line 262: Maybe for consistency we should also have a `recip_matrix_cart_to_red` ? (Of course that would just call the `comatrix` function)

src/symmetry.jl
```diff
@@ -243,19 +243,22 @@ Symmetrize a density by applying all the basis (by default) symmetries and formi
...
+# symmetrize the stress tensor, a mtrix in cartesian coordinates
+function symmetrize_stresses(model::Model, stresses, symmetries)
+    # see (A.28) of https://arxiv.org/pdf/0906.2569.pdf
     stresses_symmetrized = zero(stresses)
     for symop in symmetries
-        S_reduced = inv(lattice) * symop.S * lattice
-        stresses_symmetrized += S_reduced' * stresses * S_reduced
+        W_red, _ = get_Ww(symop)
+        W = matrix_red_to_cart(model, W_red)
+        stresses_symmetrized += W * stresses / W
```
mfherbst, line 253: I'd rather go for consistency for:
```suggestion
        W, _ = get_Ww(symop)
        W_cart = matrix_red_to_cart(model, W)
        stresses_symmetrized += W_cart * stresses / W_cart
```

src/PlaneWaveBasis.jl
```diff
@@ -359,11 +359,9 @@ G_vectors(::PlaneWaveBasis, kpt::Kpoint) = kpt.G_vectors
...
 The list of ``G`` vectors of a given `basis` or `kpt`, in cartesian coordinates.
 """
-function G_vectors_cart(basis::PlaneWaveBasis)
-    map(G -> basis.model.recip_lattice * G, G_vectors(basis))
-end
+G_vectors_cart(basis::PlaneWaveBasis) = recip_vector_red_to_cart.(basis.model, G_vectors(basis))
 function G_vectors_cart(basis::PlaneWaveBasis, kpt::Kpoint)
-    map(G -> basis.model.recip_lattice * G, G_vectors(basis, kpt))
+    recip_vector_red_to_cart.(basis.model, G_vectors(basis, kpt))
 end
```
mfherbst, line 365: Let's hope this won't become too much of a hassle with the autodiff.

src/Model.jl
antoine-levitt: good idea

src/Model.jl
antoine-levitt: that kind of tests don't actually test anything, because I know they are correct now, we're not likely to break it in future refactoring, and it's tested indirectly by other tests. Also tests are code and as such have a nontrivial maintenance cost that I think it's wise to avoid for this type of things

src/Model.jl
antoine-levitt: recip_matrix is weird. This routine is used for transformations of (real and reciprocal) space, so when using that it's good to think of reciprocal space vectors as covectors, and so having the same routine helps

src/PlaneWaveBasis.jl
antoine-levitt: no particular reason why it should: it's exactly equivalent to the previous version

src/Model.jl
mfherbst: well it's not tested. That's the point. It's not covered.

jaemolihm:
I had a look on the symmetrization part and it looks good. I also like the new `red_to_cart` and `cart_to_red` routines! Thanks @antoine-levitt !

antoine-levitt:
Cool, thanks! I was not too sure about the vector / covector terminology, but it seems clearer to me than the covariant / contravariant one (because I can never remember which is which, and because it seems to come with a bagage which is a bit different from our context)

PR #604, opened 2022-02-23
Add explicit error if SCF is used with no kinetic term

epolack:


antoine-levitt:
Technically it's OK to do it (just the eigensolver will do weird stuff), it's just the preconditioner that needs it, so put that check there probably?

epolack:
With `diag_full`, even with `scaling` arbitrarily put to `1`, it still crashes at line `Afull = Hermitian(Array(A))`.

antoine-levitt:
You mean if you've got no hamiltonian contribution at all? OK yeah but in that case you deserve what happens to you :-p I don't think we should have a check for that level of craziness. So let's just have a check that there is indeed a kinetic term in the preconditioner, and then we just have a crash when there's no hamiltonian at all

src/eigen/preconditioners.jl
```diff
@@ -33,7 +33,13 @@ mutable struct PreconditionerTPA{T <: Real}
...
 
 function PreconditionerTPA(basis::PlaneWaveBasis{T}, kpt::Kpoint; default_shift=1) where T
-    scaling = only([t for t in basis.model.term_types if t isa Kinetic]).scaling_factor
+    kinetic_term = [t for t in basis.model.term_types if t isa Kinetic]
+    if length(kinetic_term) >= 1
+        scaling = only(kinetic_term).scaling_factor
+    else
+        # If user do strange computations without a kinetic term.
+        scaling = 1
+    end
```
antoine-levitt, line 42: This is a type instability. But here just raise an error; the preconditioner is not relevant in this case, users should disable preconditioning. So just `kinetic_terms = []; length() >= 1 || error(); scaling = only(kinetic_terms).scaling`

antoine-levitt (commented):
no linebreak and I'm happy :-)

src/eigen/preconditioners.jl
```diff
@@ -33,7 +33,9 @@ mutable struct PreconditionerTPA{T <: Real}
 end
 
 function PreconditionerTPA(basis::PlaneWaveBasis{T}, kpt::Kpoint; default_shift=1) where T
-    scaling = only([t for t in basis.model.term_types if t isa Kinetic]).scaling_factor
+    kinetic_term = [t for t in basis.model.term_types if t isa Kinetic]
+    length(kinetic_term) >= 1 || error("Preconditioner should be disabled when no kinetic term is used.")
```
mfherbst, line 37:
```suggestion
    isempty(kinetic_term) && error("Preconditioner should be disabled when no Kinetic term is used.")
```

mfherbst (commented):
To make me happy :)

Issue #579, opened 2022-02-03
Convergence monitoring


antoine-levitt:
When eyeballing convergence, the usual thing is to look at a string of numbers like "1.24e-4" and trying to decypher from it whether it converges or not. It's pretty annoying to try to guess patterns from these numbers. Much more useful would be the log (and this is indeed how we plot them). In scripts I've taken to printing them as "1e" * log10(err), resulting in things like 1e-3.72. Then it becomes much easier to spot patterns. I feel it's also pretty readable for people who don't know about this convention, and am edging towards thinking this should be the universal thing to do for monitoring convergence. Maybe we should do that, eg for the default SCF convergence monitoring?

mfherbst:
Sounds like a good idea to me, actually. Feel free to add that to the default callback.

antoine-levitt:
```
n     Energy            Eₙ-Eₙ₋₁     ρout-ρin   α      Diag
---   ---------------   ---------   --------   ----   ----
  1   -7.926703358142               1e-00.70   0.80    4.4
  2   -7.931684890245   -1e-02.30   1e-01.52   0.80    1.0
  3   -7.931992462663   -1e-03.51   1e-02.43   0.80    3.2
  4   -7.931993443241   -1e-06.01   1e-03.24   0.80    2.7
  5   -7.931993461920   -1e-07.73   1e-04.48   0.80    2.1
  6   -7.931993463234   -1e-08.88   1e-05.34   0.80    3.3
  7   -7.931993463237   -1e-11.51   1e-06.40   0.80    2.2
  8   -7.931993463237   -1e-13.30   1e-07.14   0.80    2.4
  9   -7.931993463237    1e-14.75   1e-08.12   0.80    2.2
```
Do we like that? The 1e are a bit annoying. Alternative is printing log10(err) directly. That makes it harder to signal the sign of the dE though.

PR #606, opened 2022-02-28
Change printing

antoine-levitt:
Fix https://github.com/JuliaMolSim/DFTK.jl/issues/579

### DIFF
--- a/src/scf/scf_callbacks.jl
+++ b/src/scf/scf_callbacks.jl
@@ -40,28 +40,30 @@ function ScfDefaultCallback()
 
         # TODO We should really do this properly ... this is really messy
         if info.n_iter == 1
-            E_label = haskey(info.energies, "Entropy") ? "Free energy" : "Energy"
             label_magn = show_magn ? ("   Magnet", "   ------") : ("", "")
             label_damp = show_damp ? ("   α   ", "   ----") : ("", "")
             label_diag = show_diag ? ("   Diag", "   ----") : ("", "")
-            @printf "n     %-12s      Eₙ-Eₙ₋₁     ρout-ρin" E_label
+            @printf " n         Energy       log10(ΔE)   log10(Δρ)"
             println(label_magn[1], label_damp[1], label_diag[1])
-            @printf "---   ---------------   ---------   --------"
+            @printf "---   ---------------   ---------   ---------"
             println(label_magn[2], label_damp[2], label_diag[2])
         end
         E    = isnothing(info.energies) ? Inf : info.energies.total
         Δρ   = norm(info.ρout - info.ρin) * sqrt(info.basis.dvol)
         magn = sum(spin_density(info.ρout)) * info.basis.dvol
 
+        format_log(e) = @sprintf "%9.2f" log10(abs(e))
+
         Estr    = (@sprintf "%+15.12f" round(E, sigdigits=13))[1:15]
-        ΔE      = isnan(prev_energy) ? "      NaN" : @sprintf "% 3.2e" E - prev_energy
+        ΔE      = isnan(prev_energy) ? "         " : format_log(E - prev_energy)
+        Δρstr   = format_log(Δρ)
         Mstr    = show_magn ? "   $((@sprintf "%6.3f" round(magn, sigdigits=4))[1:6])" : ""
         diagstr = show_diag ? "  $(@sprintf "% 5.1f" diagiter)" : ""
 
         αstr = ""
         show_damp && (αstr = isnan(info.α) ? "    NaN" : @sprintf "  % 4.2f" info.α)
 
-        @printf "% 3d   %s   %s   %2.2e" info.n_iter Estr ΔE Δρ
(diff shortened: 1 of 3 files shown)


src/scf/scf_callbacks.jl
```diff
@@ -40,28 +40,30 @@ function ScfDefaultCallback()
...
         if info.n_iter == 1
-            E_label = haskey(info.energies, "Entropy") ? "Free energy" : "Energy"
             label_magn = show_magn ? ("   Magnet", "   ------") : ("", "")
             label_damp = show_damp ? ("   α   ", "   ----") : ("", "")
             label_diag = show_diag ? ("   Diag", "   ----") : ("", "")
-            @printf "n     %-12s      Eₙ-Eₙ₋₁     ρout-ρin" E_label
+            @printf " n         Energy       log10(ΔE)   log10(Δρ)"
             println(label_magn[1], label_damp[1], label_diag[1])
-            @printf "---   ---------------   ---------   --------"
+            @printf "---   ---------------   ---------   ---------"
```
mfherbst, line 48: I prefer the heading to be uniformly left-aligned rather than centred but not really.

src/scf/scf_callbacks.jl
```diff
@@ -40,28 +40,30 @@ function ScfDefaultCallback()
...
             println(label_magn[2], label_damp[2], label_diag[2])
         end
         E    = isnothing(info.energies) ? Inf : info.energies.total
         Δρ   = norm(info.ρout - info.ρin) * sqrt(info.basis.dvol)
         magn = sum(spin_density(info.ρout)) * info.basis.dvol
 
+        format_log(e) = @sprintf "%9.2f" log10(abs(e))
```
mfherbst, line 55: Especially for the energy I am not convinced here: How do you detect that the energy increases? I think it is pretty important so see this easily the steps where the energy goes up and where down.

src/scf/scf_callbacks.jl
antoine-levitt: We can't left align the labels and right align the numbers

src/scf/self_consistent_field.jl
```diff
@@ -101,7 +101,7 @@ Solve the Kohn-Sham equations with a SCF algorithm, starting at ρ.
 
         # Update info with results gathered so far
         info = (ham=ham, basis=basis, converged=converged, stage=:iterate, algorithm="SCF",
-                ρin=ρin, ρout=ρout, α=damping, n_iter=n_iter, n_ep_extra=n_ep_extra,
+                ρin=ρin, ρout=ρout, n_iter=n_iter, n_ep_extra=n_ep_extra,
```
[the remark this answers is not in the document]
antoine-levitt: Why? It's just noise.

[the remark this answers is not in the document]
mfherbst: sure, but I don't have to check in a callback function whether the α is set or not ...

src/scf/scf_callbacks.jl
mfherbst: For me this is very crucial. Similar to the order of magnitude of decrease, it's the first thing I check.

mfherbst: what about the 1e-5.7 thingy

src/scf/self_consistent_field.jl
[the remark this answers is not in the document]
antoine-levitt: OK, I see what you mean, will work around

src/scf/scf_callbacks.jl
antoine-levitt: It's actually super ugly, I'd rather leave it as we have now

src/scf/scf_callbacks.jl
mfherbst: Why not? Looks ok: https://github.com/JuliaMolSim/DFTK.jl/runs/5334962739?check_suite_focus=true#step:8:113

src/scf/scf_callbacks.jl
antoine-levitt: doesn't really have to be intuitive: you can just ignore it if you're not interested

mfherbst: You would be suprised what can put people off a code. For sure if they don't understand the convergence output on a first look.

antoine-levitt: Well the energy is there plain, and the heading is correct: it's just that there is possibly one extra character of information. Feels OK to me

mfherbst: yeah sure, do it and we see what it looks :smile:.

antoine-levitt:
I think this looks good. It's sufficiently infrequent, and when it happens at the end it's a good visual sign that noise is coming in

mfherbst:
I agree. Looks good!

src/scf/self_consistent_field.jl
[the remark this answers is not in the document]
mfherbst: I patched that up a little.

PR #607, opened 2022-02-28
Switch maximum iteration count to 100

antoine-levitt:
If it doesn't converge in 50 iterations, it's not going to converge in 200


mfherbst:
100 and I'm in. 50 is too little sometimes.

antoine-levitt:
Really? If 50 is too low then probably the toleranecs are much too low or something is going wrong

mfherbst:
Yes our tolerance selection is not yet great, I agree with that. Let's just go with 100 and see what happens. With 50 I'm simply uncomfortable with respect to practical calculations for now.

antoine-levitt:
OK

Issue #529, opened 2021-10-06
Segfault in preconditioning


antoine-levitt:
Just remembered there was a problem with this and wanted to create an issue so we don't forget. @gkemlin can you post more details?

mfherbst:
Ok ... interesting. In which preconditioner? SCF or diagonalisation?

antoine-levitt:
Is this with MPI or not? Does adding GC.gc() increase the frequency of the bug?

mfherbst:
Since @gkemlin uses threads, he does not use MPI.

But as a side note: You should use MPI. With threads the scaling in DFTK is very bad at the moment. I would bet your calculations are actually even *faster* if you only use 8 threads instead of 32.

gkemlin:
Is juliampi install on the cluster ? This is why I used Threads ^^" I'm not aware about the scaling though, I can change that.

@antoine-levitt I try with GC.gc() and I tell you.

PR #608, opened 2022-02-28
Fix unsafe use of unpack

antoine-levitt:
Hopefully fix https://github.com/JuliaMolSim/DFTK.jl/issues/529, CC @gkemlin

### DIFF
--- a/src/orbitals.jl
+++ b/src/orbitals.jl
@@ -39,11 +39,12 @@ function pack_ψ(ψ)
     vcat([vec(ψk) for ψk in ψ]...)
 end
 
+# Returns pointers into the unpacked ψ
+# /!\ The resulting array is only valid as long as the original x is still in live memory.
 function unpack_ψ(x, sizes_ψ)
     lengths = prod.(sizes_ψ)
     ends = cumsum(lengths)
-    # We unsafe_wrap the resulting array to avoid a complicated type for ψ.
-    # The resulting array is valid as long as the original x is still in live memory.
+    # We unsafe_wrap the resulting array to avoid a complicated type for ψ.    
     map(1:length(sizes_ψ)) do ik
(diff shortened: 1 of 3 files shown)


src/orbitals.jl
```diff
@@ -39,14 +39,16 @@ function pack_ψ(ψ)
...
-    # We unsafe_wrap the resulting array to avoid a complicated type for ψ.
-    # The resulting array is valid as long as the original x is still in live memory.
+    # We unsafe_wrap the resulting array to avoid a complicated type for ψ.    
     map(1:length(sizes_ψ)) do ik
         unsafe_wrap(Array{complex(eltype(x))},
                     pointer(@views x[ends[ik]-lengths[ik]+1:ends[ik]]),
                     sizes_ψ[ik])
     end
 end
+unpack_ψ(x, sizes_ψ) = deepcopy(unsafe_unpack_ψ, sizes_ψ)
```
mfherbst, line 54:
```suggestion
unpack_ψ(x, sizes_ψ) = deepcopy(unsafe_unpack_ψ(x, sizes_ψ))
```

antoine-levitt: duh

gkemlin:
Given the results of tests, I would go for `unsafe_unpack_ψ` here too ? https://github.com/JuliaMolSim/DFTK.jl/blob/8eadc5556bd44938baa9116b9f14fcf86facbe5f/src/scf/direct_minimization.jl#L92

I have launched the tests that were yielding the segfault, I keep you update.

gkemlin:
Well the working example from https://github.com/JuliaMolSim/DFTK.jl/issues/529 seems to work now, without raising segfault.

antoine-levitt:
Yeah, the previous version was just wrong, I don't know what we were thinking at the time

PR #613, opened 2022-03-03
Remove coordinate_cart field from Kpoint

niklasschmitz:
This PR removes the `coordinate_cart` field from `Kpoint`, which is only used in band structure. For our AD work this will mean that we can conveniently mark `build_kpoints` as non-differentiable altogether.


antoine-levitt:
You should change the constructor also but otherwise looks good!

PR #615, opened 2022-03-04
Small optimizations

antoine-levitt:


src/fft.jl
```diff
@@ -66,7 +66,7 @@ In-place version of `r_to_G!`.
 NOTE: If `kpt` is given, not only `f_fourier` but also `f_real` is overwritten.
 """
 function r_to_G!(f_fourier::AbstractArray3, basis::PlaneWaveBasis, f_real::AbstractArray3)
-    if isreal(f_real)
+    if eltype(f_real) <: Real
         f_real = complex.(f_real)
     end
```
mfherbst, line 71: Could catch this via dispatch.

antoine-levitt: We already have way too many dispatches for these functions, this is simpler

mfherbst: not sure I agree you can ever have too many dispatches, but I won't fight over it.

PR #619, opened 2022-03-04
Symop refactoring (again!)

antoine-levitt:


antoine-levitt:
TODO: fix probably failing tests, and possibly be careful about normalization (we normalize w now, which is possibly a bit different than before)

antoine-levitt:
Hm this is failing in a weird way : the stresses are nans before symmetrization, and I don't see what could have changed before then. That will have to wait...

antoine-levitt:
This is super weird. The bug only appears when the following three ingredients are present : stresses with forwarddiff, symmetries and PBE

antoine-levitt:
Got it, that was a fun one: somehow the exact density has a point where sigma is zero, which messes up the differentiation of energy_per_particle_x_pbe (which takes the sqrt of sigma). This is spurious because it is in fact differentiable (we square the sqrt of sigma immediately after). That's one of the not-so-niceties about forwarddiff unfortunately... (we had a similar issue with the Fermi-Dirac function) Fortunately the fix here is trivial.

src/SymOp.jl
```diff
@@ -11,44 +11,45 @@
...
+# Time-reversal symmetries are the anti-unitaries
+# (Uu)(x) = conj(u(Wx+w))
+# or in Fourier space
+# (Uu)(G) = e^{i G τ} conj(u(-S^-1 G))
+
 # Tolerance to consider two atomic positions as equal (in relative coordinates)
 const SYMMETRY_TOLERANCE = 1e-5
 
-# Represents a symmetry (S,τ)
+# Represents a symmetry (W,w)
```
mfherbst, line 22: add: in real space.

src/SymOp.jl
```diff
@@ -11,44 +11,45 @@
...
 # group composition and inverse.
-# Derived either from the formulas for the composition/inverse of W/w
-# then passing to reciprocal, or directly from the symmetry operation in reciprocal space
 function Base.:*(op1, op2)
-    S = op1.S * op2.S
-    τ = op1.τ + op1.S' \ op2.τ
-    SymOp(S, τ)
+    W = op1.W * op2.W
+    w = op1.w + op1.W * op2.w
+    SymOp(W, w)
```
mfherbst, line 50: Make this a one-liner, too?

src/symmetry.jl
```diff
@@ -50,7 +50,7 @@ function symmetry_operations(lattice, atoms, magnetic_moments=[]; tol_symmetry=S
     for isym = 1:length(Ws)
         S = Ws[isym]'                  # in fractional reciprocal coordinates
         τ = -Ws[isym] \ ws[isym]  # in fractional real-space coordinates
-        push!(symmetries, SymOp(S, τ))
+        push!(symmetries, SymOp(Ws[isym], ws[isym]))
     end
```
mfherbst, line 54: could be simplified into a list comprehension

src/symmetry.jl
```diff
@@ -270,7 +267,7 @@ function symmetrize_forces(model::Model, forces; symmetries)
     symmetrized_forces = zero.(forces)
     for (iel, (element, positions)) in enumerate(atoms)
         for symop in symmetries
-            W, w = get_Ww(symop)
+            W = symop.W; w = symop.w
```
mfherbst, line 270: Don't like `;`, better:
```suggestion
            W, w = symop.W, symop.w
```

src/xc/gga_x_pbe.jl
```diff
@@ -51,10 +51,11 @@ function energy_per_particle(::Val{:gga_x_pbefe}, ρ, σ)
 end
 
 function energy_per_particle_x_pbe(ρ, σ; κ, μ) # eqns refer to PBE paper
-    pbe_x_f(s) = 1 + κ - κ^2 / (κ + μ * s^2)   # (14)
+    pbe_x_f(s²) = 1 + κ - κ^2 / (κ + μ * s²)   # (14)
```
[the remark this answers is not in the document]
antoine-levitt: Yes, see comment above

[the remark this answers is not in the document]
mfherbst: ok, sorry ... did not read ... then let's keep it.

src/SymOp.jl
antoine-levitt: these are non-trivial formulas, better as separate lines

src/SymOp.jl
```diff
@@ -19,21 +19,19 @@
 # Tolerance to consider two atomic positions as equal (in relative coordinates)
 const SYMMETRY_TOLERANCE = 1e-5
 
-# Represents a symmetry (W,w)
 struct SymOp{T <: Real}
     W::Mat3{Int}
     w::Vec3{T}
+
+    S::Mat3{Int}
+    τ::Vec3{T}
```
mfherbst, line 27: I think for completeness, just add real and fourier space here

src/SymOp.jl
```diff
@@ -19,21 +19,19 @@
...
     function SymOp(W, w)
         w = mod.(w, 1)
-        new{eltype(w)}(W, w)
+        S = symop.W'
+        τ = -symop.W \symop.w
+        new{eltype(τ)}(W, w, S, τ)
     end
```
mfherbst, line 33: why is this actually needed in-struct?

antoine-levitt: it's not but for simple constructors it's simpler to have it there no?

mfherbst: I prefer the out-of-struct constructions because it is nicer to add multiple dispatch etc. and we basically only have those at the moment.

antoine-levitt: In this particular case it's better like this because having an inner constructor removes the default constructor (which takes as many inputs as in the struct), so here the invariants are guaranteed by construction

mfherbst: ok

Issue #621, opened 2022-03-05
Remove ksymops


antoine-levitt:
We don't actually use ksymops anymore, so we should excise them completely. However it's currently pretty enmeshed with the lower-level pwbasis constructors, which are mostly used for tests. We need to figure out what is the lowest-level reasonable interface we want to allow, and remove the rest.

PR #622, opened 2022-03-05
Excise ksymops

antoine-levitt:
This does the easy part of https://github.com/JuliaMolSim/DFTK.jl/issues/621, basically not using ksymops for anything. Now we have to remove it from pwbasis, overhaul the constructor and fix the tests


antoine-levitt:
(on top of https://github.com/JuliaMolSim/DFTK.jl/pull/619)

antoine-levitt:
This is ready for review

PR #623, opened 2022-03-05
[WIP] Time-reversal symmetry for kpoint reduction

antoine-levitt:


antoine-levitt:
Looks like this is actually working, except for forces. I think the formula for the symmetrization is correct though, so I'll have to investigate...

PR #627, opened 2022-03-10
Cell to supercell routines

LaurentVidal95:
All the routines contained in the former "ifft on the supercell and plotting wannier functions" without the plotting part.


LaurentVidal95:
Hello everyone, I tried to follow your recommendations Michael hope I didn't break everything !

mfherbst:
Much nicer @LaurentVidal95 , thanks. I'll take a look.

LaurentVidal95:
I'm putting back here the description of the last commit:
I cleaned and documented (as clearly as I could) a `supercell.jl` file that basically provides a `cell_to_supercell` function applying to `basis` and `scfres`. These routines are also covered by tests that work on my laptop.

Unfortunately theses routines take quite a long time for real cases (with big `kgrid` the number of electrons in the supercell is skyrocketing) and I cannot use them to plot as I wanted to. (This was the first motivation for this PR).
I have a faster way to do iFFT and plot Wannier functions on the supercell but it doesn't use the general ``cell_to_supercell`` function and is more a personal code than a real addition to the DFTK.jl code. 

So I have removed all the ``plot_wannier_function`` functions appearing in the previous PR and only kept ``supercell.jl``, so that you can use these routines if you're still interested.

(PS: I was able to plot Wannier function with Makie, so that it appears in the wannier90 example, but it keeps somehow squaring up the supercell, is quite heavy to use and the result is not very nice. Maybe the responsibility is on me, not knowing the proper commands, so I can keep digging if needed).

test/supercell.jl
```diff
@@ -0,0 +1,19 @@
...
+    model = model_DFT(silicon.lattice, [Si => silicon.positions], [:lda_xc_teter93])
+    basis = PlaneWaveBasis(model; Ecut=Ecut, kgrid=kgrid, kshift=kshift)
+    basis_supercell = DFTK.cell_to_supercell(basis)
+    # SCF
+    scfres = self_consistent_field(basis, tol=tol)
+    scfres_supercell_manual = self_consistent_field(basis_supercell, tol=tol)
+    scfres_supercell = DFTK.cell_to_supercell(scfres)
+
+    @test norm(scfres.energies.total*8 - scfres_supercell.energies.total) < 1e-5
+    @test scfres.energies.total*8 == scfres_supercell.energies.total
```
mfherbst, line 18: Ok, but this is not really testing most of the complicated things you do (e.g. unfolding the \psi).

Maybe build a new density from the unfolded \psi and check it agrees with the unfolded ρ or check the residual of the unfolded \psi wrt. the Hamiltonian built from the unfolded basis is stilll reasonable (identical to the not unfolded residual).

mfherbst (commented):
Thanks for doing that. Indeed that would be useful to have.

I'm not yet entirely sure I understand parts of your code right. Also you need something that tests the \psi and ρ unfolding.

LaurentVidal95:
Dear all, I'm sorry I forgot this branch was involved in a pull request and I pushed it for one of my own code... I hope the tests are not too long. I didn't find how to cancel the push !

mfherbst:
@LaurentVidal95 We / you can also close the PR if you're not really working on this any more.

LaurentVidal95:
In the last commit Étienne fixed the test. I still have to adapt to the other supercell PR.

antoine-levitt:
LGTM, let's merge once Etienne adds his supercell construction routine

LaurentVidal95:
@antoine-levitt I merge the PR of @epolack so you can review the PR and merge if happy with it !

antoine-levitt:
LGTM

mfherbst:
@LaurentVidal95 There is an issue with MPI. Either disable the routines for MPI or fix it (probably as easy as using a `gather_kpts_scfres`).

Will do a more careful review later.

mfherbst:
Just nits. Otherwise LGTM.

mfherbst:
> Just a note to add checks for MPI in the main routines (tbd by Michael or Laurent)

Just realised that will actually be taken care off automatically by the PlaneWaveBasis constructor for only one k-point (which will throw in each MPI process, so no deadlock)

PR #630, opened 2022-03-11
Change default kshift to zero

mfherbst:
The previous behaviour has only been causing trouble for small `kgrid`s ...


antoine-levitt:
I always get confused where the BZ starts. I seem to remember -1/2? So this choice does not necessarily include the Gamma point. Seems reasonable but I have no particular idea why one or the other should be preferred

mfherbst:
Actually it's the reverse. This choice ensures to always include the gamma point.

For large enough kgrids it makes no difference. But with Gaspard we noted very strange energy-volume curves on small k-grids on silicon with the old default `kshift`, but a much more reasonable curve when the Gamma point is included. That's the motivation for this PR.

PR #631, opened 2022-03-14
Remove kpoint fallback

mfherbst:
This code has been untested for quite a while and I have not noticed it in the wild being triggered.


antoine-levitt:
Good idea, and nice we don't need it anymore.

PR #633, opened 2022-03-20
Integrate with DftFunctionals.jl

mfherbst:
Insert a clear-cut interface between our code and Libxc. Simplifies the AD efforts as discussed with @mtowara and @niklasschmitz. Should be adapted as needed to make adjoint-derivatives as easy as possible, then upstreamed / moved to a separate package.


mtowara (commented):
make sure densities can not go below zero to avoid sqrt(x<=0)

src/terms/xc.jl
```diff
@@ -434,40 +434,19 @@ function add_kernel_gradient_correction!(δV, terms, density, perturbation, cros
...
-    #    Tuple(union(keys(nt1),keys(nt2))))
-
-    function merge_key(key)
-        if haskey(nt1, key) && !haskey(nt2, key)
-            return nt1[key]
-        elseif haskey(nt2, key) && !haskey(nt1, key)
-            return nt2[key]
+    values = map(all_keys) do key
+        if haskey(nt1, key)
+            nt1[key] .+ get(nt2, key, false)
```
mtowara, line 444: Thats way nicer and zygote actually likes it :)

Sorry for the missing include ealier, was a very poor mans merge...

mfherbst: No worries.

mfherbst: No worries.

src/DispatchFunctional.jl
```diff
@@ -0,0 +1,155 @@
...
+    Vρ = reshape(terms.vrho, s_ρ, n_p)
+    (; e, Vρ)
+end
+function DftFunctionals.potential_terms(func::LibxcFunctional{:gga}, ρ::Matrix{Float64},
+                                        σ::Matrix{Float64})
+    s_ρ, n_p = size(ρ)
+    s_σ = size(σ, 1)
+    fun = Libxc.Functional(func.identifier; n_spin=s_ρ)
+    derivatives = filter(i -> i in Libxc.supported_derivatives(fun), 0:1)
+    terms = Libxc.evaluate(fun; rho=ρ, sigma=σ, derivatives)
```
antoine-levitt, line 72: that's a lot of copy/paste, can't you make an internal function or something here?

test/phonons.jl
```diff
@@ -5,110 +5,104 @@ using LinearAlgebra
...
-        push!(positions, i*ones(3)/n_scell)
+        push!(positions, i * ones(3) / n_scell)
     end
 
     a = 5. * length(positions)
     lattice = a * [[1 0 0.]; [0 0 0.]; [0 0 0.]]
 
-    s = DFTK.compute_inverse_lattice(lattice)
-    directions = [[s * [i==j,0,0] for i in 1:n_scell] for j in 1:n_scell]
+    L⁻¹ = DFTK.compute_inverse_lattice(lattice)
```
antoine-levitt, line 23: Not super fan of this, it renders badly eg on android phones, and doesn't really help much compared to Linv (which we use at other places I believe)

antoine-levitt (commented):
Some nits, feel free to ignore!

mtowara:
sorry for the lack of review, I'm on the road right now.
I'll try to merge the updated master into my zygote branch and see what brakes.

Do I understand it correctly that the xc fallbacks have now moved again to a standalone package?

mfherbst:
@antoine-levitt Thanks for the review. I'll make some changes in master regarding your comments.

@mtowara Yes indeed it means the xc fallbacks have moved again. But this time for good (hopefully) as they really do not belong to DFTK but to something separate. I hope breakage is minimal, if not we can still easily rearrange what needs to be fixed. I'll wait to refactor Libxc.jl towards this interface until we know it works.

src/DispatchFunctional.jl
mfherbst: Don't think that's worth it. The old Libxc interface will die anyway.

PR #634, opened 2022-03-20
Remove dead and unused code

mfherbst:


antoine-levitt:
Hm the rough estimate is kind of useful... Would be nice to preserve the information in some form, but I don't really know where...

mfherbst:
And the file has not even been included. We could put it into the docs as a code block? Into the timing section?

antoine-levitt:
In the section on performance could be nice

PR #637, opened 2022-03-22
Zygote xc

mtowara:
Working XC LDA fallback implementation for use with Zygote.

Due to the crudeness and the warnings generated I would suggest leaving it on a branch zygote-xc, but github does not seem to allow that in a PR.
So if you give the go ahead I'll create a new branch and pull it into that.


mfherbst:
I'll take a more detailed look later, but this indeed looks pretty horrible. I don't mind leaving it on a separate branch, but I fear having too many development branches things can get out of hand.

What was your plan with respect to #633? For that sake it might be a good idea to keep the XC separate for a second to be able to tune the best way to design the interface?

mtowara:
Haven't really looked into https://github.com/JuliaMolSim/DFTK.jl/pull/633 yet, lets discuss on Thursday?!
We can also leave it on my fork for now.

src/terms/xc.jl
```diff
@@ -82,46 +85,49 @@ end
...
-        end
-        if haskey(terms, :vlapl) && any(x -> abs(x) > term.potential_threshold, terms.vlapl)
-            @warn "Meta-GGAs with a Vlapl term have not yet been thoroughly tested." maxlog=1
-            mG² = [-sum(abs2, G) for G in G_vectors_cart(basis)]
-            Vl_fourier = r_to_G(basis, terms.vlapl[s, :, :, :])
-            potential[:, :, :, s] .+= G_to_r(basis, mG² .* Vl_fourier)  # ΔVl
-        end
-    end
+    potential = vrho #zero(ρ)
+    #@views for s in 1:n_spin
```
mfherbst, line 89: Well this is block of commented code is clearly annoying and my PR will also not solve this issue. We should discuss ways to change primal here to make this more easily differentiable.

Problem is that keeping multiple potentials in memory is quite a bit of a memory overhead. Perhaps the easiest here is to formulate this as a map over `s` returning a `Vector{Array{3}}` and then placing that into the `RealSpaceMultiplication` further below?

src/terms/xc.jl
```diff
@@ -451,38 +457,164 @@ end
...
+        zki, vrhoi, vsigmai, v2rho2i = evaluate(xc, density; kwargs...)
+        zk     = !isnothing(zk)     ? zk + zki : zki
+        vrho   = !isnothing(vrho)   ? vrho + vrhoi : vrhoi
+        vsigma = !isnothing(vsigma) ? vsigma + vsigmai : vsigmai
+        v2rho2 = !isnothing(v2rho2) ? v2rho2 + v2rho2i : v2rho2i
+    end
+    return zk, vrho, vsigma, v2rho2
+end
+
+function Libxc.evaluate(func::Functional; derivatives=0:1, rho::AbstractArray, kwargs...)
```
mfherbst, line 502: This stuff should not be needed any more with my PR.

src/terms/xc.jl
```diff
@@ -451,38 +457,164 @@ end
...
-    result = evaluate(xcs[1], density; kwargs...)
-    for i in 2:length(xcs)
-        other = evaluate(xcs[i], density; kwargs...)
-        for (key, data) in pairs(other)
-            if haskey(result, key)
-                result[key] .+= data
-            else
-                result = merge(result, (key => data, ))
+    # conversion from NamedTuple and back can hopefully be skipped once mergewith is
+    # implemented for NamedTuple in Julia (open PR)
```
mfherbst, line 487: I have the same issue in my PR, but I think for now it is fair to hard-code it for the case of just two functionals. The set of keys even for more general cases (beyond LDA) is large, but manageable, so one could perhaps unroll everything manually using some code generation tricks. But for sure this is messy.

I think the advantage with my PR is that you can specialise on `Vector{<:FallbackFunctional}` and thereafter assume that the returned `terms` may at most have `zk`, `vrho`, `vsigma` and `v2rho2`. That could help.

src/xc/lda_c_vwn.jl
```diff
@@ -2,13 +2,17 @@
 VWN5 LDA correlation according to Vosko, Wilk, and Nusair, (DOI 10.1139/p80-159).
 """
 function energy_per_particle(::Val{:lda_c_vwn}, ρ)
+    if ρ <= 1e-8
```
mfherbst, line 5: That's a little large. Can you do 1e-14 or 1e-16? We should also put this into the primal (with a tolerance depending on the floating-point type).

src/xc/lda_c_vwn.jl
```diff
@@ -2,13 +2,17 @@
...
+        return 0.0
```
mfherbst, line 6: Type-instability: ρ might not be Float64, use `zero(ρ)`

src/xc/lda_c_vwn.jl
```diff
@@ -2,13 +2,17 @@
...
+    end
     # From https://math.nist.gov/DFTdata/atomdata/node5.html
     A = 0.0310907
     x0 = -0.10498
     b = 3.72744
     c = 12.9352
+    #println("rho ",ρ)
     rₛ = cbrt(3/(4π*ρ)) # τ in the above link
-    x = sqrt(rₛ)
+    x = rₛ > 0 ? sqrt(rₛ) : 0.0
```
mfherbst, line 15: Should not be needed if ρ is positive like you check now.

src/xc/lda_x.jl
```diff
@@ -1,4 +1,10 @@
 """
 LDA Slater exchange (DOI: 10.1017/S0305004100016108 and 10.1007/BF01340281)
 """
-energy_per_particle(::Val{:lda_x}, ρ) = -3/4 * cbrt(3/π * ρ)
+function energy_per_particle(::Val{:lda_x}, ρ) 
+    #@warn "lda_x energy per particle"
+    if ρ <= 1e-8
+        return 0.0
+    end
```
mfherbst, line 8: Same thing as before. Actually we should probably put this one level up (to where these functions are called) anyway.

mfherbst (commented):
A good start. I think before going to PBE, we should first think of restructuring the primal to get rid of as many of the workarounds you need here as needed.

examples/zygote_lda_c.jl
```diff
@@ -0,0 +1,69 @@
...
+function dipole(basis, ρ)
+    #@assert isdiag(basis.model.lattice)
+    a  = basis.model.lattice[1, 1]
+    rr = [a * (r[1] - 1/2) for r in r_vectors(basis)]
+    sum(rr .* ρ) * basis.dvol
+end
+
+## Function to compute the dipole for a given field strength
+function compute_dipole(ε; tol=1e-8, kwargs...)
+    scfres = self_consistent_field(make_basis_model(ε; kwargs...), tol=tol)
```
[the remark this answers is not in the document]
mtowara: there is still an issue with the model_DFT constructor and ChainRules, Niklas and I are on it.

src/terms/xc.jl
```diff
@@ -68,10 +69,12 @@ end
...
     # It may happen that a functional does only provide a potenital and not an energy term
     # Therefore skip_unsupported_derivatives=true to avoid an error.
-    terms = evaluate(term.functionals, density; skip_unsupported_derivatives=true)
-    @assert haskey(terms, :vrho)
-    if haskey(terms, :zk)
-        E = sum(terms.zk .* ρ) * basis.dvol
+    #terms = evaluate(term.functionals, density; skip_unsupported_derivatives=true)
+    zk, vrho, vsigma, v2rho2 = evaluate(term.functionals, density; skip_unsupported_derivatives=true)
+    #println("vrho ",vrho[1])
+    @assert !isnothing(vrho)
```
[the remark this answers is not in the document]
mtowara: maybe this one he can determine statically? or maybe there is a nondiff for isnothing

src/terms/xc.jl
```diff
@@ -82,46 +85,49 @@ end
...
-            end
-        end
-        if haskey(terms, :vlapl) && any(x -> abs(x) > term.potential_threshold, terms.vlapl)
-            @warn "Meta-GGAs with a Vlapl term have not yet been thoroughly tested." maxlog=1
-            mG² = [-sum(abs2, G) for G in G_vectors_cart(basis)]
-            Vl_fourier = r_to_G(basis, terms.vlapl[s, :, :, :])
-            potential[:, :, :, s] .+= G_to_r(basis, mG² .* Vl_fourier)  # ΔVl
-        end
-    end
+    potential = vrho #zero(ρ)
```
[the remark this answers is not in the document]
mtowara: that should work as well, but do we even need the permutation if we can just slice the arrays differently in the function calls?

src/xc/lda_c_vwn.jl
mtowara: sure, but it already is the primal? we could of course put it one level higher

src/terms/xc.jl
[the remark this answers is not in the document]
mfherbst: Ah yes that might be. `nothing` has a special type, so this is probably a compile-time constant.

src/xc/lda_c_vwn.jl
mfherbst: Sorry by primal I meant `master`. yes putting it one level higher is better, I think.

src/terms/xc.jl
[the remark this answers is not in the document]
mfherbst: I'm not sure what you mean. My point for suggesting this was to stay as close to the uncommented code as possible. But with the `map` approach I suggested below this will be obsolete anyway. Let's talk tomorrow.

PR #641, opened 2022-03-28
Add and use cis2pi functions

### LINKED ISSUE #400: Define and use cis2pi
Just had a look through the release notes for 1.6 (which will be LTS).

Things worth using in DFTK once we move to 1.6 are:
- `sincospi`
- `cispi`

kvnoct:
Close #400 

Details:
1. defined cis2pi/cos2pi/sin2pi based on the built-in functions
2. changed any scripts I could find that previously used cis(2 pi * something) to cis2pi(something)

### DIFF
--- a/src/common/cis2pi.jl
+++ b/src/common/cis2pi.jl
@@ -1,0 +1,12 @@
+"""Functions to evaluate exp(), cos(), or sin() involving 2π"""
+function cis2pi(z)
+    return cispi(2*z)
+end
+
+function sin2pi(z)
+    return sinpi(2*z)
+end
+
+function cos2pi(z)
+    return cospi(2*z)
+end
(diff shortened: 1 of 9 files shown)


src/common/cis2pi.jl
```diff
@@ -0,0 +1,12 @@
+"""Functions to evaluate exp(), cos(), or sin() involving 2π"""
```
antoine-levitt, line 1: That's a bit generic. Having the formula cis2pi(x) = exp(2π i x) in the comment makes it clearer, esp since `cis` is not super standard outside julia

antoine-levitt (commented):
Nice! You missed two spots in symmetry.jl.

kvnoct:
It seems the error comes from cispi computations

image

antoine-levitt:
That's IntervalArithmetic being buggy:  `sincospi(2±1)` is an overflow. Can you open an issue there? In the meantime, we can define a workaround `cis2pi(x::Interval) = exp(2 * (pi * (im * x)))`, and same for sin2pi and cos2pi in workarounds/interval.jl with a reference to the IntervalArithmetic issue number (note I think the parentheses are needed to avoid implicit conversions to float64)

antoine-levitt:
Merged, thanks a lot!

Issue #625, opened 2022-03-06
Investigate forces test failure


antoine-levitt:
Eg https://github.com/JuliaMolSim/DFTK.jl/runs/5439547530?check_suite_focus=true
with `println(abs(diff_findiff - diff_forces))` on the first forces test I can't get it to go down far below 1e-5 with symmetry. Without symmetry everything is fine and I get 1e-10. Possibly linked with SYMMETRY_TOLERANCE. We should investigate and set the tolerance far below 1e-5.

antoine-levitt:
Decreasing SYMMETRY_TOLERANCE does not change anything

antoine-levitt:
Reproducer
```
# Very basic setup, useful for testing
using DFTK

a = 10.26  # Silicon lattice constant in Bohr
lattice = a / 2 * [[0 1 1.];
                   [1 0 1.];
                   [1 1 0.]]
Si = ElementPsp(:Si, psp=load_psp("hgh/lda/Si-q4"))
atoms = [Si => [[1.01, 1.02, 1.03]/8, -ones(3)/8]]

Ecut = 7
kgrid = [2, 2, 2]
kshift = [0, 0, 0]
is_converged = DFTK.ScfConvergenceDensity(1e-12)

model = model_LDA(lattice, atoms)
basis = PlaneWaveBasis(model; Ecut, kgrid, kshift)
scfres = self_consistent_field(basis; is_converged)
F1 = compute_forces(scfres)

model = model_LDA(lattice, atoms, symmetries=false)
basis = PlaneWaveBasis(model; Ecut, kgrid, kshift)
scfres = self_consistent_field(basis; is_converged)
F2 = compute_forces(scfres)

println(norm(F1-F2))
```
The error goes down when using larger supersampling factors, and the error is zero when using a model without xc

antoine-levitt:
It's not the fault of lowpass_for_symmetry! (which actually is probably completely useless and should be removed...)

antoine-levitt:
A simpler test: run a computation without symmetry, and then reload it in a symmetric computation and compare the density to the symmetrized one. There's a similar difference. This is because the xc term depends on the actual fft_size, which itself is not preserved by symmetries. The Hartree is not affected because it is exact (does not depend on fft_size as long as supersampling >= 2), and the variational basis sets for the psi are symmetry-invariant. Still not completely clear to me if there's something we can do or if we should just write it up in the docs and live with it.

[...]

PR #642, opened 2022-03-30
Filter rgrid symmetries

antoine-levitt:
Fix https://github.com/JuliaMolSim/DFTK.jl/issues/625. On top of https://github.com/JuliaMolSim/DFTK.jl/pull/640. @jaemolihm could you check that this looks OK? (look at https://github.com/JuliaMolSim/DFTK.jl/pull/642/commits/c5771708aeec9eb7b0d8d1d24539d2deca868e2a to get only the relevant diff)


antoine-levitt:
Hm so this is too brutal: on silicon Ecut 15, the FFT algorithm chooses an odd size for the FFT, which invalidates too much symmetries. I'm not too sure what to do here...

jaemolihm:
I think the implementation looks fine.

One thing I find a bit strange is that symmetries are filtered within `bzmesh_ir_wedge`. I thought about filtering symmetries using `fft_size` before calling `bzmesh_ir_wedge`, and then calling `bzmesh_ir_wedge` (without the `fft_size` argument). But I expect there may be some reason you implemented it the other way.

Regarding the brutality, I discussed with QE developers and found that the QE solution is to force the fft grid to be a multiple of the denominators of the fractional translations (only for 2, 3, 4, 6).
https://gitlab.com/QEF/q-e/-/blob/develop/PW/src/symm_base.f90#L558-575

Also QE explicitly discards symmetries if the fractional translation have denominator different from 2, 3, 4, 6.

antoine-levitt:
Thanks, this is very helpful! 

> Also QE explicitly discards symmetries if the fractional translation have denominator different from 2, 3, 4, 6.

There's no particular theoretical reason to do this, right? It's just to ensure reasonable FFTs?

jaemolihm:
> There's no particular theoretical reason to do this, right? It's just to ensure reasonable FFTs?

Yes, I think there is no fundamental reason, maybe the reasoning is that 2, 3, 4, 6 covers all the ordinary cases. (Not sure how to define "ordinary" though.)

antoine-levitt:
Thanks! OK that sounds sensible and I can't think of a better and more principled way of doing things so I'll just do that.

antoine-levitt:
OK, this is now done. Unless the fft_size is passed explicitly (or the symmetries_rgrid argument is passed), the fft grid is adapted to the symmetries. As a result, the exploitation of symmetries should (finally) be a pure optimization and give identical results (this does not mean that passing symmetries=true or false to the model will yield identical results, since the FFT grid is potentially increased in the true case, but not in the false case).

antoine-levitt:
Btw
> One thing I find a bit strange is that symmetries are filtered within bzmesh_ir_wedge. I thought about filtering symmetries using fft_size before calling bzmesh_ir_wedge, and then calling bzmesh_ir_wedge (without the fft_size argument). But I expect there may be some reason you implemented it the other way.

you were completely right of course. Hopefully the code is a bit clearer now (I'm still not 100% happy about these constructors though)

PR #645, opened 2022-04-06
Integrate with InteratomicPotentials.jl

jrdegreeff:
This PR adds an implementation of the [InteratomicPotentials.jl](https://github.com/cesmix-mit/InteratomicPotentials.jl) `ArbitraryPotential` interface. This allows us to use DFTK for AIMD simulations using [Atomistic.jl](https://github.com/cesmix-mit/Atomistic.jl). [This script](https://github.com/cesmix-mit/Atomistic.jl/blob/main/examples/argon/molly_dftk_ab_initio_simulation.jl) is a toy example of (an older experimental version of) this integration in action.

The PR adds the `DFTKPotential` type (open to suggestions if this name seems vague) and implements the only required method in the interface for MD, `InteratomicPotentials.energy_and_force`. This adds a dependency on InteratomicPotentials.jl which is just an interface package with a couple trivial examples such as Lennard-Jones. The package is very lightweight (similar size to AtomsBase.jl), so the added dependency should not be a concern.

The implementation was designed in consultation with Michael.


test/external_interatomicpotentials.jl
```diff
@@ -0,0 +1,24 @@
+using AtomsBase
```
antoine-levitt, line 1: make a new subfolder external/ in tests (and move the others there as well)?

antoine-levitt (commented):
LGTM! I'm kind of nervous about too many interfacing codes going into the DFTK repo, but for this stage of maturity and complexity it's the best option.

test/external/interatomicpotentials.jl
```diff
@@ -0,0 +1,35 @@
...
+        Atom(:Ar, [21.0, 7.0, 21.0]u"bohr"; pseudopotential=pspkey),
+        Atom(:Ar, [7.0, 7.0, 21.0]u"bohr"; pseudopotential=pspkey),
+        Atom(:Ar, [21.0, 21.0, 7.0]u"bohr"; pseudopotential=pspkey),
+        Atom(:Ar, [7.0, 21.0, 7.0]u"bohr"; pseudopotential=pspkey),
+        Atom(:Ar, [21.0, 7.0, 7.0]u"bohr"; pseudopotential=pspkey),
+        Atom(:Ar, [7.0, 7.0, 7.0]u"bohr"; pseudopotential=pspkey)
+    ]
+    box = [[28.0, 0.0, 0.0], [0.0, 28.0, 0.0], [0.0, 0.0, 28.0]]u"bohr"
+    boundary_conditions = [Periodic(), Periodic(), Periodic()]
+    system = FlexibleSystem(particles, box, boundary_conditions)
```
mfherbst, line 27: Use `AtomsBase.periodic_system`.

jrdegreeff: Thanks these are good to know about. I've updated the test case.

PR #650, opened 2022-04-13
Support isolated systems

antoine-levitt:
First attempt at support for isolated systems. Right now this tackles only neutral molecules. 

TODO
- [ ] systems with dipole moments
- [ ] ions
- [ ] tests

### DIFF
--- a/src/Model.jl
+++ b/src/Model.jl
@@ -12,6 +12,8 @@ struct Model{T <: Real}
     recip_lattice::Mat3{T}
     # Dimension of the system; 3 unless `lattice` has zero columns
     n_dim::Int
+    # Whether the problem is periodic along each direction
+    periodic::Vec3{Bool}
     # Useful for conversions between cartesian and reduced coordinates
     inv_lattice::Mat3{T}
     inv_recip_lattice::Mat3{T}
@@ -62,7 +64,7 @@ _is_well_conditioned(A; tol=1e5) = (cond(A) <= tol)
 
 """
     Model(lattice, atoms, positions; n_electrons, magnetic_moments, terms, temperature,
-          smearing, spin_polarization, symmetries)
+          smearing, spin_polarization, symmetries, periodic)
 
 Creates the physical specification of a model (without any discretization information).
 
@@ -98,6 +100,7 @@ function Model(lattice::AbstractMatrix{T}, atoms=Element[], positions=Vec3{T}[];
                spin_polarization=default_spin_polarization(magnetic_moments),
                symmetries=default_symmetries(lattice, atoms, positions, magnetic_moments,
                                              spin_polarization, terms),
+               periodic=Vec3(true, true, true)
                ) where {T <: Real}
     lattice = Mat3{T}(lattice)
     temperature = T(austrip(temperature))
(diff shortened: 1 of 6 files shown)


antoine-levitt:
Currently the way I do it is with an improved Poisson solver. If I understand correctly it's similar in spirit to what ABINIT does with `icoulomb=1` (using the BigDFT Poisson solver). QE has a lot of correction schemes which I don't really understand.

antoine-levitt:
Hm the above test is meaningless, with a model_LDA the default thing appears to work correctly, and the better Poisson solver doesn't actually appear to change anything. Need to think about this more carefully.

antoine-levitt:
Benchmark: (correct this time)
```
using DFTK
using LinearAlgebra
using PyPlot

RES = Dict()
for a in (10, 20, 30)
    for per in (false, true)
        lattice = [a 0 0;
                   0 a 0;
                   0 0 a]
        atoms     = [ElementPsp(:Li, psp=load_psp("hgh/lda/Li-q3"))]
        atoms     = [ElementPsp(:He, psp=load_psp("hgh/lda/He-q2"))]
        positions = [[1/2, 1/2, 1/2]]

        kgrid = [1, 1, 1]  # no k-point sampling for an isolated system
        Ecut = 40
        tol = 1e-6

        model = model_LDA(lattice, atoms, positions, periodic=[per, per, per])
        res   = self_consistent_field(basis, is_converged=DFTK.ScfConvergenceDensity(tol))

        rr = vec([norm(a * (r .- 1/2)) for r in r_vectors(basis)])

        function quadrupole(basis, ρ)
            rr = [norm(a * (r .- 1/2)) for r in r_vectors(basis)]
            sum(rr .^ 2 .* ρ) * basis.dvol
        end;
        quad = quadrupole(basis, res.ρ)
        println(quad)
        RES[per, a] = (quad, res.energies.total)
    end
end
```
sort(RES) gives
```
  (false, 10) => (2.63994, -2.80773)
  (false, 20) => (2.6244, -2.80169)
  (false, 30) => (2.62615, -2.80087)
  (true, 10)  => (2.63995, -2.80069)
  (true, 20)  => (2.6244, -2.80081)
  (true, 30)  => (2.62615, -2.80061)
```
So it seems the naive approach is actually not that bad for the density, although the energy does look better with the isolated Poisson solver. Still need to think about why the naive supercell method is not that bad here.

antoine-levitt:
https://arxiv.org/pdf/cond-mat/9709234.pdf is a good reference. They have a clever technique to solve the Poisson equation by truncating the Coulomb kernel and using a bigger supercell to avoid aliasing, which should probably be more accurate but (kind of a lot) more expensive than the one I'm using here. I think what I see is broadly consistent with their conclusions though. I'll implement dipole matching and do some more experiments.

antoine-levitt:
See also https://arxiv.org/pdf/cond-mat/0601031.pdf which extends this technique to other dimensionalities

antoine-levitt:
Results are pretty disappointing: for He, results are more or less the same. For LiH, they are somewhat different, but the isolated scheme is not really better. I should probably implement the scheme above and compare...

PR #655, opened 2022-04-29
Handling complex numbers for PairwisePotential

epolack:


src/terms/pairwise.jl
```diff
@@ -1,3 +1,8 @@
+function norm_cplx(x)
+    # TODO: ForwardDiff bug (https://github.com/JuliaDiff/ForwardDiff.jl/issues/324)
+    sqrt(sum(x.*x))
+end
```
antoine-levitt, line 4: Explain why this is needed (analytic continuation of norm)

src/terms/pairwise.jl
```diff
@@ -10,10 +15,10 @@ Lennard—Jones terms.
 The potential is dependent on the distance between to atomic positions and the pairwise
 atomic types:
 For a distance `d` between to atoms `A` and `B`, the potential is `V(d, params[(A, B)])`.
-The parameters `max_radius` is of `100` by default, and gives the maximum (reduced) distance
-between nuclei for which we consider interactions.
+The parameters `max_radius` is of `1000` by default, and gives the maximum (full,
+non-reduced) distance between nuclei for which we consider interactions.
```
antoine-levitt, line 19: distance (in cartesian coordinates)

src/terms/pairwise.jl
```diff
@@ -36,8 +41,10 @@ end
 @timing "forces: Pairwise" function compute_forces(term::TermPairwisePotential,
                                                    basis::PlaneWaveBasis{T}, ψ, occ;
                                                    kwargs...) where {T}
-    forces = zero(basis.model.positions)
-    energy_pairwise(basis.model, term.V, term.params; term.max_radius, forces)
+    TT = eltype(lattice)
```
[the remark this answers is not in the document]
epolack: Don't remember why, but the lattice needs to be converted to Dual type when using ForwardDiff.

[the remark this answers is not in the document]
antoine-levitt: In that case, you probably want to promote

src/terms/pairwise.jl
```diff
@@ -96,32 +109,40 @@ function energy_pairwise(lattice, symbols, positions, V, params;
...
+                Δr = lattice * (ti .- tj)
+                dist = norm_cplx(Δr)
 
                 # the potential decays very quickly, so cut off at some point
-                dist > max_radius && continue
+                abs(dist) > max_radius && continue
 
                 any_term_contributes = true
-                energy_contribution = V(dist, param_ij)
+                energy_contribution = real(V(dist, param_ij))
```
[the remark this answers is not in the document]
epolack: Yep, I didn't do the tests yet.

src/terms/pairwise.jl
```diff
@@ -96,32 +109,40 @@ function energy_pairwise(lattice, symbols, positions, V, params;
...
                 sum_pairwise += energy_contribution
                 if forces !== nothing
-                    # We use ForwardDiff for the forces
-                    dE_ddist = ForwardDiff.derivative(d -> V(d, param_ij), dist)
-                    dE_dti = lattice' * dE_ddist / dist * Δr
+                    dE_ddist = ForwardDiff.derivative(real(zero(eltype(dist)))) do ε
```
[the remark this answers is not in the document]
epolack: https://github.com/JuliaDiff/ForwardDiff.jl/pull/583

antoine-levitt:
For this PR and others: can you mark the PRs as draft, and then mark it as finished when you want our review? Or indicate it in the issue or something.

src/terms/pairwise.jl
```diff
@@ -1,3 +1,9 @@
+# We cannot use `LinearAlgebra.norm` with complex numbers due to the need to use its
+# analytic continuation
```
antoine-levitt, line 2: Comment not great. "Complex-analytic extension of norm(x) from real to complex inputs."

antoine-levitt: Also explain a bit the reasoning behind this. Eg "We want to perform a matrix-vector product f'(x) * h, where f is a real-to-real function, and h a complex vector. To do this using automatic differentiation, we can extend analytically f to accept complex inputs, then differentiate t -> f(x+t h). This will fail if non-analytic functions like `norm` are used for complex inputs, and therefore we have to redefine it."

antoine-levitt: This will probably be needed for other functions, so longer-term it should live in common/, but I would say fine for now.

src/terms/pairwise.jl
```diff
@@ -37,7 +43,8 @@ end
                                                    basis::PlaneWaveBasis{T}, ψ, occ;
                                                    kwargs...) where {T}
     forces = zero(basis.model.positions)
-    energy_pairwise(basis.model, term.V, term.params; term.max_radius, forces)
+    energy_pairwise(basis.model, term.V, term.params; max_radius=term.max_radius,
```
antoine-levitt, line 46: why? We use the previous convention quite a lot, it's pretty nice in simplifying code. If your concern is that it somewhat hides the nature of the arguments, you can line-break on the `;`

src/terms/pairwise.jl
```diff
@@ -56,11 +63,19 @@ end
 
 # This could be factorised with Ewald, but the use of `symbols` would slow down the
 # computationally intensive Ewald sums. So we leave it as it for now.
+# `q` is the phonon `q`-point (`Vec3`), and `ph_disp` a list of `Vec3` displacements to
```
antoine-levitt, line 66: This does not really explain much, you need an extended comment here. Explain what this function actually computes when passed a non-nothing q/ph_disp. Write down the formulas? And also that only `forces` is valid in this case?

src/terms/pairwise.jl
```diff
@@ -56,11 +63,19 @@ end
...
+# compute the Fourier transform of the force constant matrix.
 function energy_pairwise(lattice, symbols, positions, V, params;
-                         max_radius=100, forces=nothing)
-    T = eltype(lattice)
+                         max_radius=100, forces=nothing, ph_disp=nothing, q=nothing)
     @assert length(symbols) == length(positions)
 
+    T = eltype(positions[1])
+    if ph_disp !== nothing
+        @assert q !== nothing
```
antoine-levitt, line 74: Also assert that if one of ph_disp and q is nothing, the other has to be also

src/terms/pairwise.jl
```diff
@@ -95,33 +110,40 @@ function energy_pairwise(lattice, symbols, positions, V, params;
                 rsh == 0 && i == j && continue
 
                 ti = positions[i]
-                tj = positions[j]
+                tj = positions[j] + R
+                # Phonons `q` points
+                if !isnothing(ph_disp)
+                    ti += ph_disp[i] # * cis(2T(π)*dot(q, zeros(3))) ≡ 1
```
antoine-levitt, line 116: Comment pretty obscure...

src/terms/pairwise.jl
```diff
@@ -95,33 +110,40 @@ function energy_pairwise(lattice, symbols, positions, V, params;
                 rsh == 0 && i == j && continue
 
                 ti = positions[i]
-                tj = positions[j]
+                tj = positions[j] + R
+                # Phonons `q` points
```
antoine-levitt, line 114: Doesne't really add value?

src/terms/pairwise.jl
```diff
@@ -95,33 +110,40 @@ function energy_pairwise(lattice, symbols, positions, V, params;
...
                 sum_pairwise += energy_contribution
                 if forces !== nothing
-                    # We use ForwardDiff for the forces
-                    dE_ddist = ForwardDiff.derivative(d -> V(d, param_ij), dist)
-                    dE_dti = lattice' * dE_ddist / dist * Δr
+                    dE_ddist = ForwardDiff.derivative(real(zero(eltype(dist)))) do ε
+                        V(dist + ε, param_ij)
+                    end
+                    dE_dti = lattice' * ((dE_ddist / dist) * Δr)
+                    # For the phonons, we compute the forces only in the unit cell.
```
antoine-levitt, line 137: Also quite obscure...

src/workarounds/forwarddiff_rules.jl
```diff
@@ -272,3 +272,32 @@ function Smearing.occupation(S::Smearing.FermiDirac, d::ForwardDiff.Dual{T}) whe
     end
     ForwardDiff.Dual{T}(Smearing.occupation(S, x), ∂occ * ForwardDiff.partials(d))
 end
+
+# Workarounds for issue https://github.com/JuliaDiff/ForwardDiff.jl/issues/324
+ForwardDiff.derivative(f, x::Complex) = throw(DimensionMismatch("derivative(f, x) expects that x is a real number (does not support Wirtinger derivatives). Separate real and imaginary parts of the input."))
```
antoine-levitt, line 277: These fixes are not for the issue above. Is the fix not already upstream in a released version? If yes, only define the functions if the version is below the one that has the fix (and then at some point we remove the code and depend on a new version)

test/phonons.jl
```diff
@@ -0,0 +1,124 @@
+using Test
+using DFTK
+using LinearAlgebra
+using ForwardDiff
+using StaticArrays
+
+# ## Helper functions
```
antoine-levitt, line 7: This comment doesn't add value. Something like "convert back and forth from array of Vec3 to columnwise matrix"

test/phonons.jl
```diff
@@ -0,0 +1,124 @@
...
+# Some functions that will be helpful for this example.
+fold(x)   = hcat(x...)
+unfold(x) = Vec3.(eachcol(x))
+
+const MAX_RADIUS = 1e3
```
antoine-levitt, line 12: Don't have these as constants, and include in a testset like in the other tests (so it doesn't pollute the namespace)

test/phonons.jl
```diff
@@ -0,0 +1,124 @@
...
+const TOLERANCE = 1e-6
+const N_POINTS = 10
+
+function prepare_system(; case=1)
```
antoine-levitt, line 16: is `case` not `supercells`?

test/phonons.jl
```diff
@@ -0,0 +1,124 @@
...
+    positions = [[0.,0,0]]
+    for i in 1:case-1
+        push!(positions, i*ones(3)/case)
+    end
+
+    a = 5. * length(positions)
+    lattice = a * [[1 0 0.]; [0 0 0.]; [0 0 0.]]
+
+    s = DFTK.compute_inverse_lattice(lattice)
+    if case === 1
```
antoine-levitt, line 26: Just do it programmatically for any number of cells

src/terms/pairwise.jl
epolack: Removed.

src/workarounds/forwarddiff_rules.jl
epolack: This https://github.com/JuliaDiff/ForwardDiff.jl/pull/577 seems to fix the exponentiation problem. So maybe remove it altogether?

antoine-levitt: Does it fix https://github.com/JuliaDiff/ForwardDiff.jl/issues/514#issuecomment-1113274482? Since my message was after that PR was merge I don't think so?

epolack: Right, I thought is was after. This is strange, something seems to have fixed the issue, because I am not running into it anymore, even for this test:
```
using ForwardDiff, FiniteDifferences, Random, Test

v = randn()
p, m = randn(ComplexF64), randn(ComplexF64)

for f in (x -> (x*m)^p,
          x -> m^(p*x),
          x -> (x*m)^(p*x))
  @test ≈(ForwardDiff.derivative(f, v), central_fdm(5,1)(f, v), atol=1e-8)
end
```

src/terms/pairwise.jl
epolack: Changed it a bit.

src/workarounds/forwarddiff_rules.jl
antoine-levitt: What about the code in my comment in the forwarddiff tracker above?

test/phonons.jl
```diff
@@ -0,0 +1,124 @@
...
+    end
+
+    params = Dict((:X, :X) => (; ε=1, σ=a / length(positions) /2^(1/6)))
+    V(x, p) = 4*p.ε * ((p.σ/x)^12 - (p.σ/x)^6)
+
+    (positions=positions, lattice=lattice, directions=directions, params=params, V=V)
+end
+
+# Compute phonons for a one-dimensional pairwise potential for a set of `q = 0` using
+# supercell method
```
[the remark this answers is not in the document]
epolack: Maybe talk tomorrow, as it is in 1d, I do not even compute the Fourier transform. Maybe keep the explanation for an example file?

antoine-levitt (commented):
Sorry, thought I posted these thursday, but it went into a pending review...

src/terms/pairwise.jl
```diff
@@ -54,74 +66,78 @@ function energy_pairwise(model::Model{T}, V, params; kwargs...) where {T}
 end
 
 
-# This could be factorised with Ewald, but the use of `symbols` would slow down the
-# computationally intensive Ewald sums. So we leave it as it for now.
+"""
+This could be factorised with Ewald, but the use of `symbols` would slow down the
+computationally intensive Ewald sums. So we leave it as it for now.
```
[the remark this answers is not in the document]
antoine-levitt: Docstring is something you can look at in julia's help mode (`?`). I don't use it but @mfherbst likes it. Where it is useful is that it gives some information of what comments are for "outside consumption" (description of the routine, how to use it, etc) and what comments are for internal use (implementation details etc).

src/terms/pairwise.jl
```diff
@@ -56,11 +68,23 @@ end
 
 # This could be factorised with Ewald, but the use of `symbols` would slow down the
 # computationally intensive Ewald sums. So we leave it as it for now.
+# `q` is the phonon `q`-point (`Vec3`), and `ph_disp` a list of `Vec3` displacements to
+# compute the Fourier transform of the force constant matrix. Only the computations of the
```
antoine-levitt, line 72: remove this comment

PR #657, opened 2022-05-24
Fix cart

epolack:
Should the structure factor in the file also comment about the normalization?


epolack:
In the comment, `structure factor e^{-i Gr}`, and `struct_factor(G) = cis(-2π G⋅r)`. Maybe have the same notation, and if there is the `2π`, then have also the `1/sqrt(Ω)`?

antoine-levitt:
Yeah `# where struct_factor(G) = cis(-2π G⋅r)` should just be e^-iGr

PR #658, opened 2022-05-27
Forward-diff support for pseudo parameters

mfherbst:


src/pseudo/NormConservingPsp.jl
```diff
@@ -7,19 +7,21 @@ abstract type NormConservingPsp end
 
 #    lmax::Int                   # Maximal angular momentum in the non-local part
 #    h::Vector{Matrix{Float64}}  # Projector coupling coefficients per AM channel: h[l][i1,i2]
-#    Zion::Int                   # Ionic charge (Z - valence electrons)
 #    identifier::String          # String identifying the PSP
 #    description::String         # Descriptive string
 
 #### Methods:
+# Ionic charge (Z - valence electrons)
+# charge_ionic(psp::NormConservingPsp) = error()
```
antoine-levitt, line 15: why are these commented out?

src/pseudo/NormConservingPsp.jl
```diff
@@ -7,19 +7,21 @@ abstract type NormConservingPsp end
 
 #    lmax::Int                   # Maximal angular momentum in the non-local part
 #    h::Vector{Matrix{Float64}}  # Projector coupling coefficients per AM channel: h[l][i1,i2]
-#    Zion::Int                   # Ionic charge (Z - valence electrons)
```
antoine-levitt, line 10: why?

src/pseudo/PspHgh.jl
```diff
@@ -193,39 +188,34 @@ and `Q` is a polynomial. This function returns `Q`.
     (l == 2 && i == 2) && return common * 2 / 3sqrt(T( 105)) * t^2 * ( 7 -   t^2)
     #
     (l == 3 && i == 1) && return common * 1 /  sqrt(T( 105)) * t^3
+
+    return zero(t)
```
antoine-levitt, line 192: NaN/error?

src/pseudo/NormConservingPsp.jl
mfherbst: Because it's recommended to julia itself error on non-implemented functions rather than explicitly throw an error. I can also completely remove them if you want or make them functions without methods with some docstrings.

src/pseudo/NormConservingPsp.jl
mfherbst: Better to have this as a function rather than a field. We already have a function of that name. Might was well use it also for psp structs.

src/pseudo/NormConservingPsp.jl
antoine-levitt: Yup, thanks!

Issue #656, opened 2022-05-19
Rewrite ewald


antoine-levitt:
As discussed today, that code is a bit of a mess, we should try to clean it up. Here's the relevant computation (see also bounding_rectangle in fft.jl where we do something similar): let A be the lattice matrix, then if ||Ax|| <= R, then xi = <ei, A^-1 Ax> = <A^-T ei, Ax> <= ||A^-T ei|| R. So we just have to compute these bounds, and that gives us the iteration range we need, without having to do this shell business.

niklasschmitz:
I took a first step for rewriting ewald and precomputing bounds here:
https://github.com/JuliaMolSim/DFTK.jl/blob/nfs/ewald-bounds/src/terms/ewald.jl#L91-L93
It passes tests but the bounds are overestimated a lot currently. For comparison:

on master:
```
julia> include("test/ewald.jl")
gsh = 6
rsh = 4
Test Summary: | Pass  Total
Hydrogen atom |    1      1
gsh = 5
rsh = 7
Test Summary:             | Pass  Total
Silicon diamond structure |    1      1
gsh = 6
rsh = 4
Test Summary:  | Pass  Total
Boron molecule |    1      1
gsh = 6
rsh = 4
Test Summary:     | Pass  Total
Hydrogen molecule |    1      1
gsh = 5
rsh = 7
gsh = 5
rsh = 7
Test Summary: | Pass  Total
Forces        |    1      1
Test.DefaultTestSet("Forces", Any[], 1, false, false)
```
on nfs/ewald-bounds
```
julia> include("test/ewald.jl")
Glims = [30, 30, 30]
Rlims = [18, 18, 18]
Test Summary: | Pass  Total
Hydrogen atom |    1      1
Glims = [21, 21, 21]
Rlims = [31, 31, 31]
Test Summary:             | Pass  Total
Silicon diamond structure |    1      1
Glims = [30, 30, 30]
Rlims = [18, 18, 18]
Test Summary:  | Pass  Total
Boron molecule |    1      1
Glims = [30, 30, 30]
Rlims = [18, 18, 18]
Test Summary:     | Pass  Total
Hydrogen molecule |    1      1
Glims = [21, 21, 21]
Rlims = [31, 31, 31]
Glims = [21, 21, 21]
Rlims = [31, 31, 31]
Test Summary: | Pass  Total
Forces        |    1      1
```

niklasschmitz:
[...]

PR #659, opened 2022-05-31
Precompute Ewald summation bounds

niklasschmitz:
This PR adresses #656, precomputing the summation bounds (@antoine-levitt) in reciprocal- and real-space and thus replacing the while-loop over shells by dense cubic for-loops. 

This doesn't yet include functionalization of the loops or non-mutating forces, which are left to a possible follow-up PR when needed.

### DIFF
--- a/src/terms/ewald.jl
+++ b/src/terms/ewald.jl
@@ -60,27 +60,12 @@ function energy_ewald(lattice, charges, positions; η=nothing, forces=nothing)
     energy_ewald(lattice, compute_recip_lattice(lattice), charges, positions; η, forces)
 end
 
-function compute_Glims_Rlims_ewald(lattice, recip_lattice, positions, max_exp_arg, max_erfc_arg, η; ceil_int=true)
-    # As a general statement, with A a lattice matrix, then if ||Ax|| <= R, 
-    # then xi = <ei, A^-1 Ax> = <A^-T ei, Ax> <= ||A^-T ei|| R.
-    #
-    # In the reciprocal-space term we have exp(-||B G||^2 / 4η^2), 
-    # where B is the reciprocal-space lattice, and 
-    # thus use the bound  ||B G|| / 2η ≤ sqrt(max_exp_arg)
-    Glims = [norm(inv(recip_lattice')[:, i]) * sqrt(max_exp_arg) * 2η  for i in 1:3]
-
-    # In the real-space term we have erfc(η ||A(rj - rk - R)||), 
-    # where A is the real-space lattice, rj and rk are atomic positions and
-    # thus use the bound  ||A(rj - rk - R)|| * η ≤ max_erfc_arg
-    poslims = [maximum(rj[i] - rk[i] for rj in positions for rk in positions) for i in 1:3]
-    Rlims = [norm(inv(lattice')[:, i]) * max_erfc_arg / η + poslims[i] for i in 1:3]
-    
-    if ceil_int
-        Glims = ceil.(Int, Glims)
-        Rlims = ceil.(Int, Rlims)
-    end
-
-    Glims, Rlims
+function estimate_integer_lattice_bounds(M, δ, shift=zeros(3))
+    # As a general statement, with M a lattice matrix, then if ||Mx|| <= δ, 
+    # then xi = <ei, M^-1 Mx> = <M^-T ei, Mx> <= ||M^-T ei|| δ.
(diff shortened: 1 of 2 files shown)


src/terms/ewald.jl
```diff
@@ -82,56 +82,45 @@ function energy_ewald(lattice, recip_lattice, charges, positions; η=nothing, fo
     max_exp_arg = -log(eps(T)) + 5  # add some wiggle room
     max_erfc_arg = sqrt(max_exp_arg)  # erfc(x) ~= exp(-x^2)/(sqrt(π)x) for large x
 
+    # Precomputing summation bounds from cutoffs.
+    # Let A be the lattice matrix, then if ||Ax|| <= R,
```
antoine-levitt, line 86: As a general statement, with A a lattice matrix, then if...

src/terms/ewald.jl
```diff
@@ -82,56 +82,45 @@ function energy_ewald(lattice, recip_lattice, charges, positions; η=nothing, fo
     max_exp_arg = -log(eps(T)) + 5  # add some wiggle room
     max_erfc_arg = sqrt(max_exp_arg)  # erfc(x) ~= exp(-x^2)/(sqrt(π)x) for large x
 
+    # Precomputing summation bounds from cutoffs.
+    # Let A be the lattice matrix, then if ||Ax|| <= R, 
+    # then xi = <ei, A^-1 Ax> = <A^-T ei, Ax> <= ||A^-T ei|| R.
+    #
+    # Reciprocal space:  ||A^-1 G|| / 2η ≤ max_erfc_arg
```
antoine-levitt, line 89: with A the reciprocal lattice

src/terms/ewald.jl
```diff
@@ -82,56 +82,45 @@ function energy_ewald(lattice, recip_lattice, charges, positions; η=nothing, fo
...
+    # Real space: ||A(ti - tj - R)|| * η ≤ max_erfc_arg
```
antoine-levitt, line 90: with A the real-space lattice

src/terms/ewald.jl
```diff
@@ -82,56 +82,45 @@ function energy_ewald(lattice, recip_lattice, charges, positions; η=nothing, fo
...
+    Glims = ceil.(Int, [norm(inv(recip_lattice')[:, i]) * max_erfc_arg * 2η  for i in 1:3])
+    poslims = [maximum(rj[i] - rk[i] for rj in positions for rk in positions) for i in 1:3]
+    Rlims = ceil.(Int, [norm(inv(lattice')[:, i]) * max_erfc_arg / η + poslims[i] for i in 1:3])
```
antoine-levitt, line 93: Ooh clever with the poslims, I would have just ignored it :-)

antoine-levitt (commented):
Very nice! Maybe even put the three for loops in one line?

src/terms/ewald.jl
mfherbst: It's a bit of a mouth full. Also the lines are too long. Maybe restructure a bit?

mfherbst (commented):
Nice work!

src/terms/ewald.jl
mfherbst: (i know, but the reader of the code does not at that point :smile:)

src/terms/ewald.jl
niklasschmitz: The ceil(Int, x) is also making problems with IntervalArithmetic (failing the test), I think you had a workaround for a similar case before, should this be explicitly overloaded?

src/terms/ewald.jl
niklasschmitz: I took the sqrt on both sides of the inequality, replacing max_exp_arg by max_erfc_arg; maybe squares would be more natural though?

I now renamed the recip lattice in the comment as B.

Those in the comments are already the expressions we were bounding IIUC ?

mfherbst: Ah I see, but this is a bit "by accident" as `max_erfc_arg` only happens to be `sqrt(max_exp_arg)` because of the way we approximate it. I think explicit is better here.

Yeah, what I meant is that similar to the ti, tj thing it is not obvious to a code reader at that point why these are the expressions that matter. I think just explain this (stating that collected terms are of the form exp(- stuff) for real space and erfc(stuff) for reciprocal.

PR #662, opened 2022-06-02
Testing RealSpaceMultiplication consistency as well

epolack:


src/terms/local.jl
```diff
@@ -95,7 +95,7 @@ end
     ρ_fourier = r_to_G(basis, total_density(ρ))
 
     # energy = sum of form_factor(G) * struct_factor(G) * rho(G)
-    # where struct_factor(G) = cis(-2π G⋅r)
+    # where struct_factor(G) = e^{-i Gr}
```
mfherbst, line 98: Why? The original was correct, no?

test/hamiltonian_consistency.jl
```diff
@@ -51,6 +51,9 @@ function test_consistency_term(term; rtol=1e-4, atol=1e-8, ε=1e-6, kgrid=[1, 2,
         diff_predicted = 0.0
         for ik in 1:length(basis.kpoints)
             Hψ = ham.blocks[ik]*ψ[ik]
+            if ham.blocks[ik].operators isa DFTK.RealSpaceMultiplication
```
mfherbst, line 54: This is a bit more tricky as
- This will never be true as operators is a Vector of Operators.
- Some terms add two operators.
Just check there is only a RealSpaceMultiplication and then run the check.

epolack:
Don't understand why my comments went into "pending" state this time.

antoine-levitt:
Cool! Could you also test the other types? Eg do a try/catch and check that the catch is in a whitelist of functions for which we haven't implemened the method yet (MagneticFieldOperator and DivAgrad)

epolack:
Did the change. However, I am not sure how to treat the case where there are multiple operators for `RealSpaceMultiplication`.
`DFTK.RealSpaceMultiplication{Float64, Array{Float64, 3}}` is tested but not `DFTK.RealSpaceMultiplication{Float64, SubArray{Float64, 3, Array{Float64, 4}, Tuple{Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, Int64}, true}}`.

antoine-levitt:
? There's just one array, it's just that the array in question is a view

epolack:
Ah, but it does not seem to be recognised as a subtype of RealSpaceMultiplication.

epolack:
(with `isa` keyword)

epolack:
Err, the `operator_matrix = Matrix(block)` does not work. `isa` works.

src/terms/local.jl
mfherbst: Oh, I see. I would keep the `\cdot` and than say explicitly that `r` and `G` are in cartesian coordinates (because usually we use `G` with fractional).

test/hamiltonian_consistency.jl
```diff
@@ -7,6 +7,27 @@ include("testcases.jl")
 using Random
 Random.seed!(0)
 
+function test_matrix_repr_opererator(block, ψ, Hψ; atol=1e-8)
```
mfherbst, line 10: Our convention here is `hamk` and `ψk` as well as `Hψk`

test/hamiltonian_consistency.jl
```diff
@@ -7,6 +7,27 @@ include("testcases.jl")
...
+        @assert operator isa known_operators
+        operator_matrix = nothing
+        try
+            operator_matrix = Matrix(block)
+        catch
+            @info "Matrix of operator $(typeof(operator)) not yet supported"
+        end
+        if operator_matrix ≠ nothing
+            @test norm(operator_matrix * ψ - Hψ) < atol
+        end
```
mfherbst, line 27: Why not put into try ... catch?

epolack: I have put `δψkδHψk` also, but it's a bit a mouthful.

test/hamiltonian_consistency.jl
epolack: I have put `δψkδHψk` also, but it's a bit a mouthful.

mfherbst:
I was faster ;)

src/terms/local.jl
epolack: `G` is in Cartesian for `compute_forces` but not always for `AtomicLocal`. (https://github.com/JuliaMolSim/DFTK.jl/pull/662/commits/b9f1432cb51ead3a641ee8ad1f1319bd314fb9d9)

mfherbst: Nah we don't follow that consistently any more.

epolack:
This one is done too.

PR #663, opened 2022-06-08
Refactor Pairwise term to precompute summation bounds

niklasschmitz:
This PR follows up on #656 now analogously on the Pairwise term, replacing the shell_indices & while loop by a dense for-loop over space. Later, this can also be now easily modified further into mapreduce-like patterns for GPU or AD.

### DIFF
--- a/src/structure.jl
+++ b/src/structure.jl
@@ -42,3 +42,15 @@ function diameter(lattice::AbstractMatrix)
     end
     diam
 end
+
+"""
+Estimate integer bounds for dense space loops from a given inequality ||Mx|| ≤ δ.
+For 1D and 2D systems the limit will be zero in the auxiliary dimensions.
+"""
+function estimate_integer_lattice_bounds(M, δ, shift=zeros(3))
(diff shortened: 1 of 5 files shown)


src/structure.jl
```diff
@@ -42,3 +42,15 @@ function diameter(lattice::AbstractMatrix)
     end
     diam
 end
+
+"""
+Estimate integer bounds for dense space loops from a given inequality ||Mx|| ≤ δ.
+For 1D and 2D systems the limit will be zero in the auxiliary dimensions.
+"""
+function estimate_integer_lattice_bounds(M, δ, shift=zeros(3))
```
antoine-levitt, line 50: Can you possibly also use it in fft.jl? See bounding_rectangle.jl

antoine-levitt (commented):
Looks good! @epolack ?

src/structure.jl
epolack: Will look into that.

epolack:
Had to modify it a bit to not get NaN, otherwise good. I have rebased the complex version on this change.

PR #665, opened 2022-06-15
Fix performance gotcha

antoine-levitt:
Never trust anything that's not profiled :-) symmetrize_rho is still not ultra fast, but at least we get rid of the massive penalty I accidentally introduced. We really should setup performance tracking...

### DIFF
--- a/src/SymOp.jl
+++ b/src/SymOp.jl
@@ -40,7 +40,6 @@ end
 function Base.convert(::Type{SymOp{T}}, S::SymOp{U}) where {U <: Real, T <: Real}
     SymOp{T}(S.W, T.(S.w), S.S, T.(S.τ))
 end
-symop_type(T) = SymOp{T}
 
 Base.:(==)(op1::SymOp, op2::SymOp) = op1.W == op2.W && op1.w == op2.w
 function Base.isapprox(op1::SymOp, op2::SymOp; atol=SYMMETRY_TOLERANCE)


src/bzmesh.jl
```diff
@@ -44,8 +44,6 @@ given the crystal symmetries `symmetries`. Returns the list of irreducible ``k``
 the grid.
 """
 function bzmesh_ir_wedge(kgrid_size, symmetries; kshift=[0, 0, 0])
-    all(isequal.(kgrid_size, 1)) && return bzmesh_uniform(kgrid_size; kshift)
-
```
mfherbst, line 48: Why? This actually helps in many cases with special floating-point types.

antoine-levitt: I did this because bzmesh_uniform does not know about floating point types and therefore returns symmetries as boolean type, which causes trouble later. We should probably refactor these so these functions don't return symmetries and just filter them later, anyway

mfherbst: Does not seem to be a problem any more. What was the issue exactly?

mfherbst:
Ok, I made an attempt to fix it. Essentially my approach is to enforce that some things in the model cannot be dual numbers, like the symmetries, so we should just never store them as such.

antoine-levitt:
Hm I don't like this too much, it kind of leaks the forwarddiff abstraction. Let's discuss it in person!

PR #669, opened 2022-06-20
Add documentation about conducting a convergence study

killah-t-cell:
I was playing with doing a convergence study of a small system, and I remember you mentioned Abinit has a good doc for that, and I thought maybe DFTK should have it too. Feel free to say this isn't good enough to be part of the docs, but I just wanted to share the code :)


mfherbst:
Looks very nice, thank you very much @killah-t-cell! And I agree, we should have some docs on this.

Some ideas from a quick look:
- There are a few undefined variables. Be careful that `jl` files by the docs are interpreted like cells in a jupyter notebook and executed in order. Basically any comment creates a new cell, so that might rip statements apart if you are not careful. Have not checked the details, but as a pointer towards debugging.
- It would be nice to show the convergence graphically (i.e. just generate a log-plot of log(error) or `log(value[i+1] - value[i])` versus the parameter for Ecut or kgrid.
- Since the code is running on the CI, it's probably a good idea to not run all the way up to fully converged Ecut (would take too long). So you sort of have to stop early. Perhaps you can just include a static graph of the convergence plot to the document instead.

killah-t-cell:
kgrid
I added this type of graph, but I'm not sure how you guys store images for the docs, so this might still be wrong. Happy to change it (Documenter.jl's image settings confuse me haha)

killah-t-cell:
@mfherbst This should be ready for a review

mfherbst:
Hi @killah-t-cell thanks for your efforts!

Regarding adding the image to Documenter ... yes that's a bit tricky. Some examples for how we do this are in:
- https://github.com/JuliaMolSim/DFTK.jl/blob/master/docs/src/guide/parallelization.md
- https://github.com/JuliaMolSim/DFTK.jl/blob/master/examples/ase.jl
the former probably applies more to your case.

mfherbst:
I saw you're just using the assets ... that's of course also fine. Maybe give it a more descriptive name (like include a prefix of the file name of the example)? Also I would make the plots on a semilog scale (i.e. use logscale for the y axis and don't plot energies but energy differences (against a reference for large Ecut/kgrid).

examples/convergence_study.jl
```diff
@@ -0,0 +1,82 @@
+# # Performing a convergence study
+#
+# This example shows how to perform a convergence study to find the 
+# appropriate kgrid and Ecut for a desired convergence and error tolerance.
+#
+# Generally, to wisely vary kgrid and Ecut, one can perform a convergence study by
+# starting with a reasonable grid or Ecut, picking a property such as energy, and increasing the grid
```
mfherbst, line 7: Line a bit long ;)

examples/convergence_study.jl
```diff
@@ -0,0 +1,82 @@
...
+# until the property converges – its value does not change much. 
+
+# This study must be performed separately for each value you wish to study (kgrid, Ecut, and even temperature).
+# In this example we will use a Pt system to perform a convergence study for kgrid and Ecut in DFTK.
```
mfherbst, line 11: If this is related to an ASE example, you might as well link there.

examples/convergence_study.jl
```diff
@@ -0,0 +1,82 @@
...
+
+# Then we define a function to compute the total energy of the system for varying values of `Ecut` and `kgrid`
+function compute_Ecut_convergence(Ecut)
+    model = model_LDA(lattice, atoms,position; temperature)
+    basis = PlaneWaveBasis(model; Ecut, kgrid)
+    scfres = self_consistent_field(basis; tol=tol / 10)
+    return scfres.energies.total
+end
+
+function compute_kgrid_convergence(k; lat=:sc)
```
mfherbst, line 40: lat is unused

examples/convergence_study.jl
```diff
@@ -0,0 +1,82 @@
...
+Ecut = 490u"eV"
+temperature = 0.01
+Pt = ElementPsp(:Pt, psp=load_psp("hgh/lda/Pt-q10"));
+atoms = [Pt]
+position = [zeros(3)]
+a = 5.00 # in Bohr
+lattice = a * I(3)
+
+# Then we define a function to compute the total energy of the system for varying values of `Ecut` and `kgrid`
+function compute_Ecut_convergence(Ecut)
```
mfherbst, line 33: Why not have one function to construct the basis (which takes Ecut, kgrid, tol etc as input parameters) and move all the global variables inside? That's a lot cleaner.

examples/convergence_study.jl
```diff
@@ -0,0 +1,82 @@
...
+end
+
+# And since kgrid only takes discrete values of `k` in `[k,k,k]`, we define a custom function to define its convergence for a given desired tolerance `tol`
+function get_converged_k(energies; tol)
+    for i in eachindex(energies)
+        if abs(energies[i] - energies[i+1]) < tol
+            return i
+        end
+    end
+end
```
mfherbst, line 54: I would the `findfirst` function for that ... this makes this a one-liner.

examples/convergence_study.jl
```diff
@@ -0,0 +1,82 @@
...
+function compute_Ecut_convergence(Ecut)
+    model = model_LDA(lattice, atoms,position; temperature)
+    basis = PlaneWaveBasis(model; Ecut, kgrid)
+    scfres = self_consistent_field(basis; tol=tol / 10)
+    return scfres.energies.total
+end
+
+function compute_kgrid_convergence(k; lat=:sc)
+    model = model_LDA(lattice, atoms,position; temperature)
+    basis = PlaneWaveBasis(model; Ecut, kgrid=[Int64(ceil(k)) for _ in 1:3])
```
mfherbst, line 42: Why not just enforce `k` to be an Integer and then just say `kgrid=fill(k, 3)`?

examples/convergence_study.jl
```diff
@@ -0,0 +1,82 @@
...
+    end
+end
+
+# Now we can use `get_converged_k()` and `compute_kgrid_convergence()` to get the converging kgrid
+energies_w_diff_k = [compute_kgrid_convergence(k) for k in 1:7] # change to 10 to get full convergence like in plot
+k_conv = get_converged_k(energies_w_diff_k; tol) # 5.0
+
+# and plot it:
+plot(energies_w_diff_k, dpi=300, lw=3, xlabel="k-grid", ylabel="Energy/Atom", label="k")
+scatter!(energies_w_diff_k,label="Data points")
```
mfherbst, line 62: This will output the plot. Maybe thus add a sentence that if one does this properly (i.e. until larger kgrids) than the other plot you include statically results.

examples/convergence_study.jl
```diff
@@ -0,0 +1,82 @@
...
+
+#md # ```@raw html
+#md # <img src="../assets/kgrid.png" width=600 height=400 />
+#md # ```
+#nb # <img src="../assets/kgrid.png" width=600 height=400 />
+
+# And then use `optimize()` from `Optim.jl` to do the same for `Ecut`
+opt_res = optimize(compute_Ecut_convergence, 7, 22)
```
mfherbst, line 70: Nice idea! You can do this with the function you would define above by binding one argument using an annonymous function, e.g. `Ecut -> compute_Pt_sc(Ecut, tol, kgrid, ...)`

examples/convergence_study.jl
```diff
@@ -0,0 +1,82 @@
...
+# And then use `optimize()` from `Optim.jl` to do the same for `Ecut`
+opt_res = optimize(compute_Ecut_convergence, 7, 22) 
+E_cut_conv = Optim.minimizer(opt_res)
+E_cut_conv_H = auconvert(u"eV", E_cut_conv)
+
+# and finally plot Ecut:
+Ecut_convergence = [compute_Ecut_convergence(i) for i in  7:14] # change to 22 to get full convergence like in plot
+plot(Ecut_convergence, dpi=300, lw=3, xlabel = "Energy cutoff (Ha)", ylabel="Energy/Atom", label="Ecut")
+scatter!(Ecut_convergence,label="Data points")
+
```
mfherbst, line 78: Again same thing, add a comment that a more careful convergence leads to a different plot.

Another thing you can do is have the live-generated plots inline and then have another subsection 
```
## More realistic Ecut and kgrids
```
where you then show the more realistic plots one after another and state the differences ?

mfherbst:
@killah-t-cell Are you still working on this one? I'd be really glad to have this. Let me know if you need any help.

mfherbst:
Hi @killah-t-cell Would you mind if I take over polishing and merging this PR? I would love to have this ready in the docs for our summer school end of August. Of course more than happy if you prefer to do it yourself.

antoine-levitt:
@LaurentVidal95 also wanted to do something like that I think

LaurentVidal95:
Hi everyone! Indeed I had the project to do a note about the structures of the code for people in the lab and a tutorial about convergence study (after discussing with @antoine-levitt). However I had trouble not to repeat things already said in other tutorials so the notes I have are for now pretty informal (I did a small notebook). I will adapt considering this PR !

examples/convergence_study.jl
killah-t-cell: It's an example from Sholl's book https://www.amazon.com/Density-Functional-Theory-Practical-Introduction/dp/0470373172

examples/convergence_study.jl
killah-t-cell: Added a line about it.

killah-t-cell:
I wrote this during my first week with DFT, so I used optim to find the ideal converging Ecut. This might be a bit of an overkill, given that I can run this more like nkpts and just test `200:50:900` for example.

killah-t-cell:
Hey @mfherbst I don't mind you taking over it at all! Sorry it took so long to do this (I spent a bunch of time learning ASE, GPAW in the past few weeks). Feel free to add anything to this. I answered to your comments, but happy to take this anywhere!

killah-t-cell:
@antoine-levitt I'm running into an error with makedocs. Any chance you know what is going on?

antoine-levitt:
Yes, your diff has a large set of changes that is unrelated to your example, I'm assuming you messed up with git somehow? You should try to merge against master and resolve conflicts, and if that doesn't work just save your example to a file and make another PR in another branch and destroy this one...

killah-t-cell:
oh you know what happened? I pulled from origin/main instead of origin/master... fixing

mfherbst:
Superseeded by #701.

PR #670, opened 2022-06-23
Add more type annotations to model constructors

### LINKED ISSUE #668: Stronger type annotations for model construction
I was playing around with some atom positions and came across this error

```
ERROR: LoadError: MethodError: no method matching spglib_atoms(::Vector{Vector{Int64}}, ::Vector{Array{Float64}}, ::Vector{StaticArrays.SVector{3, Float64}})
```

Whenever I defined positions like so `positions = [[0. 0. 0.],[1/2,1/2,1/2]]` instead of like so `positions = [ones(3) * 0.0,ones(3) * 0.5]`. Because the former is a `Vector{Array{Float64}}` and the latter is a `Vector{Vector{Float64}}`. I wouldn't expect this to matter as a new user, but maybe you have a reason for it, so I just wanted to flag.

Example code
```julia
temperature = 1e-3
Sc = ElementPsp(:Sc, psp=load_psp("hgh/lda/Sc-q3"))
Al = ElementPsp(:Al, psp=load_psp("hgh/lda/Al-q3"))
atoms = [Sc, Al]
lattice = 1.0 * I(3)

positions = [[0. 0. 0.],[1/2,1/2,1/2]] # this will fail
model = model_LDA(lattice, atoms, positions; temperature)

positions = [[0. 0. 0.],[1/2,1/2,1/2]] # this will work
positions = [ones(3) * 0.0,ones(3) * 0.5]

```

mfherbst:
Fixes #668.

### DIFF
--- a/src/Model.jl
+++ b/src/Model.jl
@@ -90,12 +90,14 @@ If you want to pass custom symmetry operations (e.g. a reduced or extended set) 
 external potential breaks some of the passed symmetries. Use `false` to turn off
 symmetries completely.
 """
-function Model(lattice::AbstractMatrix{T}, atoms=Element[], positions=Vec3{T}[];
+function Model(lattice::AbstractMatrix{T},
+               atoms::Vector{<:Element}=Element[],
+               positions::Vector{<:AbstractVector}=Vec3{T}[];
                model_name="custom",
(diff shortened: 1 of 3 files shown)


antoine-levitt:
Great!

PR #675, opened 2022-06-28
Update constructor for positions and lattice in Model and basis

mfherbst:


src/PlaneWaveBasis.jl
```diff
@@ -316,6 +316,29 @@ Creates a new basis identical to `basis`, but with a custom set of kpoints
...
 end
 
+"""
+    PlaneWaveBasis(basis, lattice, positions)
+    PlaneWaveBasis(basis, lattice)
+    PlaneWaveBasis(basis, positions)
+
+Creates a new basis, identical to `basis`, but different `lattice` or `positions`.
+"""
+function PlaneWaveBasis(basis::PlaneWaveBasis, lattice::AbstractMatrix,
```
antoine-levitt, line 326: How about just PlaneWaveBasis(basis, model)?

src/Model.jl
```diff
@@ -175,6 +175,26 @@ function Model(lattice::AbstractMatrix{<:Quantity}, atoms::Vector{<:Element},
...
 
+"""
+    Model(model, lattice, positions)
+    Model(model, lattice)
+    Model(model, positions)
+
+Construct an identical model to `model` with the option to change `lattice` or `positions`
+(useful for geometry or lattice optimisations).
+"""
+function Model(model::Model{T}, lattice::AbstractMatrix{U},
```
antoine-levitt, line 186: In general, not a huge fan of this type of constructor (ie why change one field instead of another?). How about `Model(model; lattice, positions)`? (which we can then extend to other things if needed)

src/PlaneWaveBasis.jl
```diff
@@ -316,6 +316,29 @@ Creates a new basis identical to `basis`, but with a custom set of kpoints
...
+                        positions::Vector{<:AbstractVector}=basis.model.positions)
+    model = Model(basis.model, lattice, positions)
+    PlaneWaveBasis(model,
+                   basis.Ecut, basis.fft_size, basis.variational,
+                   basis.kcoords_global, basis.kweights_global,
+                   basis.kgrid, basis.kshift, basis.symmetries_respect_rgrid,
+                   basis.comm_kpts)
```
antoine-levitt, line 333: Hm this can have unintended consequences, ie the model can change symmetries, lattice leading to different fft_size, etc...

antoine-levitt: There are two possible intended behaviors here: either keep the basis as much the same as possible, or keep it as close to a new basis as possible. This sort of does neither. Not sure what to do here.

mfherbst: I think it should be as close to the new basis as possible, i.e. it should be a new consistent basis with the new structure. So yes, I think this should change the FFT size if needed. I did not think about that.

src/Model.jl
mfherbst: The advantage of not taking kwargs is to do slightly different think depending on what is passed. E.g. for changing kcoords we certainly do not want to change the FFT size, but for updating positions or lattice we certainly do. If we move all to a kwargs-based update constructor it's not so clear to do.

antoine-levitt: pwbasis is something else, but Model is a relatively simple container with not much going on in the constructor, so I think it definitely makes sense to have a Model(model; kwargs). I don't get your `Model(model)` argument?

src/PlaneWaveBasis.jl
antoine-levitt: Then I'm not sure this can be done sensibly, because we have different ways of constructing a pwbasis. I think we should move towards encapsulating the kpoint grid (and possibly the G grid ?) better,and then have a unique pwbasis(::Model, ::Kgrid) constructor, which would then make what you want to do here more straightforward.

antoine-levitt: By Kgrid I mean the specification (uniform/custom/symmetrized/shifted/..., with symmetries or not, not the actual instantiation) of the kgrid. Same for the G grid (fft_size, variational, )

mfherbst:
Outdated.

PR #676, opened 2022-06-28
Initial ideas for Aiida integration

mfherbst:
@azadoks Some initial fruit for discussion.

### DIFF
--- a/src/PlaneWaveBasis.jl
+++ b/src/PlaneWaveBasis.jl
@@ -283,7 +283,7 @@ end
         else
             factors = (1,)
         end
-        fft_size = compute_fft_size(model, Ecut, kcoords; factors=factors)
+        fft_size = compute_fft_size(model, Ecut, kcoords; factors)
     end
     PlaneWaveBasis(model, Ecut, fft_size, variational, kcoords, kweights,
                    kgrid, kshift, symmetries_respect_rgrid, comm_kpts)
--- a/src/external/aiida.jl
+++ b/src/external/aiida.jl
@@ -1,0 +1,79 @@
+module Aiida  # Special submodule for Aiida integration
+
+# TODO This should be integrated with the interatomic potentials stuff
+
+# Error handling:
+#    - unit tests for std error messages
+#
+# Geometry optimisation
+# Equation of state
+
+@kwdef struct DftkParameters
+    system::AbstractSystem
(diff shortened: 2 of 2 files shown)


mfherbst:
Not sure if DftkParameters should even exist. But let's lock down on some names for the parameters first and some basic structure for passing them and then we see how we fix that into concrete code.

mfherbst:
Maybe also makes sense to make use of as much of https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/external/interatomicpotentials.jl or to exploit similarities in the implementation.

azadoks:
It looks like you're doing a bit of sanity checking in code (e.g. that the smearing type is known).
This makes sense if the AiiDA plugin writes a `.jl` file for input (i.e. it _could_ write a file which creates a struct instance and then calls a sort of `main` function), but it might eventually lead to some fairly complicated logic to make sure that everything is consistent.

A different avenue would be to define a TOML/JSON schema which deals with making sure inputs are in a defined range, the correct shape, from a restricted list of keywords, etc. Then a TOML/JSON schema library will check all of this automatically and raise an appropriate error wherever inconsistencies/incorrectness is found. This would mean that the DFTK AiiDA interface could assume everything is correct and would just have to concern itself with setting up the system and running the requested calculation.

antoine-levitt:
> This makes sense if the AiiDA plugin writes a .jl file for input (i.e. it could write a file which creates a struct instance and then calls a sort of main function)

I know nothing about AiiDA and Michael will have a better idea about this than me, but have you considered using PyJulia or https://github.com/cjdoris/PythonCall.jl?

mfherbst:
Nice idea, but unfortunately that's not sufficient because where you manage Aiida workflows and where they run could be different machines (e.g. laptop versus cluster).

antoine-levitt:
OK; in that case there is packagecompiler and https://github.com/dmolina/DaemonMode.jl

mfherbst:
indeed, but deamonmode is also tricky for other reasons (when should the daemon be shut down?) For the first version we will be stupid and recompile DFTK from scratch in each calculation. Should not be too bad. The idea of Aiida is anyway to have larger chunks of work (geometry optimisation, band structure calculation etc.) as one complete building block.

mfherbst:
Closing this as this will be superseeded by an upcoming project going in a slightly different direction.

PR #677, opened 2022-07-03
DFTK Juliacon demo code

mtowara:
proposed code for zygote juliacon demo.

Still complains about the zygote over forward diff issue (I think this warning is only issued once per REPL session)


mfherbst:
Nice. What are the files in particular where you would want some feedback on?

mtowara:
I don't think there are that many changes compared to zygote-scf and the changes for abstractfunctional you already reviewed (the non mutating mergesum)
We just have to make sure that this branch is in place by the time of juliacon, such that the QR code for the demo works.
(the branch right now is only a stub so I could create a cross project PR)

Work still to do after JuliaCon (I'll be gone next two weeks)
- Parts of terms/xc still have to be refactored, but these are only triggered by gga.
- Investigate Zygote over ForwardDiff warnings

mfherbst:
Also the relation to #633 should be explored as this is the way forward.

Issue #652, opened 2022-04-13
Get rid of n_ep_extra


antoine-levitt:
Discussed today with @gkemlin. We should get rid of n_ep_extra and have n_bands take the place of what is currently n_bands+n_ep_extra.

What we should do is that self_consistent_field takes as input n_bands and min_occ. LOBPCG should compute on the n_bands, but only checks convergence for the occupied bands, ie those with an occupation greater than min_occ. scfres should not promise convergence except for the occupied bands. For DFPT, we should only compute the dpsi_n for the occupied bands. 

Compared to what's in the code now, we should have size(psi, 2) == n_bands (unlike what we have now which is size(psi, 2) == n_bands+n_ep_extra). Should be relatively straightforward, but LOBPCG will need to know about the occupation function (maybe via a callback like check_convergence)

(this is somewhat orthogonal to https://github.com/JuliaMolSim/DFTK.jl/issues/562, which is about the choice of n_band, and its possible adaptivity)

antoine-levitt:
I thought about this some more, it's a bit tricky. If we make convergence decisions based on occupation, we risk being pretty bad in the first SCF cycles (eg not check the convergence at all because the unconverged eigenvalues are high). We could recompute occupations and Fermi level at each step of the eigensolver to make convergence decisions, but that couples the different kpoints, which is annoying, and probably introduces nasty edge cases. So not completely clear what to do here...

gkemlin:
Couldn't we use what is implemented now for the first iterations of the SCF ? Because it seems to work well already. And then, once the first cycles are over, in a sense to be made precise, we can go on with what you proposed above ?

PR #679, opened 2022-07-05
Response calculations

gkemlin:
Small, but hopefully useful, PR which changes the way we deal with extra bands for response calculations. It should answer your question @niklasschmitz about this TODO https://github.com/JuliaMolSim/DFTK.jl/blob/8d1adf58f9ea6e88f31db1fbb96e0017f2f5bad7/src/workarounds/chainrules.jl#L226

What we wanted to do in https://github.com/JuliaMolSim/DFTK.jl/issues/652, with `n_ep_extra` that gets adapted on the fly during the SCF, is more tricky. So we moved to another idea that should help. 

It relies on the fact that we need these extra bands only in the sternheimer solver, so the later we discard the extra bands from `ψ`, the better.  We thus discard them in `apply_χ0_4P`, where we separate between "occupied" and "extras" bands according to the occupation number, given some occupation threshold (the same than in ResponseOptions from `self_consistent_field.jl`) 

For insulators, this is equivalent to `select_occupied_orbitals`. For metals, to ensure that the Schur decomposition in the sternheimer solver is valid, we need to ensure that orbitals with occupation number higher than the threshold are converged, which is already done in the SCF (when we call `compute_occupation`, where I changed the assertion `if temperature > 0 && minocc > 1e-5` with the threshold). The returned `δψ` is then zero-padded to have the same size than `ψ`.

I had to apply small changes in a lot of different places so there still might be some bugs or missing stuff but, as far as I tested, everything runs smoothly (I adapted `forwarddiff_rules.jl` too to get rid of `ψ_extra`). Here is a list of the main changes:
- moved `chi0.jl` and `omega.jl` to a folder called `response/`, and renamed `omega.jl` to `hessian.jl` TODO: find a better name for  `solve_ΩplusK`
- removed all `ψ_extra` and `ε_extra` arguments, except for sternheimer
- added this `occupation_threshold` where it is required, and `apply_χ0` now needs `occupation` as argument
- added tests for the self-adjointness of `solve_ΩplusK_split` and that it matches with `solve_ΩplusK` at convergence

The conclusion to this is that now we can forget about extra bands and call directly `solve_ΩplusK_split` :)


niklasschmitz:
Thanks for looking into this @gkemlin !  If I understand correctly this will simplify the SCF rrule considerably, looking forward to discuss tomorrow.

src/response/chi0.jl
```diff
@@ -255,16 +254,43 @@ end
...
     Nk = length(basis.kpoints)
 
+    # We first select orbitals with occupation number higher than
+    # occupation_threshold for which we compute the associated response δψn,
+    # the others being discarded to ψ_extra / ε_extra.
+    #
+    # Note:
+    # By default the `self_consistent_field` routine of `DFTK`
+    # returns `3` extra bands, which are not converged by the eigensolver
+    # (see `n_ep_extra` parameter) and have 0 occupation number. These should
```
antoine-levitt, line 264: not exactly 0

src/response/chi0.jl
```diff
@@ -255,16 +254,43 @@ end
...
+    # be discarded from `ψ`, along with bands that have small occupation
+    # numbers for metals, e.g. 1e-12 or less (and therefore small contributions
+    # to `δψ`), before computing the response `δψ` for the remaining states.
+    # However, one can still use additional information from these extra bands,
+    # stored in `ψ_extra`, in particular to enhance the convergence of the
+    # Sternheimer solver with a Schur complement in the orthogonal of `ψ`.
```
antoine-levitt, line 270: comment a bit verbose

src/response/chi0.jl
```diff
@@ -255,16 +254,43 @@ end
...
+
+    ψ_occ, occ_no_extra, ε_occ = select_occupied_orbitals(basis, ψ, occ, eigenvalues;
```
antoine-levitt, line 272: just occ_occ for consistency

src/response/hessian.jl
```diff
@@ -1,3 +1,24 @@
...
+# defined below (cf. [1] for more details).
+#
+# In particular, by linearizing the Kohn-Sham equations, we have
+#
+# δP = -(Ω+K)⁻¹ δH
+#
+# which can be solved either directly (solve_ΩplusK) or by a splitting method
+# (solve_ΩplusK_split), the latter being preferable as it is well defined for
+# both metals and insulators. Solving this equation is a necessary to compute
+# response properties as well as AD chain rules (Ω+K being self-adjoing, it can
```
antoine-levitt, line 11: right, I forgot you're from the south :-)

src/occupation.jl
```diff
@@ -24,6 +24,7 @@ end
 Find the occupation and Fermi level.
 """
 function compute_occupation(basis::PlaneWaveBasis{T}, eigenvalues;
+                            occupation_threshold=1e-10,
```
antoine-levitt, line 27: or require it explicit?

src/response/hessian.jl
```diff
@@ -102,12 +123,14 @@ end
 # where χ02P = R χ04P E and K2P = R K E
 function solve_ΩplusK_split(ham::Hamiltonian, ρ::AbstractArray{T}, ψ, occupation, εF,
                             eigenvalues, rhs; tol=1e-8, tol_sternheimer=tol/10,
-                            verbose=false, kwargs...) where T
+                            verbose=false, occupation_threshold=1e-10,
```
antoine-levitt, line 126: require it explicitly

src/scf/self_consistent_field.jl
```diff
@@ -102,7 +105,7 @@ Solve the Kohn-Sham equations with a SCF algorithm, starting at ρ.
         # Diagonalize `ham` to get the new state
         nextstate = next_density(ham; n_bands, ψ, eigensolver,
                                  miniter=1, tol=determine_diagtol(info),
-                                 n_ep_extra)
+                                 n_ep_extra, response.occupation_threshold)
```
antoine-levitt, line 108: Didn't we decide to put it as a self_consistent_field kwarg and not in response?

gkemlin:
Now that `occupation_threshold` is a kwarg not assigned by default (except in SCF routines with `default_occupation_threshold()` which we set to 1e-10, I had to modifiy all the tests that use `compute_occupation`. Note that we've lowered the test from 1e-5 to 1e-10 here
https://github.com/JuliaMolSim/DFTK.jl/blob/61a6845fc115fdcab2d392680000a3a56d446035/src/occupation.jl#L93
which causes some warnings in the test log files, but I'm not sure it is an issue (or we should just increase the number of bands int these tests).

antoine-levitt:
Or just use a high threshold in these tests to disable the warning, we don't really care (we use a low number of bands to have fast tests, but we don't care that it's not converged)

gkemlin:
> Or just use a high threshold in these tests to disable the warning, we don't really care (we use a low number of bands to have fast tests, but we don't care that it's not converged)

Yes, that's another possibility, I'll do that !

gkemlin:
Haha forgot to remove this in sternheimer, reruning the tests
```
    #
    # /!\ This is only implemented for insulators at the moment, WIP for
    # metals, in which case we use empty arrays and all computations are
    # equivalent to invert the Sternheimer system in Ran(1-P) without the Schur
    # trick.

    # TODO
    # finite temperature not supported yet => remove extra bands
    if !iszero(temperature)
        ψk_extra = zeros(size(ψk, 1), 0)
        εk_extra = zeros(0)
    end
```

src/scf/self_consistent_field.jl
```diff
@@ -33,7 +35,8 @@ function next_density(ham::Hamiltonian;
     eigres.converged || (@warn "Eigensolver not converged" iterations=eigres.iterations)
 
     # Update density from new ψ
-    occupation, εF = compute_occupation(ham.basis, eigres.λ)
+    occupation, εF = compute_occupation(ham.basis, eigres.λ;
+                                        occupation_threshold)
```
mfherbst, line 39: Why linebreak?

src/scf/self_consistent_field.jl
```diff
@@ -4,22 +4,24 @@ include("scf_callbacks.jl")
...
     verbose = false
-    occupation_threshold = 1e-10
 end
 
 function default_n_bands(model)
     min_n_bands = div(model.n_electrons, filled_occupation(model), RoundUp)
     n_extra = model.temperature == 0 ? 0 : max(4, ceil(Int, 0.2 * min_n_bands))
     min_n_bands + n_extra
 end
+default_occupation_threshold() = 1e-10
```
mfherbst, line 14: What if we are using single precision or other precisions? It feels there needs to be a floating-point type argument passed here.

src/scf/self_consistent_field.jl
```diff
@@ -59,7 +62,8 @@ Solve the Kohn-Sham equations with a SCF algorithm, starting at ρ.
                                        is_converged=ScfConvergenceEnergy(tol),
                                        callback=ScfDefaultCallback(; show_damping=false),
                                        compute_consistent_energies=true,
-                                       response=ResponseOptions()  # Dummy here, only for AD
+                                       response=ResponseOptions(),  # Dummy here, only for AD
+                                       occupation_threshold=default_occupation_threshold() # 1e-10
```
mfherbst, line 66: switch round. I think the response options should be last.

src/scf/self_consistent_field.jl
```diff
@@ -118,7 +122,7 @@ Solve the Kohn-Sham equations with a SCF algorithm, starting at ρ.
         info = merge(info, (energies=energies, ))
 
         # Apply mixing and pass it the full info as kwargs
-        δρ = mix_density(mixing, basis, ρout - ρin; info...)
+        δρ = mix_density(mixing, basis, ρout - ρin; occupation_threshold, info...)
```
mfherbst, line 125: just put occupation_threshold into the info. It might be useful for other callbacks (e.g. the regular one where we do printing) as well.

src/workarounds/forwarddiff_rules.jl
```diff
@@ -245,7 +238,7 @@ function self_consistent_field(basis_dual::PlaneWaveBasis{T};
 
     # TODO Compute eigenvalue response (return dual eigenvalues and dual εF)
 
-    merge(truncated, (; ham=ham_dual, basis=basis_dual, energies=energies_dual, ψ=ψ_out,
+    merge(scfres, (; ham=ham_dual, basis=basis_dual, energies=energies_dual, ψ=ψ_out,
                         occupation=occupation_dual, ρ=ρ_out, response))
```
mfherbst, line 242: indention

test/chi0.jl
```diff
@@ -36,9 +37,11 @@ function test_chi0(testcase; symmetries=false, temperature=0,
         basis = PlaneWaveBasis(model; basis_kwargs...)
         ρ0 = guess_density(basis, magnetic_moments)
         energies, ham0 = energy_hamiltonian(basis, nothing, nothing; ρ=ρ0)
-        res = DFTK.next_density(ham0; tol, n_ep_extra, eigensolver)
-        occ, εF = DFTK.compute_occupation(basis, res.eigenvalues)
-        scfres = (ham=ham0, res..., n_ep_extra)
+        res = DFTK.next_density(ham0; tol, n_ep_extra, eigensolver,
+                                occupation_threshold)
+        occ, εF = DFTK.compute_occupation(basis, res.eigenvalues;
+                                          occupation_threshold)
```
mfherbst, line 43: no linebreaks

test/chi0.jl
```diff
@@ -36,9 +37,11 @@ function test_chi0(testcase; symmetries=false, temperature=0,
...
+        scfres = (ham=ham0, res..., n_ep_extra,occupation_threshold)
```
mfherbst, line 44: space

test/compute_density.jl
```diff
@@ -17,21 +17,22 @@ if mpi_nprocs() == 1  # not easy to distribute
             kwargs = (temperature=testcase.temperature, smearing=DFTK.Smearing.FermiDirac())
             n_bands = div(testcase.n_electrons, 2, RoundUp) + 4
         end
+        occupation_threshold = 1e-5
```
mfherbst, line 20: That's rather large. Is this on purpose?

test/interval_arithmetic.jl
```diff
@@ -47,7 +47,8 @@ end
 
     eigenvalues = [[-0.17268859, 0.26999098, 0.2699912, 0.2699914, 0.35897297, 0.3589743],
                    [-0.08567941, 0.00889772, 0.2246137, 0.2246138, 0.31941655, 0.3870046]]
-    occupations, εF = DFTK.compute_occupation(basis, Vector{Interval{Float64}}.(eigenvalues))
+    occupations, εF = DFTK.compute_occupation(basis, Vector{Interval{Float64}}.(eigenvalues);
+                                              occupation_threshold=1e-5)
```
mfherbst, line 51: again surprisingly large

test/occupation.jl
```diff
@@ -34,6 +34,7 @@ if mpi_nprocs() == 1 # can't be bothered to convert the tests
     Ecut = 5
     n_bands = 10
     fft_size = [15, 15, 15]
+    occupation_threshold = 1e-5
```
mfherbst, line 37: large ?

test/omegaplusk.jl
```diff
@@ -10,10 +11,10 @@ include("testcases.jl")
     fft_size = [9, 9, 9]
     model = model_DFT(silicon.lattice, silicon.atoms, silicon.positions, [:lda_xc_teter93])
     basis = PlaneWaveBasis(model, Ecut, silicon.kcoords, silicon.kweights; fft_size)
-    scfres_start = self_consistent_field(basis; tol=10)
+    scfres = self_consistent_field(basis; tol=10)
 
-    ψ, occupation = DFTK.select_occupied_orbitals(basis, scfres_start.ψ,
-                                                  scfres_start.occupation)
+    ψ, occupation = select_occupied_orbitals(basis, scfres.ψ,
+                                             scfres.occupation)
```
mfherbst, line 17: also no too sure about this line break

test/omegaplusk.jl
```diff
@@ -49,4 +50,61 @@ include("testcases.jl")
...
+        ψ = scfres.ψ
+        rhs = compute_projected_gradient(basis, scfres.ψ, scfres.occupation)
+        ϕ = rhs + ψ
+
+        @testset "self-adjointness of solve_ΩplusK_split" begin
+            @test isapprox(
+                           real(dot(ϕ, solve_ΩplusK_split(scfres, rhs).δψ)),
+                           real(dot(solve_ΩplusK_split(scfres, ϕ).δψ, rhs)),
+                           atol=1e-7
+                          )
```
mfherbst, line 70: compactify this a little

test/omegaplusk.jl
```diff
@@ -49,4 +50,61 @@ include("testcases.jl")
...
+        end
+
+        @testset "solve_ΩplusK_split <=> solve_ΩplusK" begin
+            scfres = self_consistent_field(basis; tol=1e-10)
+            ψ, occupation = select_occupied_orbitals(basis, scfres.ψ,
+                                                     scfres.occupation)
```
mfherbst, line 76: linebreak

test/omegaplusk.jl
```diff
@@ -49,4 +50,61 @@ include("testcases.jl")
...
+        ψ = scfres.ψ
+        rhs = compute_projected_gradient(basis, scfres.ψ, scfres.occupation)
+        ϕ = rhs + ψ
+
+        @testset "self-adjointness of solve_ΩplusK_split" begin
+            @test isapprox(
+                           real(dot(ϕ, solve_ΩplusK_split(scfres, rhs).δψ)),
+                           real(dot(solve_ΩplusK_split(scfres, ϕ).δψ, rhs)),
+                           atol=1e-7
+                          )
```
mfherbst, line 104: compactify

mfherbst (commented):
Nice!

src/occupation.jl
```diff
@@ -24,7 +19,8 @@ end
 Find the occupation and Fermi level.
 """
 function compute_occupation(basis::PlaneWaveBasis{T}, eigenvalues;
-                            temperature=basis.model.temperature) where {T}
+                            temperature=basis.model.temperature,
+                            occupation_threshold) where {T}
```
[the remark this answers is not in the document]
gkemlin: Well, this is what was done before but @antoine-levitt did not seem so satisfied with it and suggested to force the user to specify the threshold. Also, setting 1e-5 by default is quite high to compute responses no ?

src/response/chi0.jl
```diff
@@ -255,16 +244,43 @@ end
...
+    # the others being discarded to ψ_extra / ε_extra.
+    # We then use the extra information we have from these additional bands,
+    # non-necessarily converged, to split the sternheimer_solver with a Schur
+    # complement.
+
+    ψ_occ   = []
+    ε_occ   = []
+    occ_occ = []
+    ψ_extra = []
+    ε_extra = []
```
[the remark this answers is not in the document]
gkemlin: Yes, indeed for the types, I'll see what I can do. Is it really necessary to use @ view with the  @ views macro in the function definition ?

src/occupation.jl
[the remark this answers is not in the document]
mfherbst: Ok, true. To be honest what the default is does not matter so much. I just think that in top-level functions which are useful to quickly run for debugging and to get insight having to worry about such a numerical detail is annoying. We also don't force the user to specify a tolerance each time for running a solver, but have defaults everywhere.

src/response/chi0.jl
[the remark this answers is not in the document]
mfherbst: >  I'll see what I can do. I

Sure. I'm happy to do a pass if you struggle. 

> Is it really necessary to use @ view with the @ views macro in the function definition ?

No it's not. I overlooked that :sweat_smile:

src/occupation.jl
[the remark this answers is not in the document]
gkemlin: Yes, I see what you meen for the cohen-bergstrasser example. 

I don't have a preference, so if it's suits you @antoine-levitt too, we can keep occupation_threshold as a kwarg with default value, set for instance to 1e-7 ? (or back to 1e-5 as it was before)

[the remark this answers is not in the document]
antoine-levitt: My reasoning was that it's really annoying to hardcode defaults at several places, so you're never really sure what values are used... also these are relatively internal functions (the Cohen-Bergstresser example is directly plugging into internal functions). Longer-term we might want to store this kind of parameters in a struct...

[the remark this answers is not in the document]
mfherbst: I agree on the struct bit. I think we should have some "ScfOptions" or something alike. 

Ok, I'm ok with forcing the user to specify something on internal functions, but I find it weird on exported functions. I'd be ok for now to draw the line there.

test/compute_density.jl
gkemlin: well, it was the previous value harcoded in `compute_occupation` :sweat_smile: and yes it is on purpose, for testing we don't really need a strong threshold, do we ?

src/occupation.jl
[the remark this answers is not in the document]
gkemlin: Okay, so I have set 1e-7 default value for apply_\chi0, compute_\chi0 and the and the different SCF routines, and left it as unassigned kwarg for the others.

test/occupation.jl
```diff
@@ -87,6 +88,7 @@ if mpi_nprocs() == 1 # can't be bothered to convert the tests
     Ecut = 5
     fft_size = [15, 15, 15]
     kgrid  = [2, 3, 4]
+    occupation_threshold = 1e-5
```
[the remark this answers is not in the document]
gkemlin: woups sorry

PR #687, opened 2022-07-12
Fix bug on magnetization in guess_density

gkemlin:
Short fix for two things : 
- cleaning some leftovers in sternheimer + some useful doc
- fixing a bug in guess_density https://github.com/JuliaMolSim/DFTK.jl/blob/4f810fa0e9806659a12ea7c4d15579475d343b10/src/guess_density.jl#L65 where it can happen that we have `atoms = [A, B, C]` with `magnetic_moments = [m, 0, m]` with `m>0`. then, the filter will remove the 0 from `magnetic_moments` and raise an error later because the correspondance between `atoms` and `magnetic_moments` lists is not valid anymore.

### DIFF
--- a/src/guess_density.jl
+++ b/src/guess_density.jl
@@ -62,7 +62,7 @@ function _guess_spin_density(basis::PlaneWaveBasis{T}, atoms, positions, magneti
 
     # If no magnetic moments start with a zero spin density
     magmoms = Vec3{T}[normalize_magnetic_moment(magmom) for magmom in magnetic_moments]
-    if isempty(filter(!iszero, magmoms))
+    if all(iszero, magmoms)
         @warn("Returning zero spin density guess, because no initial magnetization has " *
               "been specified in any of the given elements / atoms. Your SCF will likely " *
               "not converge to a spin-broken solution.")
@@ -78,6 +78,7 @@ function _guess_spin_density(basis::PlaneWaveBasis{T}, atoms, positions, magneti
         )
         magmom[3], T(atom_decay_length(atom))::T, position
     end
+    gaussians = filter(g -> !iszero(g[1]), gaussians)
     gaussian_superposition(basis, gaussians)
 end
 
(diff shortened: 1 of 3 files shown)


src/guess_density.jl
```diff
@@ -62,15 +62,14 @@ function _guess_spin_density(basis::PlaneWaveBasis{T}, atoms, positions, magneti
 
     # If no magnetic moments start with a zero spin density
     magmoms = Vec3{T}[normalize_magnetic_moment(magmom) for magmom in magnetic_moments]
-    magmoms = filter(!iszero, magmoms)
-    if isempty(magmoms)
+    if isempty(filter(!iszero, magmoms))
```
antoine-levitt, line 65: `all(iszero, magmoms)`

src/response/chi0.jl
```diff
@@ -107,13 +107,14 @@ precondprep!(P::FunctionPreconditioner, ::Any) = P
 # Solves (1-P) (H-εn) (1-P) δψn = - (1-P) rhs
 # where 1-P is the projector on the orthogonal of ψk
 # n is used for the preconditioning with ψk[:,n] and the optional callback
+# /!\ for the solver to return the good output, εk_extra needs to be equal to
+# the diagonal of ψk_extra'Hk*ψk_extra (when not empty), which is the case by
+# default when using the extra eigenvalues
```
antoine-levitt, line 112: This is also true for the eps_nk, right? Say something like "It is assumed (and not checked) that psi'Hpsi = Diagonal(eps)" (leaving implicit that psi must be understood as the union of the psik and psik_extra...)

antoine-levitt (commented):
Good catch!

src/response/chi0.jl
gkemlin: (what I mean is that eps_nk = psi_nk'Hkpsi_nk matters for the physics but nor for the algebra)

mfherbst:
Good catch. Maybe also add a filter before

https://github.com/JuliaMolSim/DFTK.jl/blob/2ecff4236c735db20d66446bf086e7da181f3916/src/guess_density.jl#L81

that filters out the zero Gaussian coefficients.

src/response/chi0.jl
antoine-levitt: Ah, right, sorry. Then yeah OK. Your formulation still says diag(psi'Hpsi) = eps, which is weaker than psi'Hpsi = Diagonal(eps).

PR #688, opened 2022-07-14
Band energy derivatives using ForwardDiff

mfherbst:


src/workarounds/forwarddiff_rules.jl
```diff
@@ -235,10 +234,20 @@ function self_consistent_field(basis_dual::PlaneWaveBasis{T};
...
+    ## Build consistent Hamiltonian and band energies
+    energies_out, ham_out = energy_hamiltonian(basis_dual, ψ_out, occupation_dual;
+                                               ρ=ρ_out, eigenvalues=eigenvalues_dual,
+                                               εF=εF_dual)
+
+    eigenvalues_out = map(scfres.eigenvalues, ψ_out, ham_out * ψ_out) do εk, ψk, Hψk
+        # Primal and 1st order change due to updated orbitals
+        # The second term accounts for the loss of orthogonality in the
+        # orbitals due to the perturbation (We don't enforce δψnk ⟂ ψnk)
+        eigvals(Hermitian(ψk'Hψk + Diagonal(εk) * (I - ψk'ψk)))
```
antoine-levitt, line 246: That looks convoluted, are you sure this is correct and if yes can you expand on the rationale? Otherwise you can do `eigvals(psi' H psi, psi' psi)`. In any case this should be done at the level of the response solver, since that's where we decide on conventions. We can either decide <dpsink, psink> = 0, depsnk = <psink, dH psink>, or depsnk=0 and <dpsink, psink> != 0. I'd favor the first solution; in which case the response solvers should output depsnk as well as dpsi

antoine-levitt: Also here you're taking the Hermitian of a non-hermitian matrix, which is tricky

antoine-levitt: cc @gkemlin

mfherbst: > `eigvals(psi' H psi, psi' psi)`

You can, but only using GenericLinearAlgebra, which I'd like to avoid. The answer is the same, though.

> should be done at the level of the response solver,

Yes, I agree, but the whole point of our response setup is exactly to have <dpsink, psink> != 0, right? Or do you mean to "undo" that convention before returning the response result in this function.

Also why is the matrix not Hermitian? I don't see that (but it might be too early).

gkemlin: Yes indeed, the whole point of the response solver is to make another gauge choice than <dpsink, psink> = 0 for metals

antoine-levitt: > Also why is the matrix not Hermitian? I don't see that (but it might be too early).

It's not unless the psik are eigenvectors (but it might also be to first order, I haven't checked)

OK, you're right, I think I was confusing with the dfnk, sorry. In that case we can still return the deps_nk from the response solver, with the understanding that psink+dpsink might not be an eigenvector. I would be more comfortable if we just computed the deps_nk from <psink, dH psink> than from this relatively expensive and nontrivial formula. (note dH has to be the total hamiltonian, not the external one)

gkemlin: I think that we also defined deps_nk as <psink, dH psink> back in Aachen, which also comes from the first order perturbation no ?

mfherbst: @gkemlin Yes, but that's not what you get with finite differences unfortunately. My intention was to match the finite differences result (see the updated test) and for that I had to update the Hamiltonian once more.

mfherbst: Ok scratch that. I went the other way to use the result from solve_Omega_plus_K

src/workarounds/forwarddiff_rules.jl
```diff
@@ -191,54 +191,69 @@ function construct_value(basis::PlaneWaveBasis{T}) where {T <: ForwardDiff.Dual}
...
+
+    ## Compute external perturbing potential δV_{ext} (contained in δH)
+    ## and form the matrix-vector product with the bands
+    δH, δHψ = let
+        occupation_dual = [T.(occk) for occk in scfres.occupation]
+        ψ_dual = [Complex.(T.(real(ψk)), T.(imag(ψk))) for ψk in scfres.ψ]
+        ρ_dual = DFTK.compute_density(basis_dual, ψ_dual, occupation_dual)
+        εF_dual = T(scfres.εF)  # Only needed for entropy term
+        eigenvalues_dual = [T.(εk) for εk in scfres.eigenvalues]
+        _, δH = energy_hamiltonian(basis_dual, ψ_dual, occupation_dual;
```
antoine-levitt, line 212: So all this is tricky because this is not really a dH by itself, right? It's a Dual(H, dH). So, in keeping with the rest, dH -> H_dual?

src/workarounds/forwarddiff_rules.jl
```diff
@@ -191,54 +191,69 @@ function construct_value(basis::PlaneWaveBasis{T}) where {T <: ForwardDiff.Dual}
...
-        δHψ_α = [ForwardDiff.partials.(δHψk, α) for δHψk in hamψ_dual]
+        δHψ_α = [ForwardDiff.partials.(δHψk, α) for δHψk in δHψ]
+
+        δψ, resp = solve_ΩplusK_split(scfres, -δHψ_α; tol=scfres.norm_Δρ, response.verbose)
+        δρ = compute_δρ(basis_primal, scfres.ψ, δψ, scfres.occupation)
+
+        # TODO Stuff this back into δH to build a δH, by amending the stored total local
+        #      potential. This forms a δH, which has the total variation, which could be
+        #      returned.
+        δVresp   = apply_kernel(basis_primal, δρ; scfres.ρ)  # response potential ε^{-1} V_{ext}
```
antoine-levitt, line 228: Hm, I'd rather have this be in solve_omegaplusk_split, because that logic belongs there

antoine-levitt: also this is eps^-1 dV - dV, right?

src/workarounds/forwarddiff_rules.jl
```diff
@@ -191,54 +191,69 @@ function construct_value(basis::PlaneWaveBasis{T}) where {T <: ForwardDiff.Dual}
...
+
+        δevalues = map(basis_primal.kpoints, scfres.ψ, δHψ_α) do kpt, ψk, δHψk
```
antoine-levitt, line 230: evalues is not something we use elsewhere

src/workarounds/forwarddiff_rules.jl
```diff
@@ -191,54 +191,69 @@ function construct_value(basis::PlaneWaveBasis{T}) where {T <: ForwardDiff.Dual}
...
+            # compute δε_{nk} = <ψnk | δVtot | ψnk> = <ψnk | δVext + δVresp | ψnk>
+            map(eachcol(ψk), eachcol(δHψk)) do ψnk, δHψnk
+                # TODO Neither memory efficient nor properly parallelised over k-points
+                ψnk_real = G_to_r(basis_primal, kpt, ψnk)
+                sum(δVresp .* abs2.(ψnk_real)) * basis_primal.dvol + real(dot(ψnk, δHψnk))
```
antoine-levitt, line 235: same, can't this be done in solve_omegaplusK?

antoine-levitt: we already return a namedtuple there, we can just add drho, dVresp, devals, etc

src/workarounds/forwarddiff_rules.jl
```diff
@@ -191,54 +191,69 @@ function construct_value(basis::PlaneWaveBasis{T}) where {T <: ForwardDiff.Dual}
...
-                                                 ρ=ρ_dual, eigenvalues=eigenvalues_dual,
-                                                 εF=εF_dual)
-
-    response.verbose && println("Solving response problem")
+    basis_primal  = construct_value(basis_dual)
+    scfres = self_consistent_field(basis_primal; kwargs...)
+
+    ## Compute external perturbing potential δV_{ext} (contained in δH)
+    ## and form the matrix-vector product with the bands
+    δH, δHψ = let
```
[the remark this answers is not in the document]
antoine-levitt: the fact that you use let is a sign this function is too long and too involved

antoine-levitt (commented):
Content looks OK, but I think it'd be nicer to move some of the code around

mfherbst:
I like this now very much. This structure makes a lot more sense. In particular since pretty much all quantities were implicitly computed inside the solve_\Omega+K anyway, so this version effectively causes no computational overhead over master.

antoine-levitt:
Great! I like the name "induced"

src/workarounds/forwarddiff_rules.jl
```diff
@@ -191,54 +191,53 @@ function construct_value(basis::PlaneWaveBasis{T}) where {T <: ForwardDiff.Dual}
...
+    end
 
-    # TODO Compute eigenvalue response (return dual eigenvalues and dual εF)
+    # TODO Could add δresults[α].δVind the dual part of the total local potential in ham_dual
+    # and in this way return a ham that represents also the total change in Hamiltonian
 
-    merge(scfres, (; ham=ham_dual, basis=basis_dual, energies=energies_dual, ψ=ψ_out,
-                   occupation=occupation_dual, ρ=ρ_out, response))
+    merge(scfres, (; ψ, ρ, eigenvalues, basis=basis_dual,
+                   response=getfield.(δresults, :history)))
```
niklasschmitz, line 240: This is missing ham_dual, energies_dual, occupation_dual ?

PR #689, opened 2022-07-18
Add code for paper by Cazalis on 2D-confined graphene

antoine-levitt:
It would be cool to have a repository of small code samples that are used in various papers to 1) show off DFTK's uses 2) add to testsuite coverage 3) ensure low-effort reproducibility (no need to create separate repos). This would be reserved for small projects, larger ones would have a separate repo (maybe we could create a file in the examples/ folder with a link to it)

(CC @gkemlin @LaurentVidal95 @lgarrigue )

### DIFF
--- a/examples/papers/cazalis.jl
+++ b/examples/papers/cazalis.jl
@@ -10,7 +10,7 @@ struct Hartree2D end
 struct Term2DHartree <: DFTK.TermNonlinear end
 (t::Hartree2D)(basis) = Term2DHartree()
 function DFTK.ene_ops(term::Term2DHartree, basis::PlaneWaveBasis{T},
-                 ψ, occ; ρ, kwargs...) where {T}
+                      ψ, occ; ρ, kwargs...) where {T}
     ## 2D Fourier transform of 3D Coulomb interaction 1/|x|
     poisson_green_coeffs = 2T(π) ./ [norm(G) for G in G_vectors_cart(basis)]
     poisson_green_coeffs[1] = 0  # DC component
@@ -23,7 +23,7 @@ function DFTK.ene_ops(term::Term2DHartree, basis::PlaneWaveBasis{T},
     (E=E, ops=ops)
 end
 
-## define el-nucl interaction
+## define electron-nuclei interaction
 struct Element2DCoulomb <: DFTK.Element end
 function DFTK.local_potential_fourier(el::Element2DCoulomb, q::Real)
     ## = ∫ V(r) exp(-ix⋅q) dx
@@ -35,14 +35,14 @@ function DFTK.local_potential_fourier(el::Element2DCoulomb, q::Real)
 end
 
 L = 1.0
-kgrid = [10, 10, 1] # increase this for production
-Ecut = 500 # increase this for production
+kgrid = [10, 10, 1]  # increase this for production
+Ecut = 500  # increase this for production
 height = 0
 
 ## Define the geometry and pseudopotential
-a1 = L*[1/2, -sqrt(3)/2, 0]
-a2 = L*[1/2,  sqrt(3)/2, 0]
-a3 = height*[0, 0, 1]
+a1 = L * [1/2, -sqrt(3)/2, 0]
+a2 = L * [1/2,  sqrt(3)/2, 0]
+a3 = height * [0, 0, 1]
 lattice = [a1 a2 a3]
 C1 = [1/3, -1/3, 0.0]  # in reduced coordinates
 C2 = -C1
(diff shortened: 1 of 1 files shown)


gkemlin:
I think that large projects (those that have been created specifically for a paper) are listed in https://juliamolsim.github.io/DFTK.jl/stable/publications/
What you mean would be a separate repo, with smaller projects but that are maybe too advanced for the example sections of the documentation ?

antoine-levitt:
For this we would eg create a examples/papers/cances-dusson-kemlin-levitt.txt file that would simply have a link towards https://github.com/gkemlin/paper-forces-estimator, with possibly a one-line description of what's going on there. So small projects go straight into the DFTK repo (and are included with CI), and larger projects are centralized somewhere. Then we modify that section on publications to point to the examples/papers/

mfherbst:
> have a repository of small code samples that are used in various papers

I'm very critical of that, because:
- either you take the point of view of ensuring all examples work with DFTK forever, then the examples might as well go to the DFTK repo itself to ensure via DFTK's CI they always work. Then they also contribute to coverage as you have mentioned.
- or you only guarantee to have them working with DFTK at some point. Then the example should go to its separate repo with Project.toml and Manifest.toml. 
- As a compromise one could have an example repository with subfolders for each paper. But to be honest I think it's more realistic that people either contribute their code directly to DFTK or to their own repo (where they also do other stuff) than to a separate third repository of DFTK examples. At least for the moment where we still have a rather limited set of users I'd just have DFTK for small examples and separate repos for more involved stuff, which are linked from DFTK's docs.

examples/papers/cazalis.jl
```diff
@@ -0,0 +1,91 @@
...
+
+    ρtot_fourier = r_to_G(basis, total_density(ρ))
+    pot_fourier = poisson_green_coeffs .* ρtot_fourier
+    pot_real = G_to_r(basis, pot_fourier)
+    E = real(dot(pot_fourier, ρtot_fourier) / 2)
+    ops = [DFTK.RealSpaceMultiplication(basis, kpt, pot_real) for kpt in basis.kpoints]
+    (E=E, ops=ops)
+end
+
+## define el-nucl interaction
```
mfherbst, line 26: writing that out won't kill you :smile:.

examples/papers/cazalis.jl
```diff
@@ -0,0 +1,91 @@
...
+
+L = 1.0
+kgrid = [10, 10, 1] # increase this for production
+Ecut = 500 # increase this for production
+height = 0
+
+## Define the geometry and pseudopotential
+a1 = L*[1/2, -sqrt(3)/2, 0]
+a2 = L*[1/2,  sqrt(3)/2, 0]
+a3 = height*[0, 0, 1]
```
mfherbst, line 45: space around `*`

examples/papers/cazalis.jl
```diff
@@ -0,0 +1,91 @@
...
+## Run SCF
+scfres = self_consistent_field(basis, tol=1e-10)
+
+## Choose the points of the band diagram, in reduced coordinates (in the (b1,b2) basis)
+Γ = [0, 0, 0]
+K = [ 1, 1, 0]/3
+Kp= [-1, 2, 0]/3
+M = (K + Kp)/2
+kpath_coords = [Γ, K, M, Γ]
+kpath_names = ["Γ", "K", "M", "Γ"]
```
mfherbst, line 71: the formatting of the current dftk graphene example is nicer here.

antoine-levitt:
> either you take the point of view of ensuring all examples work with DFTK forever, then the examples might as well go to the DFTK repo itself to ensure via DFTK's CI they always work. Then they also contribute to coverage as you have mentioned.

That's what I meant, and that's what's done in this PR, no? (modulo setupping CI to also run these if it's not the case currently) I just took it to a separate folder to better delineate them from the other, more pedagogical, examples. We can just discuss on Thursday.

mfherbst:
ok

examples/papers/cazalis.jl
```diff
@@ -0,0 +1,91 @@
...
+
+using DFTK
+using Plots
+
+## redefine Hartree interaction
+struct Hartree2D end
+struct Term2DHartree <: DFTK.TermNonlinear end
+(t::Hartree2D)(basis) = Term2DHartree()
+function DFTK.ene_ops(term::Term2DHartree, basis::PlaneWaveBasis{T},
+                 ψ, occ; ρ, kwargs...) where {T}
```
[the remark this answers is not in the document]
antoine-levitt: Because it's not really clear what "2D Hartree" should mean (is it 1/|x| kernel, or is it the solution of Poisson's equation?), so it's better to be explicit here

PR #695, opened 2022-07-22
Automatically adapt number of bands during SCF

### LINKED ISSUE #562: Dynamically adjust number of targeted bands in SCF
As it says, move away from keeping `n_bands` fixed during the SCF.

### LINKED ISSUE #652: Get rid of n_ep_extra

mfherbst:
Closes #562 and closes #652.

- [x] More tests (Works well on pretty much everything I tried including transition metal alloys and surfaces)
- ~~Different number of bands per k-Point.~~ (moved to #733)

### DIFF
--- a/src/occupation.jl
+++ b/src/occupation.jl
@@ -19,8 +19,7 @@ end
 Find the occupation and Fermi level.
 """
 function compute_occupation(basis::PlaneWaveBasis{T}, eigenvalues;
-                            temperature=basis.model.temperature,
-                            occupation_threshold) where {T}
+                            temperature=basis.model.temperature) where {T}
     n_electrons = basis.model.n_electrons
     n_spin      = basis.model.n_spin_components
 
@@ -94,10 +93,5 @@ function compute_occupation(basis::PlaneWaveBasis{T}, eigenvalues;
     end
 
     occupation = compute_occupation(basis, eigenvalues, εF)
-    minocc = maximum(minimum, occupation)
-    if !iszero(temperature) && minocc > occupation_threshold
-        @warn "One k-point has a high minimum occupation $minocc. You should probably increase the number of bands."
-    end
-
     (; occupation, εF)
 end
--- a/src/scf/self_consistent_field.jl
+++ b/src/scf/self_consistent_field.jl
@@ -147,9 +147,9 @@ function next_density(ham::Hamiltonian, bands=AdaptiveBands(ham.basis.model);
     #      way to deal with such things in LOBPCG.
     if !increased_n_bands && minocc > bands.occupation_threshold
         @warn("Detected large minimal occupation $minocc. SCF could be unstable. " *
-              "Try switching to adaptive band selection (`bands=AdaptiveBands(model)`) " *
+              "Try switching to adaptive band selection (`bands=AdaptiveBands(basis)`) " *
               "or request more converged bands than $n_bands_converge (e.g. " *
-              "`bands=AdaptiveBands(model; n_bands_converge=$(n_bands_converge + 3)`)")
+              "`bands=AdaptiveBands(basis; n_bands_converge=$(n_bands_converge + 3)`)")
     end
 
     # TODO We should probably set occupation values below occupation_threshold explicitly
--- a/src/scf/self_consistent_field.jl
+++ b/src/scf/self_consistent_field.jl
@@ -10,10 +10,12 @@ end
 Obtain new density ρ by diagonalizing `ham`. Follows the policy imposed by the `bands`
 data structure to determine and adjust the number of bands to be computed.
 """
-function next_density(ham::Hamiltonian, bandspol::BandsPolicy=AdaptiveBands(ham.basis.model);
+function next_density(ham::Hamiltonian,
+                      nbandsalg::NbandsAlgorithm=AdaptiveBands(ham.basis.model);
                       eigensolver=lobpcg_hyper, ψ=nothing, eigenvalues=nothing,
                       occupation=nothing, kwargs...)
-    n_bands_converge, n_bands_compute = determine_n_bands(bandspol, occupation, eigenvalues, ψ)
+    n_bands_converge, n_bands_compute = determine_n_bands(nbandsalg, occupation,
+                                                          eigenvalues, ψ)
 
     if isnothing(ψ)
         increased_n_bands = true
@@ -27,8 +29,8 @@ function next_density(ham::Hamiltonian, bandspol::BandsPolicy=AdaptiveBands(ham.
     #      computed for each k-Point
     n_bands_compute = mpi_max(n_bands_compute, ham.basis.comm_kpts)
 
-    eigres = diagonalize_all_kblocks(eigensolver, ham, n_bands_compute; ψguess=ψ,
-                                     n_conv_check=n_bands_converge, kwargs...)
+    eigres = diagonalize_all_kblocks(eigensolver, ham, n_bands_compute;
+                                     ψguess=ψ, n_conv_check=n_bands_converge, kwargs...)
     eigres.converged || (@warn "Eigensolver not converged" iterations=eigres.iterations)
 
     # Check maximal occupation of the unconverged bands is sensible.
@@ -38,23 +40,23 @@ function next_density(ham::Hamiltonian, bandspol::BandsPolicy=AdaptiveBands(ham.
     # TODO This is a bit hackish, but needed right now as we increase the number of bands
     #      to be computed only between SCF steps. Should be revisited once we have a better
     #      way to deal with such things in LOBPCG.
-    if !increased_n_bands && minocc > bandspol.occupation_threshold
+    if !increased_n_bands && minocc > nbandsalg.occupation_threshold
         @warn("Detected large minimal occupation $minocc. SCF could be unstable. " *
-              "Try switching to adaptive band selection (`bandspol=AdaptiveBands(basis)`) " *
+              "Try switching to adaptive band selection (`nbandsalg=AdaptiveBands(basis)`) " *
               "or request more converged bands than $n_bands_converge (e.g. " *
-              "`bandspol=AdaptiveBands(basis; n_bands_converge=$(n_bands_converge + 3)`)")
+              "`nbandsalg=AdaptiveBands(basis; n_bands_converge=$(n_bands_converge + 3)`)")
     end
 
     ρout = compute_density(ham.basis, eigres.X, occupation)
     (ψ=eigres.X, eigenvalues=eigres.λ, occupation, εF, ρout, diagonalization=eigres,
-     n_bands_converge, bandspol.occupation_threshold)
+     n_bands_converge, nbandsalg.occupation_threshold)
 end
 
 
 """
 Solve the Kohn-Sham equations with a SCF algorithm, starting at `ρ`.
 
-- `bandspol`: By default DFTK uses `bandspol=AdaptiveBands(basis)`, which adaptively determines
+- `nbandsalg`: By default DFTK uses `nbandsalg=AdaptiveBands(basis)`, which adaptively determines
   the number of bands to compute. If you want to influence this algorithm or use a predefined
   number of bands in each SCF step, pass a [`FixedBands`](@ref) or [`AdaptiveBands`](@ref).
 """
@@ -71,7 +73,7 @@ Solve the Kohn-Sham equations with a SCF algorithm, starting at `ρ`.
                                        damping=0.8,  # Damping parameter
                                        mixing=LdosMixing(),
                                        is_converged=ScfConvergenceEnergy(tol),
-                                       bandspol::BandsPolicy=AdaptiveBands(basis),
+                                       nbandsalg::NbandsAlgorithm=AdaptiveBands(basis),
                                        callback=ScfDefaultCallback(; show_damping=false),
                                        compute_consistent_energies=true,
                                        response=ResponseOptions(),  # Dummy here, only for AD
@@ -79,9 +81,10 @@ Solve the Kohn-Sham equations with a SCF algorithm, starting at `ρ`.
     if !isnothing(n_bands) || !isnothing(n_ep_extra)
         # TODO Backwards compatibility ... emulates exactly how bands worked before
         @warn("The options n_bands and n_ep_extra of self_consistent_field are deprecated. " *
-              "Use `bands` instead.")
-        n_bands = something(n_bands, FixedBands(basis.model).n_bands)
-        bandspol = FixedBands(;n_bands, n_bands_compute=n_bands + something(n_ep_extra, 3))
+              "Use `nbandsalg` instead to influence number of bands to compute.")
+        n_bands_converge = something(n_bands, FixedBands(basis.model).n_bands_converge)
+        nbandsalg = FixedBands(; n_bands_converge,
+                               n_bands_compute=n_bands_converge + something(n_ep_extra, 3))
     end
 
     # All these variables will get updated by fixpoint_map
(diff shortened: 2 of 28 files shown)


mfherbst:
To be tested further (i.e. beyond silicon :smile:).

src/eigen/diag.jl
```diff
@@ -28,18 +28,25 @@ function diagonalize_all_kblocks(eigensolver, ham::Hamiltonian, nev_per_kpoint::
...
                 ψguessk = ψguess[ik]
-            elseif interpolate_kpoints && ik > 1
-                # use information from previous k-point
-                X0 = interpolate_kpoint(results[ik - 1].X, ham.basis, kpoints[ik - 1],
-                                        ham.basis, kpoints[ik])
-                ψguessk = ortho_qr(X0)  # Re-orthogonalize and renormalize
             else
-                ψguessk = random_orbitals(ham.basis, kpt, nev_per_kpoint)
+                X0 = similar(ψguess[ik], n_Gk, nev_per_kpoint)
+                X0[:, 1:nev_guess] = ψguess[ik]
```
antoine-levitt, line 39: this was factored out to random_orbitals to be more flexible in the future (eg if we want to support symmetries of the wfs, where we possibly need to impose some specific symmetry on the initial orbitals)

src/eigen/diag.jl
```diff
@@ -28,18 +28,25 @@ function diagonalize_all_kblocks(eigensolver, ham::Hamiltonian, nev_per_kpoint::
...
+                X0[:, nev_guess+1:end] = randn(eltype(X0), n_Gk, nev_per_kpoint - nev_guess)
+                ψguessk = ortho_qr(X0)
             end
+        elseif interpolate_kpoints && ik > 1
+            # use information from previous k-point
+            X0 = interpolate_kpoint(results[ik - 1].X, ham.basis, kpoints[ik - 1],
+                                    ham.basis, kpoints[ik])
+            ψguessk = ortho_qr(X0)  # Re-orthogonalize and renormalize
```
antoine-levitt, line 47: this orthogonalization should probably go into interpolate_kpoint

src/occupation.jl
```diff
@@ -89,11 +88,6 @@ function compute_occupation(basis::PlaneWaveBasis{T}, eigenvalues;
     end
 
     occupation = compute_occupation(basis, eigenvalues, εF)
-    minocc = maximum(minimum, occupation)
-    if temperature > 0 && minocc > occupation_threshold
-        @warn "One k-point has a high minimum occupation $minocc. You should probably increase the number of bands."
-    end
-
```
antoine-levitt, line 96: that's great if it works, but it'd be also good to support the "old" way of doing things, without adaptativity

antoine-levitt: this also really highlights the need to put "algorithmic details" into a struct

mfherbst: > that's great if it works, but it'd be also good to support the "old" way of doing things, without adaptativity
 
I agree, but this is a strict improvement as you can still specify `n_bands` (and are guaranteed to get at least that many converged). The difference is just that now the code might decide to converge more automatically.

antoine-levitt: It's still good to be able to specify a simple scheme (ie when the heuristics go crazy, or when you want to do perf benchmarking...)

mfherbst: Ok, I changed that now. Turns out already on the magnesium example our default choice was flawed (some bands with rather reasonable occupation were not guaranteed to be fully converged).

Also I'll implement k-point specific `n_bands` in this PR, I think.

mfherbst: > Also I'll implement k-point specific n_bands in this PR, I think.

Nah. That'll be a follow-up.

mfherbst:
@antoine-levitt Ready for another look if you get the time.

src/scf/self_consistent_field.jl
```diff
@@ -6,74 +6,44 @@ include("scf_callbacks.jl")
...
-    (ψ=eigres.X, eigenvalues=eigres.λ, occupation=occupation, εF=εF,
-     ρout=ρout, diagonalization=eigres)
-end
+Solve the Kohn-Sham equations with a SCF algorithm, starting at `ρ`.
 
-
-"""
-Solve the Kohn-Sham equations with a SCF algorithm, starting at ρ.
+- `bands`: By default DFTK uses an adaptive algorithm to determine the number of bands
+  to compute. If `bands` is an integer, it specifies the minimal number of bands to be
```
antoine-levitt, line 13: that seems potentially confusing. Isn't it better to do the fixed thing when bands is an integer?

src/scf/self_consistent_field.jl
```diff
@@ -6,74 +6,44 @@ include("scf_callbacks.jl")
...
+  computed. Also directly accepts `FixedBands` or `AdaptiveBands` objects, which allow full
+  configuration of how DFTK adjusts the bands (including always running with a fixed number.
```
antoine-levitt, line 15: paren

src/scf/self_consistent_field.jl
```diff
@@ -6,74 +6,44 @@ include("scf_callbacks.jl")
...
 """
-@timing function self_consistent_field(basis::PlaneWaveBasis;
-                                       n_bands=default_n_bands(basis.model),
+@timing function self_consistent_field(basis::PlaneWaveBasis{T};
+                                       n_bands=nothing,  # TODO For backwards compatibility.
+                                       n_ep_extra=nothing,
+                                       bands=AdaptiveBands(basis.model),
```
antoine-levitt, line 20: are we sure it doesn't want to use information from the basis?

src/scf/self_consistent_field.jl
```diff
@@ -6,74 +6,44 @@ include("scf_callbacks.jl")
...
+- `bands`: By default DFTK uses an adaptive algorithm to determine the number of bands
+  to compute. If `bands` is an integer, it specifies the minimal number of bands to be
+  computed. Also directly accepts `FixedBands` or `AdaptiveBands` objects, which allow full
+  configuration of how DFTK adjusts the bands (including always running with a fixed number.
 """
-@timing function self_consistent_field(basis::PlaneWaveBasis;
-                                       n_bands=default_n_bands(basis.model),
+@timing function self_consistent_field(basis::PlaneWaveBasis{T};
+                                       n_bands=nothing,  # TODO For backwards compatibility.
+                                       n_ep_extra=nothing,
```
antoine-levitt, line 19: does the above comment apply to these two or just the first one?

src/scf/self_consistent_field.jl
```diff
@@ -133,25 +93,152 @@ Solve the Kohn-Sham equations with a SCF algorithm, starting at ρ.
...
+
+
+"""
+Obtain new density ρ by diagonalizing `ham`. Follows the policy imposed by the `bands`
+data structure to determine and adjust the number of bands to be computed.
+"""
+function next_density(ham::Hamiltonian, bands=AdaptiveBands(ham.basis.model);
+                      eigensolver=lobpcg_hyper, ψ=nothing, eigenvalues=nothing,
+                      occupation=nothing, kwargs...)
+    n_bands, n_bands_compute = determine_n_bands(bands, occupation, eigenvalues, ψ)
```
antoine-levitt, line 125: it's a bit confusing to have n_bands, n_bands_compute... Would n_bands_converge be good? Or would it be too long? We don't use it that much...

src/scf/self_consistent_field.jl
```diff
@@ -133,25 +93,152 @@ Solve the Kohn-Sham equations with a SCF algorithm, starting at ρ.
...
+
+    if !isnothing(ψ)
+        @assert length(ψ) == length(ham.basis.kpoints)
+        n_bands_compute = max(n_bands_compute, maximum(ψk -> size(ψk, 2), ψ))
+    end
+    increased_n_bands = isnothing(ψ) || n_bands_compute > size(ψ[1], 2)
```
antoine-levitt, line 131: a bit tricky to understand. Set increased_n_bands to false before the if and to true inside?

src/scf/self_consistent_field.jl
```diff
@@ -133,25 +93,152 @@ Solve the Kohn-Sham equations with a SCF algorithm, starting at ρ.
...
+
+    # TODO Synchronise since right now it is assumed that the same number of bands are
```
antoine-levitt, line 133: ize :-p

examples/energy_cutoff_smearing.jl
```diff
@@ -42,12 +38,12 @@ function compute_ground_state_energy(a; Ecut, kgrid, kinetic_blowup, kwargs...)
...
     basis = PlaneWaveBasis(model; Ecut, kgrid)
-    self_consistent_field(basis; kwargs...).energies.total
+    self_consistent_field(basis; callback=identity, kwargs...).energies.total
 end
 
-callback = info->nothing # set SCF to non verbose
-E0_naive = compute_ground_state_energy.(a_list; kinetic_blowup=BlowupIdentity(),
-                                        Ecut, kgrid, n_bands, callback);
+Ecut  = 5         # Very low Ecut to display big irregularities
+kgrid = (2, 2, 2) # Very sparse k-grid to fasten convergence
```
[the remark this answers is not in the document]
mfherbst: not from me, but I'll still change it ;)

src/eigen/diag.jl
```diff
@@ -28,26 +27,31 @@ function diagonalize_all_kblocks(eigensolver, ham::Hamiltonian, nev_per_kpoint::
                   "$nev_per_kpoint eigenvalues. Increase Ecut.")
         end
         # Get ψguessk
-        @timing "QR orthonormalization" begin
```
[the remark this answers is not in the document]
mfherbst: let's move it to ortho if you want to keep it. It's basically neglibile.

src/scf/self_consistent_field.jl
mfherbst: it's not always true inside the `if`, but yes I'll move it inside as this will make the logic more clear.

src/scf/self_consistent_field.jl
mfherbst: I'm not sure. Most people will want the adaptive thing.

I view the FixedBands more as a debug / preformance-tracking option, so more expert. On the other hand I'm pretty confident that manual adjustments of the bands should now be needed a lot less regularly (i.e. basically only if you want to do something special), so perhaps we should drop the integer thing alltogether?

Project.toml
```diff
@@ -49,7 +49,7 @@ spglib_jll = "ac4a9f1e-bdb2-5204-990c-47c8b2f70d4e"
 AbstractFFTs = "1"
 AtomsBase = "0.2.2"
 BlockArrays = "0.16.2"
-Brillouin = "0.5 - 0.5.8"  # Upper bound temporary until memory bug resolved.
+Brillouin = "0.5 - 0.5.8"
```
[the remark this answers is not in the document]
mfherbst: Ah yes, that should not be removed.

src/scf/self_consistent_field.jl
```diff
@@ -133,25 +91,156 @@ Solve the Kohn-Sham equations with a SCF algorithm, starting at ρ.
...
+    #      way to deal with such things in LOBPCG.
+    if !increased_n_bands && minocc > bands.occupation_threshold
+        @warn("Detected large minimal occupation $minocc. SCF could be unstable. " *
+              "Try switching to adaptive band selection (`bands=AdaptiveBands(basis)`) " *
+              "or request more converged bands than $n_bands_converge (e.g. " *
+              "`bands=AdaptiveBands(basis; n_bands_converge=$(n_bands_converge + 3)`)")
+    end
+
+    # TODO We should probably set occupation values below occupation_threshold explicitly
+    #      to zero and ignore these bands in the density computation
```
antoine-levitt, line 156: we should not pretend they are zero if they are not, but we could pass the threshold to compute_density

src/scf/self_consistent_field.jl
```diff
@@ -133,25 +91,156 @@ Solve the Kohn-Sham equations with a SCF algorithm, starting at ρ.
...
+    n_extra = iszero(model.temperature) ? 0 : ceil(Int, 0.15 * min_n_bands)
+    min_n_bands + n_extra
+end
+default_occupation_threshold() = 1e-6
+
+"""
+In each SCF step converge exactly `n_bands_converge`, computing along the way exactly
+`n_bands_compute` (usually a few more to ease convergence in systems with small gaps).
+"""
+struct FixedBands
```
antoine-levitt, line 176: make them a subtype of NbandsPolicy or something

antoine-levitt: This is quite a bit to do in self_consistent_field.jl, move it to a separate file?

src/scf/self_consistent_field.jl
```diff
@@ -133,25 +91,156 @@ Solve the Kohn-Sham equations with a SCF algorithm, starting at ρ.
...
+    n_bands_converge::Int # Number of bands to converge
+    n_bands_compute::Int  # Number of bands to compute (but not necessarily converge)
+    occupation_threshold::Float64  # Threshold for orbital to be counted as occupied
+end
+function FixedBands(; n_bands_converge, n_bands_compute=n_bands_converge+3,
+                    occupation_threshold=default_occupation_threshold())
```
antoine-levitt, line 182: Use `Base.@kwdef`?

src/scf/self_consistent_field.jl
```diff
@@ -133,25 +91,156 @@ Solve the Kohn-Sham equations with a SCF algorithm, starting at ρ.
...
+    FixedBands(n_bands_converge, n_bands_compute, occupation_threshold)
+end
+function FixedBands(model::Model)
+    n_bands_converge = default_n_bands(model)
+    n_bands_converge += iszero(model.temperature) ? 0 : ceil(Int, 0.05 * n_bands_converge)
+    FixedBands(; n_bands_converge, n_bands_compute=n_bands_converge + 3)
+end
+FixedBands(basis::PlaneWaveBasis; kwargs...) = FixedBands(basis.model; kwargs...)
+function determine_n_bands(bands::FixedBands, occupation, eigenvalues, ψ)
+    # TODO Could return different bands per k-Points
```
antoine-levitt, line 192: not really in the purview of FixedBands, so I'd remove that comment

src/scf/self_consistent_field.jl
```diff
@@ -133,25 +91,156 @@ Solve the Kohn-Sham equations with a SCF algorithm, starting at ρ.
...
+    n_bands_converge = floor(Int, (bands.n_bands_converge + bands.n_bands_compute) / 2)
+    (; n_bands_converge, n_bands_compute)
+end
+function determine_n_bands(bands::AdaptiveBands, occupation::AbstractVector,
+                           eigenvalues::AbstractVector, ψ::AbstractVector)
+    # TODO Could return different bands per k-Points
+
+    # Determine number of bands to be actually converged
+    n_bands_occ = maximum(occupation) do occk
+        something(findlast(onk -> onk ≥ bands.occupation_threshold, occk), length(occk) + 1)
```
antoine-levitt, line 232: we call these fnk

antoine-levitt (commented):
Looks good! Only minor nits

PR #696, opened 2022-07-25
Add blowup functions to Kinetic term for cutoff smearing

LaurentVidal95:
Dear Antoine and Michael,

Here are in a nutshell the modifications I did to DFTK to add a modified kinetic energy term, allowing to have targeted smoothness of the energy band at a little cost of precision on the energies. I have two questions on this PR:
- 1) I added a file `src/terms/modified_kinetic_blow_up_function.jl` in `src/terms` which defines a function used in `src/terms/kinetic.jl` but I'm not sure this directory is the best place to put it.
- 2) I did an example to show how to compute the variations of total energy vs the volume of the unit cell of silicon. The final plot show big discontinuities for standard kinetic energies and a smooth curve for modified ones which is pretty nice. I'm not sure however that the way I did it is the best way to do so. For now I didn't put much commentary in case the example has to change. Tell me what you think about it !

PS: the paper that led to this implementation (with Eric and Hassan) is to be submitted this week if everything goes to plan.

Bests,
Laurent


src/DFTK.jl
```diff
@@ -74,6 +74,7 @@ export Hamiltonian
 export HamiltonianBlock
 export energy_hamiltonian
 export Kinetic
+export ModifiedKinetic
```
antoine-levitt, line 77: Just have it as an option to Kinetic()?

src/PlaneWaveBasis.jl
```diff
@@ -112,8 +112,17 @@ Base.Broadcast.broadcastable(basis::PlaneWaveBasis) = Ref(basis)
 
 Base.eltype(::PlaneWaveBasis{T}) where {T} = T
 
+function select_G_vector(model, G, k, Ecut)
+    # Strict inequality for ModifiedKinetic
+    if (model.model_name == "ModifiedKinetic")
+        return  (sum(abs2, model.recip_lattice*(G+k))/2 < Ecut)
+    end
+    (sum(abs2, model.recip_lattice*(G+k))/2 ≤ Ecut)
+end
```
antoine-levitt, line 121: that's a hack... just have your blow up function return a very large value at Ecut if you're really concerned about it? (see eg floatmax)

src/eigen/preconditioners.jl
```diff
@@ -33,10 +33,13 @@ mutable struct PreconditionerTPA{T <: Real}
...
-    kinetic_term = [t for t in basis.model.term_types if t isa Kinetic]
-    isempty(kinetic_term) && error("Preconditioner should be disabled when no Kinetic term is used.")
-    scaling = only(kinetic_term).scaling_factor
-    kin = Vector{T}([scaling * sum(abs2, q) for q in Gplusk_vectors_cart(basis, kpt)] ./ 2)
+    kinetic_term = only([t for t in basis.model.term_types
+                         if typeof(t) ∈ (Kinetic, ModifiedKinetic)])
+    scaling = kinetic_term.scaling_factor;
+    Ecut = basis.Ecut
+    # Choose between standard and modified kinetic energies.
+    Gp = x->x'x; (kinetic_term isa ModifiedKinetic) && (Gp=kinetic_term.blow_up_function)
```
antoine-levitt, line 41: we usually use sum(abs2)

antoine-levitt: and Gp(x) instead of Gp = x ->

antoine-levitt: Also that's a very bad name

src/terms/kinetic.jl
```diff
@@ -22,6 +22,36 @@ function TermKinetic(basis::PlaneWaveBasis{T}, scaling_factor) where {T}
...
+@doc raw"""
+For a given cutoff energy ``E_{\rm cut}`` and for all ``k``-point and
+``G``-vector such that ``|k+G|^2 < 2E_{\rm cut}``, modified kinetic energies are given by
+``E_{\rm cut} G_p(|G+k|/\sqrt(2E_{\rm cut}))``
+where ``G_p`` is a blow-up function defined in `modified_kinetic_blow_up_function.jl`.
+"""
+struct ModifiedKinetic
+    scaling_factor::Real
+    blow_up_function
+end
```
antoine-levitt, line 34: just merge that with Kinetic, and have an identity blow up function by default

src/terms/modified_kinetic_blow_up_function.jl
```diff
@@ -0,0 +1,92 @@
...
+#  - g1(x) = x^2 defined on [0,x_1),
+#  - g3(x, p) ≥ x^2 ∀p, defined on [x_2,1), that blows-up at x=1 with blow-up rate |⋅|^{-p},
+#  - g2, a C^2 interpolation between g1 and g3_p on [x_1,x_2].
+#
+# In practice we define:
+#
+# g3(x,p) = g3(x, a, p) = C(a,p)/(1-x)^p, where C(a,p) = (3/2)*(a^2)*(1-a)^(p)
+#
+# and where the parameter a is optimized so that ||Gp - x^2||_{L^1((0,1))} is minimized.
+
```
antoine-levitt, line 14: Looks like it's something that was used to generate the blowup function and does not need to be in DFTK? Just hardcode the coefficients. Eg you can have in kinetic.jl things like blowup_function_CHV (for cancès hassan vidal), and blowup_function_abinit

jaemolihm:
Hi, just out of curiosity, is this related to the "energy cutoff smearing" of ABINIT (https://docs.abinit.org/variables/rlx/#ecutsm, Appendix B of https://journals.aps.org/prb/abstract/10.1103/PhysRevB.93.205147)? I have always wondered how useful it is (I mostly use QE and QE doesn't have it).

(Maybe I should wait a bit for the paper to come out :) )

antoine-levitt:
I'll let @LaurentVidal95 answer in more details but yes it's the same. I'd imagine it's useful to plot nice smooth curves, but I don't think it improves accuracy.

LaurentVidal95:
> Basic idea sound but a bit too much new code for my taste

I tried to merge ModifiedKinetic to Kinetic, tell me if it looks a bit more like it !
I still have to redact a full text for the example and to add references.  Also for testing I launch a quick scf computation on silicon LDA with a blow-up function and check convergencebut maybe there's something else to check !

LaurentVidal95:
> Hi, just out of curiosity, is this related to the "energy cutoff smearing" of ABINIT (https://docs.abinit.org/variables/rlx/#ecutsm, Appendix B of https://journals.aps.org/prb/abstract/10.1103/PhysRevB.93.205147)? I was curious about how useful it is (I haven't tried it - I mostly use QE and QE doesn't have it).
> 
> Maybe I should wait a bit for the paper to come out :)

Hi @jaemolihm !

As I understand it now the cutoff smearing is helpful when you cannot afford to take a high `Ecut`. Typically for unit-cell relaxation or geometry optimization. In that case the size of your plane-wave basis varies a lot with the variation of the unit-cell and the energy w.r. to the cell volume display irregularities. The use of an appropriate cutoff smearing ensures regularity of your bands, while introducing a slight over-evaluation of your energies.

Indeed Abinit also features an energy cutoff smearing ! I implemented it in the new version of the PR by the way but its behavior is pretty strange. The blow-up function is really discontinuous if the parameters are not chosen wisely so I might have done something wrong that I don't see for now !

The advantage with the other blow-up function is that we have a mathematical proof that it provides ``C^m``, ``m`` depending of your parameters. If you are interested the paper is to be submitted this week, I can forward it to you as soon as it is online :+1:  !

jaemolihm:
Thanks for the kind explanation! It sounds like a nice mathematical property, and it's also nice that DFTK allows much flexibility to play with all those details. It would be great if you could forward the paper when it's online. Thanks a lot 😄

src/terms/kinetic.jl
```diff
@@ -1,22 +1,26 @@
 """
-Kinetic energy: 1/2 sum_n f_n ∫ |∇ψn|^2.
+Kinetic energy: 1/2 sum_n f_n ∫ Blowup_function(|∇ψn|).
 """
 struct Kinetic
     scaling_factor::Real
+    Blowup_function # blow-up fonction to smooth energy bands. Defaut is x↦x^2.
```
mfherbst, line 6: We don't capitalise variables.

mfherbst: Also I would rename this to `blowup` and make this struct an `@kwdef` like
```julia
@kwdef struct Kinetic{F}
    scaling_factor::Real = 1
    blowup::F = BlowupKinetic()  # Default kinetic energy
end
```

src/terms/kinetic.jl
```diff
@@ -1,22 +1,26 @@
...
+(kin::Kinetic)(basis) = TermKinetic(basis, kin.scaling_factor, kin.Blowup_function)
 function Base.show(io::IO, kin::Kinetic)
     fac = isone(kin.scaling_factor) ? "" : ", scaling_factor=$scaling_factor"
     print(io, "Kinetic($fac)")
 end
 
 struct TermKinetic <: Term
     scaling_factor::Real  # scaling factor, absorbed into kinetic_energies
-    kinetic_energies::Vector{<:AbstractVector}  # kinetic energy 1/2|G+k|^2 for every kpoint
+    # kinetic energy Ecut*Blowup_function(|G+k|/√(2Ecut)) for every kpoint
```
mfherbst, line 18: spaces around `/` helps here

src/terms/kinetic.jl
```diff
@@ -40,3 +44,44 @@ end
...
+    g3(x) = Ca(a_opti)/( (1-x)^(2) )
+
+    # Assemble all parts
+    function Blowup_function(x)
+        x1, x2 = [0.7, 0.8]
+        (0 ≤ x < x1)   && (return g1(x))
+        (x1 ≤ x < x2)  && (return g2(x))
+        (x2 ≤ x < 1)   && (return g3(x))
+        (x==1) && (return 1e6) # Handle |G+k|^2 = E_cut case
+        error("The blow-up function is defined on [0,1). Did you devide by √Ecut ?")
```
mfherbst, line 69: Pretty low-level here, should be an assertion.

src/terms/kinetic.jl
```diff
@@ -40,3 +44,44 @@ end
...
+    a_opti = 0.1792973510040141
+    g3(x) = Ca(a_opti)/( (1-x)^(2) )
+
+    # Assemble all parts
+    function Blowup_function(x)
+        x1, x2 = [0.7, 0.8]
+        (0 ≤ x < x1)   && (return g1(x))
+        (x1 ≤ x < x2)  && (return g2(x))
+        (x2 ≤ x < 1)   && (return g3(x))
+        (x==1) && (return 1e6) # Handle |G+k|^2 = E_cut case
```
mfherbst, line 68: no brackets needed

src/terms/kinetic.jl
```diff
@@ -40,3 +44,44 @@ end
 
     (E=E, ops=ops)
 end
+
+"""
+Blow-up function as proposed in [REF paper Cancès, Hassan, Vidal to be submitted]
+"""
+function Blowup_function_CHV()
```
mfherbst, line 51: Be careful. This code is generally not typestable. E.g. $x$ could be a Float32 or an interval or a dual number.

mfherbst: Also it feels weird to have this function returning a function. Why is this needed? Can't the outer be merged into the inner?

src/terms/kinetic.jl
```diff
@@ -40,3 +44,44 @@ end
...
+        (x==1) && (return 1e6) # Handle |G+k|^2 = E_cut case
+        error("The blow-up function is defined on [0,1). Did you devide by √Ecut ?")
+    end
+    Blowup_function
+end
+
+"""
+Blow-up function as implemented in the Abinit code [ref code].
+"""
+function Blowup_function_Abinit(Ecut, Ecutsm)
```
mfherbst, line 77: Ah, now I get it, because you need to pass some extra information.

Ok in that case I would have the blowup stuff be a `struct` (i.e. one standard `BlowupKineticEnergy`, `BlowupCHV` and `BlowupAbinit`). The `BlowupAbinit` one has an `Ecutsm` field, the others have no fields. Then you multiple dispatch on the blowup as in have three methods `TermKinetic(basis::PlaneWaveBasis{T}, blowup::BlowupAbinit)` etc. There you have all the info to build the `TermKinetic` appropriately.

src/terms/kinetic.jl
mfherbst: ok I have a better idea, see my comment below.

mfherbst (commented):
Nice to have this, but I would restructure a bit.

LaurentVidal95:
Hello everyone, about the last commit I applied the changes you suggested @mfherbst. Tell me if that fits what you ment !
I still have to finish the documentation. The graph is a bit long to generate so I'm planning to just link to a pdf file.

Also, I have removed the Abinit blow-up function because it is equivalent to the CHV one for a blow-up of order 1. So the Abinit one only ensures C^1 regularity, where the CHV with the current hard-coded coefficients gives C^2 regularity. Also I wasn't completely sure about the implementation of the Abinit function with was discontinuous for certain coefficients..

src/terms/kinetic.jl
```diff
@@ -1,22 +1,69 @@
...
+struct BlowupKineticEnergy end
+struct BlowupCHV end
+
+"""
+Chooses between standard and modified kinetic energies, acording to the blow-up type.
+"""
+function blowup_function(x, blowup)
+    @assert( typeof(blowup) ∈ [BlowupKineticEnergy, BlowupCHV] )
+    (blowup isa BlowupKineticEnergy) && (return sum(abs2, x)) # Standard kinetic term
+    (blowup isa BlowupCHV) && (return blowup_function_CHV(x)) # Modified kinetic term
```
mfherbst, line 29: Do this with multiple dispatch.

src/DFTK.jl
```diff
@@ -89,6 +89,9 @@ export PairwisePotential
 export Anyonic
 export apply_kernel
 export compute_kernel
+export BlowupKineticEnergy
+export BlowupCHV
+export BlowupAbinit
```
mfherbst, line 94: Be careful: Removed!

examples/energy_cutoff_smearing.jl
```diff
@@ -0,0 +1,88 @@
...
+                              kwargs...).energies.total
+    (;compute_GS=compute_GS, basis=basis_silicon, model=model_silicon)
+end
+
+# We can now compute the wanted energies for standard and modified kinetic term.
+# We use the CHV blow-up function introduced in [REF] which ensures ``C^2``
+# regularity of the energy bands.
+
+Ecut = 5 # very low Ecut to display big irregularities
+kgrid = [4,4,4]
```
mfherbst, line 66: If the computation is too long, maybe reduce the kgrid?

test/energy_cutoff_smearing.jl
```diff
@@ -0,0 +1,19 @@
...
+@testset "Energy cutoff smearing on silicon LDA" begin
+    # LDA term with modified kinetic term
+    blowup=BlowupCHV()
+    terms = [Kinetic(;blowup), AtomicLocal(), AtomicNonlocal(),
+             Ewald(), PspCorrection(), Hartree(), Xc([:lda_x, :lda_c_pw])]
+    # Launch scf for a low Ecut = 5 Ha
+    model = Model(silicon.lattice, atoms, silicon.positions; terms)
+    basis = PlaneWaveBasis(model, 5, silicon.kcoords, silicon.kweights)
+    scfres = self_consistent_field(basis; n_bands=8)
+    @test scfres.converged
```
mfherbst, line 18: yeah ok, but actually this does not really test anything about your implementation specifically. I.e. this will pass even if I refactor code and make an error in the blowup function you implemented.

mfherbst:
> Also, I have removed the Abinit blow-up function because it is equivalent to the CHV one for a blow-up of order 1. So the Abinit one only ensures C^1 regularity, where the CHV with the current hard-coded coefficients gives C^2 regularity. Also I wasn't completely sure about the implementation of the Abinit function with was discontinuous for certain coefficients..
If need I can just switch back to the commit where it was implemented.

I don't think it's bad to have the Abinit version as well, unless the cost of your implementation is always the same and always provides better regularity. In any case I would have a comment about this relationship in the code and in the example as people might be more familiar with the Abinit tricks.

mfherbst (commented):
Hi Laurent. Thanks for the improvements, but not fully what I was after. Let's discuss next week in person.

LaurentVidal95:
Dear all,
I pushed a more precise test as discussed in person last week ! I think the PR is ready to be reviewed as such.  Let me know if that's suits you better !
Bests

src/terms/kinetic.jl
```diff
@@ -1,22 +1,66 @@
...
+"""
+Blow-up function as proposed in [REF paper Cancès, Hassan, Vidal to be submitted]
+Parameters have been computed to ensure C^2 regularity of the energies bands
+away from crossings and Lipschitz continuity at crossings.
+TODO: find a better C^2 interpolation.
+"""
+function (blowup::BlowupCHV)(x)
+    @assert (0 ≤ x < 1) "The blow-up function is defined on [0,1)"
+    # Define the three parts of the blow-up function
+    g1(x) = sum(abs2, x)
```
antoine-levitt, line 37: isn't x a scalar? also just inline that function

src/terms/kinetic.jl
```diff
@@ -1,22 +1,66 @@
...
+    # C^2 interpolation part. The coefficients have been obtained by solving a
+    # linear system.
+    C = [7449.468883604184, -50230.90030429237, 135207.0860245428,
+         -181568.08750464107, 121624.06499834878, -32503.74733526006]
```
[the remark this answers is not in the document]
antoine-levitt: (look up polynomial interpolation on wikipedia for better ways)

src/terms/kinetic.jl
```diff
@@ -1,22 +1,66 @@
...
+    g2(x) = C'*((x * ones(6)) .^(0:5))
+    # Blow-up part
+    Ca(a) = (3/2)*(a^2)*(1-a)^(2)
+    a_opti = 0.1792973510040141
+    g3(x) = Ca(a_opti)/( (1-x)^(2) )
```
antoine-levitt, line 46: Just write C_opt = ... ? Also don't parenthesize the 2

src/terms/kinetic.jl
```diff
@@ -1,22 +1,66 @@
...
+
+    # Assemble all parts
+    x1, x2 = 0.7, 0.8
+    (0 ≤ x < x1)   && return g1(x)
+    (x1 ≤ x < x2)  && return g2(x)
+    (x2 ≤ x < 1)   && return g3(x)
+    (x==1) && return 1e6 # Handle |G+k|^2 = E_cut case
```
antoine-levitt, line 53: Would `Inf(T)` or `floatmax(T)` do the trick?

src/terms/kinetic.jl
```diff
@@ -1,22 +1,66 @@
...
     scaling_factor::Real  # scaling factor, absorbed into kinetic_energies
-    kinetic_energies::Vector{<:AbstractVector}  # kinetic energy 1/2|G+k|^2 for every kpoint
+    # kinetic energies 1/2(k+G)^2 (or Ecut*blowup(|k+G|/√(2Ecut)) for energy
+    # cutoff smearing methods) for each k-point.
+    kinetic_energies::Vector{<:AbstractVector}
 end
-function TermKinetic(basis::PlaneWaveBasis{T}, scaling_factor) where {T}
-    kinetic_energies = [[T(scaling_factor) * sum(abs2, Gk) / 2
+function TermKinetic(basis::PlaneWaveBasis{T}, scaling_factor, blowup) where {T}
+    kinetic_energies = [[T(scaling_factor) * basis.Ecut * blowup(norm(Gk)/√(2Ecut))*
```
antoine-levitt, line 63: why is there a basis.Ecut * here, and why is there a * at the end (this is probably the source of the CI issues)

src/terms/kinetic.jl
```diff
@@ -1,22 +1,66 @@
...
+Define different energy cutoff strategies.
+ADD REF when the paper is published.
+"""
+struct BlowupKineticEnergy end
+struct BlowupCHV end
+
+"""
+Default blow-up, x->x^2, corresponding to the standard kinetic energies.
+"""
+(blowup::BlowupKineticEnergy)(x) = sum(abs2, x)
```
antoine-levitt, line 26: Isn't x scalar?

src/terms/kinetic.jl
```diff
@@ -1,22 +1,66 @@
...
+
+"""
+Blow-up function as proposed in [REF paper Cancès, Hassan, Vidal to be submitted]
+Parameters have been computed to ensure C^2 regularity of the energies bands
+away from crossings and Lipschitz continuity at crossings.
+TODO: find a better C^2 interpolation.
+"""
+function (blowup::BlowupCHV)(x)
```
antoine-levitt, line 34: I think I would just have it be a multiplicative thing, ie the kinetic energy is g^2/2 * blowup(g)

src/terms/kinetic.jl
```diff
@@ -1,22 +1,66 @@
...
     fac = isone(kin.scaling_factor) ? "" : ", scaling_factor=$scaling_factor"
     print(io, "Kinetic($fac)")
 end
 
+"""
+Define different energy cutoff strategies.
+ADD REF when the paper is published.
+"""
+struct BlowupKineticEnergy end
+struct BlowupCHV end
```
antoine-levitt, line 21: Add Abinit's?

src/eigen/preconditioners.jl
```diff
@@ -35,8 +34,11 @@ end
 function PreconditionerTPA(basis::PlaneWaveBasis{T}, kpt::Kpoint; default_shift=1) where T
     kinetic_term = [t for t in basis.model.term_types if t isa Kinetic]
     isempty(kinetic_term) && error("Preconditioner should be disabled when no Kinetic term is used.")
-    scaling = only(kinetic_term).scaling_factor
-    kin = Vector{T}([scaling * sum(abs2, q) for q in Gplusk_vectors_cart(basis, kpt)] ./ 2)
+    kinetic_term = only(kinetic_term)
+    scaling = kinetic_term.scaling_factor
+    blowup = kinetic_term.blowup     # blowup for energy cut-off smearing
+    kin = Vector{T}([scaling * basis.Ecut * blowup(norm(q)/√(2*Ecut))
+                     for q in Gplusk_vectors_cart(basis, kpt)])
```
antoine-levitt, line 41: sucks that you have to recompute it. Can the Preconditioner maybe get the `HamiltonianBlock` object and use it to get the kinetic energies?

src/terms/kinetic.jl
```diff
@@ -1,22 +1,66 @@
...
+"""
+function (blowup::BlowupCHV)(x)
+    @assert (0 ≤ x < 1) "The blow-up function is defined on [0,1)"
+    # Define the three parts of the blow-up function
+    g1(x) = sum(abs2, x)
+    # C^2 interpolation part. The coefficients have been obtained by solving a
+    # linear system.
+    C = [7449.468883604184, -50230.90030429237, 135207.0860245428,
+         -181568.08750464107, 121624.06499834878, -32503.74733526006]
+    g2(x) = C'*((x * ones(6)) .^(0:5))
```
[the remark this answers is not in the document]
antoine-levitt: (and use tuples instead of arrays to store small immutable arrays, it's faster and doesn't allocate memory)

src/eigen/preconditioners.jl
mfherbst: Yes I'd argue that makes sense as the preconditioner can be thought of as a black box trying to construct an approximation of the discretised operator.

antoine-levitt:
Not sure 1.6 vs 1.8 can explain, it looks more that your test is very brittle and fails under a different random seed or something?

LaurentVidal95:
> Not sure 1.6 vs 1.8 can explain, it looks more that your test is very brittle and fails under a different random seed or something?

Okay well I'll just raise the test "x<10" to "x<20" then !

antoine-levitt:
Please don't: it means that it'll break 2 years from now because of an unrelated change, and then I'll have to debug it :-p Make sure your test is consistent physically, ie your "x large" (translated to x >= 10) should probably actually be "x large compared to y" (translated to x >= 10 y)

LaurentVidal95:
> Please don't: it means that it'll break 2 years from now because of an unrelated change, and then I'll have to debug it :-p Make sure your test is consistent physically, ie your "x large" (translated to x >= 10) should probably actually be "x large compared to y" (translated to x >= 10 y)

Indeed that will be better ! 
There's another error that occurs just in the mpi test which says that I'm trying to take element 51 of a 50 element table.. It happens when I'm computing the finite difference derivative of the bands with something like `[(x[i+1] - x[i])/delta for i in ...]`. 
I'll try to find a fix but if it already happened to you I'll take any advice !

antoine-levitt:
With MPI the bands are distributed across the processors so you have to get everything on one processor (or be clever) if you want to make this check. Just disable the test when running under MPI (see mpi_is_master in the tests)

mfherbst:
@LaurentVidal95 I made a few modifications (mostly type stability and I also added a convenience flag to switch between blowups as I think that will be very useful).

One thing I wonder is: Why did you add `@inline` to all the functions?

LaurentVidal95:
> @LaurentVidal95 I made a few modifications (mostly type stability and I also added a convenience flag to switch between blowups as I think that will be very useful).
> 
> One thing I wonder is: Why did you add `@inline` to all the functions?

Thanks for the modifications !

About the `@inline` I think I simply missunderstood a previous comment by Antoine  !  I don't fully understand what that does so I don't mind removing it.

Issue #690, opened 2022-07-18
Move LOBPCG out


antoine-levitt:
We should move LOBPCG to the broader ecosystem. Mohamed Tarek was open to moving it to IterativeSolvers if we can demonstrate it's superior to the existing implementation. We should benchmark it against the existing implementation for a range of (M,N). Ideally using a diagonal A, so that doesn't skew the timings.

mfherbst:
Main difficulty in my opinion is GPU compatibility

antoine-levitt:
Benchmarks:
```
using DFTK
using IterativeSolvers
using Random

function bench(N, density, tol)
    Random.seed!(0)
    M = ceil(Int, density*N)

    d = vcat(rand(M), 2 .+ rand(N-M))
    A = Diagonal(d)
    X = randn(N, M)
    t1 = @elapsed DFTK.LOBPCG(A, X, I, I, tol)
    t2 = @elapsed IterativeSolvers.lobpcg(A, I, false, X, tol=tol)
    (t1, t2)
end

for (N, density) in Iterators.product((100, 1000, 4000, 10000), (.01, .1))
    println("$N, $density")
    display(bench(N, density, 1e-3))
end
```
Results
```
100, 0.01
(0.000709827, 0.000113247)
1000, 0.01
(0.007919946, 0.006568147)
4000, 0.01
(0.102116052, 0.118252976)
10000, 0.01
(0.74366979, 0.906752279)
100, 0.1
(0.002315039, 0.00134439)
1000, 0.1
(0.098804512, 0.104305223)
4000, 0.1
(3.08999895, 4.546904686)
10000, 0.1
(46.322246197, 88.606136158)
```

PR #697, opened 2022-07-25
Bringing GPU programming to DFTK

GVigne:
This is a work in progress and probably shouldn't be merged immediately into DFTK.
As part of my GSoC (you can see some additional information [here](https://summerofcode.withgoogle.com/programs/2022/projects/GSUS0kxC) ), I am working on a GPU version of DFTK. One of the goal is to keep overall code changes as low as possible (I do not aim to build an other DFTK package using GPUs).
This is the first step in my project: so far I managed to implement the Kinetic, Local and NonLocal terms and am running computations through the `self_consistent_field` function. I have not yet managed to make the SCF solvers work, so I disabled them for now.

Here is a MWE:
```
using DFTK
using CUDA
a = 10.263141334305942  # Lattice constant in Bohr
lattice = a / 2 .* [[0 1 1.]; [1 0 1.]; [1 1 0.]]
atoms     = [Si, Si]
positions = [ones(3)/8, -ones(3)/8];
terms_LDA = [Kinetic(), AtomicLocal(), AtomicNonlocal()]

# Setup an LDA model and discretize using
# a single k-point and a small `Ecut` of 5 Hartree.
mod = Model(lattice, atoms, positions; terms=terms_LDA,symmetries=false)
basis = PlaneWaveBasis(mod; Ecut=30, kgrid=(1, 1, 1))
basis_gpu = PlaneWaveBasis(mod; Ecut=30, kgrid=(1, 1, 1), array_type = CuArray)

scfres = self_consistent_field(basis; tol=1e-3, solver=scf_damping_solver(1.0))
scfres_gpu = self_consistent_field(basis_gpu; tol=1e-3, solver=scf_damping_solver(1.0))
```
Any feedback is appreciated, be it on the code in itself or the implementation of new features!


mfherbst:
Super awesome! I have not had a look at the code, but the interface is pretty much exactly how I'd like to have it.

src/PlaneWaveBasis.jl
```diff
@@ -67,8 +67,8 @@ struct PlaneWaveBasis{T, VT} <: AbstractBasis{T} where {VT <: Real}
     G_to_r_normalization::T  # G_to_r = G_to_r_normalization * BFFT
 
     # "cubic" basis in reciprocal and real space, on which potentials and densities are stored
-    G_vectors::Array{Vec3{Int}, 3}
-    r_vectors::Array{Vec3{VT }, 3}
+    G_vectors::AbstractArray{Vec3{Int}, 3}
+    r_vectors::AbstractArray{Vec3{VT }, 3}
```
antoine-levitt, line 71: That's bad, as the actual type will be dynamic. pwbasis should be parametric on this type

src/PlaneWaveBasis.jl
```diff
@@ -191,7 +191,8 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Number, fft_size, variational,
     kweights_global = kweights
 
     # Setup FFT plans
-    (ipFFT, opFFT, ipBFFT, opBFFT) = build_fft_plans(T, fft_size)
+    G_vects = G_vectors(fft_size, array_type)
```
antoine-levitt, line 194: we usually use vec instead of vect when we need abbrevations (or just Gs)

src/PlaneWaveBasis.jl
```diff
@@ -255,15 +256,14 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Number, fft_size, variational,
...
+        G_vects, r_vectors,
         kpoints, kweights_thisproc, kgrid, kshift,
         kcoords_global, kweights_global, comm_kpts, krange_thisproc, krange_allprocs,
         symmetries, symmetries_respect_rgrid, terms)
-
     # Instantiate the terms with the basis
     for (it, t) in enumerate(model.term_types)
         term_name = string(nameof(typeof(t)))
-        @timing "Instantiation $term_name" basis.terms[it] = t(basis)
+        @timing "Instantiation $term_name" basis.terms[it] = t(basis, array_type = array_type)
```
antoine-levitt, line 266: this array type should be deduced from the basis array type

src/PlaneWaveBasis.jl
```diff
@@ -331,6 +331,11 @@ end
 The wave vectors `G` in reduced (integer) coordinates for a cubic basis set
 of given sizes.
 """
+function G_vectors(fft_size::Union{Tuple,AbstractVector}, array_type::UnionAll)
```
antoine-levitt, line 334: UnionAll is an implementation detail I think. Type it as ::Type if you really want to type it (ie if you need it for dispatch or for clarity) or just don't type it at all

src/densities.jl
```diff
@@ -33,17 +33,27 @@ grid `basis`, where the individual k-points are occupied according to `occupatio
 
     @sync for (ichunk, chunk) in enumerate(Iterators.partition(ik_n, chunk_length))
         Threads.@spawn for (ik, n) in chunk  # spawn a task per chunk
-            ψnk_real = ψnk_real_chunklocal[ichunk]
-            ρ_loc = ρ_chunklocal[ichunk]
-
             kpt = basis.kpoints[ik]
-            G_to_r!(ψnk_real, basis, kpt, ψ[ik][:, n])
-            ρ_loc[:, :, :, kpt.spin] .+= occupation[ik][n] .* basis.kweights[ik] .* abs2.(ψnk_real)
+            #TODO: is this the right way to got? Probably rewrite compute_density for GPUArrays
```
antoine-levitt, line 37: I think you can just leave the code as it is and just have the chunklocal array be of the correct type? threads+GPU is a fun interaction but let's assume people don't mix those for now

src/densities.jl
```diff
@@ -33,17 +33,27 @@ grid `basis`, where the individual k-points are occupied according to `occupatio
...
+                G_to_r!(ψnk_real, basis, kpt, ψ[ik][:, n])
+                ρ_loc[:, :, :, kpt.spin] .+= occupation[ik][n] .* basis.kweights[ik] .* abs2.(ψnk_real)
+            end
         end
     end
 
     ρ = sum(ρ_chunklocal)
     mpi_sum!(ρ, basis.comm_kpts)
+    array_type = typeof(similar(basis.G_vectors,complex(T), size(ρ)))
+    ρ = convert(array_type, ρ)
```
antoine-levitt, line 56: Can't you just do the full thing on the GPU?

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -43,20 +43,96 @@
...
-function array_mul(X::AbstractArray{T}, Y) where T
-    Z = Array{T}(undef, size(X, 1), size(Y, 2))
-    mul!(Z, X, Y)
+
+# For now, BlockVector can store arrays of different types (for example, an element of type views and one of type Matrix). Maybe for performance issues it should only store arrays of the same type?
+
+struct BlockVector
+    blocks::Tuple
+    size::Tuple{Int64,Int64}
+end
```
antoine-levitt, line 55: that's a bit too generic I think. For our purposes you can get by with a simplified version. Check out the history of this file, we used to have a "bmul" thing that did the job reasonably well and was a bit simpler

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -43,20 +43,96 @@
...
 end
 
+@timing function rayleigh_ritz(XAX::CuArray, N)
+    if eltype(XAX) <: Complex
+        vals, vects = CUDA.CUSOLVER.heevd!('V','U',XAX)
+    else
+        vals, vects = CUDA.CUSOLVER.syevd!('V','U',XAX)
+    end
+    vects[:,1:N], vals[1:N]
+end
```
antoine-levitt, line 135: Is this really needed? Did you open an issue over at CUDA.jl?

antoine-levitt: Also for this you can just type pirate it (at least for now...) and add it to workarounds/

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -170,16 +246,16 @@ end
     niter = 1
     ninners = zeros(Int,0)
     while true
-        BYX = BY'X
-        # XXX the one(T) instead of plain old 1 is because of https://github.com/JuliaArrays/BlockArrays.jl/issues/176
-        mul!(X, Y, BYX, -one(T), one(T)) # X -= Y*BY'X
+        BYX = block_overlap(BY,X) # = BY' X
+        X .-= block_mul(Y, BYX) #X = X -Y * BYX
```
antoine-levitt, line 250: this is a bit less efficient than the previous version. Maybe have the custom type override `mul!(a,b,c,alpha,beta)`?

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -213,13 +289,12 @@ end
 function final_retval(X, AX, resid_history, niter, n_matvec)
     λ = real(diag(X' * AX))
     residuals = AX .- X*Diagonal(λ)
-    (λ=λ, X=X,
+    (λ=Array(λ), X=X,
      residual_norms=[norm(residuals[:, i]) for i in 1:size(residuals, 2)],
-     residual_history=resid_history[:, 1:niter+1],
-     n_matvec=n_matvec)
+     residual_history=resid_history[:, 1:niter+1], n_matvec=n_matvec)
+    #λ doesn't have to be on the GPU, but X does (ψ should always be on GPU throughout the code).
```
antoine-levitt, line 295: No reference to DFTK from inside LOBPCG (see https://github.com/JuliaMolSim/DFTK.jl/issues/690)

antoine-levitt: LOBPCG should just return lambda on the GPU if everything is provided there. If you want to get it to an array, do it outside of LOBPCG

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -230,12 +305,14 @@ end
                         miniter=1, ortho_tol=2eps(real(eltype(X))),
                         n_conv_check=nothing, display_progress=false)
     N, M = size(X)
+    typearray = typeof(X)
```
antoine-levitt, line 308: you use array_type above

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -271,6 +348,7 @@ end
     nlocked = 0
     niter = 0  # the first iteration is fake
     λs = @views [(X[:,n]'*AX[:,n]) / (X[:,n]'BX[:,n]) for n=1:M]
+    λs = oftype(X[:,1], λs) #Offload to GPU if needed
```
antoine-levitt, line 351: We need a good solution for this type of array construction. Let's discuss it

src/eigen/lobpcg_hyper_impl.jl
antoine-levitt: Also from this you get the specific array type I believe, like Array{T,N}

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -431,4 +511,4 @@ end
     end
 
     final_retval(full_X, full_AX, resid_history, maxiter, n_matvec)
-end
+end
```
antoine-levitt, line 514: We need newlines at the end of files, otherwise many tools panic. It's generally accepted practice I believe. I've setup my editor (emacs) to enforce it automatically long ago and never had trouble.

src/eigen/preconditioners.jl
```diff
@@ -27,7 +27,7 @@ PreconditionerNone(basis, kpt) = I
 mutable struct PreconditionerTPA{T <: Real}
     basis::PlaneWaveBasis
     kpt::Kpoint
-    kin::Vector{T}  # kinetic energy of every G
+    kin::AbstractVector{T}  # kinetic energy of every G
```
antoine-levitt, line 30: this incurs dynamic dispatch

src/eigen/preconditioners.jl
```diff
@@ -37,6 +37,8 @@ function PreconditionerTPA(basis::PlaneWaveBasis{T}, kpt::Kpoint; default_shift=
     isempty(kinetic_term) && error("Preconditioner should be disabled when no Kinetic term is used.")
     scaling = only(kinetic_term).scaling_factor
     kin = Vector{T}([scaling * sum(abs2, q) for q in Gplusk_vectors_cart(basis, kpt)] ./ 2)
+    array_type = typeof(similar(basis.G_vectors, T, size(kin)))
```
antoine-levitt, line 40: this builds a new array just for the type computation

src/eigen/preconditioners.jl
```diff
@@ -37,6 +37,8 @@ function PreconditionerTPA(basis::PlaneWaveBasis{T}, kpt::Kpoint; default_shift=
...
+    kin = convert(array_type, kin)
```
antoine-levitt, line 41: `map` over Gplusk_vectors_cart (which should be on the GPU)?

src/eigen/preconditioners.jl
```diff
@@ -68,4 +70,11 @@ end
 
 function precondprep!(P::PreconditionerTPA, X)
     P.mean_kin = [real(dot(x, Diagonal(P.kin), x)) for x in eachcol(X)]
```
antoine-levitt, line 72: Interesting, what do you get if you map over eachcol(::GPUArray)? An Array or a GPUArray?

src/eigen/preconditioners.jl
```diff
@@ -68,4 +70,11 @@ end
...
+    array_type = typeof(similar(X,eltype(X),size(P.mean_kin)))
+    P.mean_kin = convert(array_type, P.mean_kin)
 end
+
+#TODO: remove this if it implemented in GPUArrays
+import LinearAlgebra.dot
+using GPUArrays
+LinearAlgebra.dot(x::AbstractGPUArray, D::Diagonal,y::AbstractGPUArray) = x'*(D*y)
```
antoine-levitt, line 80: 🏴‍☠️! Put that in workarounds/ so we at least isolate places where we're bad

src/fft.jl
```diff
@@ -253,14 +255,16 @@ _fftw_flags(::Type{Float64}) = FFTW.MEASURE
...
+#Removed the flags as CUDA's plan_fft doesn't need flags. If this is a performance issue, we should check array_type's type then call either FFTW.plan_fft(tmp, flags = ...) or CUDA.plan_fft(tmp)
+function build_fft_plans(array_type::AbstractArray{T}, fft_size) where {T<:Union{Float32,Float64}}
+    tmp = similar(array_type, Complex{Float64}, fft_size...)
+    ipFFT = AbstractFFTs.plan_fft!(tmp)
+    opFFT = AbstractFFTs.plan_fft(tmp)
     # backward by inverting and stripping off normalizations
     ipFFT, opFFT, inv(ipFFT).p, inv(opFFT).p
 end
 
+LinearAlgebra.mul!(Y::AbstractGPUArray, p::AbstractFFTs.Plan, X::AbstractGPUArray) = Y .= p *X
```
antoine-levitt, line 267: Ooof that's very bad for performance. Is this really needed?

src/guess_density.jl
```diff
@@ -66,7 +66,8 @@ function _guess_spin_density(basis::PlaneWaveBasis{T}, atoms, positions, magneti
         @warn("Returning zero spin density guess, because no initial magnetization has " *
               "been specified in any of the given elements / atoms. Your SCF will likely " *
               "not converge to a spin-broken solution.")
-        return zeros(T, basis.fft_size)
+        array_type = typeof(similar(basis.G_vectors, T, basis.fft_size))
+        return convert(array_type,zeros(T, basis.fft_size))
```
antoine-levitt, line 70: to be discussed

src/guess_density.jl
```diff
@@ -93,23 +94,23 @@ which follow the functional form
 and are placed at `position` (in fractional coordinates).
 """
 function gaussian_superposition(basis::PlaneWaveBasis{T}, gaussians) where {T}
-    ρ = zeros(complex(T), basis.fft_size)
-    isempty(gaussians) && return G_to_r(basis, ρ)
-
-    # Fill ρ with the (unnormalized) Fourier transform, i.e. ∫ e^{-iGx} f(x) dx,
-    # where f(x) is a weighted gaussian
```
antoine-levitt, line 100: you removed the nice comment :(

src/guess_density.jl
```diff
@@ -93,23 +94,23 @@ which follow the functional form
...
-    #
-    # is formed from a superposition of atomic densities, each scaled by a prefactor
-    for (iG, G) in enumerate(G_vectors(basis))
-        Gsq = sum(abs2, basis.model.recip_lattice * G)
+    ρ = deepcopy(basis.G_vectors)
+    #These copies are required so that recip_lattice and gaussians are isbits (GPU compatibility)
+    recip_lattice = basis.model.recip_lattice
```
antoine-levitt, line 99: This is not a copy?

antoine-levitt: You might be running into the big bad "captured variable in closure" performance bug

src/orbitals.jl
```diff
@@ -51,6 +51,9 @@ function unsafe_unpack_ψ(x, sizes_ψ)
 end
 unpack_ψ(x, sizes_ψ) = deepcopy(unsafe_unpack_ψ(x, sizes_ψ))
 
+using Random
 function random_orbitals(basis::PlaneWaveBasis{T}, kpt::Kpoint, howmany) where {T}
-    ortho_qr(randn(Complex{T}, length(G_vectors(basis, kpt)), howmany))
-end
+    orbitals = similar(basis.G_vectors, Complex{T}, length(G_vectors(basis, kpt)), howmany)
+    randn!(TaskLocalRNG(), orbitals) #Force the use of GPUArrays.jl's random function if using the GPU
```
antoine-levitt, line 57: Interesting and a bit annoying. Make a note to discuss it with Valentin?

src/terms/local.jl
```diff
@@ -66,24 +66,25 @@ end
...
     model = basis.model
 
     # pot_fourier is <e_G|V|e_G'> expanded in a basis of e_{G-G'}
     # Since V is a sum of radial functions located at atomic
     # positions, this involves a form factor (`local_potential_fourier`)
     # and a structure factor e^{-i G·r}
 
-    pot_fourier = map(G_vectors(basis)) do G
+    #This operation needs to be done only once, so let's try to make it happen on CPU (else we needs to isbitsify the pseudopotentials)
+    pot_fourier = map(Array(G_vectors(basis))) do G
```
antoine-levitt, line 78: that's a bit wasteful (the copy)

antoine-levitt (commented):
This is awesome!! A lot of minor comments, but I'm really impressed with how simple it turns out to be (which of course you should take as a compliment: making things look simple is very hard). Most important is we should talk about a good API for defining new arrays.

antoine-levitt:
It'd be good to merge some bits before we get the full GPU story, so that it's easier for you to stay and sync and for us to review. Eg you could have in separate PRs the uncontroversial fixes (eg zeros->similar, etc) that we can just merge easily now, and the LOBPCG fixes to not use blockarrays

src/fft.jl
GVigne: I don't think it's bad for performance, as it is actually what is recommended by the [AbstractFFT documentation](https://juliamath.github.io/AbstractFFTs.jl/stable/api/#AbstractFFTs.plan_fft). We should either use `*` or use the `mul! `function: the thing is there are a few restrictions to use `mul!` (the input array must be a complex floating-point array like the output). 
Initially I hadn't understood that, so that's why there is this weird workaround. But now I fixed the code, this is no longer useful, and I am removing it. Thanks for pointing that out!

antoine-levitt: Great! Basically any time there's a `!` it should mean it's a place where performance and allocations matter (otherwise, people should just use the non-mutating version, as it's clearer, easier to reason with and hides fewer bugs), so what you do here (which allocates a new array for p*X) breaks that implicit contract

src/PlaneWaveBasis.jl
GVigne: I think I'm going to change it to ::Type just for code clarity purpose.

src/eigen/lobpcg_hyper_impl.jl
GVigne: I had never heard of this convention before, but after some reading it does make sense, and there is also an option in VSCode for it to do automatically. Thanks!

src/eigen/preconditioners.jl
GVigne: So far this type of workaround happens twice (for `dot` and for the equality test `B!=I` in LOBPCG), but it should really go in CUDA or GPUArrays. I put it in workarounds/ for now but it should soon go away.

src/guess_density.jl
GVigne: Oh no!

src/densities.jl
GVigne: Yes, that was one of the part of the code which was functional but I must admit that I didn't understand everything that was happening. It was a good first shot though.
If I understand correctly, you suggest I make an other function which does not rely on threads, so that it can be immediately GPU compatible?

antoine-levitt: The function allocates per-thread buffers, then each thread does its stuff, and the result is accumulated. You can assume that the code is single-threaded, and put the buffer on the GPU. Just add an assert that GPU code is run single-threaded to be safe.

GVigne: Ok, that makes sense. Done!

src/densities.jl
GVigne: Indeed: this was fixed following the conversation about multithreading/CUDA (it came down to the same issue).

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -43,20 +43,100 @@
 vprintln(args...) = nothing
```
mfherbst, line 43: I think the modifications here should be a separate PR that we merge in first.

src/workarounds/gpu_computations.jl
```diff
@@ -0,0 +1,15 @@
+#TODO: remove this when it is implemented in GPUArrays
```
mfherbst, line 1: Just call the file gpu_arrays.jl ?

mfherbst (commented):
Some more detailed comments from my side.

Overall a very solid start @GVigne !

src/PlaneWaveBasis.jl
GVigne: Ok, but if I am not mistaken, we are going to loose some information about the G_vectors and r_vectors.
Doing 
```
G_vectors::AT
r_vectors::AT
```
will work, but not something like this
```
G_vectors::AT{Vec3{Int}, 3}
r_vectors::AT{Vec3{VT }, 3}
```

mfherbst: Yes, that's true. In that case you would need two type arguments (one for `G_vectors::Garray` and one for `r_vectors::Rarray`) and then the `AT` could be implicitly obtained from the `G_vectors`.

mfherbst: I'm not sure we want the two type arguments, however. @antoine-levitt thoughts?

antoine-levitt: you can do {GT, RT, AT} where {GT <: AT, RT <: AT}. But yeah not great. OTOH we do want gvecs and rvecs to be statically typed, so yeah I don't think there's a way around it. What you can do is push the not interesting types further down the list, so you can do function f(basis::pwbasis{T, AT})

src/guess_density.jl
GVigne: Actually, I realised we don't need that copy at all. Later on, I am forced to use `map` instead of `map!` (we can't use the inplace version as we are converting an Array of Vec3 to an Array of Complex), so a new array is build there. So I can just remove this line and put  `ρ = map(build_ρ, basis.G_vectors)` instead of  `ρ = map(build_ρ, ρ)`.

src/terms/local.jl
GVigne: As I tried to explain in the comment, either we use directly the G_vectors and then the entire map function needs to be a kernel, or we do it on CPU (it's only going to happen once, when we build the PlaneWaveBasis). If we don't want to make a copy, then all the element in the map need to be isbits, so we have to convert the pseudopotentials to a isbits structure. 
I agree it's a bit ugly, but for performance purposes it only happens once so I thought it was ok.

src/DFTK.jl
```diff
@@ -13,6 +13,10 @@ using spglib_jll
 using Unitful
 using UnitfulAtomic
 using ForwardDiff
+using AbstractFFTs
+using GPUArrays
+using CUDA
+using Random
```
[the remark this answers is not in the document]
GVigne: I think we will need to discuss dependencies (especially if we want to move LOBPCG out of DFTK, that can take some work): I also didn't really know where to put my imports and how they were managed in a big package, so there is room for improvement.

src/PlaneWaveBasis.jl
```diff
@@ -331,6 +331,11 @@ end
 The wave vectors `G` in reduced (integer) coordinates for a cubic basis set
 of given sizes.
 """
+function G_vectors(fft_size::Union{Tuple,AbstractVector}, array_type::Type)
+    #This functions allows to convert the G_vectors (currently being built on the CPU) to a GPU Array.
```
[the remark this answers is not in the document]
GVigne: Thanks for the note on `#`: I have my IDE configured to write on multiple lines if a piece of code would be bigger than my screen size, so I tend to forget this.

src/common/ortho.jl
```diff
@@ -1,2 +1,3 @@
 # Orthonormalize
-ortho_qr(φk) = Matrix(qr(φk).Q)
+ortho_qr(φk::AbstractArray) = Matrix(qr(φk).Q) #LinearAlgebra.QRCompactWYQ -> Matrix
```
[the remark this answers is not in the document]
GVigne: Couldn't we simply do this?
```
ortho_qr(φk::Array) = Matrix(qr(φk).Q) 
ortho_qr(φk::T) where T <: AbstractGPUArray = T(qr(φk).Q) 
```
Another way to do it would be to have only one function and to get the the "base type" of `φk`, then convert `qr(φk).Q` to this type: this can be done by calling T.name.wrapper (or maybe one day a dedicated function in Base). We would then have the following code:
`ortho_qr(φk::T) where T <: AbstractArray = T.name.wrapper(qr(φk).Q)`

src/densities.jl
```diff
@@ -19,25 +19,28 @@ grid `basis`, where the individual k-points are occupied according to `occupatio
...
+    Threads.nthreads() !=1 && G_vectors(basis) isa AbstractGPUArray && error("Can't mix multi-threading and GPU computations yet.") #We assume there is only 1 thread
     T = promote_type(eltype(basis), real(eltype(ψ[1])))
 
     # we split the total iteration range (ik, n) in chunks, and parallelize over them
     ik_n = [(ik, n) for ik = 1:length(basis.kpoints) for n = 1:size(ψ[ik], 2)]
     chunk_length = cld(length(ik_n), Threads.nthreads())
 
     # chunk-local variables
-    ρ_chunklocal = Array{T,4}[zeros(T, basis.fft_size..., basis.model.n_spin_components)
+    array_type = typeof(similar(G_vectors(basis),T, basis.fft_size..., basis.model.n_spin_components))
```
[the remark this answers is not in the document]
GVigne: For all of these array_type questions, I found out I could call `typeof(...).name.wrapper`, and it works in our case. I am trying to get this to be [added to Base](https://github.com/JuliaLang/julia/pull/46213), seeing that I am clearly not the first to have this problem: worst case scenario we can define our own function.

src/densities.jl
```diff
@@ -19,25 +19,28 @@ grid `basis`, where the individual k-points are occupied according to `occupatio
...
+    ρ_chunklocal = [convert(array_type, zeros(T, basis.fft_size..., basis.model.n_spin_components))
                                for _ = 1:Threads.nthreads()]
-    ψnk_real_chunklocal = Array{complex(T),3}[zeros(complex(T), basis.fft_size)
-                                               for _ = 1:Threads.nthreads()]
+    array_type = typeof(similar(G_vectors(basis),complex(T), basis.fft_size))
+    ψnk_real_chunklocal = [convert(array_type, zeros(complex(T), basis.fft_size)) 
+                                for _ = 1:Threads.nthreads()]
```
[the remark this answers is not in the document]
GVigne: Oops, fixing that!

src/fft.jl
```diff
@@ -253,15 +256,15 @@ _fftw_flags(::Type{Float64}) = FFTW.MEASURE
 Plan a FFT of type `T` and size `fft_size`, spending some time on finding an
 optimal algorithm. (Inplace, out-of-place) x (forward, backward) FFT plans are returned.
 """
-function build_fft_plans(T::Union{Type{Float32}, Type{Float64}}, fft_size)
-    tmp = Array{Complex{T}}(undef, fft_size...)
-    ipFFT = FFTW.plan_fft!(tmp, flags=_fftw_flags(T))
-    opFFT = FFTW.plan_fft(tmp, flags=_fftw_flags(T))
```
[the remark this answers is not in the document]
GVigne: We still might need FFTW when doing multi-threading. But for the fft file, I guess we could remove the dependency, although it would have to be replaced by AbstractFFTs.

src/eigen/diag_lobpcg_hyper.jl
```diff
@@ -9,8 +9,10 @@ function lobpcg_hyper(A, X0; maxiter=100, prec=nothing,
     result = LOBPCG(A, X0, I, prec, tol, maxiter; n_conv_check=n_conv_check, kwargs...)
 
     n_conv_check === nothing && (n_conv_check = size(X0, 2))
+
     converged = maximum(result.residual_norms[1:n_conv_check]) < tol
     iterations = size(result.residual_history, 2) - 1
+    λ = Array(result.λ) #TODO: offload this to gpu? Careful then, as self_consistent_field's eigenvalues will be a CuArray -> due to the Smearing.occupation function, occupation will also be a CuArray, so no scalar indexing (in ene_ops, in compute_density...)
```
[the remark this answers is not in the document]
GVigne: Can't we just add this at the end of `diagonalize_all_kblocks`? High level objects go in and a tuple goes out with all necessary fields: we simply transfer to CPU when all blocks have been diagonalized.
Or we could say we want `λ` to be on the GPU but I'm not sure it's necessary, and it will require quite some work as this means that `occupation` will be on the GPU as well (so no scalar indexing, which is something done a lot for this array).

src/eigen/preconditioners.jl
GVigne: We get a GPUArray. The is the same type of problem as the issue you raised [here](https://github.com/JuliaMolSim/DFTK.jl/pull/697#discussion_r929222161). List comprehension is really handy for CPU but sometimes can't be directly used when building GPU Arrays.

src/eigen/preconditioners.jl
GVigne: I'm having some trouble writing `Gplusk_vectors_cart ` as a GPU-compatible function.
Currently, we are broadcasting functions such as `vector_red_to_cart(model::Model, rred)` onto the `G_vectors`. However:
-it is not possible to change these functions to have the signature (recip_lattice::Mat3, rred) as broadcast fails to find common dimensions
-we shouldn't change the function's signature, as sometimes we need the `recip_lattice`, sometimes we need the inverse, and sometimes we need to compute their adjoint
-for GPUs, we can't map or broadcast using a function taking model as an argument as it is not `isbits` (and many of its fields shouldn't be on the GPU anyway).

So far, the best thing I can think of is to replace the functions such as `vector_red_to_cart` into functions returning a static matrix `A` (basically extracting the model's lattice and maybe computing the adjoint) and then mapping the `A*G` operation onto the `G_vectors`, but it's not very satisfying.

PR #698, opened 2022-07-28
Change default thread setup

mfherbst:
You learn every day regarding the peculiarities of threading. Short story is: The default setup in pretty much all blas libraries is suboptimal.

### DIFF
--- a/src/common/threading.jl
+++ b/src/common/threading.jl
@@ -1,8 +1,20 @@
 import FFTW
 using LinearAlgebra
 
-function setup_threading(;n_fft=1, n_blas=Threads.nthreads())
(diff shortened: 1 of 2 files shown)


antoine-levitt:
That doesn't seem to correspond with the thing you link? Since we don't call (much) blas inside of julia threaded regions, it's better to set nblas=nthreads, no ?

mfherbst:
Actually you're right I confused myself. Hmm. I still wonder whether `n_blas=1` is a better default. It almost always is faster from my experience. But I certainly don't have proper benchmark data to back that up.

mfherbst:
Ach whatever let's not change it for now. We probably should test this more rigorously first.

PR #701, opened 2022-07-30
 Add documentation about conducting a convergence study

killah-t-cell:
just fixing the old PR after a bad merge


mfherbst:
@killah-t-cell Just corrected an erroneously commited file.

killah-t-cell:
Thank you!

mfherbst:
I did a final refactoring to make it easier to regenerate the static plots in case this becomes necessary in the future (which I almost bet it will).

PR #710, opened 2022-08-22
Add a few NaN checks in LOBPCG

GVigne:
This PR adds a few NaN checks in LOBPCG. So far, I have mostly seen issues with NaNs around the cholesky decomposition, so this is where I focused most of the checks. Whenever we do a cholesky decomposition, we now make sure that the result does not have any NaNs in it.
On a side note, I run the tests on `R` (ie the result of the cholesky decomposition) instead of `O` (the input) as `O` is Hermitian: somehow, running `any(isnan,O)` when `O` is the Hermitian view of a GPU array fails due to scalar indexing. This is probably a bug/missed implementation. It doesn't change much in our case: if `O` has NaNs, then `R` also has NaNs.

### DIFF
--- a/src/eigen/lobpcg_hyper_impl.jl
+++ b/src/eigen/lobpcg_hyper_impl.jl
@@ -41,6 +41,8 @@
 
 # vprintln(args...) = println(args...)  # Uncomment for output
 vprintln(args...) = nothing
+# check_nan(X, msg) = any(isnan,X) && error(msg) # Uncomment for NaN checks.
+check_nan(args...) = nothing
 
 using LinearAlgebra
 using BlockArrays # used for the `mortar` command which makes block matrices
@@ -65,7 +67,7 @@ end
 function B_ortho!(X, BX)
     O = Hermitian(X'*BX)
     U = cholesky(O).U
-    any(isnan, U) && error("B_ortho!(X,BX) fails because cholesky decomposition has NaNs.")
+    check_nan(U, "B_ortho! fails because cholesky decomposition has NaNs.")
     rdiv!(X, U)
     rdiv!(BX, U)
 end
@@ -95,7 +97,6 @@ normest(M) = maximum(abs.(diag(M))) + norm(M - Diagonal(diag(M)))
         catch err
             @assert isa(err, PosDefException)
             vprintln("fail")
-            any(isnan, R) && error("Ortho(X) fails because cholesky decomposition has NaNs.")
             # see https://arxiv.org/pdf/1809.11085.pdf for a nice analysis
             # We are not being very clever here; but this should very rarely happen
             # so it should be OK
@@ -111,7 +112,6 @@ normest(M) = maximum(abs.(diag(M))) + norm(M - Diagonal(diag(M)))
                 catch err
                     @assert isa(err, PosDefException)
                 end
-                any(isnan, R) && error("Cholesky shifting fails because cholesky decomposition has NaNs.")
                 nbad += 1
                 if nbad > 10
                     error("Cholesky shifting is failing badly, this should never happen")
@@ -120,6 +120,7 @@ normest(M) = maximum(abs.(diag(M))) + norm(M - Diagonal(diag(M)))
             success = false
         end
         invR = inv(R)
+        check_nan(invR, "Ortho(X) fails because invR has NaNs.")
         rmul!(X, invR) # we do not use X/R because we use invR next
 
         # We would like growth_factor *= opnorm(inv(R)) but it's too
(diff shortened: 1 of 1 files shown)


src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -132,6 +135,7 @@ normest(M) = maximum(abs.(diag(M))) + norm(M - Diagonal(diag(M)))
         # condR = 1/LAPACK.trcon!('I', 'U', 'N', Array(R))
         condR = normest(R)*norminvR # in practice this seems to be an OK estimate
 
+        any(isnan, condR) && error("The condition number of R is NaN. Check R's norm.")
```
antoine-levitt, line 138: condR is a scalar

antoine-levitt:
Hm not sure all of these are really needed... feels a bit haphazard now, can you maybe remove some of them? Another place where NaN can appear is through the A/B/P applies (which are under control of the user), can you add checks there instead? I'm not sure these checks will not have a performance penalty but it shouldn't be too huge, can you check with profiling? (on CPU)

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -41,6 +41,8 @@
 
 # vprintln(args...) = println(args...)  # Uncomment for output
 vprintln(args...) = nothing
+# check_nan(X, msg) = any(isnan,X) && error(msg) # Uncomment for NaN checks.
```
mfherbst, line 44: space before comma

mfherbst: two spaces before comment

mfherbst (commented):
Modulo the performance checking Antoine mentioned LGTM.

GVigne:
I ran the computations with a supercell with parameters [4,4,2], so a medium-small example. Every NaN check took more or less 10 ms, for a total of approximately 0.5s. BenchmarkTools tells me that the code ran for 108s. So overall, we are looking at a 0.5% increase in time: that's why we gated it behind a function, so as to only use it for debugging purpose and not every single time.

antoine-levitt:
Possibly a better solution is to only run the checks on the "small" quantities (those that have dimension MxM, not NxM), eg XAX but not AX. That should eliminate the performance penalty while still being a useful diagnostic

mfherbst:
@antoine-levitt I like that idea. In particular that would allow us to enable this check by default.

GVigne:
Ok, why not. Just a few things though:
- should we also remove the check after the preconditioner is applied? I guess the user will often want to define his own preconditionner, so that will happen a lot - and we are currently doing the NaN check on a NxM matrix
- the only place where we explicitly compute XAX is in the rayleigh-ritz function, so I guess this is where we could put the NaN check.

Anyways, by keeping 3 NaN checks (in `B_ortho!` and `ortho!` on the result of the cholesky decomposition and in `rayleigh-ritz`), the time spent doing these tests amounts to about 3 milliseconds for the same example as above, so I guess it's fine. We are doing a lot more tests because `ortho!` and `rayleigh-ritz` happen more often, but they take much less time.

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -41,6 +41,7 @@
 
 # vprintln(args...) = println(args...)  # Uncomment for output
 vprintln(args...) = nothing
+check_nan(X, msg) = any(isnan, X) && error(msg)
```
mfherbst, line 44: I would completely inline this function. It is basically no shorter if you make it explicit (which is clearer).

mfherbst:
@GVigne  Sounds fair to me.

antoine-levitt:
OK, so typically with these, the error message doesn't actually add anything compared to the stacktrace. I know @mfherbst you don't like "internal" errors showing up to the user, but can't we just do a `@assert` without error message here? My point is that reading stacktraces is a life skill that we should assume people have, rather than trying to have nice error messages.

mfherbst:
Assert is generally discouraged, because it can be disabled. But perhaps this is what we want here. I feel an error is more explicit here, but overall I don't have a strong opinion. Probably we need to adapt anyways to what the IterativeSolver people are doing.

GVigne:
I took a look at IterativeSolvers, and I don't feel there is a clear-cut answer. It's a bit hard to know who is responsible for catching this type of error.

From my testing and what I saw in the code, there is barely any NaN check in IterativeSolvers: when using their version of LOBPCG or their SVDL method with NaNs, I get errors from LAPACK. Even these errors from LAPACK aren't dealt with the same way: when using `eigen` with NaNs, you get a user-friendly error (`ERROR: ArgumentError: matrix contains Infs or NaNs`), but when using the SVD, you get a not so friendly error (`ERROR: LAPACKException(18)`).

It really seems there isn't any real norm for this, and I couldn't really decide which version to choose (error or assert). In the end, I think assert might be slightly better for code clarity and to show immediately to the end-user what the assumptions are: if the assert fails, the code who triggered the assert is shown, so at least the user knows exactly what kind of test fails. I updated the PR.

mfherbst:
Interesting @GVigne, so you prefer a `throw(ArgumentError, "nice message")` over the cryptic Lapack error but in your code you go for an assertion, which in my opinion is less clear than the `ArgumentError` ... but whatever I don't think it matters too much here.

PR #711, opened 2022-08-23
Make LOBPCG GPU-compatible

GVigne:
This draft PR implements GPU compatibility for LOBPCG. 
At the highest level, the goal is to be able to call LOBPCG with either Arrays or any type of GPU Arrays. This means the user should be able to make the following calls (A is a Hermitian matrix and X is vector)
`LOBPCG(A,X) #CPU`
`LOBPCG(CuArray(A),CuArray(X)) #GPU`
`LOBPCG(ROCArray(A),ROCArray(X)) #GPU`
LOBPCG returns a named tuple. One of the field is an array of the eigenvalues found, and another is the corresponding eigenvectors. These two fields have the same array type as the input: if the user calls LOBPCG with CuArrays, then the eigenvalues and eigenvectors will be stored in a CuArray.

I tried to make the code as abstract as possible so as not to rely on a hardware-specific library. The only part where I needed to call specific CUDA functions (I am using an NVIDIA GPU) is when computing eigenvalues. If I am not mistaken, there is no generic `eigen` function in CUDA.jl, so I had to manually call the eigendecomposition functions (CUDA.syevd! and CUDA.heevd!).

### DIFF
--- a/src/eigen/lobpcg_hyper_impl.jl
+++ b/src/eigen/lobpcg_hyper_impl.jl
@@ -43,17 +43,83 @@
 vprintln(args...) = nothing
 
 using LinearAlgebra
-using BlockArrays # used for the `mortar` command which makes block matrices
+import Base: *
+include("../workarounds/gpu_arrays.jl")
 
-# when X or Y are BlockArrays, this makes the return value be a proper array (not a BlockArray)
-function array_mul(X::AbstractArray{T}, Y) where T
-    Z = Array{T}(undef, size(X, 1), size(Y, 2))
-    mul!(Z, X, Y)
+# For now, BlockMatrix can store arrays of different types (for example, an element 
+# of type views and one of type Matrix). Maybe for performance issues it should only
+# store arrays of the same type?
+
+struct BlockMatrix
+    blocks::Tuple
+    size::Tuple{Int64,Int64}
+end
+
+"""
+Build a BlockMatrix containing the given arrays, from left to right.
+This function will fail (for now) if:
+    -the arrays do not all have the same "height" (ie size[1] must match).
+"""
+function make_block_vector(arrays::AbstractArray...)
+    length(arrays) ==0 && error("Empty BlockMatrix is not currently implemented")
+    n_ref= size(arrays[1])[1]
+    m=0
+    for array in arrays
+        n_i, m_i = size(array)
+        n_ref != n_i && error("The given arrays do not have matching 'height': cannot build a BlockMatrix out of them.")
+        m += m_i
+    end
+    BlockMatrix(arrays, (n_ref,m))
+end
+
+
+"""
+Given A and B as two BlockMatrixs [A1, A2, A3], [B1, B2, B3] form the matrix
+A'B (which is not a BlockMatrix). block_overlap also has compatible versions with two Arrays. 
+block_overlap always compute some form of adjoint, ie the product A'*B.
+"""
+@views function block_overlap(A::BlockMatrix, B::BlockMatrix)
+    rows = A.size[2]
+    cols = B.size[2]
+    ret = similar(A.blocks[1], rows, cols)
+
+    orow = 0  # row offset
+    for (iA, blA) in enumerate(A.blocks)
+        ocol = 0  # column offset
+        for (iB, blB) in enumerate(B.blocks)
+            ret[orow .+ (1:size(blA, 2)), ocol .+ (1:size(blB, 2))] = blA' * blB
+            ocol += size(blB, 2)
+        end
+        orow += size(blA, 2)
+    end
+    ret
+end
+
+block_overlap(blocksA::BlockMatrix, B) = block_overlap(blocksA, make_block_vector(B))
+block_overlap(A, B) = A' * B # Default fallback method. Note the adjoint.
+
+"""
+Given A as a BlockMatrix [A1, A2, A3] and B a Matrix, compute the matrix-matrix product
+A * B avoiding a concatenation of the blocks to a dense array. 
+"""
+@views function *(Ablock::BlockMatrix, B)
+    res = Ablock.blocks[1] * B[1:size(Ablock.blocks[1], 2), :]  # First multiplication
+    offset = size(Ablock.blocks[1], 2)
+    for block in Ablock.blocks[2:end]
+        mul!(res, block, B[offset .+ (1:size(block, 2)), :], 1, 1)
+        offset += size(block, 2)
+    end
+    res
+end
+
+function LinearAlgebra.mul!(res,A::BlockMatrix,B::AbstractArray,α,β)
+    # Has slightly better performances than a naive res = α*A*B - β*res
+    mul!(res, A*B, I, α, β)
 end
 
 # Perform a Rayleigh-Ritz for the N first eigenvectors.
-@timing function rayleigh_ritz(X, AX, N)
-    F = eigen(Hermitian(array_mul(X', AX)))
+@timing function rayleigh_ritz(X::BlockMatrix, AX::BlockMatrix, N)
+    F = eigen(Hermitian(block_overlap(X, AX))) # block_overlap(X,AX) is an AbstractArray, not a BlockMatrix
     F.vectors[:,1:N], F.values[1:N]
 end
 
@@ -170,16 +236,16 @@ end
     niter = 1
     ninners = zeros(Int,0)
     while true
-        BYX = BY'X
-        # XXX the one(T) instead of plain old 1 is because of https://github.com/JuliaArrays/BlockArrays.jl/issues/176
+        BYX = block_overlap(BY,X) # = BY' X
         mul!(X, Y, BYX, -one(T), one(T)) # X -= Y*BY'X
         # If the orthogonalization has produced results below 2eps, we drop them
         # This is to be able to orthogonalize eg [1;0] against [e^iθ;0],
         # as can happen in extreme cases in the ortho!(cP, cX)
         dropped = drop!(X)
         if dropped != []
-            @views mul!(X[:, dropped], Y, BY' * (X[:, dropped]), -one(T), one(T)) # X -= Y*BY'X
+            X[:, dropped] .-= Y * block_overlap(BY,X[:, dropped]) #X = X - Y'*BY*X
         end
+
         if norm(BYX) < tol && niter > 1
             push!(ninners, 0)
             break
@@ -215,10 +281,9 @@ function final_retval(X, AX, resid_history, niter, n_matvec)
     residuals = AX .- X*Diagonal(λ)
     (λ=λ, X=X,
      residual_norms=[norm(residuals[:, i]) for i in 1:size(residuals, 2)],
-     residual_history=resid_history[:, 1:niter+1],
-     n_matvec=n_matvec)
+     residual_history=resid_history[:, 1:niter+1], n_matvec=n_matvec)
+    #λ doesn't have to be on the GPU, but X does (ψ should always be on GPU throughout the code).
 end
-
 
 ### The algorithm is Xn+1 = rayleigh_ritz(hcat(Xn, A*Xn, Xn-Xn-1))
 ### We follow the strategy of Hetmaniuk and Lehoucq, and maintain a B-orthonormal basis Y = (X,R,P)
@@ -230,12 +295,13 @@ end
                         miniter=1, ortho_tol=2eps(real(eltype(X))),
                         n_conv_check=nothing, display_progress=false)
     N, M = size(X)
+
     # If N is too small, we will likely get in trouble
     error_message(verb) = "The eigenproblem is too small, and the iterative " *
-                           "eigensolver $verb fail; increase the number of " *
-                           "degrees of freedom, or use a dense eigensolver."
-    N > 3M    || error(error_message("will"))
-    N >= 3M+5 || @warn error_message("might")
+                            "eigensolver $verb fail; increase the number of " *
+                            "degrees of freedom, or use a dense eigensolver."
+     N > 3M    || error(error_message("will"))
+     N >= 3M+5 || @warn error_message("might")
 
     n_conv_check === nothing && (n_conv_check = M)
     resid_history = zeros(real(eltype(X)), M, maxiter+1)
@@ -271,6 +337,7 @@ end
     nlocked = 0
     niter = 0  # the first iteration is fake
     λs = @views [(X[:,n]'*AX[:,n]) / (X[:,n]'BX[:,n]) for n=1:M]
+    λs = oftype(X[:,1], λs) #Offload to GPU if needed
     new_X = X
     new_AX = AX
     new_BX = BX
@@ -286,13 +353,13 @@ end
 
             # Form Rayleigh-Ritz subspace
             if niter > 1
-                Y = mortar((X, R, P))
-                AY = mortar((AX, AR, AP))
-                BY = mortar((BX, BR, BP))  # data shared with (X, R, P) in non-general case
+                Y = make_block_vector(X, R, P)
+                AY = make_block_vector(AX, AR, AP)
+                BY = make_block_vector(BX, BR, BP)  # data shared with (X, R, P) in non-general case
             else
-                Y  = mortar((X, R))
-                AY = mortar((AX, AR))
-                BY = mortar((BX, BR))  # data shared with (X, R) in non-general case
+                Y  = make_block_vector(X, R)
+                AY = make_block_vector(AX, AR)
+                BY = make_block_vector(BX, BR)  # data shared with (X, R) in non-general case
             end
             cX, λs = rayleigh_ritz(Y, AY, M-nlocked)
 
@@ -300,9 +367,9 @@ end
             # wait on updating P because we have to know which vectors
             # to lock (and therefore the residuals) before computing P
             # only for the unlocked vectors. This results in better convergence.
-            new_X  = array_mul(Y, cX)
-            new_AX = array_mul(AY, cX)  # no accuracy loss, since cX orthogonal
-            new_BX = (B == I) ? new_X : array_mul(BY, cX)
+            new_X  = Y * cX
+            new_AX = AY * cX  # no accuracy loss, since cX orthogonal
+            new_BX = (B == I) ? new_X : BY * cX
         end
 
         ### Compute new residuals
@@ -356,20 +423,23 @@ end
             # orthogonalization, see Hetmaniuk & Lehoucq, and Duersch et. al.
             # cP = copy(cX)
             # cP[Xn_indices,:] .= 0
-            e = zeros(eltype(X), size(cX, 1), M - prev_nlocked)
-            for i in 1:length(Xn_indices)
-                e[Xn_indices[i], i] = 1
-            end
+
+            lenXn = length(Xn_indices)
+            e = zero(similar(X, size(cX, 1), M - prev_nlocked))
+            lower_diag = one(similar(X, lenXn, lenXn))
+            #e has zeros everywhere except on one of its lower diagonal
+            e[Xn_indices[1] : last(Xn_indices), 1 : lenXn] = lower_diag
+
             cP = cX .- e
             cP = cP[:, Xn_indices]
             # orthogonalize against all Xn (including newly locked)
             ortho!(cP, cX, cX, tol=ortho_tol)
 
             # Get new P
-            new_P  = array_mul( Y, cP)
-            new_AP = array_mul(AY, cP)
+            new_P  = Y * cP
+            new_AP = AY * cP
             if B != I
-                new_BP = array_mul(BY, cP)
+                new_BP = BY * cP
             else
                 new_BP = new_P
             end
@@ -414,8 +484,8 @@ end
 
         # Orthogonalize R wrt all X, newly active P
         if niter > 0
-            Z  = mortar((full_X, P))
-            BZ = mortar((full_BX, BP))  # data shared with (full_X, P) in non-general case
+            Z  = make_block_vector(full_X, P)
+            BZ = make_block_vector(full_BX, BP) # data shared with (full_X, P) in non-general case
         else
             Z  = full_X
             BZ = full_BX
--- a/src/workarounds/gpu_arrays.jl
+++ b/src/workarounds/gpu_arrays.jl
@@ -1,0 +1,16 @@
+#TODO: remove this when it is implemented in GPUArrays
+import LinearAlgebra.dot, LinearAlgebra.eigen, LinearAlgebra.RealHermSymComplexHerm
+using LinearAlgebra
+using GPUArrays
+using CUDA
+
+LinearAlgebra.dot(x::AbstractGPUArray, D::Diagonal,y::AbstractGPUArray) = x'*(D*y)
+
+function LinearAlgebra.eigen(A::RealHermSymComplexHerm{T,AT}) where {T,AT <: CuArray}
+    if eltype(A) <: Complex
+        vals, vects = CUDA.CUSOLVER.heevd!('V','U', A.data)
+    else
+        vals, vects = CUDA.CUSOLVER.syevd!('V','U',A.data)
+    end
+    (vectors = vects, values = vals)
+end


src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -230,12 +295,13 @@ end
...
     # If N is too small, we will likely get in trouble
     error_message(verb) = "The eigenproblem is too small, and the iterative " *
-                           "eigensolver $verb fail; increase the number of " *
-                           "degrees of freedom, or use a dense eigensolver."
-    N > 3M    || error(error_message("will"))
-    N >= 3M+5 || @warn error_message("might")
+                            "eigensolver $verb fail; increase the number of " *
+                            "degrees of freedom, or use a dense eigensolver."
+     N > 3M    || error(error_message("will"))
+     N >= 3M+5 || @warn error_message("might")
```
mfherbst, line 304: That looks like an unintentional whitespace commit.

src/workarounds/gpu_arrays.jl
```diff
@@ -0,0 +1,16 @@
+#TODO: remove this when it is implemented in GPUArrays
+import LinearAlgebra.dot, LinearAlgebra.eigen, LinearAlgebra.RealHermSymComplexHerm
+using LinearAlgebra
+using GPUArrays
+using CUDA
+
+LinearAlgebra.dot(x::AbstractGPUArray, D::Diagonal,y::AbstractGPUArray) = x'*(D*y)
+
+function LinearAlgebra.eigen(A::RealHermSymComplexHerm{T,AT}) where {T,AT <: CuArray}
+    if eltype(A) <: Complex
```
mfherbst, line 10: Use multiple dispatch on element type of CuArray.

src/workarounds/gpu_arrays.jl
```diff
@@ -0,0 +1,16 @@
+#TODO: remove this when it is implemented in GPUArrays
+import LinearAlgebra.dot, LinearAlgebra.eigen, LinearAlgebra.RealHermSymComplexHerm
+using LinearAlgebra
+using GPUArrays
+using CUDA
+
```
mfherbst, line 6: You should in each case add a comment with the issue numbers in GpuArrays, CUDA etc. where you track these missing functionalities.

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -43,17 +43,83 @@
...
+    BlockMatrix(arrays, (n_ref,m))
+end
+
+
+"""
+Given A and B as two BlockMatrixs [A1, A2, A3], [B1, B2, B3] form the matrix
+A'B (which is not a BlockMatrix). block_overlap also has compatible versions with two Arrays. 
+block_overlap always compute some form of adjoint, ie the product A'*B.
+"""
+@views function block_overlap(A::BlockMatrix, B::BlockMatrix)
```
mfherbst, line 81: Why is this function not covered by tests? Did you check you end up dispatching correctly?

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -43,17 +43,83 @@
...
+    blocks::Tuple
+    size::Tuple{Int64,Int64}
+end
+
+"""
+Build a BlockMatrix containing the given arrays, from left to right.
+This function will fail (for now) if:
+    -the arrays do not all have the same "height" (ie size[1] must match).
+"""
+function make_block_vector(arrays::AbstractArray...)
```
mfherbst, line 63: Just make this a constructor?

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -43,17 +43,83 @@
 vprintln(args...) = nothing
 
 using LinearAlgebra
-using BlockArrays # used for the `mortar` command which makes block matrices
```
mfherbst, line 46: You can also nuke BlockArrays from our dependencies.

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -43,17 +43,83 @@
...
-# when X or Y are BlockArrays, this makes the return value be a proper array (not a BlockArray)
-function array_mul(X::AbstractArray{T}, Y) where T
-    Z = Array{T}(undef, size(X, 1), size(Y, 2))
-    mul!(Z, X, Y)
+# For now, BlockMatrix can store arrays of different types (for example, an element 
+# of type views and one of type Matrix). Maybe for performance issues it should only
+# store arrays of the same type?
+
+struct BlockMatrix
+    blocks::Tuple
```
mfherbst, line 54: This is not type stable, which matters here because of performance. You have to make `Tuple` a concrete type.

src/eigen/lobpcg_hyper_impl.jl
mfherbst: Also not too big a fan of having this function. Why not define
```julia
Base.:*(A::Adjoint{T, <: BlockMatrix}, B::BlockMatrix) where {T})
```
instead?

mfherbst (commented):
A few comments. Also check the formatting.

src/eigen/lobpcg_hyper_impl.jl
GVigne: Yes, I thought this would also be an issue. The thing is, we sometimes build a BlockMatrix with arrays which are views of other arrays. And taking the view of an array builds a `SubArray` which derives from AbstractArray.
So mixed in BlockMatrix we can have either arrays or subarrays which do not have a concrete type in common: the only thing they have in common is that they derive from AbstractArray. I may be wrong, but I can't think of an easy way to make this field concrete by keeping the code the way it is: maybe we shouldn't give a BlockMatrix views?

src/eigen/lobpcg_hyper_impl.jl
GVigne: You're right: no need for a function with a weird name.

src/eigen/lobpcg_hyper_impl.jl
mfherbst: Like this?
```julia
struct BlockMatrix{T <: Tuple}
    blocks::T
end
```

mfherbst: btw. Not sure we should have the `size` as a member. It's easily computed on the fly from the constituents of `blocks`.

GVigne: You're right: I was trying to describe more precisely what was hidden behind the Tuple, but doing `D <: Tuple` will work.
As for the size, I created this member so we don't have to recompute every time the size of the BlockMatrix. As it turns out, we are not using it much and it can be computed very easily, so we can drop it.

src/eigen/lobpcg_hyper_impl.jl
GVigne: At first I thought defining it this way would be clearer: as it turns out, it's just weird and confusing. I re-wrote the `block_overlap` functions, now we only overload `Base.*`.

GVigne:
Thanks for the feedback! I did some modifications to take into account what you said. I also added a few small tests for the BlockMatrix structure in `test/lobpcg` to check if nothing is broken.

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -43,19 +43,96 @@
...
-function array_mul(X::AbstractArray{T}, Y) where {T}
-    Z = Array{T}(undef, size(X, 1), size(Y, 2))
-    mul!(Z, X, Y)
+include("../workarounds/gpu_arrays.jl")
+
+# For now, BlockMatrix can store arrays of different types (for example, an element 
+# of type views and one of type Matrix). Maybe for performance issues it should only
+# store arrays of the same type?
+
+struct BlockMatrix{T <: Number, D <: Tuple} <: AbstractMatrix{T}
```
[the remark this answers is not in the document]
antoine-levitt: In particular it would be nice if it was more clear that it is column blocks, but I don't find a good name, so just make it clear from a comment that it is a horizontal concatenation of blocks

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -43,19 +43,96 @@
...
+    blocks::D
+end
+
+"""
+Build a BlockMatrix containing the given arrays, from left to right.
+This function will fail (for now) if:
+    -the arrays do not all have the same "height" (ie size[1] must match).
```
antoine-levitt, line 62: You're beginning an enumeration with one item

antoine-levitt: height -> number of rows

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -43,19 +43,96 @@
...
+"""
+function BlockMatrix(arrays::AbstractArray...)
+    length(arrays) ==0 && error("Empty BlockMatrix is not currently implemented")
+    n_ref= size(arrays[1], 1)
+    for array in arrays
+        n_i = size(array, 1)
+        n_ref != n_i && error("The given arrays do not have matching 'height': "*
```
antoine-levitt, line 69: this is very internal, no need to have a nice error message, just say @assert n_ref == n_i

src/eigen/lobpcg_hyper_impl.jl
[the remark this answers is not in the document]
antoine-levitt: note that these are relatively simple wrappers only supporting some multiplication routines, and in particular operations with them materialize to plain arrays

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -43,19 +43,96 @@
...
+    (n,m)
+end
+
+Base.Array(A::BlockMatrix)  = hcat(A.blocks...)
+
+Base.adjoint(A::BlockMatrix) = Adjoint(A)
+
+"""
+Given A and B as two BlockMatrices [A1, A2, A3], [B1, B2, B3] form the matrix
+A'B. Return an array, not a BlockMatrix.
```
antoine-levitt, line 90: remove docstring (* does what * does)

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -43,19 +43,96 @@
...
+    A = Aadj.parent
+    rows = size(A)[2]
+    cols = size(B)[2]
+    ret = similar(A.blocks[1], rows, cols)
+
+    orow = 0  # row offset
+    for (iA, blA) in enumerate(A.blocks)
+        ocol = 0  # column offset
+        for (iB, blB) in enumerate(B.blocks)
+            ret[orow .+ (1:size(blA, 2)), ocol .+ (1:size(blB, 2))] = blA' * blB
```
antoine-levitt, line 102: .= for this type of stuff (possibly equivalent here, but better practice for this type of operaitons)

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -43,19 +43,96 @@
...
+    end
+    ret
+end
+
+Base.:*(Aadj::Adjoint{T, <: BlockMatrix}, B::AbstractMatrix) where {T} = Aadj * BlockMatrix(B)
+
+"""
+Given A as a BlockMatrix [A1, A2, A3] and B a Matrix, compute the matrix-matrix product
+A * B avoiding a concatenation of A's blocks to a dense array.
+"""
```
antoine-levitt, line 115: remove docstring (see above)

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -43,19 +43,96 @@
...
+    for block in Ablock.blocks[2:end]
+        mul!(res, block, B[offset .+ (1:size(block, 2)), :], 1, 1)
+        offset += size(block, 2)
+    end
+    res
+end
+
+function LinearAlgebra.mul!(res::AbstractMatrix, Ablock::BlockMatrix,
+                            B::AbstractVecOrMat, α::Number, β::Number)
+    # Has slightly better performances than a naive res = α*A*B - β*res
```
antoine-levitt, line 128: really? even with dots? I find that surprising

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -43,19 +43,96 @@
...
+    mul!(res, Ablock*B, I, α, β)
 end
 
 # Perform a Rayleigh-Ritz for the N first eigenvectors.
-@timing function rayleigh_ritz(X, AX, N)
-    XAX = array_mul(X', AX)
-    @assert all(!isnan, XAX)
-    F = eigen(Hermitian(XAX))
+@timing function rayleigh_ritz(X::BlockMatrix, AX::BlockMatrix, N)
+    # Multiplying two BlockMatrices yields an array, not a BlockMatrix
```
antoine-levitt, line 134: remove comment, remove typing of variables

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -174,16 +251,16 @@ end
     niter = 1
     ninners = zeros(Int,0)
     while true
-        BYX = BY'X
-        # XXX the one(T) instead of plain old 1 is because of https://github.com/JuliaArrays/BlockArrays.jl/issues/176
```
antoine-levitt, line 178: you can just use 1 here now, right?

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -174,16 +251,16 @@ end
...
-        mul!(X, Y, BYX, -one(T), one(T)) # X -= Y*BY'X
+        BYX = BY' * X
+        mul!(X, Y, BYX, -one(T), one(T))  # X -= Y*BY'X
         # If the orthogonalization has produced results below 2eps, we drop them
         # This is to be able to orthogonalize eg [1;0] against [e^iθ;0],
         # as can happen in extreme cases in the ortho!(cP, cX)
         dropped = drop!(X)
         if dropped != []
-            @views mul!(X[:, dropped], Y, BY' * (X[:, dropped]), -one(T), one(T)) # X -= Y*BY'X
+            X[:, dropped] .-= Y * (BY' * X[:, dropped])  # X = X - Y*BY'*X
```
antoine-levitt, line 261: remove comment

antoine-levitt: technically here we should probably do as above, but who cares

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -360,20 +437,23 @@ end
             # orthogonalization, see Hetmaniuk & Lehoucq, and Duersch et. al.
             # cP = copy(cX)
             # cP[Xn_indices,:] .= 0
-            e = zeros(eltype(X), size(cX, 1), M - prev_nlocked)
-            for i in 1:length(Xn_indices)
-                e[Xn_indices[i], i] = 1
-            end
+
+            lenXn = length(Xn_indices)
+            e = zero(similar(X, size(cX, 1), M - prev_nlocked))
```
antoine-levitt, line 442: `zero(similar(` allocates twice. Should we have a `zeros_like` function or something, that could fallback to `zero(similar(` but be better optimized in specific cases?

antoine-levitt: ping @vchuravy who might have an opinion

src/eigen/lobpcg_hyper_impl.jl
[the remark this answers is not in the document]
antoine-levitt: also explain relationship to BlockArrays (it's a lightweight subset of functionality, it materializes to plain arrays, it supports GPU)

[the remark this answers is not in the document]
antoine-levitt: Re name : LazyHcat? (because it's basically equivalent to `hcat` but better optimized)

antoine-levitt (commented):
Super nice! Very cute little struct and nicer code than we had before.

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -43,19 +43,96 @@
...
 
-# when X or Y are BlockArrays, this makes the return value be a proper array (not a BlockArray)
-function array_mul(X::AbstractArray{T}, Y) where {T}
-    Z = Array{T}(undef, size(X, 1), size(Y, 2))
-    mul!(Z, X, Y)
+include("../workarounds/gpu_arrays.jl")
+
+# For now, BlockMatrix can store arrays of different types (for example, an element 
+# of type views and one of type Matrix). Maybe for performance issues it should only
+# store arrays of the same type?
```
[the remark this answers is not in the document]
antoine-levitt: There's no conversion at all here. You just use the tuple elements in `mul!` functions, which dispatches to the correct function. I don't think there's any need to change what you have now

src/eigen/lobpcg_hyper_impl.jl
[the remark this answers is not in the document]
GVigne: I think LazyHcat is not too bad: it's clearer to have "hcat" than to just have the generic "block" which doesn't give a lot of information. And we probably never will have to implement a true block-matrix structure for LOBPCG.

src/eigen/lobpcg_hyper_impl.jl
GVigne: I did some more testing, and you are right: it is indeed the same when using dots. I removed the comment.

src/eigen/lobpcg_hyper_impl.jl
GVigne: Yes, 1 works fine now.

src/eigen/lobpcg_hyper_impl.jl
GVigne: I rewrote the documentation, all of this has been summed up when defining the struct.

GVigne:
I've updated the PR following @antoine-levitt's comments. Is there anything you want to add @mfherbst ?

Project.toml
```diff
@@ -6,14 +6,15 @@ version = "0.5.8"
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
 AtomsBase = "a963bdd2-2df7-4f54-a1ee-49d51e6be12a"
-BlockArrays = "8e7c35d0-a365-5155-bbbb-fb81a777f24e"
 Brillouin = "23470ee3-d0df-4052-8b1a-8cbd6363e7f0"
+CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
```
vchuravy, line 10: Would be good if we can avoid this direct dependency

Project.toml
```diff
@@ -6,14 +6,15 @@ version = "0.5.8"
...
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 Conda = "8f4d0f93-b110-5947-807f-2305c1781a2d"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 DftFunctionals = "6bd331d2-b28d-4fd3-880e-1a1c7f37947f"
 FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
+GPUArrays = "0c68f7d7-f131-5f86-a1c3-88cf8149b2d7"
```
vchuravy, line 17: Is `GPUArraysCore` sufficient?

src/workarounds/gpu_arrays.jl
```diff
@@ -0,0 +1,19 @@
...
+import LinearAlgebra.dot, LinearAlgebra.eigen
+using LinearAlgebra
+using GPUArrays
+using CUDA
+
+# https://github.com/JuliaGPU/CUDA.jl/issues/1565
+LinearAlgebra.dot(x::AbstractGPUArray, D::Diagonal,y::AbstractGPUArray) = x'*(D*y)
+
+# https://github.com/JuliaGPU/CUDA.jl/issues/1572
+function LinearAlgebra.eigen(A::Hermitian{T,AT}) where {T <: Complex,AT <: CuArray}
```
vchuravy, line 11: We should open a PR to CUDA.jl to implement this there.

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -43,19 +43,81 @@
...
+LazyHcat is a lightweight subset of BlockArrays.jl's functionalities, but has the
+advantage to be able to store GPU Arrays (BlockArrays is heavily built on Julia's CPU Array).
+"""
+struct LazyHcat{T <: Number, D <: Tuple} <: AbstractMatrix{T}
+    blocks::D
+end
+
+function LazyHcat(arrays::AbstractArray...)
+    @assert length(arrays) != 0
+    n_ref= size(arrays[1], 1)
```
mfherbst, line 65: one space before `=`

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -43,19 +43,81 @@
...
+
+    T = promote_type(map(eltype, arrays)...)
+
+    LazyHcat{T, typeof(arrays)}(arrays)
+end
+
+function Base.size(A::LazyHcat)
+    n = size(A.blocks[1], 1)
+    m = sum(size(block, 2) for block in A.blocks)
+    (n,m)
```
mfherbst, line 76: space after `,`

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -43,19 +43,81 @@
...
+function LinearAlgebra.mul!(res::AbstractMatrix, Ablock::LazyHcat,
+                            B::AbstractVecOrMat, α::Number, β::Number)
+    mul!(res, Ablock*B, I, α, β)
 end
 
 # Perform a Rayleigh-Ritz for the N first eigenvectors.
 @timing function rayleigh_ritz(X, AX, N)
-    XAX = array_mul(X', AX)
-    @assert all(!isnan, XAX)
-    F = eigen(Hermitian(XAX))
```
mfherbst, line 58: Any reason you are removing the NaN check again?

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -360,20 +422,25 @@ end
...
-                e[Xn_indices[i], i] = 1
-            end
+
+            lenXn = length(Xn_indices)
+            # TODO: two allocations needed for zero(similar(...)). Create a zero_like
+            # function which only does one allocation.
+            e = zero(similar(X, size(cX, 1), M - prev_nlocked))
+            lower_diag = one(similar(X, lenXn, lenXn))
+            # e has zeros everywhere except on one of its lower diagonal
+            e[Xn_indices[1] : last(Xn_indices), 1 : lenXn] = lower_diag
```
mfherbst, line 432: We don't put spaces around the `:`.

test/lobpcg.jl
```diff
@@ -122,3 +122,23 @@ end
         @test res1.λ[ik] ≈ res2.λ[ik] atol=1e-6
     end
 end
+
+@testset "LOBPCG Internal data structures" begin
+    a1 = rand(10,5)
```
mfherbst, line 127: spaces after `,`.

mfherbst (commented):
Only some nits with spaces left on my end.

src/eigen/lobpcg_hyper_impl.jl
antoine-levitt: Hm. OK then let's just have our own `zeros_like` function that does this, and later as we figure out the patterns we use we can design our own mini-API (eg something to replace comprehensions). Sucks there's no standard way to do this.

Project.toml
GVigne: Don't we need it because of the functions in `workaround/gpu_arrays.jl`? For the `eigen` workaround for example, we explicitly use CUDA functions, so if I understand correcly, we need to have CUDA as a dependency for DFTK. As soon as the workarounds are implemented in CUDA, we could remove this.

src/eigen/lobpcg_hyper_impl.jl
GVigne: Absolutely no reason: this is a mistake I made when merging the master branch into this one. Sorry!

Project.toml
GVigne: I made a few tests, and it is sufficient. This is the same problem as with the dependency with CUDA: we only really need functionalities from GPUArrays for our workarounds. They shouldn't be in DFTK, but rather merged into the GPU packages: when they'll be implemented, we can remove GPUArrays and CUDA from the dependencies.

src/workarounds/gpu_arrays.jl
```diff
@@ -17,3 +17,11 @@ function LinearAlgebra.eigen(A::Hermitian{T,AT}) where {T <: Real,AT <: CuArray}
...
     (vectors = vects, values = vals)
 end
+
+# Create an array of same type as X filled with zeros, minimizing the number
+# of allocations.
+function zeros_like(X, n, m)
+    Z = similar(X, n, m)
+    Z .= 0
+    Z
+end
```
mfherbst, line 27: Not really GPU-specific. I'd move it to other.

But to be completely honest: I don't think we even need it at all, because
`Z = similar(X, n, m) .= 0`
Does exactly the same thing, is more general (as other forms of similar could be used) and is also just one line.

@antoine-levitt Don't you think that's sufficient for now?

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -71,7 +71,7 @@ function LazyHcat(arrays::AbstractArray...)
 end
 
 function Base.size(A::LazyHcat)
-    n = size(A.blocks[1], 1)
+    n = size(A.blocks[1] , 1)
```
mfherbst, line 74: No space before `,`

src/workarounds/gpu_arrays.jl
GVigne: The reason I put it in the GPU workarounds is that I had to use `zeros(similar())` to distinguish between the cases where `X` is a GPU array and `X` a CPU array. For the CPU version, there is no need for this, as we can directly call `zeros`: but when mixing CPU and GPU, what we really want to do is call Base.zeros, or CUDA.zeros (or AMDGPU.zeros...) depending on the input type of `X`. The only way I found to have this type of behavior is to first use similar.

antoine-levitt: > But to be completely honest: I don't think we even need it at all, because Z = similar(X, n, m) .= 0 does exactly the same thing, is more general (as other forms of similar could be used) and is also just one line.

This is a bit ugly, and also it's possible that this can be optimized in the future (eg for arrays `zeros` does exactly this right now, but could conceivably use `calloc` in future versions of julia), so having a function in src/common makes it easier. You can have this version in src/common, then overload for X::CuArray in the workarounds if you want

mfherbst: ok, let's have a `zeros_like` which works like `similar` but allocates zeros and put it in common. If we want to avoid direct dependency on CUDA (which we should!) we have no other option than avoiding the use of CUDA.zeros. But of course we could still specialise for `<: Array` arguments.

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -45,7 +45,7 @@ vprintln(args...) = nothing
 using LinearAlgebra
 import Base: *
 import Base.size, Base.adjoint, Base.Array
-
+import DFTK:zeros_like
```
mfherbst, line 48: Should not be needed.

src/common/zeros_like.jl
```diff
@@ -0,0 +1,7 @@
+# Create an array of same type as X filled with zeros, minimizing the number
+# of allocations.
+function zeros_like(X::AT, n, m) where AT <: AbstractArray
+    Z = similar(X, n, m)
+    Z .= 0
+    Z
+end
```
mfherbst, line 7: I would make this more general directly:
```suggestion
function zeros_like(X::AbstractArray, T::Type=eltype(X), dims::Integer=size(array)...)
    Z = similar(X, T, dims...)
    Z .= 0
    Z
end
zeros_like(X::AbstractArray, dims::Integer...) = zeros_like(X, eltype(X), dims...)
zeros_like(X::Array, T::Type=eltype(X), dims::Integer=size(array)...) = zeros(T, dims...)
```

GVigne: This doesn't seem to work because of  the default value for `dims`.
I'm a bit confused by the slurping operator. I understand it's purpose, but I don't find the syntax very clear. From what I understand by reading the docs, writing `dims::Integer...` means we want n integers to be merged into a tuple called dims: so implicitly, `dims` is here a Tuple of Integers. Is this correct? If it is, then it would explain why we can't write `dims::Integer=size(X)` as `size(X)` is a Tuple, not an Integer.
What syntax can use to explicitly say that we want `dims` (the tuple) to be `size(X)` (a tuple) by default? Or is there a way to flatten `size(X)`?

mfherbst: > This doesn't seem to work because of the default value for dims.

Indeed, I had some typos in my code, corrected now.

> so implicitly, dims is here a Tuple of Integers. Is this correct?

yes

> Or is there a way to flatten size(X)?

Yes, more `...` ;), see the updated code snippet.

src/eigen/lobpcg_hyper_impl.jl
GVigne: Hum, you are right. But now I'm confused as to why this type of import is required in the tests (such as [here](https://github.com/JuliaMolSim/DFTK.jl/blob/a96f551ed78a47f6cbd5280fedf08f6a7443d31e/test/adaptive_damping.jl#L3)). Is this because the file `lobpcg_hyper_imp` is under `src`, which is also where we define all the functions which be accessed through the DFTK package (in the [DFTK.jl file](https://github.com/JuliaMolSim/DFTK.jl/blob/a96f551ed78a47f6cbd5280fedf08f6a7443d31e/src/DFTK.jl)), whereas the tests are in an other, possibly independent directory?

mfherbst: The tests run the code "from the outside" in a new environment to emulate how DFTK would be used. That's why you need to import not exported functions, but inside DFTK itself you don't need to import module-internal stuff.

mfherbst: The directory is just for us humans. Julia does not care about this.

src/common/zeros_like.jl
GVigne: Oh, ok, makes sense!

src/eigen/lobpcg_hyper_impl.jl
GVigne: Alright, I see now. I think I was expecting some sort of internal imports inside DFTK, but they are not needed. Thanks!

mfherbst:
I will merge this now, even though we don't yet have a way to test the GPU stuff. This will hopefully follow soon.

mfherbst:
Thanks for your efforts @GVigne !

PR #712, opened 2022-08-24
Make some computations in DFTK GPU-compatible

GVigne:
This PR is a followup of [this one](https://github.com/JuliaMolSim/DFTK.jl/pull/711), which implements GPU compatibility for LOBPCG. If you have any questions/remarks as to how LOBPCG works, please refer to this other PR. 

The goal of the following PR is to implement GPU compatibility for some computations made by DFTK. This mainly means modifying the PlaneWaveBasis so it can store GPUArrays, and extending the `apply!` functions to allow the Hamiltonian and its operators to be applied to GPUArrays.

From an end user perspective, the only thing that changes is when he builds the basis. There is now an optional argument `array_type` which tells the code which type of array structure should be used. For example :
   ```
 basis = PlaneWaveBasis(model; Ecut=30, kgrid=(1, 1, 1)) # Computations will happen on CPU
 basis_gpu = PlaneWaveBasis(model; Ecut=30, kgrid=(1, 1, 1), array_type = CuArray) #Computations will happen on GPU using CUDA
```
The end-user can then call the SCF with either `basis` or `basis_gpu`.

I used CUDA since I have an NVIDIA GPU, but this part of the code should also work with other GPUs, since I did not use any CUDA-specific function.
    

Things that I already know could be greatly improved:

-   The preconditionners. For now they work, but two things could be done. The first one is offload `mean_kin` to the GPU. That would require some work, as it means we would have to rewrite `ldiv!` and `mul!` (which right now does a lot of scalar indexing). ~~The other thing would be to build `kin` directly on the GPU instead of building it on CPU then offloading it (which is currently being done). In order to do this, we would need `Gplusk_vectors_cart` to return a GPUArray: this means that `G_vectors(basis, kpoint)` should return a GPUArray, ie that `kpt.G_vectors` should be on GPU. And this is going to be quite hard, as it means that we would have to rewrite every function calling `G_vectors(basis, kpoint)` to be GPU-compatible.~~
-  The solvers. I didn't manage to make the NLSolve solvers work, so I had to use the ones implementend in DFTK. `scf_damping_solver` works fine, but not `scf_anderson_solver` as I didn't manage to write it in a GPU-compatible way.
-  The terms. I have implemented the "easy" terms (Kinetic, AtomicLocal, AtomicNonlocal and Hartree), but we could add the Magnetic, PairWisePotential, Anyonic and XC terms. These terms will be much harder to implement, as they either vastly use scalar indexing or rely on other librairies (libxc).

**Edit:**  Two big changes:
- The G_vectors for each kpoint and the occupation have been offloaded to the GPU. However, to do this, some functions (like `compute_density`) have to bring those arrays back on the CPU, as they do scalar indexing. This could be improved if it is performance-critical.
- The Anderson solver will work in CUDA once [this bug](https://github.com/JuliaGPU/CUDA.jl/pull/1584) has been solved. We will only have to allow scalar indexing on `βs`, which should really be fine as it isn't a big vector.

### CODE REVIEW

**Comment by GVigne** [comment]
This PR is now up to date with master.
For now, here is what's supported
- terms: Kinetic, AtomicLocal, AtomicNonlocal, Ewald, PspCorrection, Entropy and Hartree
- solver: scf_damping_solver and scf_anderson_solver
- mixing: KerkerMixing, SimpleMixing, DielectricMixing, KerkerDosMixing, HybridMixing

I think we have a few things we need to talk about @mfherbst and @antoine-levitt. One of the things I am not very happy with is the way we call the `force_real!` function. If I am not mistaken, it's a bit tricky to have it work on GPU, so everytime we call it, we need first to put the arrays on the CPU, compute `force_real!` then go back on GPU. The good news is that it's not performance critical as it only happens when building objects (for example when building the ```TermHartree```).

**Comment by mfherbst** [comment]
@GVigne Take a look at the failing tests. It seems `TaskLocalRNG` cannot be found and I think you still have a `using CUDA` somewhere in the main code. Finally in `orbitals.jl` I noticed you put a `using` right in the middle of the file. That can be easily overlooked. Best put `using` statements at the top of the file, after some introductory comment in case there is any.

File: src/DFTK.jl [hunk 1]
```diff
@@ -13,6 +13,10 @@ using spglib_jll
 using Unitful
 using UnitfulAtomic
 using ForwardDiff
+using AbstractFFTs
+using GPUArraysCore
+using CUDA
```
mfherbst, line 18: This should not be here

GVigne:
Ok, that's on me: there's definitely a few things I don't understand in the creation of a package in Julia (for now, it's the classic: "why does it work on my computer and not in the CI?").
As for `TaskLocalRNG`, it's more or less the same problem as the discussion with the zeros_like function. We need a function which takes a CPU array or a GPU array and fills it with something random. Currently it's done with `TaskLocalRNG`, as it's been overloaded in the GPUArrays package: this makes it so we don't have to explicitly call `CUDA.randn` (or `AMDGPU.randn`...)

src/DFTK.jl
```diff
@@ -13,9 +13,6 @@ using spglib_jll
 using Unitful
 using UnitfulAtomic
 using ForwardDiff
-using AbstractFFTs
-using GPUArraysCore
```
mfherbst: Not sure about removing these ... sorry I was only talking about the `using CUDA`.

mfherbst:
Re TaskLocalRNG: I don't think it's a problem that you use it. It seems reasonable to do so, but for some reason the tests don't find it. Could be Julia version issue.

GVigne:
Hum, you're right. It seems TaskLocalRng is only available in Julia 1.7 and higher, and not in Julia 1.6.
I guess we can always build the array on the CPU then offload it if needed on the GPU: that will always work, but we're not really using the GPU's features.

mfherbst:
> Hum, you're right. It seems TaskLocalRng is only available in Julia 1.7 and higher, and not in Julia 1.6.
I guess we can always build the array on the CPU then offload it if needed on the GPU: that will always work, but we're not really using the GPU's features.

There is a few ways to deal with this:
1.  We bump our minimal Julia version to 1.7. I would like to avoid that since 1.6 is LTS.
2. We always use the fallback, which is not so elegant, but in this case probably has no real performance implications.
3.  We have a check for the Julia version and do as you suggest as a fallback for <1.7. This can be completely compiled out using an `@static` check, e.g.
  ```julia
  @static if VERSION < v"1.7"
      # do fallback
  end
  ```
I'd favour solution 3.

mfherbst:
Also take a look at the test output of the nighly test. There are some more issues in your tests beyond the RNG stuff.

GVigne:
> 1. We bump our minimal Julia version to 1.7. I would like to avoid that since 1.6 is LTS.

I entirely agree with you. As long as 1.6 is supported, we should try to maintain compatibility. 

> 3\. We have a check for the Julia version and do as you suggest as a fallback for <1.7. This can be completely compiled out using an `@static` check, e.g.

Probably the best way to go: when 1.6 gets deprecated, we will know which part of the code to remove and the reason there were two versions.

src/orbitals.jl
```diff
@@ -53,8 +53,17 @@ function unsafe_unpack_ψ(x, sizes_ψ)
 end
 unpack_ψ(x, sizes_ψ) = deepcopy(unsafe_unpack_ψ(x, sizes_ψ))
 
-function random_orbitals(basis::PlaneWaveBasis{T}, kpt::Kpoint, howmany) where {T}
-    orbitals = similar(basis.G_vectors, Complex{T}, length(G_vectors(basis, kpt)), howmany)
-    randn!(TaskLocalRNG(), orbitals)  # Force the use of GPUArrays.jl's random function if using the GPU
-    ortho_qr(orbitals; array_type = array_type(basis))
+@static if VERSION < v"1.7"
```
mfherbst, line 56: You can actually use this in the middle of a function. This saves you from duplicating 3 lines of code.

mfherbst:
> Probably the best way to go: when 1.6 gets deprecated, we will know which part of the code to remove and the reason there were two versions.

Exactly!

src/orbitals.jl
GVigne: Oh, neat! I underestimated the `@static` macro.

GVigne:
There's still a few bugs lying around. That was bound to happen, as I changed the signature of quite a lot of functions: apparently I wasn't careful enough when making those changes. Hopefully I'll manage to fix them quickly.

mfherbst:
> Hopefully I'll manage to fix them quickly.

You should be able to reproduce these locally. Just `include( )` the test files of the failing tests in a julia shell then you don't have to recompile each time.

mfherbst:
@GVigne You still have an issue in the code saving an `scfres` (and thus a `PlaneWaveBasis`) to disk. Can you take a look?

GVigne:
I haven't forgotten, it's just that I don't understand yet what sort of bug I'm facing. From what I'm understanding, there is a problem with the new PlaneWaveBasis's type when trying to save one with JLD2.: specifically, the `AT` type we're using to store the type of array used for computations. Currently, it's the `save_scfres` function which is failing, and I think it's because of the [`gather_kpts`](https://github.com/JuliaMolSim/DFTK.jl/blob/2a41de91ebbcddb833b211a130042e93d50a6164/src/PlaneWaveBasis.jl#L463-L490) function.

mfherbst:
@GVigne  There are a few things missing, e.g. there is some potential to trust the Julia compiler more to infer types and some cases (e.g. with the Kpoints) you have not pushed through the extra type argument, which means that e.g. in the PlaneWaveBasis the kpoints are not yet stored as a concrete type. I have not yet had the time to fix this. I'll try to point you to these places in a more detailed PR as soon as possible, but perhaps you can take already a first pass.

mfherbst:
Also maybe GT and RT are not the best type names. Essentially one is a vector of real vector type and one an vector of integer vector type. Perhaps there is a way to easily incorporate that.

examples/gpu.jl
```diff
@@ -0,0 +1,28 @@
...
+            Ewald(),
+            PspCorrection(),
+            Entropy(),
+            Hartree()]
+# Now, build a supercell to have a larger system
+pystruct = pymatgen_structure(lattice, atoms, positions)
+pystruct.make_supercell([4,2,2])
+lattice   = load_lattice(pystruct)
+positions = load_positions(pystruct)
+atoms     = fill(Si, length(positions))
```
mfherbst, line 21: We dropped pymatgen support some time ago. Just replace by doing the same thing in ASE (see https://docs.dftk.org/stable/examples/supercells/):
```suggestion
supercell = ase_atoms(lattice, atoms, positions) * (repeat, 1, 1)
lattice   = load_lattice(supercell)
positions = load_positions(supercell)
atoms     = fill(Si, length(positions))
```

src/architecture.jl
```diff
@@ -0,0 +1,13 @@
...
+abstract type AbstractArchitecture end
+
+struct CPU <: AbstractArchitecture end
+
+get_array_type(::CPU) = Array
+
+"""
+Generic, hardware independent architecture for DFTK.
+"""
+abstract type GPU <: AbstractArchitecture end
```
mfherbst, line 13: @vchuravy Are you aware of a more wide-spread solution for this type of "device / computing architecture selection". It seems currently everyone is rolling their own ... including naming conflicts for `GPU`, `CPU` etc. I would like to avoid this, but this leads to a pretty powerful and extensible interface to specify hardware setup.

src/terms/nonlocal.jl
```diff
@@ -63,8 +63,8 @@ end
         C = build_projection_coefficients_(T, element.psp)
         for (ik, kpt) in enumerate(basis.kpoints)
             # we compute the forces from the irreductible BZ; they are symmetrized later
-            qs_cart = Gplusk_vectors_cart(basis, kpt)
             qs = Gplusk_vectors(basis, kpt)
+            qs_cart = Array(Gplusk_vectors_cart(basis, kpt))  # Get on the CPU
```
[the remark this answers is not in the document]
antoine-levitt: and `to_gpu`

antoine-levitt (commented):
Very nice!

src/architecture.jl
vchuravy: Yeah, I have some of that in KA, https://github.com/JuliaGPU/KernelAbstractions.jl/pull/320 and there is https://docs.juliahub.com/GPUArraysCore/qiYUe/0.1.2/autodocs/#GPUArraysCore.backend-Tuple{Type}

@oschulz was working on something https://github.com/JuliaGPU/KernelAbstractions.jl/pull/297

cc: @maleadt @jpsamaroo We do need some sort of "Query compute backend for <:AbstractArray" and "allocate array for backend (similar is not always enough)"

src/PlaneWaveBasis.jl
```diff
@@ -15,16 +15,16 @@ Discretization information for ``k``-point-dependent quantities such as orbitals
 More generally, a ``k``-point is a block of the Hamiltonian;
 eg collinear spin is treated by doubling the number of kpoints.
 """
-struct Kpoint{T <: Real}
+struct Kpoint{T <: Real, GT <: AbstractArray}
     spin::Int                     # Spin component can be 1 or 2 as index into what is
-                                  # returned by the `spin_components` function
+    #                             # returned by the `spin_components` function
```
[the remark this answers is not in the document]
mfherbst: I made it :smile:.

src/Model.jl
```diff
@@ -300,7 +312,14 @@ s_cart = L s_red = L A_red r_red = L A_red L⁻¹ r_cart, thus A_cart = L A_red
...
-matrix_red_to_cart(model::Model, Ared)    = model.lattice * Ared * model.inv_lattice
-matrix_cart_to_red(model::Model, Acart)   = model.inv_lattice * Acart * model.lattice
-comatrix_red_to_cart(model::Model, Bred)  = model.inv_lattice' * Bred * model.lattice'
-comatrix_cart_to_red(model::Model, Bcart) = model.lattice' * Bcart * model.inv_lattice'
+_closure_matmatmul(M, Minv) = mat -> M * mat * Minv
+
+matrix_red_to_cart(model::Model)   = _closure_matmatmul(model.lattice,      model.inv_lattice)
+matrix_cart_to_red(model::Model)   = _closure_matmatmul(model.inv_lattice,  model.lattice)
+comatrix_red_to_cart(model::Model) = _closure_matmatmul(model.inv_lattice', model.lattice')
+comatrix_cart_to_red(model::Model) = _closure_matmatmul(model.lattice',     model.inv_lattice')
```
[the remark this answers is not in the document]
mfherbst: No does not work because of avoiding closure on the model.

src/PlaneWaveBasis.jl
```diff
@@ -322,23 +337,26 @@ Creates a new basis identical to `basis`, but with a custom set of kpoints
     PlaneWaveBasis(basis.model, basis.Ecut,
                    basis.fft_size, basis.variational,
                    kcoords, kweights, kgrid, kshift,
-                   basis.symmetries_respect_rgrid, basis.comm_kpts)
+                   basis.symmetries_respect_rgrid, basis.comm_kpts, basis.G_vectors)
 end
 
 """
-    G_vectors(fft_size::Tuple)
+    G_vectors([array_type=Vector], fft_size::Tuple)
```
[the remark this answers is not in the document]
mfherbst: I explicitly suggested to have this be the first because usually it's done like this in julia (see similar, ones, zeros etc).

src/fft.jl
```diff
@@ -257,23 +257,30 @@ function compute_Glims_fast(lattice::AbstractMatrix{T}, Ecut; supersampling=2, t
...
+    ipFFT, opFFT, inv(ipFFT).p, inv(opFFT).p
+end
+function build_fft_plans!(tmp::Array{Complex{Float32}})
+    # For Float32 there are issues with aligned FFTW plans, so we
+    # fall back to unaligned FFTW plans (which are generally discouraged).
+    ipFFT = FFTW.plan_fft!(tmp; flags=FFTW.MEASURE | FFTW.UNALIGNED)
+    opFFT = FFTW.plan_fft(tmp;  flags=FFTW.MEASURE | FFTW.UNALIGNED)
+    # backwards-FFT by inverting and stripping off normalizations
+    ipFFT, opFFT, inv(ipFFT).p, inv(opFFT).p
+end
```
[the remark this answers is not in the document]
mfherbst: Because it's not needed ... just creates another layer of indirection.

src/guess_density.jl
```diff
@@ -66,7 +66,7 @@ function _guess_spin_density(basis::PlaneWaveBasis{T}, atoms, positions, magneti
         @warn("Returning zero spin density guess, because no initial magnetization has " *
               "been specified in any of the given elements / atoms. Your SCF will likely " *
               "not converge to a spin-broken solution.")
-        return zeros(T, basis.fft_size)
+        return zeros_like(basis.G_vectors, T, basis.fft_size...)
```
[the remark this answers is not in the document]
mfherbst: Not sure ... not super clear which size that should be. But something to keep in mind ... these whole array functions right now become a bit of a zoo. We have to be a bit careful to not overdo it

src/orbitals.jl
```diff
@@ -52,5 +54,12 @@ end
 unpack_ψ(x, sizes_ψ) = deepcopy(unsafe_unpack_ψ(x, sizes_ψ))
 
 function random_orbitals(basis::PlaneWaveBasis{T}, kpt::Kpoint, howmany) where {T}
-    ortho_qr(randn(Complex{T}, length(G_vectors(basis, kpt)), howmany))
+    @static if VERSION < v"1.7"  # TaskLocalRNG not yet available.
+        orbitals = randn(Complex{T}, length(G_vectors(basis, kpt)), howmany)
+        orbitals = convert_like(basis.G_vectors, orbitals)
+    else
+        orbitals = similar(basis.G_vectors, Complex{T}, length(G_vectors(basis, kpt)), howmany)
+        randn!(TaskLocalRNG(), orbitals)  # use the RNG on the device if we're using a GPU
```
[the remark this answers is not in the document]
mfherbst: This `_like` stuff is used nowhere in julia. I'm not sure we should buy into it so much.

src/scf/mixing.jl
```diff
@@ -49,8 +49,8 @@ end
 
 @timing "KerkerMixing" function mix_density(mixing::KerkerMixing, basis::PlaneWaveBasis,
                                             δF; kwargs...)
-    T      = eltype(δF)
-    G²     = [sum(abs2, G) for G in G_vectors_cart(basis)]
+    T  = eltype(δF)
+    G² = map(G -> sum(abs2, G), G_vectors_cart(basis))
```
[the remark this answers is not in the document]
mfherbst: Yeah, why not

src/symmetry.jl
```diff
@@ -198,6 +198,15 @@ function lowpass_for_symmetry!(ρ, basis; symmetries=basis.symmetries)
     end
     ρ
 end
+function lowpass_for_symmetry!(ρ::AT, basis;
+                               symmetries=basis.symmetries) where {AT <: AbstractGPUArray}
+    all(isone, symmetries) && return ρ
+    # lowpass_for_symmetry! currently uses scalar indexing, so we have to do this very ugly
+    # thing for cases where ρ sits on a device (e.g. GPU)
+    ρ_CPU = Array(ρ)
+    ρ_CPU = lowpass_for_symmetry!(ρ_CPU, basis; symmetries)
```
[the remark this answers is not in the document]
mfherbst: yes my bad

src/terms/hartree.jl
```diff
@@ -29,8 +29,13 @@ end
...
     # is multiplying elementwise by 4π / |G|^2.
-    poisson_green_coeffs = 4T(π) ./ [sum(abs2, G) for G in G_vectors_cart(basis)]
-    poisson_green_coeffs[1] = 0  # Compensating charge background => Zero DC
+    poisson_green_coeffs = map(G_vectors_cart(basis)) do G
+        if iszero(G)
+            zero(T)  # Compensating charge background => zero DC
+        else
+            4T(π) / sum(abs2, G)
+        end
+    end
```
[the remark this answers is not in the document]
mfherbst: yes there is. You can temporary disable the warning for scalar indexing with a `do ... end` style syntax

src/terms/nonlocal.jl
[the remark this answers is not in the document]
mfherbst: > maybe on all these comments move them to something like # GPU: move to the CPU so that it's clear these are GPU code paths.

Hmm. It's not just for GPUs (well for now it is ... but this comment would become outdated if we implement other architectures). That's why I removed it.

We really should buy into one of the helper infrastructures Valentin linked above for this `to_cpu` / `to_gpu` stuff before building yet our own.

src/densities.jl
```diff
@@ -23,20 +23,23 @@ using an optional `occupation_threshold`. By default all occupation numbers are
...
+    ρ_chunklocal = [zeros_like(basis.G_vectors, S, basis.fft_size..., basis.model.n_spin_components)
                                for _ = 1:Threads.nthreads()]
-    ψnk_real_chunklocal = Array{complex(S),3}[zeros(complex(S), basis.fft_size)
-                                               for _ = 1:Threads.nthreads()]
+    ψnk_real_chunklocal = [zeros_like(basis.G_vectors, complex(S), basis.fft_size...)
+                                for _ = 1:Threads.nthreads()]
 
     @sync for (ichunk, chunk) in enumerate(Iterators.partition(ik_n, chunk_length))
         Threads.@spawn for (ik, n) in chunk  # spawn a task per chunk
+            kpt = basis.kpoints[ik]
```
mfherbst, line 42: duplication

src/occupation.jl
```diff
@@ -29,8 +29,9 @@ function compute_occupation(basis::PlaneWaveBasis{T}, eigenvalues, εF;
     inverse_temperature = iszero(temperature) ? T(Inf) : 1/temperature
 
     filled_occ = filled_occupation(basis.model)
-    [filled_occ * Smearing.occupation.(smearing, (εk .- εF) .* inverse_temperature)
-     for εk in eigenvalues]
+    [to_device(basis.architecture,
+        filled_occ * Smearing.occupation.(smearing, (εk .- εF) .* inverse_temperature))
+        for εk in eigenvalues]
```
mfherbst, line 34: I think a `map` makes this clearer

src/scf/mixing.jl
```diff
@@ -49,8 +49,8 @@ end
 
 @timing "KerkerMixing" function mix_density(mixing::KerkerMixing, basis::PlaneWaveBasis,
                                             δF; kwargs...)
-    T      = eltype(δF)
-    G²     = [sum(abs2, G) for G in G_vectors_cart(basis)]
+    T  = eltype(δF)
+    G² = map(G -> norm2(G), G_vectors_cart(basis))
```
mfherbst, line 53: just `map(norm2, G_vectors_cart(basis))` and spacing

src/terms/hartree.jl
```diff
@@ -1,3 +1,5 @@
+using GPUArraysCore
```
mfherbst, line 1: I would move this to the general DFTK.jl

My rule of thumb for the `using`s is:
- Is it a package that I only use in this (or a few selected) files or where the functionality of the package is closely coupled to one file in DFTK, then I put the `using` into that file.
- If it's more "global" and affects everyone (like the `allowscalar`) that the package should be in the `DFTK.jl` file.

We don't follow this rule thoroughly, though.

src/eigen/preconditioners.jl
```diff
@@ -40,8 +40,10 @@ function PreconditionerTPA(basis::PlaneWaveBasis{T}, kpt::Kpoint; default_shift=
     kinetic_term = only(kinetic_term)
     scaling = kinetic_term.scaling_factor
     blowup = kinetic_term.blowup     # blowup for energy cut-off smearing
-    kin = Vector{T}([scaling * sum(abs2, q)/2 * blowup(norm(q), basis.Ecut)
-                     for q in Gplusk_vectors_cart(basis, kpt)])
+    Ecut = basis.Ecut
+    kin = map(Gplusk_vectors_cart(basis, kpt)) do q
+        scaling * norm2(q) /2 * blowup(norm(q), Ecut)
+    end
```
mfherbst, line 46: use `kinetic_energy` function.

src/terms/kinetic.jl
```diff
@@ -19,18 +19,27 @@ struct TermKinetic <: Term
...
-    kinetic_energies = [[T(scaling_factor) * sum(abs2, Gk)/2 * blowup(norm(Gk), basis.Ecut)
-                         for Gk in Gplusk_vectors_cart(basis, kpt)] for kpt in basis.kpoints]
+
+    kinetic_energies = [kinetic_energy(Gplusk_vectors_cart(basis, kpt), scaling_factor,
+                                        blowup, basis.Ecut, T)
+                        for kpt in basis.kpoints]
     TermKinetic(T(scaling_factor), kinetic_energies)
 end
 
+function kinetic_energy(q, scaling_factor, blowup, Ecut, ::Type{T}) where {T}
```
mfherbst, line 29: Get rid of the last argument and infer from the eltype of q, also I would reorder the arguments to
`(blowup, scaling_factor, Ecut, q)`. That allows you to also have a version `kinetic_energy(kin::Kinetic, Ecut, q)`, which simplifies the preconditioner.

src/terms/local.jl
```diff
@@ -100,7 +104,9 @@ function (::AtomicLocal)(basis::PlaneWaveBasis{T}) where {T}
     end
     enforce_real!(basis, pot_fourier)  # Symmetrize Fourier coeffs to have real iFFT
 
-    pot_real = irfft(basis, pot_fourier)
+    # Offload potential values to a device (like a GPU) and do the FFT
```
mfherbst, line 107: I think this comment is now obvious.

src/terms/nonlocal.jl
```diff
@@ -108,7 +108,7 @@ function build_projection_coefficients_(T, psps, psp_positions)
     end  # psp, r
     @assert count == n_proj
 
-    proj_coeffs
+    to_device(architecture, proj_coeffs)
```
mfherbst, line 111: I would move this to the calling side (and have the function always return on the CPU).

src/scf/nbands_algorithm.jl
```diff
@@ -68,6 +68,8 @@ function determine_n_bands(bands::AdaptiveBands, occupation::AbstractVector,
     # TODO Could return different bands per k-Points
 
     # Determine number of bands to be actually converged
+    # Bring occupation on the CPU, or findlast will fail
+    occupation = [to_cpu(occk) for occk in occupation]
```
mfherbst, line 72: When this is merged, can you open an issue @GVigne to indicate that we should reduce the transfer to CPU of the occupation array.

src/PlaneWaveBasis.jl
```diff
@@ -124,34 +131,36 @@ Base.eltype(::PlaneWaveBasis{T}) where {T} = T
...
+                  Kpoint{T,typeof(Gvecs_k)}(iσ, k, mapping, mapping_inv, Gvecs_k))
         end
     end
-
     vcat(kpoints_per_spin...)  # put all spin up first, then all spin down
 end
 function build_kpoints(basis::PlaneWaveBasis, kcoords)
     build_kpoints(basis.model, basis.fft_size, kcoords, basis.Ecut;
-                  variational=basis.variational)
+                  variational=basis.variational, architecture = basis.architecture)
```
mfherbst, line 151: No need for the `architecture = `

src/PlaneWaveBasis.jl
```diff
@@ -113,7 +119,8 @@ Base.Broadcast.broadcastable(basis::PlaneWaveBasis) = Ref(basis)
 Base.eltype(::PlaneWaveBasis{T}) where {T} = T
 
 @timing function build_kpoints(model::Model{T}, fft_size, kcoords, Ecut;
-                               variational=true) where {T}
+                               variational=true,
+                               architecture::AbstractArchitecture=CPU()) where {T}
```
mfherbst, line 123: I think I would make the `architecture` a non-optional kwarg (i.e. remove the `=CPU()`).

src/PlaneWaveBasis.jl
```diff
@@ -245,21 +256,27 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Number, fft_size, variational,
     @assert mpi_sum(sum(kweights_thisproc), comm_kpts) ≈ model.n_spin_components
     @assert length(kpoints) == length(kweights_thisproc)
 
+    if Gs isa AbstractGPUArray && Threads.nthreads() > 1
```
mfherbst, line 259: if `architecture isa GPU && ...`

src/PlaneWaveBasis.jl
```diff
@@ -322,23 +340,26 @@ Creates a new basis identical to `basis`, but with a custom set of kpoints
...
+G_vectors(fft_size::Union{Tuple,AbstractVector}) = G_vectors(CPU(), fft_size)
+function G_vectors(architecture::AbstractArchitecture, fft_size::Union{Tuple,AbstractVector})
     # Note that a collect(G_vectors_generator(fft_size)) is 100-fold slower
     # than this implementation, hence the code duplication.
     start = .- cld.(fft_size .- 1, 2)
     stop  = fld.(fft_size .- 1, 2)
     axes  = [[collect(0:stop[i]); collect(start[i]:-1)] for i in 1:3]
-    [Vec3{Int}(i, j, k) for i in axes[1], j in axes[2], k in axes[3]]
+    Gs = [Vec3{Int}(i, j, k) for i in axes[1], j in axes[2], k in axes[3]]
+    to_device(architecture, Gs)
```
mfherbst, line 360: Again unless you see any major issues with this I would make the calling party responsable to do the transfer.

mfherbst (commented):
Small nits

src/PlaneWaveBasis.jl
```diff
@@ -15,16 +15,16 @@ Discretization information for ``k``-point-dependent quantities such as orbitals
 More generally, a ``k``-point is a block of the Hamiltonian;
 eg collinear spin is treated by doubling the number of kpoints.
 """
-struct Kpoint{T <: Real}
+struct Kpoint{T <: Real, GT <: AbstractArray}
```
antoine-levitt, line 18: can't this be typed as AbstractArray{Vec3{Int}}?

src/PlaneWaveBasis.jl
```diff
@@ -39,7 +39,10 @@ Normalization conventions:
 
 `ifft` and `fft` convert between these representations.
 """
-struct PlaneWaveBasis{T, VT} <: AbstractBasis{T} where {VT <: Real}
+struct PlaneWaveBasis{T, VT, T_G_vectors, T_r_vectors, T_kpt_G_vecs} <: AbstractBasis{
+    T
+} where {VT <: Real, T_G_vectors <: AbstractArray{Vec3{Int}}, T_r_vectors <: AbstractArray,
```
antoine-levitt, line 44: split lines

antoine-levitt: can't T_r_vectors be typed more concretely?

antoine-levitt: btw you can also add the `3` in there, can't you?

src/PlaneWaveBasis.jl
antoine-levitt: AbstractVector, even

src/PlaneWaveBasis.jl
```diff
@@ -124,34 +131,36 @@ Base.eltype(::PlaneWaveBasis{T}) where {T} = T
...
                 push!(Gvecs_k, G)
             end
         end
+        Gvecs_k = to_device(architecture, Gvecs_k)
+
         mapping_inv = Dict(ifull => iball for (iball, ifull) in enumerate(mapping))
         for iσ = 1:model.n_spin_components
             push!(kpoints_per_spin[iσ],
-                  Kpoint(iσ, k, mapping, mapping_inv, Gvecs_k))
+                  Kpoint{T,typeof(Gvecs_k)}(iσ, k, mapping, mapping_inv, Gvecs_k))
```
antoine-levitt, line 144: do you really need this? Usually it just gets inferred

src/PlaneWaveBasis.jl
```diff
@@ -376,7 +399,8 @@ end
 The list of ``G + k`` vectors, in reduced coordinates.
 """
 function Gplusk_vectors(basis::PlaneWaveBasis, kpt::Kpoint)
-    map(G -> G + kpt.coordinate, G_vectors(basis, kpt))
+    coordinate = kpt.coordinate  # Avoid closure on kpt (not isbits)
```
antoine-levitt, line 402: prefix with GPU: someone reading this has no idea why not being isbits is a problem

src/PlaneWaveBasis.jl
```diff
@@ -426,6 +450,10 @@ Returns nothing if outside the range of valid wave vectors.
     end
 end
 
+@inline function index_G_vectors(basis::PlaneWaveBasis, G::AbstractVector{<:Integer})
```
antoine-levitt, line 453: is there really a need for an explicit inline?

src/terms/hartree.jl
```diff
@@ -29,9 +29,13 @@ end
 function TermHartree(basis::PlaneWaveBasis{T}, scaling_factor) where {T}
     # Solving the Poisson equation ΔV = -4π ρ in Fourier space
     # is multiplying elementwise by 4π / |G|^2.
-    poisson_green_coeffs = 4T(π) ./ [sum(abs2, G) for G in G_vectors_cart(basis)]
-    poisson_green_coeffs[1] = 0  # Compensating charge background => Zero DC
+
+    GPUArraysCore.allowscalar() do
+        poisson_green_coeffs = 4T(π) ./ [norm2(G) for G in G_vectors_cart(basis)]
```
antoine-levitt, line 34: norm2.()

src/terms/hartree.jl
```diff
@@ -29,9 +29,13 @@ end
...
+        poisson_green_coeffs[1] = 0  # Compensating charge background => Zero DC
```
antoine-levitt, line 35: only enclose this one in allowscalar? also is there no macro that avoids the explicit closure?

src/terms/xc.jl
```diff
@@ -97,7 +97,7 @@ end
         end
         if haskey(terms, :Vl) && any(x -> abs(x) > term.potential_threshold, terms.Vl)
             @warn "Meta-GGAs with a Δρ term have not yet been thoroughly tested." maxlog=1
-            mG² = [-sum(abs2, G) for G in G_vectors_cart(basis)]
+            mG² = [-norm2(G) for G in G_vectors_cart(basis)]
```
antoine-levitt, line 100: norm2.()

src/terms/xc.jl
```diff
@@ -255,7 +255,7 @@ function LibxcDensities(basis, max_derivative::Integer, ρ, τ)
     # Compute Δρ
     if max_derivative > 1
         Δρ_real = similar(ρ_real, n_spin, basis.fft_size...)
-        mG² = [-sum(abs2, G) for G in G_vectors_cart(basis)]
+        mG² = [-norm2(G) for G in G_vectors_cart(basis)]
```
antoine-levitt, line 258: norm2.()

antoine-levitt, line 258: norm2.()

src/PlaneWaveBasis.jl
mfherbst: Rather accelerator (to be more general)

src/scf/mixing.jl
```diff
@@ -137,7 +136,7 @@ end
     εr > 1 / sqrt(eps(T)) && return mix_density(KerkerMixing(; kTF), basis, δF)
 
     C0 = 1 - εr
-    Gsq = [sum(abs2, G) for G in G_vectors_cart(basis)]
+    Gsq = map(G -> norm2(G), G_vectors_cart(basis))
```
mfherbst, line 139: Also here `map(norm2, G_vectors_cart(basis)`

mfherbst: or even shorter `norm2.()` actually.

src/PlaneWaveBasis.jl
antoine-levitt: You know how I feel about being general before we actually know what we're generalizing to, but OK :-p

src/PlaneWaveBasis.jl
```diff
@@ -39,10 +39,9 @@ Normalization conventions:
 
 `ifft` and `fft` convert between these representations.
 """
-struct PlaneWaveBasis{T, VT, T_G_vectors, T_r_vectors, T_kpt_G_vecs} <: AbstractBasis{
-    T
-} where {VT <: Real, T_G_vectors <: AbstractArray{Vec3{Int}}, T_r_vectors <: AbstractArray,
-         T_kpt_G_vecs <: AbstractVector{Vec3{Int}}}
+struct PlaneWaveBasis{T, VT, T_G_vectors, T_r_vectors, T_kpt_G_vecs} <: AbstractBasis{T} where
+    {VT <: Real, T_G_vectors <: AbstractArray{Vec3{Int}},
+    T_r_vectors <: AbstractArray{Vec3}, T_kpt_G_vecs <: AbstractVector{Vec3{Int}}}
```
antoine-levitt, line 44: Can't this be AbstractArray{Vec3{T}, 3}?
Sorry for the bother, but it does help readers to know what is what
(also skip lines)

mfherbst: Getting the `T` in is tricky, but we can try the 3

mfherbst: no this seems to work

vchuravy:
Congrats! This is great :)

GVigne:
Thanks a lot!

Issue #530, opened 2021-10-07
Possibility of using hybrid functionals


exenGT:
Is it possible to use hybrid functionals like PBE0 and HSE in libxc?

Thanks,
Jingyang

antoine-levitt:
Ah, I thought we already had an issue for this. Not yet possible but the infrastructure is there so it shouldn't be hard to implement; it just needs somebody interested to actually do it.

PR #717, opened 2022-09-03
[WIP] Hybrid functionals

BKaperick:
Started work at the DFTK Summer School with @RashidRafeek and @ChRickert in response to issue #530.

- [ x]  Initialize fock exchange term and operator
- [ x] Compute energy for 1 kpoint
- [ x] Compute Hamiltonian
- [ ] Extend energy code to work for arbitrary number of kpoints
- [ ] Extend LibXc and DftFunctionals interface to allow hybrid families
- [ ] Add standard models like PBE0 and add test case


antoine-levitt:
Looks super, great work!
> Extend energy code to work for arbitrary number of kpoints

Should be OK to do: when i = (n,k) and j = (n',k'), rho_ij = psi_i^* psi_j is exp(i(k'-k)x) u_nk*(x) u_mk(x), so V_ij can be computed in Fourier space just as you do now but with a |k'-k+G|^2 instead of G^2. That should possibly be the only change needed, except for normalizations maybe.

For testing, we should be able to compare hartree-fock against eg abinit.

additional TODOs (not necessarily in the scope of this PR, I'm just adding for future reference) is to generalize SCF to mix orbitals, and to implement the ACE method for performance.

antoine-levitt:
also as a TODO: generalize to fractional occupations

src/DispatchFunctional.jl
```diff
@@ -12,8 +12,8 @@ function LibxcFunctional(identifier::Symbol)
     @assert fun.kind   in (:exchange, :correlation, :exchange_correlation)
     kind = Dict(:exchange => :x, :correlation => :c, :exchange_correlation => :xc)[fun.kind]
 
-    @assert fun.family in (:lda, :gga, :mgga)  # Hybrids not supported yet.
-    if fun.family == :mgga && Libxc.needs_laplacian(fun)
+    @assert fun.family in (:lda, :hyb_lda, :gga, :hyb_gga, :mgga, :hyb_mgga)
+    if (fun.family == :mgga || fun.family == :hyb_mgga) && Libxc.needs_laplacian(fun)
         family = :mggal
```
mfherbst, line 17: Or `:hyb_mggal` actually.

src/DispatchFunctional.jl
```diff
@@ -54,7 +54,7 @@ end
 
 libxc_energy(terms, ρ) = haskey(terms, :zk) ? reshape(terms.zk, 1, size(ρ, 2)) .* ρ : false
 
-function DftFunctionals.potential_terms(func::LibxcFunctional{:lda}, ρ::Matrix{Float64})
+function DftFunctionals.potential_terms(func::Union{LibxcFunctional{:lda},LibxcFunctional{:hyb_lda}}, ρ::Matrix{Float64})
```
mfherbst, line 57: Line too long (same as many others ... we try to limit it to something like 92).

src/DispatchFunctional.jl
```diff
@@ -129,7 +129,6 @@ function DftFunctionals.kernel_terms(func::LibxcFunctional{:gga}, ρ::Matrix{Flo
     Vσσ = libxc_unfold_spin(terms.v2sigma2, s_σ)
     (; e, Vρ, Vσ, Vρρ, Vρσ, Vσσ)
 end
-
```
mfherbst, line 132: Why?

src/standard_models.jl
```diff
@@ -72,3 +72,16 @@ function model_SCAN(lattice::AbstractMatrix, atoms::Vector{<:Element},
                     positions::Vector{<:AbstractVector}; kwargs...)
     model_DFT(lattice, atoms, positions, [:mgga_x_scan, :mgga_c_scan]; kwargs...)
 end
+
+"""
+Build an PBE0 model from the specified atoms.
+DOI:10.1103/PhysRevLett.77.3865
+https://www.vasp.at/tutorials/latest/hybrids/part1/
+"""
+function model_PBE0(lattice::AbstractMatrix, atoms::Vector{<:Element},
```
mfherbst, line 81: Also needs to be added in https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/external/atomsbase.jl

examples/silicon.jl
```diff
@@ -9,6 +9,6 @@ Si = ElementPsp(:Si, psp=load_psp("hgh/lda/Si-q4"))
 atoms     = [Si, Si]
 positions = [ones(3)/8, -ones(3)/8]
 
-model = model_LDA(lattice, atoms, positions)
+model = DFTK.model_PBE0(lattice, atoms, positions)
```
mfherbst, line 12: I would not commit that here. Better to make a new example (e.g. one where you restrict to the exact case that can be covered by the code.

src/standard_models.jl
mfherbst: Maybe also add a Hartree-Fock model?

src/terms/terms.jl
```diff
@@ -50,6 +50,7 @@ include("ewald.jl")
 include("psp_correction.jl")
 include("entropy.jl")
 include("pairwise.jl")
+include("fock_exchange.jl")
```
mfherbst, line 53: I'd call it `exact_exchange.jl`.

src/terms/xc.jl
```diff
@@ -49,7 +49,7 @@ end
 
     model    = basis.model
     n_spin   = model.n_spin_components
-    @assert all(family(xc) in (:lda, :gga, :mgga, :mggal) for xc in term.functionals)
+    @assert all(family(xc) in (:lda, :hyb_lda, :gga, :hyb_gga, :mgga, :hyb_mgga, :mggal) for xc in term.functionals)
```
mfherbst, line 52: :hyb_mggal missing

src/terms/xc.jl
```diff
@@ -187,10 +187,13 @@ for derivatives wrt. σ_αβ or σ_βα.
...
-    family(functional) == :gga   && return 1
-    family(functional) == :mgga  && return 1
-    family(functional) == :mggal && return 2
+    family(functional) == :lda      && return 0
+    family(functional) == :hyb_lda      && return 0
+    family(functional) == :gga      && return 1
+    family(functional) == :hyb_gga  && return 1
+    family(functional) == :mgga     && return 1
+    family(functional) == :hyb_mgga     && return 1
+    family(functional) == :mggal    && return 2
```
mfherbst, line 196: alignment of spacing

src/terms/operators.jl
```diff
@@ -163,3 +163,24 @@ function optimize_operators_(ops)
...
+         integral_fourier = op.poisson_green_coeffs_kpt .* ψ_n .* ψ.fourier
+         integral_real = G_to_r(op.basis, op.kpoint, integral_fourier)
+         psi_n_real = G_to_r(op.basis, op.kpoint, ψ_n)
+         Hψ.real .-= integral_real .* psi_n_real
+    end
+end
+
+function Matrix(op::ExchangeOperator)
+    # TODO - implement this for testing/debugging purposes
+end
```
mfherbst, line 186: If you don't want to implement it, don't define this function (see the divAgrad operator).

src/standard_models.jl
```diff
@@ -72,3 +72,16 @@ function model_SCAN(lattice::AbstractMatrix, atoms::Vector{<:Element},
...
+                   positions::Vector{<:AbstractVector}; kwargs...)
+    functional = DispatchFunctional(:hyb_gga_xc_pbeh)
+    libxcfun = Libxc.Functional(:hyb_gga_xc_pbeh)
+    scaling_factor = getproperty(libxcfun, :exx_coefficient)
+    model_DFT(lattice, atoms, positions, Xc([functional]), extra_terms=[FockExchange(; scaling_factor = scaling_factor)]; kwargs...)
```
mfherbst, line 86: This looks very complicated. I think in this case I would just hard-code the scaling factor (which for PBE0 is always 0.25).

src/terms/fock_exchange.jl
```diff
@@ -0,0 +1,60 @@
...
+
+@timing "ene_ops: FockExchange" function ene_ops(term::TermFockExchange, basis::PlaneWaveBasis{T},
+                                            ψ, occ; ρ, kwargs...) where {T}
+    ops = [NoopOperator(basis, kpoint)
+    for (ik, kpoint) in enumerate(basis.kpoints)]
+    isnothing(ψ) && return (E=T(0), ops=ops)
+
+    ψ, occ = select_occupied_orbitals(basis, ψ, occ; threshold=0.1)
+
+    # @assert length(ψ) == 1 # TODO: make it work for more kpoints
```
mfherbst, line 43: Good idea. I would indeed put all the assertions in here (i.e. also temperature).

src/terms/fock_exchange.jl
```diff
@@ -0,0 +1,60 @@
...
+    poisson_green_coeffs[1] = 0
+    poisson_green_coeffs_kpt[1] = 0
+
+    TermFockExchange(T(scaling_factor), poisson_green_coeffs, poisson_green_coeffs_kpt)
+end
+
+@timing "ene_ops: FockExchange" function ene_ops(term::TermFockExchange, basis::PlaneWaveBasis{T},
+                                            ψ, occ; ρ, kwargs...) where {T}
+    ops = [NoopOperator(basis, kpoint)
+    for (ik, kpoint) in enumerate(basis.kpoints)]
```
mfherbst, line 38: formatting is off.

src/terms/fock_exchange.jl
```diff
@@ -0,0 +1,60 @@
...
+    E = T(0)
+    for psi_i in eachcol(ψ[1])
+        for psi_j in eachcol(ψ[1])
+            rho_ij_real = conj(G_to_r(basis, basis.kpoints[1], psi_i)) .* G_to_r(basis, basis.kpoints[1], psi_j)
+
+            rho_ij_real_conj = conj(rho_ij_real)
+            rho_ij_four = r_to_G(basis, rho_ij_real)
+            rho_ij_four_conj = r_to_G(basis, rho_ij_real_conj)
+
+            vij_fourier = rho_ij_four_conj .* term.poisson_green_coeffs
```
mfherbst, line 54: Deserves a comment what you are doing here (solving poisson equation to obtain potential corresponding to `ϕ_i(r')ϕ_j(r) / |r - r'|`).

src/terms/fock_exchange.jl
```diff
@@ -0,0 +1,60 @@
...
+    # @assert length(ψ) == 1 # TODO: make it work for more kpoints
+
+    E = T(0)
+    for psi_i in eachcol(ψ[1])
+        for psi_j in eachcol(ψ[1])
+            rho_ij_real = conj(G_to_r(basis, basis.kpoints[1], psi_i)) .* G_to_r(basis, basis.kpoints[1], psi_j)
+
+            rho_ij_real_conj = conj(rho_ij_real)
+            rho_ij_four = r_to_G(basis, rho_ij_real)
+            rho_ij_four_conj = r_to_G(basis, rho_ij_real_conj)
```
mfherbst, line 52: It should be possible to save an FFT here (e.g. by using the correspondence of Fourier coefficients for complex conjugates or by reformulating the `dot` below)

src/terms/fock_exchange.jl
```diff
@@ -0,0 +1,60 @@
+"""
+Exact exchange term: the Hartree-Fock exchange energy of the orbitals
+
+
+-1/2 ∑ ∫∫ ϕ_i^*(r)ϕ_j^*(r')ϕ_i(r')ϕ_j(r) / |r - r'| dr dr'
+
+"""
+struct FockExchange
```
mfherbst, line 8: I'd call it `ExactExchange`

mfherbst (commented):
Thanks for this contribution! A very solid start.

I would in fact vote for rather not going all the way and merging this into DFTK asap (i.e. maybe only with the version for one k-Point and without temperature as it is implemented right now).

src/DispatchFunctional.jl
mfherbst: Actually see my comments in https://github.com/JuliaMolSim/DftFunctionals.jl/pull/8/.

src/terms/xc.jl
mfherbst: See my comments in https://github.com/JuliaMolSim/DftFunctionals.jl/pull/8, maybe remove the `:hyb` here alltogether.

src/terms/fock_exchange.jl
BKaperick: I'm not sure the clean way to do this.  For now I just rearranged indices to avoid the fft but it's messy.  What do you have in mind?

src/terms/exact_exchange.jl
```diff
@@ -0,0 +1,81 @@
...
+                ind2 = cat(1,basis.fft_size[2]:-1:2;dims=1)
+                ind3 = cat(1,basis.fft_size[3]:-1:2;dims=1)
+                
+
+                # equivalent to r_to_G(basis, kpoint, rho_ij_real_conj)
+                rho_ij_four_conj = conj(rho_ij_four[ind1,ind2,ind3])
+
+                # Solving poisson equation to obtain potential corresponding to 
+                # ϕ_i(r')ϕ_j(r) / |r - r'|
+                vij_fourier = rho_ij_four_conj .* term.poisson_green_coeffs
```
mfherbst, line 67: Should this not be `rho_ij_four` ? Also it replaces `\int ϕ_i^*(r) ϕ_j(r) / |r - r'| dr`

src/terms/fock_exchange.jl
mfherbst: Hmm. Indeed that's not so nice.

In fact I just checked the literature and there is quite a bit we still need to implement (e.g. some details with respect to discontinuities when using multiple k-points, also some codes seem to use a different k-point mesh for the EXX than for the rest of the code etc.). So short story is: Let's keep it simple for now and do the "fast" implementation once we aim for something more exhaustive.

antoine-levitt:
Note that the BZ integral is singular and needs to be regularized, either by short-ranging the Coulomb (this is what abinit does, apparently) or by a correction technique, see eg https://journals.aps.org/prb/pdf/10.1103/PhysRevB.34.4405 (this is what QE does by default). There's also https://journals.aps.org/prb/pdf/10.1103/PhysRevB.73.205119 which I haven't read yet, and probably more.

antoine-levitt:
See also https://www.vasp.at/wiki/index.php/Coulomb_singularity

zhubonan:
Hi @antoine-levitt @mfherbst,

I am interested in implementing the hybrid functional with k-point sampling. Actually have tried a bit, but there are quite a few things to be done such as getting the coulomb singularity correction computed and generating the *q* point set and corresponding wave functions.  I am wondering if this is already being actively worked on and what is the best way to contribute.

antoine-levitt:
Hi,

As far as I know nobody's working on it actively, feel free to pick it up (the simplest is probably to start another PR; unless @BKaperick objects I think it's fine to use some of the code in this PR if you want, with proper attribution in a commit message). Getting this to work would be awesome!

Regarding the Coulomb the simplest solution to get something working is probably to just do a damped Coulomb. 

For the set of q points, I'm not sure a specific set needs to be computed, can't you just use the existing kpoint grid? Or do you mean in the presence of symmetries? Note also that phonons are being actively worked on (cc @epolack), they might share some of the technology?

Feel free to get in touch (here, by email or on zulip)

zhubonan:
Thanks for the reply! I will try to pick it up and extend the functionality to multiple k points. 

> For the set of q points, I'm not sure a specific set needs to be computed, can't you just use the existing kpoint grid? Or do you mean in the presence of symmetries? 

Yes, the existing grid can be used, but if the actual kpoints are reduced by symmetry only those in the irreducible BZ are calculated. However, the double summation requires the full grid to be used for the q points. So for each q we need to get its wave function from the symmetry related k point. This requires some transformation of the wave function (see https://docs.abinit.org/theory/wavefunctions/#symmetry-properties).

> Note also that phonons are being actively worked on (cc @epolack), they might share some of the technology?

Good call, I will take a look there.

antoine-levitt:
I have not looked closely if symmetries can be used for exchange computations, presumably they can and then the symmetric structure can be used? We use things like apply_symop (in symmetries.jl) to convert when needed. If they can't (or as a first step before tackling the symmetries), it's reasonable to just do a fully non-symmetry-adapted calculation (symmetries=false in Model()). See also unfold_bz(scfres) in symmetry.jl.

BKaperick:
> Hi,
> 
> As far as I know nobody's working on it actively, feel free to pick it up (the simplest is probably to start another PR; unless @BKaperick objects I think it's fine to use some of the code in this PR if you want, with proper attribution in a commit message). Getting this to work would be awesome!
> 
> Regarding the Coulomb the simplest solution to get something working is probably to just do a damped Coulomb.
> 
> For the set of q points, I'm not sure a specific set needs to be computed, can't you just use the existing kpoint grid? Or do you mean in the presence of symmetries? Note also that phonons are being actively worked on (cc @epolack), they might share some of the technology?
> 
> Feel free to get in touch (here, by email or on zulip)

Yes, sorry for letting this drag on @mfherbst and @antoine-levitt , I just could not find the time to get it working.  Thank you for taking over the topic, @zhubonan.  And of course you can use whatever code is useful from this PR.

mfherbst:
Superseeded by  #942.

PR #719, opened 2022-09-06
Adaptive depth DIIS

msdupuy:
Implementation of the adaptive depth DIIS in [https://hal.archives-ouvertes.fr/hal-02492983/document](url) (Algorithm 4 p. 10)

### DIFF
--- a/src/scf/potential_mixing.jl
+++ b/src/scf/potential_mixing.jl
@@ -13,25 +13,32 @@ using Statistics
 #    fixed-point map          g(x)  = V + α Pf(x)
 # where the α may vary between steps.
 #
-# Finds the linear combination xₙ₊₁ = g(xₙ) + ∑ᵢ βᵢ (g(xᵢ) - g(xₙ))
-# such that |Pf(xₙ) + ∑ᵢ βᵢ (Pf(xᵢ) - Pf(xₙ))|² is minimal
+# Finds the linear combination xₙ₊₁ = g(xₙ) + ∑ᵢ βᵢ (g(xᵢ₊₁) - g(xᵢ))
+# such that |Pf(xₙ) + ∑ᵢ βᵢ (Pf(xᵢ₊₁) - Pf(xᵢ))|² is minimal
 struct AndersonAcceleration
     m::Int                  # maximal history size
     iterates::Vector{Any}   # xₙ
     residuals::Vector{Any}  # Pf(xₙ)
+    error::Vector{Real}     # \|Pf(xₙ)\|
+    δ::Real                 # Adaptive parameter in AD-DIIS
     maxcond::Real           # Maximal condition number for Anderson matrix
 end
-AndersonAcceleration(;m=10, maxcond=1e6) = AndersonAcceleration(m, [], [], maxcond)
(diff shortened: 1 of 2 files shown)


antoine-levitt:
Cool! So, how does it perform?

msdupuy:
Without preconditionners, it seems to be more stable than the restarted version that is currently implemented (in the test, it is the case for Aluminium). 
I also tried on GaAs and it converged in less iterations (again without preconditionners)

msdupuy:
I have made a mistake in the restarted parameter so the convergence of AD-DIIIS is in fact similar for aluminium and gaas
al
gaas

src/scf/potential_mixing.jl
```diff
@@ -47,25 +54,37 @@ function (anderson::AndersonAcceleration)(xₙ, αₙ, Pfxₙ)
...
+        if anderson.δ*anderson.error[i] > min_err #ensure δ|Pfxᵢ| ≤ minᵢ |Pfxᵢ| 
+            push!(throw_index,i)
+        end
+    end
+    deleteat!(anderson,throw_index)
+
+    # Special case where the history is of size 1
+    length(anderson.error) == 1 && return xₙ .+ αₙ .* Pfxₙ
+
+    M =  hcat(Pfxs[2:end]...) - hcat(Pfxs[1:end-1]...)  # Mᵢⱼ = (Pfxⱼ₊₁)ᵢ - (Pfxⱼ)ᵢ
```
mfherbst, line 71: I wonder how the numerical stability of this compares to our previous approach of using Pfxi - Pfxn. Would be good to test this on a non-trivial case just to check.

mfherbst (commented):
Thanks! We should test this a bit more before merging. E.g. on some of the cases from https://github.com/mfherbst/supporting-adaptive-damping/ and maybe also in fact using the adaptive damping scheme in combination with the adaptive depth.

src/scf/potential_mixing.jl
```diff
@@ -13,25 +13,32 @@ using Statistics
 #    fixed-point map          g(x)  = V + α Pf(x)
 # where the α may vary between steps.
 #
-# Finds the linear combination xₙ₊₁ = g(xₙ) + ∑ᵢ βᵢ (g(xᵢ) - g(xₙ))
-# such that |Pf(xₙ) + ∑ᵢ βᵢ (Pf(xᵢ) - Pf(xₙ))|² is minimal
+# Finds the linear combination xₙ₊₁ = g(xₙ) + ∑ᵢ βᵢ (g(xᵢ₊₁) - g(xᵢ))
+# such that |Pf(xₙ) + ∑ᵢ βᵢ (Pf(xᵢ₊₁) - Pf(xᵢ))|² is minimal
 struct AndersonAcceleration
     m::Int                  # maximal history size
```
mfherbst, line 19: Add comment somewhere that ideally one chooses `m` to the maximal value fitting into main memory (trusting on `δ` and `maxcond` to take care of conditioning issues). But since it is tricky to do this reliably in an automated fashion we just pick `m=10` here by default.

mfherbst:
@msdupuy ping

mfherbst:
See #939

PR #720, opened 2022-09-08
Derivative of complex erfc

epolack:


test/forwarddiff.jl
```diff
@@ -115,3 +115,16 @@ end
     @show derivative_ε derivative_fd
     @test norm(derivative_ε - derivative_fd) < 1e-4
 end
+
+@testset "Derivative of complex function" begin
+    using SpecialFunctions, FiniteDifferences
+
+    tol = 1e-10
+    fact=randn(ComplexF64)
+    value=randn()
```
mfherbst, line 124: spacing

mfherbst: also you could probably use more helpful variable names (you may use more than 5 characters :smile:)

mfherbst (commented):
We have not yet used FiniteDifferences anywhere in the tests. Feel free to add it though. It could be useful elsewhere, too.

test/forwarddiff.jl
mfherbst: Well you also have a dead line, which I just removed.

mfherbst: Certainly cleaner now, though!

PR #721, opened 2022-09-08
Changing naming convention for FFT

### LINKED ISSUE #715: G_to_r/r_to_G -> ifft/fft?
These names were always a bit gimmicky, and they are relatively long and hard to write (_ is nasty to type). Should we just replace by fft/ifft? They're defined in AbstractFFTs, we can just overload them with our own dispatched version.

epolack:
closes #715

### DIFF
--- a/src/DFTK.jl
+++ b/src/DFTK.jl
@@ -55,10 +55,10 @@ export compute_fft_size
 export G_vectors, G_vectors_cart, r_vectors, r_vectors_cart
 export Gplusk_vectors, Gplusk_vectors_cart
 export Kpoint
-export G_to_r
-export G_to_r!
-export r_to_G
-export r_to_G!
+export ifft
+export ifft!
+export fft
+export fft!
 include("Smearing.jl")
 include("Model.jl")
 include("structure.jl")
--- a/src/PlaneWaveBasis.jl
+++ b/src/PlaneWaveBasis.jl
@@ -37,7 +37,7 @@ Normalization conventions:
   This also holds for the density and the potentials.
 - Quantities expressed on the real-space grid are in actual values.
 
-`G_to_r` and `r_to_G` convert between these representations.
+`ifft` and `fft` convert between these representations.
 """
 struct PlaneWaveBasis{T, VT} <: AbstractBasis{T} where {VT <: Real}
     # T is the default type to express data, VT the corresponding bare value type (i.e. not dual)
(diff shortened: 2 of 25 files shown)


mfherbst:
> you mean on the REPL? Just tried it and seems ok

No what @antoine-levitt is referring to, is that you need to explicitly import `ifft` from `AbstractFFTs` in order to ensure that you are defining a new `method` for the `AbstractFFTs.ifft` function instead of a new function `DFTK.ifft` altogether with only one method.

> yes, although to show the warning julia has to be run with 'depwarn'

That's fine. I turn it on by default (and I don't get why it is turned off by default in julia, but ok).

src/DFTK.jl
```diff
@@ -55,10 +55,10 @@ export compute_fft_size
...
 export Gplusk_vectors, Gplusk_vectors_cart
 export Kpoint
-export G_to_r
-export G_to_r!
-export r_to_G
-export r_to_G!
+export ifft
+export ifft!
+export fft
+export fft!
```
mfherbst, line 61: Should be no need as this is now handled by AbstractFFTs.

src/PlaneWaveBasis.jl
```diff
@@ -57,14 +57,14 @@ struct PlaneWaveBasis{T, VT} <: AbstractBasis{T} where {VT <: Real}
...
+    # The normalizations are performed in ifft/fft according to
     # the DFTK conventions (see above)
     opFFT   # out-of-place FFT plan
     ipFFT   # in-place FFT plan
     opBFFT  # inverse plans (unnormalized plan; backward in FFTW terminology)
     ipBFFT
-    r_to_G_normalization::T  # r_to_G = r_to_G_normalization * FFT
-    G_to_r_normalization::T  # G_to_r = G_to_r_normalization * BFFT
+    fft_normalization::T  # fft = fft_normalization * FFT
+    ifft_normalization::T  # ifft = ifft_normalization * BFFT
```
mfherbst, line 67: alignment

src/PlaneWaveBasis.jl
```diff
@@ -194,13 +194,13 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Number, fft_size, variational,
...
+    # fft = fft_normalization * FFT
     # The convention we want is
     # ψ(r) = sum_G c_G e^iGr / sqrt(Ω)
-    # so that the G_to_r has to normalized by 1/sqrt(Ω).
+    # so that the ifft has to normalized by 1/sqrt(Ω).
     # The other constant is chosen because FFT * BFFT = N
-    G_to_r_normalization = 1/sqrt(model.unit_cell_volume)
-    r_to_G_normalization = sqrt(model.unit_cell_volume) / length(ipFFT)
+    ifft_normalization = 1/sqrt(model.unit_cell_volume)
+    fft_normalization = sqrt(model.unit_cell_volume) / length(ipFFT)
```
mfherbst, line 203: alignment (also elsewhere, please check this)

src/DFTK.jl
mfherbst: OTOH we probably want people to have access to these functions even though they're not `using AbstractFFTs`, so ok, keep as is!

mfherbst (commented):
Thanks!

src/DFTK.jl
mfherbst: Yes, otherwise `fft` will actually be a different function from `AbstractFFTs.fft`, which just happens to have the same name.

PR #722, opened 2022-09-08
Introduce `irfft` for inverse FFT of real-valued quantities

epolack:
Related to #36.

Introducing `irfft` when the inverse Fourier transform is expected to be real, with a check on imaginary part by default.
For some quantities where the Fourier coefficients are computed by hand, we apply `force_real!` to ensure that the inverse Fourier transform is real by removing coefficients for `G` that do not have a `-G` counterpart in the basis.
For some quantities the check has been lazily disabled (which was done silently by default). There are put as `TODO` in comments, as we should have a way not to cast them brutally as real.

### DIFF
--- a/src/fft.jl
+++ b/src/fft.jl
@@ -62,11 +62,11 @@ function ifft(basis::PlaneWaveBasis, f_fourier::AbstractArray)
     f_real
 end
 """
-Perform a real valued iFFT; see `ifft`.
+Perform a real valued iFFT; see [`ifft`](@ref).
 """
 function irfft(basis::PlaneWaveBasis{T}, f_fourier::AbstractArray; check=Val(true)) where {T}
     f_real = ifft(basis, f_fourier)
-    (check == Val(true)) && @assert isapprox(norm(imag(f_real)), 0.0, atol=sqrt(eps(T)))
+    (check == Val(true)) && @assert norm(imag(f_real)) < sqrt(eps(T))
     real(f_real)
 end
 function ifft(basis::PlaneWaveBasis, kpt::Kpoint, f_fourier::AbstractVector; kwargs...)
--- a/src/fft.jl
+++ b/src/fft.jl
@@ -52,17 +52,22 @@ Perform an iFFT to obtain the quantity defined by `f_fourier` defined
 on the k-dependent spherical basis set (if `kpt` is given) or the
 k-independent cubic (if it is not) on the real-space grid.
 """
-function ifft(basis::PlaneWaveBasis, f_fourier::AbstractArray; assume_real=Val(true))
-    # assume_real is true by default because this is the most common usage
-    # (for densities & potentials). Val(true) to help const-prop;
-    # see https://github.com/JuliaLang/julia/issues/44330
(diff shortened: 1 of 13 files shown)


src/fft.jl
```diff
@@ -52,17 +52,22 @@ Perform an iFFT to obtain the quantity defined by `f_fourier` defined
...
     end
-    (assume_real == Val(true)) ? real(f_real) : f_real
+    f_real
+end
+"""
+Perform a real valued iFFT; see `ifft`.
+"""
+function irfft(basis::PlaneWaveBasis{T}, f_fourier::AbstractArray; check=Val(true)) where {T}
+    f_real = ifft(basis, f_fourier)
+    (check == Val(true)) && @assert isapprox(norm(imag(f_real)), 0.0, atol=sqrt(eps(T)))
```
mfherbst, line 69: Why do you always write it like that. I find
```
norm(imag(f_real)) < sqrt(eps(T))
```
way easier to read.

src/symmetry.jl
```diff
@@ -333,3 +333,9 @@ function unfold_kcoords(kcoords, symmetries)
         normalize_kpoint_coordinate(round.(k; digits))
     end
 end
+
+# Filter the wavevectors that don't have an opposite, to ensure fourier_coeffs is real in
+# real space
```
mfherbst, line 338: Proper docstring, maybe:
```
"""
Ensure its real-space equivalent of passed fourier-space representation is entirely real by removing wavevectors `G` that don't have a `-G` counterpart in the basis.
"""
```

src/symmetry.jl
```diff
@@ -333,3 +333,9 @@ function unfold_kcoords(kcoords, symmetries)
...
+function force_real!(fourier_coeffs, basis)
+    lowpass_for_symmetry!(fourier_coeffs, basis; symmetries=[SymOp(-Mat3(I), Vec3(0, 0, 0))])
```
mfherbst, line 340: I think we have `one(SymOp)` for this use case.

src/symmetry.jl
```diff
@@ -333,3 +333,9 @@ function unfold_kcoords(kcoords, symmetries)
         normalize_kpoint_coordinate(round.(k; digits))
     end
 end
+
+# Filter the wavevectors that don't have an opposite, to ensure fourier_coeffs is real in
+# real space
+function force_real!(fourier_coeffs, basis)
```
mfherbst, line 339: Not sure about the name, because could be mistaken as a cast to a real-valued array. Don't have a good alternative in mind though right now.

Also typically we have the basis as first argument in such functions (even though we are not fully consistent).

src/terms/anyonic.jl
```diff
@@ -54,7 +54,8 @@ function make_div_free(basis::PlaneWaveBasis{T}, A) where {T}
             out[1][iG], out[2][iG] = vec
         end
     end
-    [ifft(basis, out[α]) for α = 1:2]
+    # TODO: see https://github.com/JuliaMolSim/DFTK.jl/pull/722
```
mfherbst, line 57: The PR does not really explain why this has been done. Either add explanation to the PR or put it into the code once and refer to it (better).

PR #723, opened 2022-09-12
Adding ElementGaussian

epolack:
Not sure what to do about `docs/src/guide/periodic_problems.jl` where it is explained how to create a new potential.

### DIFF
--- a/docs/src/guide/periodic_problems.jl
+++ b/docs/src/guide/periodic_problems.jl
@@ -267,35 +267,15 @@ plot_bandstructure(basis; n_bands=6, kline_density=100)
 #     (see [Gross-Pitaevskii equation in one dimension](@ref gross-pitaevskii))
 #   * even some more unusual cases like anyonic models.
 #
-# In this tutorial we will go a little more low-level and directly provide
-# an analytic potential describing the interaction with the electrons to DFTK.
+# We will use `ElementGaussian`, which is an analytic potential describing a Gaussian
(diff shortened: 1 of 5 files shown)


mfherbst:
Why do you need that in the main code ? (just curious)

I would leave that example in the docs to make it clear to people how that can be achieved. You should add a comment that DFTK has a gaussian potential by default though and that that should be used in production.

Also I believe there is another example where Gaussians have been used as elements (IIRC GPE + forces or something like that). There you should definitely use the new "builtin" ElementGaussian.

mfherbst:
> Is there some way to print a struct/method definition at the repl? Then we could use that in the docs?

Maybe, but I would not bother. Even if the two things become out of sync it's fine, no?

antoine-levitt:
Sure, it's just to not have binding conflicts (and advertise it's already in DFTK in case someone wants to use it), but it's fine if there's no binding issue

mfherbst:
By the way there is no issue ... if you define it in a script it lives in the `Main` namespace and not in `DFTK`. Also Documenter treats code examples in each page independently.

antoine-levitt:
> By the way there is no issue ... if you define it in a script it lives in the Main namespace and not in DFTK

right but if export ElementGaussian from DFTK, won't it complain if I do `using DFTK; struct ElementGaussian ... end`?

mfherbst:
Nope. Try it in a shell.

mfherbst:
but whether one should avoid it to avoid confusing people is of course a different story.

PR #725, opened 2022-09-13
Fix definition of erfc

epolack:
Sorry, messed up previous commit, `erfc(x) = Γ(0.5, x)/sqrt(π)`... and I only checked derivatives which where ok.

### DIFF
--- a/src/workarounds/forwarddiff_rules.jl
+++ b/src/workarounds/forwarddiff_rules.jl
@@ -306,6 +306,6 @@ function erfc(x::Complex{ForwardDiff.Dual{T,V,N}}) where {T,V,N}
     xx = complex(ForwardDiff.value(real(x)), ForwardDiff.value(imag(x)))
     dx = complex.(ForwardDiff.partials(real(x)), ForwardDiff.partials(imag(x)))
     dgamma = -2*exp(-xx^2)/sqrt(π) * dx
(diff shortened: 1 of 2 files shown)


mfherbst:
Actually now that you mention it, there is another issue in your code: It is not type stable. The `sqrt(\pi)` will promote to at least `Float64`. Please check that as well in a test (e.g. take a Float32 derivative and check you get a Float32 number back).

src/workarounds/forwarddiff_rules.jl
```diff
@@ -305,7 +305,8 @@ end
 function erfc(x::Complex{ForwardDiff.Dual{T,V,N}}) where {T,V,N}
     xx = complex(ForwardDiff.value(real(x)), ForwardDiff.value(imag(x)))
     dx = complex.(ForwardDiff.partials(real(x)), ForwardDiff.partials(imag(x)))
-    dgamma = -2*exp(-xx^2)/sqrt(π) * dx
+    Tx = typeof(xx)
```
mfherbst, line 308: `Tx == T`

mfherbst:
> I get correct type at the end  

Ah yes ... well then a test is not worth it. But good to have the explicit cast.

src/workarounds/forwarddiff_rules.jl
mfherbst: it might be the `V` then instead. For sure ForwardDiff has a way to get that `valtype` I believe.

PR #727, opened 2022-09-14
Allowing for fixed Fermi level

epolack:
Mode to allow fixed Fermi level computations and structure for occupation scheme.

### DIFF
--- a/src/Model.jl
+++ b/src/Model.jl
@@ -21,8 +21,14 @@ struct Model{T <: Real, VT <: Real}
     unit_cell_volume::T
     recip_cell_volume::T
 
-    # Electrons, occupation and smearing function
-    n_electrons::Int  # usually consistent with `atoms` field, but doesn't have to
+    # If not `nothing` then do computations in the canonical ensemble, where
+    # `n_electrons` is fixed; electron count usually consistent with `atoms` field.
+    n_electrons::Union{Int, Nothing}
+    # Expert option: If not `nothing` then do computations in the grand-canonical ensemble,
+    # where `εF` is fixed. Computations with Coulomb electrostatics will fail by default.
+    # Use `check_electrostatics=false` to disable these checks.
+    εF::Union{T, Nothing}
+    check_electrostatics::Bool  # Enables checks that electrostatics is consistent (default)
 
     # spin_polarization values:
     #     :none       No spin polarization, αα and ββ density identical,
@@ -94,28 +100,50 @@ function Model(lattice::AbstractMatrix{T},
                atoms::Vector{<:Element}=Element[],
                positions::Vector{<:AbstractVector}=Vec3{T}[];
                model_name="custom",
-               n_electrons::Int=sum(n_elec_valence, atoms; init=0),
+               n_electrons::Union{Int, Nothing}=nothing,
+               εF=nothing,
+               check_electrostatics=true,  # Don't disable unless you know what you do
                magnetic_moments=T[],
                terms=[Kinetic()],
                temperature=zero(T),
-               smearing=nothing,
+               smearing=default_smearing(temperature),
                spin_polarization=default_spin_polarization(magnetic_moments),
                symmetries=default_symmetries(lattice, atoms, positions, magnetic_moments,
                                              spin_polarization, terms),
                ) where {T <: Real}
-    lattice = Mat3{T}(lattice)
-    temperature = T(austrip(temperature))
+    # Ensembles and electrons
+    if isnothing(n_electrons) && isnothing(εF)  # Default: NVT with electrons given by atoms
+        n_electrons = default_n_electrons(atoms)
+    end
 
-    # Atoms and electrons
+    is_μVT = !isnothing(εF)           # Grand-canonical ensemble
+    is_NVT = !isnothing(n_electrons)  # Canonical ensemble (default)
+    is_μVT && is_NVT && error("`n_electrons` is incompatible with fixed Fermi level `εF`.")
+    something(n_electrons, 0) < 0 && error("n_electrons should be non-negative.")
+
+    if is_μVT && check_electrostatics
+        if extrema(charge_ionic, atoms) != (0, 0)
+            error("DFTK is currently unable to do Coulomb electrostratics in the " *
+                  "grand-canonical ensemble. To disable this check use " *
+                  "`check_electrostatics=false`")
+        end
+    end
+    if is_NVT && !isempty(atoms) && check_electrostatics
+        if sum(charge_ionic, atoms) != n_electrons
+            error("DFTK is currently unable to consistently simulate non-neutral cells. " *
+                  "To disable this check use `check_electrostatics=false`")
+        end
+    end
+
+    # Atoms and terms
     if length(atoms) != length(positions)
         error("Length of atoms and positions vectors need to agree.")
     end
-    n_electrons < 0 && error("n_electrons should be non-negative. Ensure to provide a " *
-                             "non-empty atoms list or an appropriate `n_electrons` kwarg.")
     isempty(terms) && error("Model without terms not supported.")
     atom_groups = [findall(Ref(pot) .== atoms) for pot in Set(atoms)]
 
     # Special handling of 1D and 2D systems, and sanity checks
+    lattice = Mat3{T}(lattice)
     n_dim = count(!iszero, eachcol(lattice))
     n_dim > 0 || error("Check your lattice; we do not do 0D systems")
     for i = n_dim+1:3
@@ -141,11 +169,8 @@ function Model(lattice::AbstractMatrix{T},
     )
     n_spin = length(spin_components(spin_polarization))
 
-    if isnothing(smearing)
-        @assert temperature >= 0
-        # Default to Fermi-Dirac smearing when finite temperature
-        smearing = temperature > 0.0 ? Smearing.FermiDirac() : Smearing.None()
-    end
+    temperature = T(austrip(temperature))
+    temperature < 0 && error("temperature must be non-negative")
 
     if !allunique(string.(nameof.(typeof.(terms))))
         error("Having several terms of the same name is not supported.")
@@ -163,7 +188,8 @@ function Model(lattice::AbstractMatrix{T},
     Model{T,value_type(T)}(model_name,
                            lattice, recip_lattice, n_dim, inv_lattice, inv_recip_lattice,
                            unit_cell_volume, recip_cell_volume,
-                           n_electrons, spin_polarization, n_spin, T(temperature), smearing,
+                           n_electrons, εF, check_electrostatics,
+                           spin_polarization, n_spin, temperature, smearing,
                            atoms, positions, atom_groups, terms, symmetries)
 end
 function Model(lattice::AbstractMatrix{<:Integer}, atoms::Vector{<:Element},
@@ -180,8 +206,14 @@ normalize_magnetic_moment(mm::Number)::Vec3{Float64}         = (0, 0, mm)
 normalize_magnetic_moment(mm::AbstractVector)::Vec3{Float64} = mm
 
 
+"""Defaults to Fermi-Dirac smearing when finite temperature."""
+default_smearing(temperature) = temperature > 0 ? Smearing.FermiDirac() : Smearing.None()
+
+"""Defaults to the number of valence electrons."""
+default_n_electrons(atoms)    = sum(n_elec_valence, atoms; init=0)
+
 """
-:none if no element has a magnetic moment, else :collinear or :full
+`:none` if no element has a magnetic moment, else `:collinear` or `:full`.
 """
 function default_spin_polarization(magnetic_moments)
     isempty(magnetic_moments) && return :none
@@ -243,6 +275,20 @@ function spin_components(spin_polarization::Symbol)
     spin_polarization == :full      && return (:undefined, )
 end
 spin_components(model::Model) = spin_components(model.spin_polarization)
+
+# Ensembles
+is_NVT(model::Model) = !isnothing(model.n_electrons)
+is_μVT(model::Model) = !isnothing(model.εF)
+
+function assert_consistent_electrostatics(model::Model)
+    # DFTK currently assumes in a number of terms that the compensating charge
+    # in the electronic and nuclear terms is equal and of opposite sign.
+    # See also the PSP correction term, where n_electrons is used synonymously
+    # for sum of charges.
+    if model.check_electrostatics && !isempty(model.atoms)
+        @assert sum(charge_ionic, model.atoms) == model.n_electrons
+    end
+end
 
 
 # prevent broadcast
--- a/src/Model.jl
+++ b/src/Model.jl
@@ -26,9 +26,7 @@ struct Model{T <: Real, VT <: Real}
     n_electrons::Union{Int, Nothing}
     # Expert option: If not `nothing` then do computations in the grand-canonical ensemble,
     # where `εF` is fixed. Computations with Coulomb electrostatics will fail by default.
-    # Use `check_electrostatics=false` to disable these checks.
     εF::Union{T, Nothing}
-    check_electrostatics::Bool  # Enables checks that electrostatics is consistent
 
     # spin_polarization values:
     #     :none       No spin polarization, αα and ββ density identical,
@@ -111,25 +109,28 @@ function Model(lattice::AbstractMatrix{T},
                symmetries=default_symmetries(lattice, atoms, positions, magnetic_moments,
                                              spin_polarization, terms),
                ) where {T <: Real}
-    # Ensembles and electrons
-    if isnothing(n_electrons) && isnothing(εF)  # Default: NVT with electrons given by atoms
+    # Electrons and Fermi level
+    if isnothing(n_electrons) && isnothing(εF)
+        # Default: fixed number of electrons given by atoms
         n_electrons = default_n_electrons(atoms)
+    else
+        is_μ_fixed = !isnothing(εF)           # Fixed Fermi level
+        is_n_fixed = !isnothing(n_electrons)  # Fixed number of electrons
+        is_μ_fixed && is_n_fixed && error("`n_electrons` is incompatible with fixed Fermi " *
+                                          "level `εF`.")
+        something(n_electrons, 0) < 0 && error("n_electrons should be non-negative.")
+
+        if check_electrostatics
+            if is_μ_fixed && any(!iszero, charge_ionic.(atoms))
+                error("DFTK is currently unable to do Coulomb electrostratics in the " *
+                      "grand-canonical ensemble. Don't use any charged atoms.")
+            end
+            if is_n_fixed && sum(charge_ionic, atoms; init=0) != n_electrons
+                error("DFTK is currently unable to consistently simulate non-neutral cells.")
+            end
+        end
     end
 
-    is_μVT = !isnothing(εF)           # Grand-canonical ensemble
-    is_NVT = !isnothing(n_electrons)  # Canonical ensemble (default)
-    is_μVT && is_NVT && error("`n_electrons` is incompatible with fixed Fermi level `εF`.")
-    something(n_electrons, 0) < 0 && error("n_electrons should be non-negative.")
-
-    if check_electrostatics
-        if is_μVT && any(!iszero, charge_ionic.(atoms))
-            error("DFTK is currently unable to do Coulomb electrostratics in the " *
-                  "grand-canonical ensemble. Don't use any charged atoms.")
-        end
-        if is_NVT && sum(charge_ionic, atoms; init=0) != n_electrons
-            error("DFTK is currently unable to consistently simulate non-neutral cells.")
-        end
-    end
 
     # Atoms and terms
     if length(atoms) != length(positions)
@@ -184,8 +185,7 @@ function Model(lattice::AbstractMatrix{T},
     Model{T,value_type(T)}(model_name,
                            lattice, recip_lattice, n_dim, inv_lattice, inv_recip_lattice,
                            unit_cell_volume, recip_cell_volume,
-                           n_electrons, εF, check_electrostatics,
-                           spin_polarization, n_spin, temperature, smearing,
+                           n_electrons, εF, spin_polarization, n_spin, temperature, smearing,
                            atoms, positions, atom_groups, terms, symmetries)
 end
 function Model(lattice::AbstractMatrix{<:Integer}, atoms::Vector{<:Element},
@@ -273,18 +273,18 @@ end
 spin_components(model::Model) = spin_components(model.spin_polarization)
 
 # Ensembles
-is_NVT(model::Model) = !isnothing(model.n_electrons)
-is_μVT(model::Model) = !isnothing(model.εF)
+is_n_fixed(model::Model) = !isnothing(model.n_electrons)
+is_μ_fixed(model::Model) = !isnothing(model.εF)
 
 function assert_consistent_electrostatics(model::Model)
     # DFTK currently assumes in a number of terms that the compensating charge
     # in the electronic and nuclear terms is equal and of opposite sign.
     # See also the PSP correction term, where n_electrons is used synonymously
     # for sum of charges.
-    if model.check_electrostatics
-        @assert !is_μVT(model)
-        @assert sum(charge_ionic, model.atoms) == model.n_electrons
-    end
+    is_consistent = ( sum(charge_ionic, model.atoms) == model.n_electrons )
+    is_n_fixed(model) && @assert is_consistent
+    is_μ_fixed(model) && @debug "You are doing electrostatics computations with fixed " *
+                                "Fermi level: don't expect computations to be meaningful."
 end
 
 
--- a/src/Model.jl
+++ b/src/Model.jl
@@ -28,7 +28,7 @@ struct Model{T <: Real, VT <: Real}
     # where `εF` is fixed. Computations with Coulomb electrostatics will fail by default.
     # Use `check_electrostatics=false` to disable these checks.
     εF::Union{T, Nothing}
-    check_electrostatics::Bool  # Enables checks that electrostatics is consistent (default)
+    check_electrostatics::Bool  # Enables checks that electrostatics is consistent
 
     # spin_polarization values:
     #     :none       No spin polarization, αα and ββ density identical,
@@ -102,7 +102,7 @@ function Model(lattice::AbstractMatrix{T},
                model_name="custom",
                n_electrons::Union{Int, Nothing}=nothing,
                εF=nothing,
-               check_electrostatics=true,  # Don't disable unless you know what you do
+               check_electrostatics=any(!iszero, charge_ionic.(atoms)),
                magnetic_moments=T[],
                terms=[Kinetic()],
                temperature=zero(T),
@@ -121,17 +121,13 @@ function Model(lattice::AbstractMatrix{T},
     is_μVT && is_NVT && error("`n_electrons` is incompatible with fixed Fermi level `εF`.")
     something(n_electrons, 0) < 0 && error("n_electrons should be non-negative.")
 
-    if is_μVT && check_electrostatics
-        if extrema(charge_ionic, atoms) != (0, 0)
+    if check_electrostatics
+        if is_μVT && any(!iszero, charge_ionic.(atoms))
             error("DFTK is currently unable to do Coulomb electrostratics in the " *
-                  "grand-canonical ensemble. To disable this check use " *
-                  "`check_electrostatics=false`")
+                  "grand-canonical ensemble. Don't use any charged atoms.")
         end
-    end
-    if is_NVT && !isempty(atoms) && check_electrostatics
-        if sum(charge_ionic, atoms) != n_electrons
-            error("DFTK is currently unable to consistently simulate non-neutral cells. " *
-                  "To disable this check use `check_electrostatics=false`")
+        if is_NVT && sum(charge_ionic, atoms; init=0) != n_electrons
+            error("DFTK is currently unable to consistently simulate non-neutral cells.")
         end
     end
 
@@ -285,7 +281,8 @@ function assert_consistent_electrostatics(model::Model)
     # in the electronic and nuclear terms is equal and of opposite sign.
     # See also the PSP correction term, where n_electrons is used synonymously
     # for sum of charges.
-    if model.check_electrostatics && !isempty(model.atoms)
+    if model.check_electrostatics
+        @assert !is_μVT
         @assert sum(charge_ionic, model.atoms) == model.n_electrons
     end
 end
--- a/src/Model.jl
+++ b/src/Model.jl
@@ -282,7 +282,7 @@ function assert_consistent_electrostatics(model::Model)
     # See also the PSP correction term, where n_electrons is used synonymously
     # for sum of charges.
     if model.check_electrostatics
-        @assert !is_μVT
+        @assert !is_μVT(model)
         @assert sum(charge_ionic, model.atoms) == model.n_electrons
     end
 end
--- a/src/Model.jl
+++ b/src/Model.jl
@@ -21,11 +21,9 @@ struct Model{T <: Real, VT <: Real}
     unit_cell_volume::T
     recip_cell_volume::T
 
-    # If not `nothing` then do computations in the canonical ensemble, where
-    # `n_electrons` is fixed; electron count usually consistent with `atoms` field.
+    # Computations can be performed at fixed `n_electrons` (`n_electrons` Int, `εF` nothing),
+    # or fixed Fermi level (expert option, `n_electrons` nothing, `εF` T)
     n_electrons::Union{Int, Nothing}
-    # Expert option: If not `nothing` then do computations in the grand-canonical ensemble,
-    # where `εF` is fixed. Computations with Coulomb electrostatics will fail by default.
     εF::Union{T, Nothing}
 
     # spin_polarization values:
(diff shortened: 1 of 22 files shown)


src/scf/self_consistent_field.jl
```diff
@@ -36,7 +37,12 @@ function next_density(ham::Hamiltonian;
     eigres.converged || (@warn "Eigensolver not converged" iterations=eigres.iterations)
 
     # Update density from new ψ
-    occupation, εF = compute_occupation(ham.basis, eigres.λ; occupation_threshold)
+    if isnothing(debug_fix_εF)
+        occupation, εF = compute_occupation(ham.basis, eigres.λ; occupation_threshold)
+    else
+        εF = debug_fix_εF
+        occupation = compute_occupation(ham.basis, eigres.λ, εF)
```
antoine-levitt, line 44: keep the threshold

antoine-levitt (commented):
Huh, it's that simple, cool! The name is not great because this is not just debug, it's also physical (it models the grand canonical ensemble). Maybe just fixed_fermi_level? Also it's a property of the model, not of the method used to solve it, so it should go in the Model struct. Then, a number of things probably have to change in response/ (the terms that refer to a variation of the Fermi level)

antoine-levitt:
Yeah it's fine, the contribution got moved to delta fn. Just set deps_F to zero in this case

src/Model.jl
```diff
@@ -58,13 +58,17 @@ struct Model{T <: Real, VT <: Real}
 
     # list of symmetries of the model
     symmetries::Vector{SymOp{VT}}
+
+    # Allows to fix the Fermi level to a certain value. By default (value of `nothing`), the
+    # level depends on the number of electrons.
```
mfherbst, line 63: Mention that this makes DFTK do simulations in the grand-canonical ensemble instead of the canonical ensemble.

src/Model.jl
```diff
@@ -89,6 +93,9 @@ If you want to pass custom symmetry operations (e.g. a reduced or extended set)
 `symmetry_operations` function. Notice that this may lead to wrong results if e.g. the
 external potential breaks some of the passed symmetries. Use `false` to turn off
 symmetries completely.
+
+`fixed_fermi_level` allows to fix the Fermi level instead of having its value depends on the
+number of electrons.
```
mfherbst, line 98: Ensemble stuff.

src/scf/self_consistent_field.jl
```diff
@@ -36,7 +36,17 @@ function next_density(ham::Hamiltonian;
...
+        occupation, εF = compute_occupation(ham.basis, eigres.λ; occupation_threshold)
+    else
+        εF = ham.basis.model.fixed_fermi_level
+        occupation = compute_occupation(ham.basis, eigres.λ, εF)
+        minocc = maximum(minimum, occupation)
+        if ham.basis.model.temperature > 0 && minocc > occupation_threshold
+            @warn "One k-point has a high minimum occupation $minocc. You should probably " *
+                  "increase the number of bands."
+        end
+    end
```
mfherbst, line 49: I'm not so happy to see all this bloat here plus it's repeated code. Maybe move to a separate function?

mfherbst (commented):
Very nice and simple indeed.

mfherbst:
> These two computations are doing literally the same thing

Maybe Julia can be more clever in one case and inline something? But indeed very strange this is not exactly identical. Perhaps just go for an `\approx` after all ?

test/PlaneWaveBasis.jl
```diff
@@ -112,20 +112,21 @@ end
     @test length(G_vectors(fft_size)) == prod(fft_size)
     @test length(r_vectors(basis))    == prod(fft_size)
 
-    @test all(G_vectors(basis)      .== G_vectors(fft_size))
-    @test all(G_vectors_cart(basis) .== map(G -> model.recip_lattice * G,
+    # Using `isapprox` due to https://github.com/JuliaMolSim/DFTK.jl/pull/727#issuecomment-1246749288
+    @test all(G_vectors(basis)      .≈ G_vectors(fft_size))
+    @test all(G_vectors_cart(basis) .≈ map(G -> model.recip_lattice * G,
```
mfherbst, line 117: @epolack Be very careful! This is not a test, this is an assignment!

Also the comment is out of sync.

mfherbst:
Yes the test is too simple. Perhaps just run exactly at the Fermi level keeping the electron count once and once slightly above or below and see whether you get the expected changes.

test/PlaneWaveBasis.jl
mfherbst: > Doesn't .≈ still do the assignation, but at ε accuracy?

Not sure what you mean. `=` does assignment, `==` equality comparison, `===` identity comparision.

> Also not sure what is out of sync.

You are no longer using `isapprox`.

mfherbst: Yes it is, sorry!

Would you mind adding a very strong abstol (e.g. 1e-12 or less) and also take a pass on the alignment.

test/occupation.jl
```diff
@@ -134,3 +134,28 @@ if mpi_nprocs() == 1 # can't be bothered to convert the tests
...
+
+    parameters = (
+        (DFTK.Smearing.FermiDirac(), 0.01, 0.2898048739348472, ≈),
+        (DFTK.Smearing.FermiDirac(), 0.01, 0.40, >),
+        (DFTK.Smearing.FermiDirac(), 0.01, 0.20, <),
+    )
+
+    for (smearing, temperature, εF_ref, op) in parameters
+        model = Model(silicon.lattice, testcase.atoms, testcase.positions;
+                      temperature, smearing, terms=[Kinetic()], fixed_fermi_level=εF_ref)
```
mfherbst, line 153: Why only kinetic?

test/occupation.jl
```diff
@@ -134,3 +134,28 @@ if mpi_nprocs() == 1 # can't be bothered to convert the tests
     end
 end
 end
+
+if mpi_nprocs() == 1 # can't be bothered to convert the tests
+@testset "Fixed Fermi level" begin
+    testcase = magnesium
+    Ecut = 5
+    fft_size = [15, 15, 15]
+    kgrid  = [2, 3, 4]
```
mfherbst, line 143: I would inline these arguments below.

test/occupation.jl
```diff
@@ -134,3 +134,28 @@ if mpi_nprocs() == 1 # can't be bothered to convert the tests
...
+
+    parameters = (
+        (DFTK.Smearing.FermiDirac(), 0.01, 0.2898048739348472, ≈),
```
mfherbst, line 146: If smearing and temperature are the same, don't put them here. Also make a comment about the 0.28 ....

mfherbst: In fact I would actually just compute it in the test.

test/occupation.jl
mfherbst: `εF_ref` is a misleading variable name (not a reference but an input argument), just call it `εF` ?

antoine-levitt (commented):
sorry I thought I posted this but apparently not

src/response/chi0.jl
```diff
@@ -267,7 +267,7 @@ end
     # First compute δεF
     δεF = zero(T)
     δocc = [zero(occ_occ[ik]) for ik = 1:Nk]  # = fn' * (δεn - δεF)
-    if temperature > 0
+    if temperature > 0 && isnothing(ham.basis.model.fixed_fermi_level)
```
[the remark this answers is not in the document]
antoine-levitt: > Should I take care of model.n_electrons?

What do you mean?

> direct_minimization and newton will fail I think. Should I try to fix them?

Nope, that's relatively hard to do (it's basically the same as doing it for finite temperature). Just assert n_electrons is not nothing.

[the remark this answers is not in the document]
antoine-levitt: Crap one other comment got lost. I meant to write at some point: with the fixed fermi level, n_electrons doesn't make sense anymore, so it has to be a Union{Nothing, Int}, and the constructor has to check that either fixed_fermi_level or n_electrons, but not both, are provided.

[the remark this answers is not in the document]
antoine-levitt: This means that a number of things (such as initial density computation, number of band detection, etc) will not work out of the box. That's completely fine: it's an expert option so it's reasonable to ask users to know what they are doing and pass explicitly the initial guess, number of bands, etc. It's better to make everything fail explicitly than have some mixed behavior where some things work but don't do the proper thing.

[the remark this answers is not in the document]
mfherbst: I think we should solve this by having a marker struct in the Model that keeps track of the type of ensemble we are working in. Than it should be cleaner for despatches, checks etc.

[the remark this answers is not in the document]
antoine-levitt: That makes a lot of sense, we could have a abstract struct OccupationScheme and have eg Canonical(n_electrons, temp), GrandCanonical(epsF, temp), MicroCanonical(n_electrons). That's for another PR though.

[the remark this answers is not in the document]
antoine-levitt: @epolack hm that's a very good question actually, hadn't thought about that. Does grand canonical even make sense with periodic electrostatics? This whole idea of fixed Fermi level actually looks nonsensical now I think about it...

test/PlaneWaveBasis.jl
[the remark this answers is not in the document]
antoine-levitt: Hm, that's very weird. Inlining or not should not change semantics, even in floating point. It may end up being a julia bug... Can you maybe reduce it to a self-contained example (without DFT)?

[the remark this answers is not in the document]
antoine-levitt: Fastmath might potentially change things, but we don't use fastmath...

src/response/chi0.jl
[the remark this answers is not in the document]
antoine-levitt: It's not just ewald, it's Hartree as well, and potentially also AtomicLocal with a Coulomb or pseudopotential element type... maybe it's better to just assert at model creation time that there are no Coulomb electrostatics at all, eg `charge_nuclear` is zero for all elements? That still allows us to do custom (non-physical) models eg to debug things. But we definitely should not advertise it too much.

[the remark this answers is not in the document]
antoine-levitt: Remove docstrings is good. Just put an assert, expert users can just remove it if they really want...

antoine-levitt:
Feels like a bug in either julia or static arrays. Can you reduce it even further (maybe without structs at all, just inline and noinline)? Then we can inspect further the generated code.

antoine-levitt:
Hm not sure. You can always do `@code_native` on the constructor, but I'm not aware of something that shows the struct layout (unless you can reverse engineer it from `sizeof`)

antoine-levitt:
See also fieldTAB at the REPL

antoine-levitt:
OK. The code above is still a bit long, can't you isolate a smaller example (not necessarily exhaustive as you did above, just two behaviors that differ) and post it as a julia issue, so that somebody can have a look at it? Also is this consistent across julia versions?

src/response/chi0.jl
```diff
@@ -283,6 +283,9 @@ end
         D = mpi_sum(D, basis.comm_kpts)  # equal to minus the total DOS
         δocc_tot = mpi_sum(sum(basis.kweights .* sum.(δocc)), basis.comm_kpts)
         δεF = δocc_tot / D
+        if !isnothing(ham.basis.model.fixed_εF)
+            δεF = zero(δεF)
+        end
```
mfherbst, line 288: Put this on the `if temperature > 0`, because the whole block does nothing in case `!isnothing(ham.basis.model.fixed_εF)`.

src/occupation.jl
```diff
@@ -88,11 +113,7 @@ function compute_occupation(basis::PlaneWaveBasis{T}, eigenvalues;
         end
     end
 
-    occupation = compute_occupation(basis, eigenvalues, εF)
-    minocc = maximum(minimum, occupation)
-    if temperature > 0 && minocc > occupation_threshold
-        @warn "One k-point has a high minimum occupation $minocc. You should probably increase the number of bands."
-    end
+    occupation = compute_occupation(basis, eigenvalues, εF; occupation_threshold)
```
mfherbst, line 116: actually we need to pass the temperature here as well.

test/PlaneWaveBasis.jl
```diff
@@ -109,23 +109,22 @@ end
     model = Model(silicon.lattice, silicon.atoms, silicon.positions)
     basis = PlaneWaveBasis(model, Ecut, silicon.kcoords, silicon.kweights; fft_size)
 
+    # `isapprox` and not `==` because of https://github.com/JuliaLang/julia/issues/46849
+    atol = 10*eps(eltype(basis))
+    isclose(x, y) = all(isapprox.(x, y; atol))
```
mfherbst, line 114: Can be simplified: isapprox works on vectors.

test/PlaneWaveBasis.jl
```diff
@@ -109,23 +109,22 @@ end
...
-
-        @test all(Gplusk_vectors(basis, kpt) .== map(G -> G + kpt.coordinate,
-                                                     G_vectors(basis, kpt)))
-        @test all(Gplusk_vectors_cart(basis, kpt) .== map(q -> model.recip_lattice * q,
-                                                          Gplusk_vectors(basis, kpt)))
+        @test length(G_vectors(basis, kpt))  == length(kpt.mapping)
+
+        @test isclose(G_vectors_cart(basis, kpt)     , map(G -> model.recip_lattice * G, G_vectors(basis, kpt)))
+        @test isclose(Gplusk_vectors(basis, kpt)     , map(G -> G + kpt.coordinate, G_vectors(basis, kpt)))
+        @test isclose(Gplusk_vectors_cart(basis, kpt), map(q -> model.recip_lattice * q, Gplusk_vectors(basis, kpt)))
```
mfherbst, line 128: keep an eye on line length.

test/chi0.jl
```diff
@@ -8,7 +8,8 @@ include("testcases.jl")
 
 function test_chi0(testcase; symmetries=false, temperature=0,
                    spin_polarization=:none, eigensolver=lobpcg_hyper, Ecut=10,
-                   kgrid=[3, 1, 1], fft_size=[15, 1, 15], compute_full_χ0=false)
+                   kgrid=[3, 1, 1], fft_size=[15, 1, 15], compute_full_χ0=false,
+                   fix_εF=false)
```
mfherbst, line 12: Just have a `fixed_εF=nothing` here.

test/chi0.jl
```diff
@@ -104,15 +110,19 @@ end
     for (case, temperatures) in [(silicon, (0, 0.03)), (magnesium, (0.01, ))]
         for temperature in temperatures, spin_polarization in (:none, :collinear)
             for symmetries in (false, true)
-                test_chi0(case; symmetries, temperature, spin_polarization)
+                for fix_εF in (false, true)
+                    test_chi0(case; symmetries, temperature, spin_polarization, fix_εF)
+                end
```
mfherbst, line 115: Not sure about the full multiplication by 2 in the number of tests. Is it not enough to test some of the combinations here? E.g. metal with collinear / non-collinear spin and fixed fermi level?

test/occupation.jl
```diff
@@ -134,3 +134,38 @@ if mpi_nprocs() == 1 # can't be bothered to convert the tests
...
+    n_electrons_ref = scfres_ref.basis.model.n_electrons
+    @test n_electrons_ref == testcase.n_electrons
+
+    δεF = εF_ref / 4
+
+    for εF in [εF_ref - δεF, εF_ref + δεF]
+        scfres = compute_scfres(εF)
+        @test εF ≈ scfres.εF
+        op = εF > εF_ref ? :> : :<
+        @test eval(op)(DFTK.weighted_ksum(scfres.basis, sum.(scfres.occupation)), n_electrons_ref)
```
mfherbst, line 161: That's pretty complicated for saving one line actually. I prefer that being written out twice.

src/Model.jl
```diff
@@ -94,24 +103,38 @@ function Model(lattice::AbstractMatrix{T},
...
                                              spin_polarization, terms),
+               fixed_εF=nothing,
+               skip_fixed_εF_sane_check=false,
                ) where {T <: Real}
+
+    if !isnothing(fixed_εF)
+        # `n_electrons` incompatible with `fixed_εF`
+        @assert isnothing(n_electrons)
+        # Coulomb electrostratics does not make sense in the grand-canonical ensemble.
+        skip_fixed_εF_sane_check || @assert minmax(charge_nuclear.(atoms)...) == (0, 0)
```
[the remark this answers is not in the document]
antoine-levitt: Yeah the flag is not great, I'd rather not have this as it's really really shaky ground to do fixed fermi level with electrostatics. If we convert it to a struct like michael said, we can have a way to force it.

src/Model.jl
```diff
@@ -94,24 +103,38 @@ function Model(lattice::AbstractMatrix{T},
...
+    elseif isnothing(n_electrons)
+        # Default value if not provided
+        n_electrons = sum(n_elec_valence, atoms; init=0)
+    elseif n_electrons < 0
+        error("`n_electrons` should be non-negative. Ensure to provide a " *
```
antoine-levitt, line 127: you can just `@assert n_electrons >= 0`, if people pass a negative electron number they can debug that themselves

src/occupation.jl
```diff
@@ -2,17 +2,33 @@
...
     # This is needed to get the right behaviour for special floating-point types
     # such as intervals.
     inverse_temperature = iszero(temperature) ? T(Inf) : 1/temperature
 
     filled_occ = filled_occupation(basis.model)
-    [filled_occ * Smearing.occupation.(smearing, (εk .- εF) .* inverse_temperature)
+    occupation = [filled_occ * Smearing.occupation.(smearing, (εk .- εF) .* inverse_temperature)
      for εk in eigenvalues]
+    check_occupation(occupation, occupation_threshold, temperature)
+
```
antoine-levitt, line 30: no newline

src/occupation.jl
```diff
@@ -21,6 +37,15 @@ Find the occupation and Fermi level.
 function compute_occupation(basis::PlaneWaveBasis{T}, eigenvalues;
                             temperature=basis.model.temperature,
                             occupation_threshold) where {T}
+
+    # Early exit if we are doing computations in the grand-canonical ensemble
+    fixed_εF = basis.model.fixed_εF
```
antoine-levitt, line 42: no need to give it a name

src/response/chi0.jl
antoine-levitt: yeah you can remove the +

src/terms/ewald.jl
```diff
@@ -34,7 +34,7 @@ function energy_ewald(model::Model{T}; kwargs...) where {T}
     # terms is equal and of opposite sign. See also the PSP correction term, where
     # n_electrons is used synonymously for sum of charges
     charges = T.(charge_ionic.(model.atoms))
-    @assert sum(charges) == model.n_electrons
+    isnothing(model.n_electrons) || @assert sum(charges) == model.n_electrons
```
antoine-levitt, line 37: I find this tricky to understand... !isnothing(nel) && assert seems more readable? (same below)

src/Model.jl
```diff
@@ -94,24 +103,38 @@ function Model(lattice::AbstractMatrix{T},
                atoms::Vector{<:Element}=Element[],
                positions::Vector{<:AbstractVector}=Vec3{T}[];
                model_name="custom",
-               n_electrons::Int=sum(n_elec_valence, atoms; init=0),
+               n_electrons=nothing,
```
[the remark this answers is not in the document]
mfherbst: Because `5 isa Int` but not `5 isa UInt`, so it's annoying for users.

src/Model.jl
mfherbst: Well you can always argue one or the other. I think this is very user facing and users deserve a nice message. Novices do all sort of crazy things (e.g. they might assume `n_electrons` behaves like a relative charge). It does not cost a lot of code and is just way nicer than an asserition.

docs/src/tricks/parallelization.md
```diff
@@ -64,7 +64,7 @@ function estimate_time_per_scf_step(basis::PlaneWaveBasis)
     (time_per_FFT_per_grid_point
      * prod(basis.fft_size)
      * length(basis.kpoints)
-     * div(basis.model.n_electrons, DFTK.filled_occupation(basis.model), RoundUp)
+     * div(basis.model.ensemble.n_electrons, DFTK.filled_occupation(basis.model), RoundUp)
```
mfherbst, line 67: I wonder if we should have a `n_electrons(model)` and an `n_electrons(basis)` to avoid such deep hierachies in this case.

mfherbst (commented):
some more quick comments

src/OccupationScheme.jl
```diff
@@ -0,0 +1,58 @@
...
+
+    _do_electrostatics::Bool  # option to force the computation of electrostatics terms;
+                              # non-physical and purely for debug purposes
+end
+function GrandCanonical(εF::T; temperature=zero(T),
+                        smearing=default_smearing(temperature)) where {T}
+    temperature < 0 && error("temperature must be non-negative")
+    εF < 0   && error("Fermi level must be non-negative")
+
+    !iszero(temperature) && (temperature = austrip(temperature))
```
mfherbst, line 56: even if temperature is zero you can austrip it.

mfherbst:
@epolack @antoine-levitt I changed the way the electrostatics check is done to unify the assumptions / exceptions we did in both the canonical and the grand-canonical ensembles. I think it is clearer now what happens and still gives the ability to disable if needed. Please take a look if you agree.

src/Model.jl
```diff
@@ -21,8 +21,14 @@ struct Model{T <: Real, VT <: Real}
...
-    # Electrons, occupation and smearing function
-    n_electrons::Int  # usually consistent with `atoms` field, but doesn't have to
+    # If not `nothing` then do computations in the canonical ensemble, where
+    # `n_electrons` is fixed; electron count usually consistent with `atoms` field.
+    n_electrons::Union{Int, Nothing}
+    # Expert option: If not `nothing` then do computations in the grand-canonical ensemble,
+    # where `εF` is fixed. Computations with Coulomb electrostatics will fail by default.
+    # Use `check_electrostatics=false` to disable these checks.
+    εF::Union{T, Nothing}
+    check_electrostatics::Bool  # Enables checks that electrostatics is consistent
```
antoine-levitt, line 31: This is a debug option, I really don't like having it as a field in model. Just remove the check_electrostatics and have a warning?

src/Model.jl
```diff
@@ -21,8 +21,14 @@ struct Model{T <: Real, VT <: Real}
     unit_cell_volume::T
     recip_cell_volume::T
 
-    # Electrons, occupation and smearing function
-    n_electrons::Int  # usually consistent with `atoms` field, but doesn't have to
+    # If not `nothing` then do computations in the canonical ensemble, where
```
antoine-levitt, line 24: this is not necessarily the canonical ensemble (as the temperature might be zero). I'm less and less convinced this vocabulary of ensembles is adequate here...

src/Model.jl
```diff
@@ -94,28 +100,46 @@ function Model(lattice::AbstractMatrix{T},
...
-               smearing=nothing,
+               smearing=default_smearing(temperature),
                spin_polarization=default_spin_polarization(magnetic_moments),
                symmetries=default_symmetries(lattice, atoms, positions, magnetic_moments,
                                              spin_polarization, terms),
                ) where {T <: Real}
-    lattice = Mat3{T}(lattice)
-    temperature = T(austrip(temperature))
+    # Ensembles and electrons
+    if isnothing(n_electrons) && isnothing(εF)  # Default: NVT with electrons given by atoms
```
antoine-levitt, line 115: not necessarily NVT

antoine-levitt: Just say fixed number of electrons

src/Model.jl
```diff
@@ -94,28 +100,46 @@ function Model(lattice::AbstractMatrix{T},
...
+        n_electrons = default_n_electrons(atoms)
+    end
 
-    # Atoms and electrons
```
antoine-levitt, line 109: put that whole block in the `else` clause of the previous thing, so it's clear it's related

src/Model.jl
```diff
@@ -244,6 +272,21 @@ function spin_components(spin_polarization::Symbol)
 end
 spin_components(model::Model) = spin_components(model.spin_polarization)
 
+# Ensembles
+is_NVT(model::Model) = !isnothing(model.n_electrons)
+is_μVT(model::Model) = !isnothing(model.εF)
```
antoine-levitt, line 277: again, not necessarily T. I'd just skip these and check manually the isnothing

src/occupation.jl
```diff
@@ -2,6 +2,23 @@
...
+function compute_occupation(basis::PlaneWaveBasis, eigenvalues;
+                            temperature=basis.model.temperature)
+    if is_μVT(basis.model)
+        εF = basis.model.εF
+    else
+        εF = compute_fermi_level(basis, eigenvalues; temperature)
+    end
+    occupation = compute_occupation(basis, eigenvalues, εF; temperature)
+
+    (; occupation, εF)
```
antoine-levitt, line 18: nice refactor

antoine-levitt:
Ah crap, I still had a pending review in  my queue... so some of these comments might be outdated, sorry

antoine-levitt:
I tried to simplify a bit (possibly overwriting things you did previously, sorry if that's the case)

src/occupation.jl
```diff
@@ -11,15 +29,12 @@ function compute_occupation(basis::PlaneWaveBasis{T}, eigenvalues, εF;
     inverse_temperature = iszero(temperature) ? T(Inf) : 1/temperature
 
     filled_occ = filled_occupation(basis.model)
-    [filled_occ * Smearing.occupation.(smearing, (εk .- εF) .* inverse_temperature)
-     for εk in eigenvalues]
+    occupation = [filled_occ * Smearing.occupation.(smearing, (εk .- εF) .* inverse_temperature)
+                  for εk in eigenvalues]
+    occupation
```
mfherbst, line 34: Why? The function is called `compute_occupation` ... sort of clear what it returns, no?

src/Model.jl
```diff
@@ -94,28 +96,43 @@ function Model(lattice::AbstractMatrix{T},
                atoms::Vector{<:Element}=Element[],
                positions::Vector{<:AbstractVector}=Vec3{T}[];
                model_name="custom",
-               n_electrons::Int=sum(n_elec_valence, atoms; init=0),
+               εF=nothing,
+               n_electrons=isnothing(εF) ? n_electrons_from_atoms(atoms) : nothing,
```
mfherbst, line 100: I really would like the type annotation here to keep people from using floating-point electrons.

src/Model.jl
```diff
@@ -94,28 +96,43 @@ function Model(lattice::AbstractMatrix{T},
...
+        if !isnothing(n_electrons)
+            error("`n_electrons` is incompatible with fixed Fermi " *
+                  "level `εF`.")
+        end
+        if !force_electrostatics && any(!iszero, charge_ionic.(atoms))
+            error("Coulomb electrostatics is incompatible with fixed Fermi level.")
+        end
+    else # fixed number of electrons
+        n_electrons < 0 && error("n_electrons should be non-negative.")
+        if !force_electrostatics && sum(charge_ionic, atoms; init=0) != n_electrons
```
mfherbst, line 122: `n_electrons_from_atoms`?

src/Model.jl
```diff
@@ -94,28 +96,43 @@ function Model(lattice::AbstractMatrix{T},
...
-               smearing=nothing,
+               smearing=temperature > 0 ? Smearing.FermiDirac() : Smearing.None(),
                spin_polarization=default_spin_polarization(magnetic_moments),
                symmetries=default_symmetries(lattice, atoms, positions, magnetic_moments,
                                              spin_polarization, terms),
                ) where {T <: Real}
-    lattice = Mat3{T}(lattice)
-    temperature = T(austrip(temperature))
+    # validate εF and n_electrons
+    if !isnothing(εF) # fixed Fermi level
```
mfherbst, line 112: Two spaces before `#`

src/Model.jl
```diff
@@ -94,28 +96,43 @@ function Model(lattice::AbstractMatrix{T},
                atoms::Vector{<:Element}=Element[],
                positions::Vector{<:AbstractVector}=Vec3{T}[];
                model_name="custom",
-               n_electrons::Int=sum(n_elec_valence, atoms; init=0),
+               εF=nothing,
+               n_electrons=isnothing(εF) ? n_electrons_from_atoms(atoms) : nothing,
+               # force electrostatics with non-neutral cells; results not guaranteed
+               force_electrostatics=false,
```
mfherbst, line 102: I had this as `check_electrostatics` because I think it's more suggestive to "disable a check" than to "force" something, don't you think?

mfherbst (commented):
Good simplificationk

src/Model.jl
```diff
@@ -94,28 +96,44 @@ function Model(lattice::AbstractMatrix{T},
...
+        if !isnothing(n_electrons)
+            error("`n_electrons` is incompatible with fixed Fermi " *
+                  "level `εF`.")
+        end
+        if !force_electrostatics && any(!iszero, charge_ionic.(atoms))
+            error("Coulomb electrostatics is incompatible with fixed Fermi level.")
+        end
+    else # fixed number of electrons
+        n_electrons < 0 && error("n_electrons should be non-negative.")
+        if !force_electrostatics && n_electrons_from_atoms(atoms) !== n_electrons
```
mfherbst, line 123: Oh also `!==` should be `!=` I'd say.

src/Model.jl
mfherbst: Well only to people who know the `--force` flags in unix commands, I'd say.

antoine-levitt: force is a pretty universal notion in computing, no? It conveys the sense that you have to be extra careful.

test/compute_bands.jl
```diff
@@ -99,7 +99,8 @@ end
 
 @testset "High-symmetry kpath construction for 1D system" begin
     lattice = diagm([8.0, 0, 0])
-    model   = Model(lattice; n_electrons=1, terms=[Kinetic()])
+    model   = Model(lattice; n_electrons=1, terms=[Kinetic()],
+                    disable_electrostatics_check=true)
```
mfherbst, line 103: I think we need a special case for this type of situation.

We have a bunch of examples of this sort where there is not even electrostatics at all (e.g. first 1D examples, free electron examples). It does not make sense for people to need to explicitly disable this check here.

mfherbst:
@epolack sorry I found one more :smile:

test/compute_bands.jl
```diff
@@ -99,7 +99,7 @@ end
 
 @testset "High-symmetry kpath construction for 1D system" begin
     lattice = diagm([8.0, 0, 0])
-    model   = Model(lattice; n_electrons=1, terms=[Kinetic()])
+    model   = Model(lattice)
```
mfherbst, line 102: I'd keep the `Kinetic()`. We might change the default later.

test/external_potential.jl
```diff
@@ -11,12 +11,14 @@ if mpi_nprocs() == 1  # Direct minimisation does not yet support MPI
...
              LocalNonlinearity(ρ -> C * ρ^α)]
-    model = Model(lattice; n_electrons=1, terms, spin_polarization=:spinless)
+    model = Model(lattice; n_electrons=1, terms, spin_polarization=:spinless,
+                  disable_electrostatics_check=true)
 
     Ecut = 15
     basis = PlaneWaveBasis(model; Ecut, kgrid=(1, 1, 1))
     scfres_dm = direct_minimization(basis, tol=1e-10)
     scfres_scf = self_consistent_field(basis, tol=1e-10)
+    @show scfres_scf.energies.total
```
mfherbst, line 21: Debug commit.

test/occupation.jl
```diff
@@ -131,3 +131,36 @@ if mpi_nprocs() == 1 # can't be bothered to convert the tests
...
 end
+
+if mpi_nprocs() == 1 # can't be bothered to convert the tests
+@testset "Fixed Fermi level" begin
+    testcase = magnesium
+    atoms = fill(ElementGaussian(1.0, 0.5), length(testcase.positions))
+    temperature = 0.01
+
+    compute_scfres(εF=nothing) = begin
+        comput = isnothing(εF) ? Dict(:n_electrons=>testcase.n_electrons) : Dict(:εF=>εF)
```
mfherbst, line 142: Why a dict? Just use a named tuple instead.

test/diag_compare.jl
```diff
@@ -11,7 +11,7 @@ using DFTK
     end
 
     lattice = Float64[5 0 0; 0 0 0; 0 0 0]
-    model = Model(lattice; n_electrons=4, terms=[Kinetic()])
```
mfherbst, line 14: Again not sure about removing `Kinetic()`.

src/Model.jl
```diff
@@ -118,9 +118,10 @@ function Model(lattice::AbstractMatrix{T},
         if !disable_electrostatics_check && any(!iszero, charge_ionic.(atoms))
             error("Coulomb electrostatics is incompatible with fixed Fermi level.")
         end
-    else # fixed number of electrons
+    else  # fixed number of electrons
         n_electrons < 0 && error("n_electrons should be non-negative.")
-        if !disable_electrostatics_check && n_electrons_from_atoms(atoms) != n_electrons
+        if !disable_electrostatics_check && any(!iszero, charge_ionic.(atoms)) &&
```
mfherbst, line 123: This shenanigans is the reason why I initialised `disable_electrostatics_check=all(iszero, charge_ionic.(atoms))` in my version of this PR. I suggest we do the same here to simplify this logic.

LaurentVidal95:
Just passing by to inaugurate the 200th comment. (Sorry for the bad joke).

mfherbst:
> Just passing by to inaugurate the 200th comment. (Sorry for the bad joke).

@LaurentVidal95 I think now we finally have a good reason to exclude you from the next meeting :D

PR #729, opened 2022-09-20
UPF pseudo format (NC only) support

unkcpz:
This is a draft of supporting UPF in DFTK. The parser is from `UPF.jl` by @azadoks. 

The test of projectors in Fourier space adapt from `PspHgh.jl` which I assume has the same implementation logic as abinit (https://github.com/abinit/abinit/blob/2ae4d8224b85559234f137ee3b5383132568d7cd/src/64_psp/m_psp1.F90#L800-L824)? 
However, the results are different (modified from https://github.com/JuliaMolSim/DFTK.jl/pull/399). 

Any comments on how to move on? @mfherbst


antoine-levitt:
Looks great! So nice there's a julia upf parser. Did you see https://github.com/JuliaMolSim/DFTK.jl/pull/399#issuecomment-1156438553? I'm not sure the potential is correct, on particular

mfherbst:
>  So nice there's a julia upf parser.

Well @azadoks made it to include the pseudo in DFTK :smile:.

@unkcpz I'll take a more detailed look tomorrow ... looks good on first sight. My initial points are:
- I'm not sure all UPF files should map to the same psp type. But that's anyway a discussion we are currently having for UPF.jl. For now it's probably ok to leave it as is, but do make a comment that this is sth. to be revisited.
- We should probably issue some sort of warning or error if the UPF contains features we don't yet support (e.g. NLCC).
- About the integration and the potential: We should check more carefully how other codes handle this.

src/pseudo/load_psp.jl
```diff
@@ -27,7 +29,7 @@ function load_psp(key::AbstractString)
         end
     else
         # Not a file: Treat as identifier, add extension if needed
-        fullpath = joinpath(datadir_psp(), lowercase(key))
+        fullpath = joinpath(datadir_psp(), key)
```
mfherbst, line 32: You can't do this, unfortunately. This is needed for Windows support.

test/silicon_upf.jl
```diff
@@ -0,0 +1,62 @@
+include("run_scf_and_compare.jl")
+include("testcases.jl")
+
+function run_silicon_pbe(T ;Ecut=5, grid_size=15, spin_polarization=:none, kwargs...)
+    # These values were computed using ABINIT with the same kpoints as testcases.jl
```
mfherbst, line 5: Abinit has UPF support ... I did not know that.

unkcpz: My fault here😅. This is copy paste from `test/silicon_pbe.jl`, I didn't run the calculation and update the reference. But Abinit __do support__ UPF `v1` which is parsed by `UPF.jl` as well. I will ask @azadoks about how to run it and maybe can use the result as a reference here.

unkcpz:
@antoine-levitt @mfherbst thanks a lot for quick reply.

> I'm not sure all UPF files should map to the same psp type. 

That's true, I think this struct is more general and can be named such as PspBKB which has the format of this paper https://journals.aps.org/prb/abstract/10.1103/PhysRevB.41.5414. Which basically has multiple projectors of each angular momentum and one local potential.

> Did you see https://github.com/JuliaMolSim/DFTK.jl/pull/399#issuecomment-1156438553? I'm not sure the potential is correct, on particular

This is really helpful, I didn't see it, will take a close look.

azadoks:
Hi guys, sorry I'm a bit late.
Great to see that @unkcpz has written this up, thanks!

> I'm not sure all UPF files should map to the same psp type.

This is definitely true; both version of UPF can represent a wide variety of different pseudopotentials (ultrasoft/paw/norm-conserving) with a variety of additional information (non-linear core correction/spin-orbit coupling). It would be good to discuss this over at UPF.jl (I've also added @antoine-levitt to the repo).

I'm just wrapping up a psp8 parser that I'll push to UPF.jl after a bit of testing. Once it's ready, we can try to run some test calculations vs. ABINIT using PseudoDojo potentials (available in UPF v2 and psp8 formats).

azadoks:
For reference, [here's the `eval_psp_local_fourier` equivalent in QE](https://github.com/QEF/q-e/blob/8cb3103b10fc5c4111250a259f857e5b74bf966e/PW/src/vloc_of_g.f90#L10). They note that

> A term erf(r) / r is subtracted in real space (thus making the function short-ranged) and added again in G space (for G!=0). The G=0 term contains \int (V_loc(r) + Ze^2/r) 4pi r^2 dr. This is the "alpha" in the so-called "alpha Z" term of the energy.

So, this does sound a lot like what @antoine-levitt proposed in #399.

Also, I think I found some unit errors still floating around in the UPF parser that need to be addressed.

antoine-levitt:
> A term erf(r) / r is subtracted in real space (thus making the function short-ranged) and added again in G space (for G!=0).

OK that makes sense. They just take erf(r), rather than some erf(c r)? I guess the optimal c is obtained by balancing the error due to finite rmax and that due to finite grid size, but hopefully these are both sufficiently converged that it doesn't matter...

azadoks:
> They just take erf(r), rather than some erf(c r)?

Yeah, it does seem that they don't use any `c` term.

I also dug up the [ABINIT implementation for computing the local potential in fourier space for the `psp8` format](https://github.com/abinit/abinit/blob/8d8531b524c0875199f8ab9f7abf75089b27fb56/src/64_psp/m_psptk.F90#L1195).
Some random notes:
- They don't seem to bet adding any correction factor.
- They do spline interpolate `r^2 * V(r) + r * Zv` when doing the Fourier transform for `q != 0` (I guess rather than doing an interpolation in real-space?)
- `q^2 * V(q)` (result of `psp8lo`) is spline interpolated after the Fourier transform

About the quadrature for the Fourier transform:
- ABINIT opts for ["corrected" trapezoidal integration](https://github.com/abinit/abinit/blob/7305b32543bfb40f4c9d84ea3b95614f439c3bc7/shared/common/src/28_numeric_noabirule/m_numeric_tools.F90#L2800)
- QE is using [Simpson's rule](https://github.com/QEF/q-e/blob/98b8236fd7e4aa6b447d5a01efc7418eecffdc79/upflib/simpsn.f90#L9)

In any case, @antoine-levitt 's comment about `quadgk` in the TM thread seems reasonable, and we may want to move towards a simpler/faster quadrature method here too.

I'm having a bit of trouble wrapping my head around all the different approaches here, it seems that both QE and ABINIT are doing a 1-dimensional sine transform of _only_ the radial part of the local potential, while @unkcpz is implementing the 3d FT with Bessel functions; is this correct?
Why do this rather than the 1d case that seems like it could be more performant?

@mfherbst It could be good to set up a meeting at some point to discuss.

antoine-levitt:
> I'm having a bit of trouble wrapping my head around all the different approaches here, it seems that both QE and ABINIT are doing a 1-dimensional sine transform of only the radial part of the local potential, while @unkcpz is implementing the FT for the radial function and spherical harmonics using Bessel functions, is this correct?

I don't understand what you mean, but 1D sine transforms should be sufficient for the local potential. Higher Bessel functions are probably needed for the KB pseudopotential form factors, though.

> They don't seem to bet adding any correction factor.

What do you mean?

> About the quadrature for the Fourier transform:
> ABINIT opts for ["corrected" trapezoidal integration](https://github.com/abinit/abinit/blob/7305b32543bfb40f4c9d84ea3b95614f439c3bc7/shared/common/src/28_numeric_noabirule/m_numeric_tools.F90#L2800)
>QE is using [Simpson's rule](https://github.com/QEF/q-e/blob/98b8236fd7e4aa6b447d5a01efc7418eecffdc79/upflib/simpsn.f90#L9)

It seems to me that plain old trapezoidal should already be super good (ie exponentially convergent), see the paper I linked somewhere. Might just have to be careful about the origin (the trick is performing the integral of the even-symmetrized function from -R to +R, so the origin counts half). It's actually possible fancier integration techniques are degrading convergence: they are usually designed for definite integrals (ie 1D functions that do not vanish at endpoints).

> They do spline interpolate r^2 * V(r) + r * Zv when doing the Fourier transform for q != 0 (I guess rather than doing an interpolation in real-space?)

This suggests that the r grid is of insufficient accuracy. Do you have some data on this, eg what are commonly used spacings (I assume the grid is uniform)? One reasonable thing to do as a first attempt is to use a simple integration routine without any interpolation, and warn if we are asked for fourier coefficients on too high momentum (we should have 2pi/qmax >> grid_spacing). Eg if we aim to support Ecut=100, that's qmax=14, so a characteristic grid size of about .5, meaning if the grid size if 0.01 or something we can rest relatively easy.

> q^2 * V(q) (result of psp8lo) is spline interpolated after the Fourier transform

This makes sense as V(q) is requested a large amount of times at similar q points

antoine-levitt:
Is the "corrected trapezoidal method" the one in https://math.stackexchange.com/questions/2823288/corrected-trapezoidal-method-resources? If so, it's pretty much equivalent to the simple trapezoidal method, with proper half-accounting at 0 (the h/2 term accounts for the origin counting as half, and f'(a), f(b) and f'(b) should be zero)

antoine-levitt:
(except that if they correct by 1/r instead of erf(r), that makes the thing non-even, so that's probably a bad idea)

unkcpz:
> 1-dimensional sine transform of only the radial part of the local potential, while @unkcpz is implementing the FT for the radial function and spherical harmonics using Bessel functions, is this correct?

I only implement the `sin(2π * q * r)/(2π * q)` for the local potential.

azadoks:
> What do you mean?
>> They don't seem to be adding any correction factor

At least they aren't playing the same `erf` game as QE.
From what I understood from Jusong's code comments, the expression to compute is
```
V(q) = 4pi * int[ sin(2pi * q * r) / (2pi * q * r) * (r^2 * V(r) + r * Zv) dr ]
```

I think this is equivalent to
```
q^2 V(q) = 2q int[ sin(2pi * q * r) * (r * V(r) + Zv) dr ]
```

The expression ABINIT calculates, as far as I can tell, is
```
q^2 V(q) = 2q int[ sin(2pi * q * r) * (r * V(r) + Zv) dr ] - Zv / pi
```

where `V(q=0) = -Zv / pi` (I guess they're just shifting the potential).

So, I don't see where their correction is if they have one.

> I don't understand what you mean, but 1D sine transforms should be sufficient for the local potential. Higher Bessel functions are probably needed for the KB pseudopotential form factors, though.

You're right, I think I was just confused by Jusong's comment.

> Do you have some data on this, eg what are commonly used spacings (I assume the grid is uniform)?

PseudoDojo pseudos seem to all have a grid step of 0.01.

> Is the "corrected trapezoidal method" the one in https://math.stackexchange.com/questions/2823288/corrected-trapezoidal-method-resources?

I'm not sure; here it is translated into Julia:

```julia
function ctrap(f::Vector{T}, dx::T)::T where {T <: Real}
    nf = length(f)
    if nf >= 10
        endpt = (
            T(23.75) * (f[1] + f[nf    ]) + 
            T(95.10) * (f[2] + f[nf - 1]) +
            T(55.20) * (f[3] + f[nf - 2]) +
            T(79.30) * (f[4] + f[nf - 3]) +
            T(70.65) * (f[5] + f[nf - 4])
        ) / 72
        
        if nf > 10
            ans = (sum(f[6:nf-5]) + endpt) * dx
        else
            ans = endpt * dx
        end
    elseif nf >= 8
        endpt = (
            17(f[1] + f[nf    ]) +
            59(f[2] + f[nf - 1]) +
            43(f[3] + f[nf - 2]) +
            49(f[4] + f[nf - 3])
        ) / 48
        if nf == 9
            ans = (f[5] + endpt) * dx
        else
            ans = endpt * dx
        end
    elseif nf == 7
        ans = (17(f[1] + f[7]) + 59(f[2] + f[6]) + 43(f[3] + f[5]) + 50(f[4])) / 48 * dx
    elseif nf == 6
        ans = (17(f[1] + f[6]) + 59(f[2] + f[5]) + 44(f[3] + f[4])           ) / 48 * dx
    elseif nf == 5
        ans = (  (f[1] + f[5]) +  4(f[2] + f[4]) +  2(f[3]       )           ) /  3 * dx
    elseif nf == 4
        ans = ( 3(f[1] + f[4]) +  9(f[2] + f[3])                             ) /  8 * dx
    elseif nf == 3
        ans = (  (f[1] + f[3]) +  4(f[2]       )                             ) /  3 * dx
    elseif nf == 2
        ans = (  (f[1] + f[2])                                               ) /  2 * dx
    elseif nf == 1
        ans = (  (f[1]       )                                               )      * dx
    end
    return ans
end
```

antoine-levitt:
> So, I don't see where their correction is if they have one.

I'm a bit confused by your expressions, but isn't it that their correction is Z/r whereas QE's is Z*erf(r)/r? So it's not really

> where V(q=0) = -Zv / pi (I guess they're just shifting the potential).

but more that they subtract -Zv/q^2 (modulo factors of pi). That's because the Fourier transform of 1/r is 1/q^2.

antoine-levitt:
>I'm not sure; here it is translated into Julia (it doesn't look like it to me, but I might be missing something):

Thanks for the translation: I admit to giving up quite quickly on reading fortran code :-) Then OK they have a very fancy quadrature, which is possibly needed by the fact that they don't have an erf, and therefore are integrating a function whose even-periodic version is not smooth at 0. Might explain why they do splines as well. I would suggest doing the erf thing with a simple trapezoidal method, with the only modification to half-weight the origin. That should be plenty accurate. If it's not I can take a closer look.

mfherbst:
> I don't think I can fix this in the parser; it will probably need to be dealt with here.
If it makes more sense, I can remove all unit conversion from the UPF parser so that the units are at least consistent and can be consistently converted in DFTK.

Ach units are always a pain. I think it's important that each package is consistent. In DFTK we go for atomic units, in UPF we can do something else. But for sure mixing Bohr and Ry in one code is a bad idea and should be avoided. 

> It could be good to set up a meeting at some point to discuss.

Yes! Let's set up a meeting. I'll write around an email tomorrow.

> (except that if they correct by 1/r instead of erf(r), that makes the thing non-even, so that's probably a bad idea)

Why? It's just a term `Z / abs(r)`, which is even, but of course not exactly smooth ... 

So yeah, from the discussion I favour also the `erf` approach. We can think a bit about a possible `c` factor as Antoine suggested. I recall there were some tricks in the GTH pseudos we might be able to steal. But in any case a `c=1` should get us started.

azadoks:
I've implemented the QE and ABINIT approaches as far as I understand them [here](https://github.com/azadoks/DFTK.jl/commit/011ce4f3a13062749e251e90337f9de347118976), along with their quadrature. I also re-wrote an implementation like what Jusong has (no correction) using the on-grid quadrature. For now, I have everything using the "corrected" trapezoidal method that Antoine linked.

Here's what they look like:
image

There's a lot of fiddly unit stuff to deal with, and I suspect that's where the difference between QE and Abinit is.

> Yes! Let's set up a meeting. I'll write around an email tomorrow.

Thanks, sounds good!

antoine-levitt:
> Why? It's just a term Z / abs(r), which is even, but of course not exactly smooth ...

yes sorry that's what I meant

> I've implemented the QE and ABINIT approaches as far as I understand them [here](https://github.com/azadoks/DFTK.jl/commit/011ce4f3a13062749e251e90337f9de347118976)

wow that's quick! Ideally it would be great to find a sample analytical form that looks more or less like the actual thing, and then compare the different approaches in terms of accuracy (against a reference on a much finer grid). If we find some inaccuracies in the way abinit or QE do things we could potentially report it to the developers. The differences are minute, but that's the kind of things the delta people are running against, so they'd probably welcome it.

mfherbst:
@azadoks Thanks for the comparison.

I guess a good way to compare these approaches is to actually check the all-electron wavefunction, which should not be too hard to compute e.g. for helium. Not sure what example element you picked @azadoks, but perhaps we are lucky and already hydrogen or helium shows this difference.

azadoks:
After a bit of fiddling and staring, I think I've gotten the Abinit and QE approaches to agree better.

For reference, here are fairly direct translations of the Abinit and QE subroutines:
```julia
# QE
function vloc_of_g(r, dr, Z, Vloc_r, g, tpiba2, omega)
    # tpiba2 = (2π / alat)^2
    # omega = unit cell volume
    # g = g(i) = i-th shell of G^2 (in units of tpiba2)
    e2 = 2.0  # Electron charge in QE units
    if iszero(g)
        aux = r .* (r .* Vloc_r .+ Z * e2)
        Vloc_q = 4π / omega * simpson_qe(aux, dr)
    else
        aux1 = r .* Vloc_r .+ Z * e2 .* erf.(r)
        fac = Z * e2 / tpiba2
        gx = sqrt(g * tpiba2)
        aux = aux1 .* sin.(gx .* r) ./ gx
        vlcp = simpson_qe(aux, dr)
        vlcp -= fac * exp.(-g * tpiba2 / 4) / g
        Vloc_q = 4π / omega * vclp
    end
    return q2vq
end
```
QE is calculating:
image
Where (and I'm guessing here on the conversion to `q`):
image
Ignoring Omega (I'm sure it cancels somewhere):
image

```julia
# ABINIT
function psp8lo(r, dr, Z, Vloc_r, q)
    rvlpz = r .* Vloc_r .+ Z
    work = r .* rvlpz
    if iszero(q)
        result = ctrap_abinit(work, dr)
        epsatm = 4π * result
        q2vq = -Z / π
    else
        # interpolate Vloc_r using cubic splines if necessary
        arg = 2π * q
        work = sin.(arg .* r) .* rvlpz
        result = ctrap_abinit(work, dr)
        q2vq = -Z / π + 2 * q * result
    end
    return q2vq
end
```
Abinit is calculating:
image
Just by re-arranging:
image

So I end up with
QE:
image
Abinit:
image

From which it's pretty obvious that Abinit is using a `1/r` correction, and QE is using a `erf(r)/r` correction.

I can then write very similar `eval_psp_local_fourier(psp, q)` functions:
```julia
function eval_psp_local_fourier_erf(psp::PspUpf, q::T)::T where {T <: Real}
    if iszero(q)
        vloc = -Inf
    else
        sin_integrand = sin.(2T(π) * q .* psp.rgrid) ./ (2T(π) * q)
        pot_integrand = psp.rgrid .* psp.vloc .+ psp.Zion .* erf.(psp.rgrid)
        integrand = sin_integrand .* pot_integrand
        correction = -psp.Zion / (T(π) * q^2) * exp(-(T(π) * q)^2)
        vloc = 4T(π) * ctrap(integrand, psp.dr) + correction
    end
    return vloc
end

function eval_psp_local_fourier_invr(psp::PspUpf, q::T)::T where {T <: Real}
    if iszero(q)
        vloc = -Inf
    else
        sin_integrand = sin.(2T(π) * q .* psp.rgrid) ./ (2T(π) * q)
        pot_integrand = psp.rgrid .* psp.vloc .+ psp.Zion
        integrand = sin_integrand .* pot_integrand
        correction = -psp.Zion / (T(π) * q^2)
        vloc = 4T(π) * ctrap(integrand, psp.dr) + correction
    end
    return vloc
end
```

Which actually agree quite well:
image

antoine-levitt:
> There is another function build_form_factors in nonlocal.jl; do we need to do the same thing in eval_psp_projector_fourier as Abinit does in psp8nl, or is part of the calculation handled in build_form_factors?

The code is nonlocal.jl is generic to the psp type: in this particular case it handles the ylm part of the projector. So if p(r) = ylm(r) f(r), the job of eval_psp_projector_fourier is to compute the Fourier transform of f (modulo normalizations and maybe a r^(l+1), we don't document this very well...

> When is the normalization 1 / sqrt(dvms) done / does it need to be done given the normalization of the projectors?

It's done in build_projection_vectors_, you don't need to care about it for the implementation of eval_psp_projector_fourier

antoine-levitt:
Re factors of pi, i^l and friends, the formula is the second one in https://docs.dftk.org/dev/developer/useful_formulas/

> Should we have r in the integrand instead of r^2?

If the psp is stored as r*beta, then I would imagine so yeah

mfherbst:
> is to compute the Fourier transform of f (modulo normalizations and maybe a r^(l+1), we don't document this very well...

I think it's ok, see https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/pseudo/NormConservingPsp.jl#L30-L37 and also the tests in https://github.com/JuliaMolSim/DFTK.jl/blob/master/test/PspHgh.jl are useful.

So if `βnl(r)` is the real-space projector (i.e. the actual projector values) than Jusong's integral is correct, i.e. if the UPF actually stores `r * βnl(r)` then you need to do the QE thing times `4\pi`.

By the way: One thing we can also do to check all this is to test the UPF versions of the HGH pseudopotentials against our internal HGH implementation.

unkcpz:
> By the way: One thing we can also do to check all this is to test the UPF versions of the HGH pseudopotentials against our internal HGH implementation.

@azadoks I checked and reran the HGH type upf pseudopotential in http://pseudopotentials.quantum-espresso.org/legacy_tables/hartwigesen-goedecker-hutter-pp/he. The output is the same, so IMO it should be safe to use for comparing.

It is converted using QE tool `upfconv.x` from CPMD format at https://raw.githubusercontent.com/cp2k/cp2k-data/master/potentials/Goedecker/cpmd/pbe/He-q2 which as I understand is exact corresponds to the cp2k format used in DFTK, with the same number for r_loc, C1, C2.

mfherbst:
It looks at least visually the same. Our data is parsed from 
https://github.com/JuliaMolSim/DFTK.jl/blob/master/data/psp/hgh/pbe/he-q2.hgh

unkcpz:
Yes, if you go to [this website tool](https://htmlpreview.github.io/?https://github.com/cp2k/cp2k-data/blob/master/potentials/Goedecker/index.html) and select the type as `pbe/cp2k`, the output will be the one in DFTK. 

By changing the type to `pbe/cpmd` (which is just a different format) and using `upfconv.x` to convert to upf it is the upf file http://pseudopotentials.quantum-espresso.org/upf_files/He.pbe-hgh.UPF. 

There is also the tool `hgh2upf.x` but it seems to need extra aux input called `rhoatom.dat` which I don't know where to get. See https://gitlab.com/QEF/q-e/-/issues/429, may need to ask the author of the code. I think we shouldn't bother ourselves at the moment.

mfherbst:
Related to #2 and #134.

azadoks:
Cool, thanks guys! I'll work with the He pseudo until the local part is ironed out.
I haven't made much progress on it today, though, and I'm a bit stuck.

The trapezoidal integrals on the grid and the "reference" integral from the test match, but they both have nasty oscillations.
Neither the `-Zerf(r)/r` nor the `-Z/r` correction terms seem to do much to fix this, but I'm not sure I have them right (i.e. in the right convention) anyway.

Abinit uses `-Z/(π * q^2)` for `-Z/r`, and QE uses `-Z/(π * q^2)exp(-(πq)^2)`, but their centered Fourier transforms use a different convention `sin(2πqr)/(2πqr)` rather than `sin(qr)/(qr)` from the equation in the documentation.

image

antoine-levitt:
The expressions look correct, and match the docstring of eval_psp_local_fourier. 

V(r)+Z/r should be short-range (ie V(r) should behave at long range as -Z/r), did you check this? If it does, then the first integral is a smooth and short-range function, whose integral with the trapezoidal rule should be very accurate (ie free of oscillations), as long as q << 2pi/(delta r), with delta r the stepsize.

The second integral is analytic with the erf(r)/r correction, because it's then essentially the Fourier transform of erf, and since erf is a primitive of the gaussian, you can compute the FT of erf using the FT of the gaussian. With the 1/r correction, it comes up at the FT of the constant function (ie a Dirac at 0).

azadoks:
For some reason the local potential is different from the UPF file vs. HGH:
image

`V(r)` does behave as `-Z/r` at long range, so subtracting `-Z/r` or, better, `-Zerf(r)/r` does make it smooth and short-range. 
image

The full integrand is `r^2 j0(q*r) R(r)`, so `r^2 sin(q*r)/(q*r) (V(r) - -Z/r)`. Here, I've added the `sin` term, and the function is still smooth and relatively short-range.
image

However, adding the `r^2` destroys this, and the function becomes long-range again. This is the full integrand, so I must have something wrong.
image

antoine-levitt:
That is not short range, there's still significant potential at rmax. Short range means exponential here, so it really should be almost numerical zero. Try plotting with 2Z, or Z/2 : only at the right value should it really be short range. The potential that you plotted for hgh also blows up at zero, which is weird. Also be careful about your y-axis, as the blow-up at zero obscures what's going on at long range

antoine-levitt:
You can also plot in loglin plot to make sure it is exponentially going to zero

unkcpz:
> For some reason the local potential is different from the UPF file vs. HGH:

That is very weird, I think there must be something missing. I also found that cp2k can convert GTH to UPF with their `ATOM` code. Please check the attachments for the input of cp2k and the UPF output. The numerical values in the are the same as the result of qe's `upfconv.x` and the UPF download from the QE website. 
[He.inp.txt](https://git
[He_HGH.UPF-1.upf.txt](https://github.com/JuliaMolSim/DFTK.jl/files/9681752/He_HGH.UPF-1.upf.txt)
hub.com/JuliaMolSim/DFTK.jl/files/9681751/He.inp.txt)

unkcpz:
I also found there is a [`gth.f90`](https://github.com/QEF/q-e/blob/develop/upflib/gth.f90) from q-e (in the same folder of `upfconv.f90`) worth to check, where the `vloc_gth` function doing exactly the same process as `eval_psp_local_fourier `. But when I align them together, they have a 2pi factor difference for integrant. Back to https://github.com/JuliaMolSim/DFTK.jl/pull/729#issuecomment-1260710670 from @azadoks, the second formula he assumed should be `q = sqrt(g (2pi/a)^2)` without 2pi. To me, this seems QE didn't use 2pi factor for the Fourier transform. I hope this can provide some inspiration.

azadoks:
> Back to https://github.com/JuliaMolSim/DFTK.jl/pull/729#issuecomment-1260710670 from @azadoks, the second formula he assumed should be q = sqrt(g (2pi/a)^2) without 2pi. To me, this seems QE didn't use 2pi factor for the Fourier transform.

I thought that factors of 2pi in q and g had more to do with the definition of the reciprocal cell than with the Fourier transform convention, though I might be wrong.

azadoks:
I found a possible reason for all of the discrepancies: the mesh in the UPF file is not evenly-spaced.

I'll work on implementing interpolation back to an evenly-spaced mesh, hopefully this fixes most of the issues.

azadoks:
I'm still working out how best to interpolate from the log grid to a linear grid. I did get curious and tried integrating on the log grid, and with the following code:
```julia
function eval_psp_local_fourier(psp::PspUpf, q::T)::T where {T <: Real}
    vloc = psp.vloc  # in Rydberg
    Z = psp.Zion * 2 # for some strange reason
    # -Z/r correction (canceling `r` in the code):
    #   F[V(r) - -Z/r] / 4π = ∫ sin(qr) / (qr) * (V(r) - -Z/r) r^2
    vloc_corr_ignd = sin.(q .* psp.rgrid) ./ q .* (psp.rgrid .* vloc .- -Z)
    # Simpson's method, irregular grid
    vloc_corr_itgl = simpson_qe(vloc_corr_ignd, psp.rab)
    # F[-Z/r] / 4π
    corr_itgl = -Z / q^2
    # Both terms have 4π factors left to multiply
    # Could remove 1/q from both terms as well, haven't for clarity
    vloc = 4π * (vloc_corr_itgl + corr_itgl)  # still in Rydberg
    return vloc / 2  # in Hartree
end
```

If I use Simpson's method as implemented in QE to integrate on the log grid, I get:
image
image

Nice!

Q: should I try to interpolate back to a linear mesh, or is this good enough?

azadoks:
Last quick update before the weekend:
On my branch, HGH and HGH in UPF have numerical agreement for the real and fourier local potential and projectors when I check them independently.
I very quickly implemented the pseudo correction basically straight from the test that Michael wrote.

When I plot and compare the fourier local and fourier projectors, they are numerically the same (<1e-8 difference across the board), but there still seems to be something fishy (maybe units, I have to scale the projectors by a factor of 1/2 when I plot them) when I run the SCF:
```
ulia> scfres_hgh.energies
Energy breakdown (in Ha):
    Kinetic             3.1404569 
    AtomicLocal         -2.1028736
    AtomicNonlocal      1.5639986 
    Ewald               -8.4004648
    PspCorrection       -0.2100813
    Hartree             0.5487809 
    Xc                  -2.4160273

    total               -7.876210594699

julia> scfres_upf.energies
Energy breakdown (in Ha):
    Kinetic             3.3850447 
    AtomicLocal         -0.6428023
    AtomicNonlocal      0.7348377 
    Ewald               -8.4004648
    PspCorrection       -0.2100813
    Hartree             0.3758415 
    Xc                  -2.3880702

    total               -7.145694608238
```

antoine-levitt:
> Q: should I try to interpolate back to a linear mesh, or is this good enough?

Nah you should use the data available and integrate it on its grid. I have to think about it and play with it but plausibly what you've done is already good (ie exponentially convergent). Is the logarithmic grid documented somewhere?

antoine-levitt:
Re projectors/potentials being good but results being wrong, have you checked the linking coefficients (Cij or whatever they are called)?

azadoks:
> Is the logarithmic grid documented somewhere?

It's mentioned in the [UPF documentation](http://pseudopotentials.quantum-espresso.org/home/unified-pseudopotential-format) on the q-e website.

It's one of the following:
"log_1"
`e^(x_min) * e^((i-1) * dx) / z` or equivalently `b * e(a * (i-1))`
"log_2"
`e^(x_min) * (e^((i-1) * dx) - 1) / z` or equivalently `b * (e^(a * (i-1)) - 1)`

It seems like the HGH UPFs are on a  "log_1" grid.

> Re projectors/potentials being good but results being wrong, have you checked the linking coefficients (Cij or whatever they are called)?

In the HGH UPFs, it seems like the projectors are in units of `Ry Bohr^-1/2`, so the `Dij` are in `Ry^-1`.
If I convert the projectors to the correct units _after_ the Fourier transform and convert `Dij` in the parser, both quantities match with HGH.

I get the following results for an SCF on Silicon:

```
HGH
----------------------------------------------------
n     Energy            log10(ΔE)   log10(Δρ)   Diag
---   ---------------   ---------   ---------   ----
  1   -7.873981368494                   -0.70    3.5
  2   -7.876842400045       -2.54       -1.49    1.0
  3   -7.877091697920       -3.60       -2.48    3.6
  4   -7.877093329975       -5.79       -3.39    2.9
  5   -7.877093358198       -7.55       -4.30    2.9

Energy breakdown (in Ha):
    Kinetic             3.1378747 
    AtomicLocal         -2.1001044
    AtomicNonlocal      1.5639087 
    Ewald               -8.4004648
    PspCorrection       -0.2100813
    Hartree             0.5470405 
    Xc                  -2.4152669

    total               -7.877093358198

UPF-HGH
-----------------------------------------------------
n     Energy            log10(ΔE)   log10(Δρ)   Diag
---   ---------------   ---------   ---------   ----
  1   -7.876007036953                   -0.70    3.5
  2   -7.878859154990       -2.54       -1.50    1.0
  3   -7.879110847285       -3.60       -2.48    3.5
  4   -7.879112518072       -5.78       -3.39    3.0
  5   -7.879112545150       -7.57       -4.32    2.9

Energy breakdown (in Ha):
    Kinetic             3.1378999 
    AtomicLocal         -2.1030688
    AtomicNonlocal      1.5646709 
    Ewald               -8.4004648
    PspCorrection       -0.2100813
    Hartree             0.5471267 
    Xc                  -2.4151951

    total               -7.879112545150

```

mfherbst:
Haven't read the whole thread (I'm on holidays today), these agreements is on the order I remember between QE and us for small Ecuts. Try doing the same calculation with QE, try playing with Ecut and see if the agreement persists.

mfherbst:
> Note that there are fewer projectors in the UPF file than in the DFTK HGH struct.

That could also be an issue. Can't you just feed the HGH struct in DFTK exactly with the data the UPF file produces?

azadoks:
## Real local potential
image
## Fourier local potential
image
## Real projectors
image
## Fourier projectors
image

azadoks:
> That could also be an issue. Can't you just feed the HGH struct in DFTK exactly with the data the UPF file produces? I.e. hide the extra projectors somehow?

I'll give it a try!

antoine-levitt:
The doc says

> rab(mesh): factor required for discrete integration: ∫ f(r) dr =∑ifi rabi.

Can't you just use that?

azadoks:
Actually, is the number of projectors used in HGH pseudos determined by the number of `h` coefficients?
I can plot more projectors from the HGH than I have in the UPF, but the number, size, and values of the `h` coefficients agree between the two.

azadoks:
> The doc says
>> rab(mesh): factor required for discrete integration: ∫ f(r) dr =∑ifi rabi.

> Can't you just use that?

Yeah, that's exactly what I have been using. I just had to modify the corrected trapezoid function so that it works on a non-homogeneous grid.

antoine-levitt:
Eg when l=1 there is one radial wf stored but 3 actually used (m=-1, m=0, m=1), is that what's confusing you?

mfherbst:
> Actually, is the number of projectors used in HGH pseudos determined by the number of h coefficients?

Yes

https://github.com/JuliaMolSim/DFTK.jl/blob/3ce3b2c/src/pseudo/NormConservingPsp.jl#L87

azadoks:
Ah ok, I don't think we have a projector problem then.
Here are results with higher energy cutoff and a tighter k-grid

```
julia> begin
           hgh = load_psp("hgh/pbe/si-q4.hgh")
           Si = ElementPsp(:Si, psp=hgh)
           atoms     = [Si, Si]
       
           model_hgh = model_PBE(lattice, atoms, positions)
           basis_hgh = PlaneWaveBasis(model_hgh; Ecut=60, kgrid=[12,12,12])
           scfres_hgh = self_consistent_field(basis_hgh);
       end;
n     Energy            log10(ΔE)   log10(Δρ)   Diag
---   ---------------   ---------   ---------   ----
  1   -7.873786946794                   -0.70    2.8
  2   -7.876916294886       -2.50       -1.49    1.0
  3   -7.877238616489       -3.49       -2.47    3.6
  4   -7.877240352994       -5.76       -3.37    3.0
  5   -7.877240383287       -7.52       -4.27    3.2

julia> scfres_hgh.energies
Energy breakdown (in Ha):
    Kinetic             3.1372277 
    AtomicLocal         -2.0994660
    AtomicNonlocal      1.5640342 
    Ewald               -8.4004648
    PspCorrection       -0.2100813
    Hartree             0.5465741 
    Xc                  -2.4150643

    total               -7.877240383287
```

```
julia> begin
           upf = load_psp("upf/pbe/si-q4.upf")
           Si = ElementPsp(:Si, psp=upf)
           atoms     = [Si, Si]
       
           model_upf = model_PBE(lattice, atoms, positions)
           basis_upf = PlaneWaveBasis(model_upf; Ecut=60, kgrid=[12,12,12])
           scfres_upf = self_consistent_field(basis_upf);
       end;
n     Energy            log10(ΔE)   log10(Δρ)   Diag
---   ---------------   ---------   ---------   ----
  1   -7.874405372554                   -0.70    2.8
  2   -7.877522458917       -2.51       -1.49    1.0
  3   -7.877836868155       -3.50       -2.47    3.7
  4   -7.877838631441       -5.75       -3.37    3.0
  5   -7.877838661220       -7.53       -4.29    3.2

julia> scfres_upf.energies
Energy breakdown (in Ha):
    Kinetic             3.1372357 
    AtomicLocal         -2.1003466
    AtomicNonlocal      1.5642613 
    Ewald               -8.4004648
    PspCorrection       -0.2100813
    Hartree             0.5466002 
    Xc                  -2.4150432

    total               -7.877838661220
```

antoine-levitt:
Looks awesome! I'd like to spend some quality time with the integrator, when you prepare a PR I can take a closer look. Let us know if you need help with performance optimization in case it's still slow.

azadoks:
Cool! I've cleaned stuff up a little bit and created a new PR from my branch #741.
We can move the discussion there for refining the approach and improving documentation and testing.

mfherbst:
I guess that we close this in favour of #741.

PR #731, opened 2022-09-21
Misc fixes for the examples

epolack:
Small fixes for the doc, mostly removing useless returns.
I ran it locally and the results seem alright.
~Don't know how to fix ugly `Apply ε† = 1 - χ0 (vc + fxc)` of `dielectric.jl`, so I didn't :p~


examples/anyons.jl
```diff
@@ -6,18 +6,19 @@ using DFTK
 using StaticArrays
 using Plots
 
-## Unit cell. Having one of the lattice vectors as zero means a 2D system
+# Unit cell. Having one of the lattice vectors as zero means a 2D system
```
antoine-levitt, line 9: Meh, these are technical comments that don't really merit a separate markdown section

antoine-levitt: More worryingly, the plot in https://docs.dftk.org/dev/examples/anyons/ doesn't show up anymore, @mfherbst do you know what's up with that?

antoine-levitt (commented):
This is super annoying to get right, so thanks, this looks good!

examples/input_output.jl
```diff
@@ -29,7 +29,7 @@ qe_input  = "Fe_afm.pwi"
 atoms     = load_atoms(qe_input)
 positions = load_positions(qe_input)
 lattice   = load_lattice(qe_input)
-magnetic_moments = load_magnetic_moments(qe_input)
+magnetic_moments = load_magnetic_moments(qe_input);
```
mfherbst, line 32: I think it's nice to see these
```suggestion
magnetic_moments = load_magnetic_moments(qe_input)
```

mfherbst (commented):
Thanks very much @epolack!

PR #739, opened 2022-09-29
Test for consistency of psp correction

mfherbst:
@azadoks This should help you to understand what to implement here. LMK if it is not clear.


antoine-levitt:
Nice!

PR #740, opened 2022-09-29
Save some matrix-vector products in Sternheimer solver

gkemlin:
Two small things I realized yesterday and that I optimized.
- The Schur complement in the Sternheimer solver was not optimized with two Hamiltonian applications per CG iteration, even when extra arrays were empty. So I changed it to one Hamiltonian application per CG iteration, with additional `ham * ψ_extra` which is required only once. This makes it about twice faster !
- Doing this, I realized that `ham * ψ` with `ψ` an array of size zero was not supported in the optimized version of `DftHamiltonian multiplication`, so I added a test to support that. We did not see this before because we had `ψ_extra' * ham * ψ` with `ψ_extra` empty, which is not optimal at all ;)

### DIFF
--- a/src/response/chi0.jl
+++ b/src/response/chi0.jl
@@ -295,10 +295,10 @@ end
 
     # compute δψnk band per band
     δψ = zero.(ψ)
-    Hψ_extra = ham * ψ_extra
     for ik = 1:Nk
         ψk = ψ_occ[ik]
         δψk = δψ[ik]
+        Hψk_extra = ham.blocks[ik] * ψ_extra[ik]
 
         εk = ε_occ[ik]
         for n = 1:length(εk)
@@ -316,7 +316,7 @@ end
             # Sternheimer contribution
             δψk[:, n] .+= sternheimer_solver(ham.blocks[ik], ψk, εk[n], δHψ[ik][:, n], n;
                                              ψk_extra=ψ_extra[ik], εk_extra=ε_extra[ik],
-                                             Hψk_extra=Hψ_extra[ik], kwargs_sternheimer...)
+                                             Hψk_extra, kwargs_sternheimer...)
         end
     end
 
--- a/src/response/chi0.jl
+++ b/src/response/chi0.jl
@@ -111,6 +111,7 @@ precondprep!(P::FunctionPreconditioner, ::Any) = P
 # included).
 function sternheimer_solver(Hk, ψk, εnk, rhs, n; callback=info->nothing,
                             ψk_extra=zeros(size(ψk,1), 0), εk_extra=zeros(0),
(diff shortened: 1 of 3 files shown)


src/terms/Hamiltonian.jl
```diff
@@ -117,6 +117,10 @@ end
                                                                            H::DftHamiltonianBlock,
                                                                            ψ::AbstractArray)
     n_bands = size(ψ, 2)
+    # return empty array if ψ is empty
+    if iszero(n_bands)
+        return zeros(size(ψ,1), 0)
```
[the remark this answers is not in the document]
gkemlin: No, this raises an error when trying to create a partition of size 0.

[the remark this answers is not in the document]
antoine-levitt: Ah, annoying. Feels like Iterators.partition(1:0, 0) should work and produce an empty iterator... wanna make a PR to julia? :-p

antoine-levitt:
Or just return Hpsi and don't bother

src/terms/Hamiltonian.jl
```diff
@@ -117,6 +117,10 @@ end
                                                                            H::DftHamiltonianBlock,
                                                                            ψ::AbstractArray)
     n_bands = size(ψ, 2)
+    # return empty array if ψ is empty
+    if iszero(n_bands)
+        return zeros(eltype(ψ), size(ψ,1), 0)
+    end
```
mfherbst, line 123: I would not bother allocating a new thing:
```suggestion
    iszero(n_bands) && return Hψ  # Nothing to do if ψ empty
```

src/response/chi0.jl
```diff
@@ -295,6 +295,7 @@ end
 
     # compute δψnk band per band
     δψ = zero.(ψ)
+    Hψ_extra = ham * ψ_extra
```
mfherbst, line 298: I think I would do this inside the `ik` loop k-point by k-point to save some memory.

test/chi0.jl
```diff
@@ -72,6 +72,17 @@ function test_chi0(testcase; symmetries=false, temperature=0,
         diff_applied_χ0 = apply_χ0(scfres, δV)
         @test norm(diff_findiff - diff_applied_χ0) < testtol
 
+        # Test apply_χ0 without extra bands
+        ψ_occ, occ_occ = DFTK.select_occupied_orbitals(basis,
+                                                       scfres.ψ,
+                                                       scfres.occupation;
+                                                       threshold=scfres.occupation_threshold)
+        ε_occ = [scfres.eigenvalues[ik][1:size(ψk,2)] for (ik, ψk) in enumerate(ψ_occ)]
```
mfherbst, line 80: space after `,`

PR #744, opened 2022-10-05
Fix forces with symmetries

antoine-levitt:
Thanks @gkemlin for the testcase and @jaemolihm for the debugging! @gkemlin can you make sure that the bug is covered by the tests?

### DIFF
--- a/src/symmetry.jl
+++ b/src/symmetry.jl
@@ -243,7 +243,9 @@ function symmetrize_forces(model::Model, forces; symmetries)
             # (but careful that our symmetries are r -> Wr+w, not R(r+f))
             other_at = W \ (position - w)
             i_other_at = findfirst(a -> is_approx_integer(a - other_at), positions_group)
-            symmetrized_forces[idx] += W * forces[group[i_other_at]]
+            # (A.27) is in cartesian coordinates, and since Wcart is orthogonal,
+            # Fsymcart = Wcart * Fcart <=> Fsymred = inv(Wred') Fred
+            symmetrized_forces[idx] += inv(W') * forces[group[i_other_at]]
         end
     end
     symmetrized_forces / length(symmetries)


gkemlin:
Yes, but we still have the problem in `test/forwardiff.jl` that the different basis do not have the same number of k-points... So I would suggest to add `symmetries=false` in the rHF model from `test/forwardiff.jl` to make sure they have the same number of k-points.

gkemlin:
Well actually we are lucky and it seems that for the specific test configuration it's not the case :smile:

antoine-levitt:
> Yes, but we still have the problem in test/forwardiff.jl that the different basis do not have the same number of k-points

How is that a problem?

gkemlin:
> How is that a problem?
Yes, no my bad, I wrote before thinking.

Actually, there is still a bug in the test.... The gmres starts with a residual which is pretty high.

```
=== gmres ===
rest	iter	resnorm
  1	  1	1.68e+06
  1	  2	4.08e+05
  1	  3	8.57e+04
  1	  4	4.06e+04
  1	  5	7.51e+03
  1	  6	1.00e+03
  1	  7	4.98e+02
  1	  8	5.18e+01
  1	  9	1.29e+01
  1	 10	6.05e+00
  1	 11	9.76e-01
  1	 12	1.34e-01
  1	 13	6.85e-03
  1	 14	1.08e-03
  1	 15	4.48e-04
  1	 16	4.39e-05
  1	 17	1.35e-05
  1	 18	2.38e-06
  1	 19	5.56e-07
  1	 20	3.21e-08
  2	  1	5.32e-06
  2	  2	2.27e-06
  2	  3	2.40e-07
  2	  4	1.13e-07
  2	  5	4.64e-08
  2	  6	1.94e-08
  2	  7	8.19e-09
  2	  8	1.47e-09
  2	  9	3.47e-10
  2	 10	1.51e-10
  2	 11	3.60e-11
Derivative for metals: Test Failed at /Users/runner/work/DFTK.jl/DFTK.jl/test/forwarddiff.jl:56
  Expression: norm(derivative_ε1 - derivative_ε1_fd) < 0.0001
   Evaluated: 2.3349364804113014e17 < 0.0001
Stacktrace:
 [1] macro expansion
   @ ~/work/DFTK.jl/DFTK.jl/test/forwarddiff.jl:56 [inlined]
 [2] macro expansion
   @ /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1151 [inlined]
 [3] macro expansion
   @ ~/work/DFTK.jl/DFTK.jl/test/forwarddiff.jl:51 [inlined]
 [4] macro expansion
   @ /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1151 [inlined]
 [5] top-level scope
   @ ~/work/DFTK.jl/DFTK.jl/test/forwarddiff.jl:10
```

antoine-levitt:
Who cares about macs.

antoine-levitt:
But yeah that's weird...

mfherbst:
I'll merge this, because it is the right direction to go even though it does not yet fix all the bugs.

Issue #742, opened 2022-10-03
Test on forces derivatives is failing unexpectedly


gkemlin:
In the recent CI from the previous PR's (see eg [here](https://github.com/JuliaMolSim/DFTK.jl/actions/runs/3156469908/jobs/5136651740#step:7:2294)), there was some bug happening with the derivative of forces for metal which I wanted to investigate, but this was merged automatically. In the test `test/forwardiff.jl`, we use rHF silicon with temperature for this test and it happens to fail sometimes, always for this specific case and not any other test. At first I thought this was an error from χ0, but actually the error seems to happen before. For instance, I've tweaked a bit the configuration from `test/forwardiff.jl` in the code below, and this results in a error from `symmetrize_ρ` https://github.com/JuliaMolSim/DFTK.jl/blob/17c64890a38b17fc8b57abe5306715aee56e9127/src/response/chi0.jl#L345 where the assertion from https://github.com/JuliaMolSim/DFTK.jl/blob/17c64890a38b17fc8b57abe5306715aee56e9127/src/fft.jl#L69 is not passed.

Any ideas what is happening here ? This really only happens for the rHF model, so maybe there is something we miss here but I can't manage to understand what...

```
using DFTK
using ForwardDiff
using Random

a = 10.26  # Silicon lattice constant in Bohr
lattice = a / 2 * [[0 1 1.];
                   [1 0 1.];
                   [1 1 0.]]
Si = ElementPsp(:Si, psp=load_psp("hgh/lda/Si-q4"))
atoms     = [Si, Si]
positions = [[1.02, 0.97, 1.04] / 8, -ones(3) / 8]
temperature = 1e-3

function compute_force(ε1, ε2)
    T = promote_type(typeof(ε1), typeof(ε2))
    # solve rHF for Silicon (metallic)
    pos = positions .+ [ zeros(3), ε1 * [1., 0, 0] + ε2 * [0, 1., 0]]
    model = model_DFT(Matrix{T}(lattice), atoms, pos, []; temperature)
    basis = PlaneWaveBasis(model; Ecut=5, kgrid=[2, 2, 2], kshift=[0, 0, 0])
    is_converged = DFTK.ScfConvergenceDensity(1e-10)
    scfres = self_consistent_field(basis; is_converged,
                                   response=ResponseOptions(verbose=true))
    # return forces
    compute_forces_cart(scfres)
end

derivative_ε1_fd = let ε1 = 1e-5
[...]

PR #746, opened 2022-10-12
Fix sternheimer tol in ForwardDiff

gkemlin:


src/response/hessian.jl
```diff
@@ -126,7 +126,7 @@ Solve the problem `(Ω+K) δψ = rhs` using a split algorithm, where `rhs` is ty
       to get `δHψ`).
 """
 function solve_ΩplusK_split(ham::Hamiltonian, ρ::AbstractArray{T}, ψ, occupation, εF,
-                            eigenvalues, rhs; tol=1e-8, tol_sternheimer=tol/10,
+                            eigenvalues, rhs; tol=1e-8, tol_sternheimer=1e-8,
```
mfherbst, line 129: That's a bad idea. That should be `tol_sternheimer=tol`.

src/workarounds/forwarddiff_rules.jl
```diff
@@ -218,7 +218,19 @@ function self_consistent_field(basis_dual::PlaneWaveBasis{T};
...
+        residual_norms = scfres.diagonalization[1].residual_norms
+        # first, select only fully converged eigenvectors above the occupation
+        # threshold
+        mask_occ   = map(occk -> isless.(scfres.occupation_threshold, occk),
+                         scfres.occupation)
+        residual_norms_occ = [residual_norms[ik][maskk]
+                              for (ik, maskk) in enumerate(mask_occ)]
+        tol_sternheimer = maximum(maximum.(residual_norms_occ))
+        solve_ΩplusK_split(scfres, -δHextψ; tol=scfres.norm_Δρ,
+                           tol_sternheimer, response.verbose)
```
mfherbst, line 233: `tol_sternheimer` should at most be `tol`, so you have to adapt tol as well.

Also this is a lot of repeated code with things that are also done in `solve_ΩplusK_split(scfres)`. I think that should be refactored.

test/forwarddiff.jl
```diff
@@ -44,7 +44,7 @@ include("testcases.jl")
         @test abs(grad[2] - derivative_ε2[1][1]) < 1e-4
 
         jac = ForwardDiff.jacobian(v -> compute_force(v...)[1], [0.0, 0.0])
-        @test norm(grad - jac[1, :]) < 1e-10
+        @test norm(grad - jac[1, :]) < 1e-9
```
mfherbst, line 47: Not sure. They should in theory even be exactly equal as numerically the same equations are solved.

src/workarounds/forwarddiff_rules.jl
gkemlin: Ok, so we should just take `tol=scfres.norm_Δρ` for both the Dyson and the Sternheimer ?

mfherbst: I'm more tempted to have the maximal residual be both `tol` and `tol_sternheimer`. See the discussion in #742.

docs/src/publications.md
```diff
@@ -46,6 +46,11 @@ Additionally the following publications describe DFTK or one of its algorithms:
 The following publications report research employing DFTK as a core component.
 Feel free to drop us a line if you want your work to be added here.
 
+- E. Cancès, M. F. Herbst, G. Kemlin, A. Levitt, B. Stamm.
+  [*Numerical stability and efficiency of response property calculations in density functional theory*](https://arxiv.org/abs/2210.04512)
+  [ArXiv:2210.04512](https://arxiv.org/abs/2210.04512)
+  ([Supplementary materials and computational scripts](https://github.com/gkemlin/response-calculations-metals)
+
```
mfherbst, line 53: Actually this goes to the other block (`Additionally the following publications describe DFTK or one of its algorithms:`).

src/response/chi0.jl
```diff
@@ -103,6 +104,47 @@ LinearAlgebra.ldiv!(y::T, P::FunctionPreconditioner, x) where {T} = P.preconditi
 LinearAlgebra.ldiv!(P::FunctionPreconditioner, x) = (x .= P.precondition!(similar(x), x))
 precondprep!(P::FunctionPreconditioner, ::Any) = P
 
+# custom cg to ensure that the iterates stay in Ran(R), taken from
+# https://github.com/JuliaLinearAlgebra/IterativeSolvers.jl/blob/master/src/cg.jl
+function cg(A, b, R=Identity();
```
mfherbst, line 109: You might want to use multiple dispatch here instead of defining a new function.

Also this feels a lot like (a) too much code duplication with what `IterativeSolvers` already has and (b) something that breaks easily when `IterativeSolvers` refactors stuff. Are we really sure this is long-term the easiest and best way to solve this problem? Should we not better make a PR upstream?

gkemlin: You mean a PR to IterativeSolvers.jl ?

gkemlin: >  too much code duplication 
Yes I agree, but I can't figure out a shorter way of doing that (except a PR to IterativeSolvers)

mfherbst: > You mean a PR to IterativeSolvers.jl ?

Yes, exactly.

antoine-levitt:
So this is probably where the tradeoff in using an external solver vs rolling our own changes. Cg (unlike, say, gmres, anderson, arnoldi etc), is trivial to code, there's nothing one implementation could do better than the other, and our custom requirement is not trivial to factor out to IterativeSolvers. So let's just do our own? It'll lead to less code in the end.

In fact we might want to start thinking about using our own custom solvers more. For instance, nlsolve has probably resulted in more trouble than it's worth (I'm the one that coded anderson in nlsolve in the first place, we have our custom anderson anyway, we do very strange dances to conform to nlsolve, and I'm pretty sure nlsolve is the number 1 reason our latency is so bad). There's not a lot of activity that goes into improving the julia packages for solvers, so we don't miss out on much. Anyway, just a random thought, I'm not suggesting we do anything about nlsolve now.

gkemlin:
Yes ! I suggest we keep this branch pending then, I create a new one with only the doc update with the paper and we talk about the cg at the meeting this afternoon.

mfherbst:
@antoine-levitt NLSolve I agree, but about the CG let's discuss.

@gkemlin Minor remark: That's the reason why it's usually a good idea to keep to one PR per feature :smile:.

gkemlin:
Ok, so I close this PR and create a new one for the CG.

PR #749, opened 2022-10-13
Creation of supercell system

epolack:
Routine to easily create a supercell system using `ASE` tools.

The `create_supercell_by_hand` function is the one from @LaurentVidal95, without `kpoints` stuff.
I created a separate PR because it's not that straightforward, as I am not using the calculator from `ASE` which seem would require some work that I'm lazy to look into, and bypassing the problem using the `info` field.
I'll take care of the merge if needed.

This is also usefull for phonons, to compare unit cell and supercell results.

### DIFF
--- a/src/DFTK.jl
+++ b/src/DFTK.jl
@@ -62,6 +62,7 @@ export irfft
 export ifft!
 export fft
 export fft!
+export create_supercell
 include("Smearing.jl")
 include("Model.jl")
 include("structure.jl")
@@ -69,6 +70,7 @@ include("PlaneWaveBasis.jl")
 include("fft.jl")
 include("orbitals.jl")
 include("show.jl")
+include("supercell.jl")
 
 export Energies
 include("Energies.jl")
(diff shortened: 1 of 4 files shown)


LaurentVidal95:
Thank you @epolack, I'll definitely use this in the "cell_to_supercell" PR.

mfherbst:
I'm reluctant to have this in DFTK.

The ASE-based parser for input files has never been included with the intention to keep it long-term, but only for short-term convenience until [AtomsBase](https://github.com/JuliaMolSim/AtomsBase.jl) has matured. This is now the case (a generic parser  library producing AtomsBase structures has been developed by some students of mine and will become public in the next weeks). As a result the next step will be to completely remove the ASE-based code in DFTK alltogether as it's only a nuisance to get this working smoothly on CI and for beginners.

So I suggest to develop this purely based on [AtomsBase](https://github.com/JuliaMolSim/AtomsBase.jl) `AbstractSystem`s and include the supercell handling directly in `AtomsBase` (throwing an error if ASE is not installed). Also @rkurchin expressed some interest on Zulip to have stuff like this working fully without python baggage long-term.

BTW: The info-hack in AtomsBase is avoided using this code:
https://github.com/JuliaMolSim/DFTK.jl/blob/2a41de91ebbcddb833b211a130042e93d50a6164/src/external/atomsbase.jl#L37-L51

rkurchin:
@epolack yes, would love to discuss pure-Julia supercell things! Feel free to hit me up on Zulip, there's some discussion already in the #atomsbase channel under the "supercells" topic!

PR #750, opened 2022-10-13
Custom CG

gkemlin:
Here is our custom implementation of CG to allow for projection into a subspace defined as the range of an orthonormal projector.
I've also added a small test to ensure the implementation is correct and changed the call of CG in `solve_ΩplusK`.

I've removed `reltol` to only deal with `abstol` and I've included some things for debugging and convergence monitoring (note that `CG` in capital letters has been chosen for consistency with `LOBPCG`).


src/response/cg.jl
```diff
@@ -0,0 +1,62 @@
+# Implementation of the preconditioned conjugate gradient method, which allows
+# for projection operations, in order to solve linear systems in fixed subspace
+# of the whole space.
+
+"""
+Implementation of the conjugate gradient method which allows for preconditioning
+and projection operations along iterations.
+"""
+function CG!(x, A, b; precon=I, proj=ϕ->ϕ,
```
mfherbst, line 9: Please use lower-case (I never liked the upper-case LOBPCG either)

mfherbst: `proj=identity` (not that it makes a difference, but for clarity)

Add a callback agrument.

src/response/cg.jl
```diff
@@ -0,0 +1,62 @@
...
+    end
+    ldiv!(c, precon, r)
+    γ = dot(r, c)
+    # p is the descent direction
+    p = copy(c)
+    niter = 0
+
+    # convergence history
+    isconverged = false
+    residual_history = []
```
mfherbst, line 31: Can be easily made type stable.

src/response/cg.jl
```diff
@@ -0,0 +1,62 @@
...
+        end
+
+        # apply preconditioner and prepare next iteration
+        ldiv!(c, precon, r)
+        γ_prev, γ = γ, dot(r, c)
+        β = γ / γ_prev
+        p .= proj(c .+ β .* p)
+    end
+
+    ch = (; isconverged, tol, residual_history, niter)
```
mfherbst, line 59: Normalise the name of the tuple with what we do in the diagonalisation. I think we only call it `converged` for example.

src/response/chi0.jl
```diff
@@ -176,8 +175,9 @@ function sternheimer_solver(Hk, ψk, εnk, rhs, n; callback=info->nothing,
         x .= R(precon \ R(y))
     end
     J = LinearMap{eltype(ψk)}(RAR, size(Hk, 1))
-    δψknᴿ, ch = cg(J, bb; Pl=FunctionPreconditioner(R_ldiv!), abstol, reltol,
-                   verbose, log=true)
+    δψknᴿ, ch = CG(J, bb; precon=FunctionPreconditioner(R_ldiv!), tol, proj=R,
+                   verbose)
+    !ch.isconverged && println(ch.residual_history)
```
mfherbst, line 180: This should issue a warning.

test/cg.jl
```diff
@@ -0,0 +1,18 @@
...
+    A = A' * A + I
+    b = rand(n)
+    tol = 1e-10
+
+    x, ch = DFTK.CG(A, b; tol, maxiter=2n)
+
+    @test norm(A*x - b) ≤ 1e-10
+    @test ch.isconverged
+    @test ch.niter == n+1
+end
```
mfherbst, line 18: Also test type stability.

mfherbst (commented):
Nice.

I still think we should move this to a `TinyCG` Julia package (because I can see myself using this elsewhere), but this can be a follow-up.

test/cg.jl
mfherbst: That's not a type stability test as Float64 will be the default type. See also https://docs.julialang.org/en/v1/stdlib/Test/#Test.@inferred

mfherbst: I would use Float32 for a small test and see if you get that back.

mfherbst: well then go for lower. It's just to ensure type stability (you will need two tests though!)

mfherbst: Oh sorry. You can do two things:
- Have a test in Float64 to full accuracy (done :smile:)
- Have a test in Float32, check you get Float32 back. Use `@inferred` on the CG to check the compiler manages to infer the Float32. If this does not work, scratch the `@inferred`.

gkemlin: Yes, but the CG actually returns several things, among which the number of iterations which have no reason to be type dependent, so I should create an intermediate function that returns for instance only the residual or something like that right ?

mfherbst: Yes, the residual or (what I initially had in mind) the solution.

src/response/cg.jl
```diff
@@ -0,0 +1,66 @@
...
+        ldiv!(c, precon, r)
+        γ_prev, γ = γ, dot(r, c)
+        β = γ / γ_prev
+        p .= proj(c .+ β .* p)
+    end
+
+    (; x, converged, tol, residual_norm, residual_history,
+     iterations=n_iter, maxiter)
+end
+cg(x::AbstractVector, A::AbstractMatrix, b::AbstractVector; kwargs...) = cg!(x, LinearMap(A), b; kwargs...)
```
mfherbst, line 64: This should be `cg!`

src/response/cg.jl
```diff
@@ -0,0 +1,66 @@
+using LinearMaps
+
+"""
+Implementation of the conjugate gradient method which allows for preconditioning
+and projection operations along iterations.
+"""
+function cg!(x::AbstractVector{T}, A::LinearMap{T}, b::AbstractVector{T};
+             precon=I, proj=identity, callback=info->nothing,
+             tol=1e-10, maxiter=100, miniter=1, verbose=false) where {T}
```
mfherbst, line 9: No need for `verbose` any more. Solve this by an appropriate callback.

src/response/cg.jl
```diff
@@ -0,0 +1,66 @@
...
+
+        # update iterate and residual while ensuring they stay in Ran(proj)
+        x .= proj(x .+ α .* p)
+        r .= proj(r .- α .* c)
+        residual_norm = norm(r)
+
+        # output
+        verbose && @printf("%3d\t%1.2e\n", n_iter, residual_norm)
+        push!(residual_history, residual_norm)
+        info = (; n_iter, x, r, A, b)
```
mfherbst, line 47: Add residual norm to info, drop residual_history field as can be extracted by callback.

test/cg.jl
```diff
@@ -0,0 +1,37 @@
...
+    x = A \ b
+
+    f(b) = DFTK.cg(A, b; tol, maxiter=2n).x
+    g(b) = DFTK.cg(A, b; tol, maxiter=2n).residual_norm
+    h(b) = eltype(DFTK.cg(A, b; tol, maxiter=2n).residual_history)
+
+    @test x ≈ @inferred f(b)
+    @test tol ≥ @inferred g(b)
+    @test T == @inferred h(b)
+end
```
mfherbst, line 37: repeated code can be saved by having a function to do the test, which takes `T` as an argument.

mfherbst: Also test that x is the right solution by re-computing `norm(A*x-b)`

PR #752, opened 2022-10-17
Dynamical matrix for local term

epolack:
Open points that can be clarified after:
- [ ] Docs of new sternheimer
- [ ] Docs of compute_δHψ_sα


src/response/phonon.jl
```diff
@@ -0,0 +1,291 @@
...
+
+    # `kpt.coordinate + q = kpt′.coordinate + ΔG`
+    # TODO: see if we use this
+    mapping::Vector{Int}           # Index of ``k``-point `kpt′`
+    ΔG_vectors::Vector{Vec3{Int}}  # Shifts ΔG ∈ ℤ³
+
+    ordering
+end
+
+QPoint = P6
```
antoine-levitt, line 17: I'm not sure we really need this to be in a struct. The response functions just need the q coordinate, and the precomputed stuff can be done at a lower level

antoine-levitt (commented):
I will have lots of comments but the basic structure looks sound to me: perturbation functions should take an additional q, defaulting to zero. All in all quite reasonable changes. It'd be nice to reduce duplication of code (eg the d2V), special casing of real/complex stuff, and the ewald, we should discuss how to do that.

antoine-levitt (commented):
Not too happy with the amount of code, let's see if we can trim it down. Also needs comments (by comments I mean: write the equations, and the specifications of the functions' job). I don't know how you can code like this: writing what a function should do *before* actually writing it is super useful.

Otherwise lots of nits, but generally seems OK, great work!

src/response/phonon/commons.jl
```diff
@@ -0,0 +1,86 @@
+# Create a ``(n_{\rm atoms}×n_{\rm dim})^2`` diagonal matrix with the atomic masses of the
+# atoms.
+function get_mass_matrix(basis)
+    model = basis.model
+    positions = model.positions
+    n_atoms = length(positions)
+    n_dim = model.n_dim
```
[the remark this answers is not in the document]
antoine-levitt: I meant it's positions::Vector{Vec3{T}} even if n_dim < 3, so for consistency we should always use 3

antoine-levitt:
I have zero issues with having the dynamical matrix be a tensor. The word matrix is pretty overloaded and can mean a lot of things in different contexts - I see it as the numerical representation of a mapping between vector spaces equipped with bases, and if these bases happen to be naturally indexed by two integers instead of one, that's fine. The important thing to keep in mind is the convention that matrices are (dof output) x (dof input), and that the dof are sorted by having the "innermost" things first.

antoine-levitt:
More precisely, index manipulation arithmetic is error-prone, ugly and hard to read, so if we can avoid it and make julia do it, we should do it as systematically as we can

antoine-levitt:
Nice! There are a number of micro-optimizations that can probably be done here, but let's not bother.

src/postprocess/phonon.jl
```diff
@@ -13,18 +13,88 @@ function dynmat_red_to_cart(model::Model, dynamical_matrix)
...
+        δψs = [zero.(ψ) for _ = 1:3, _ = 1:n_atoms]
+        for s = 1:n_atoms, α = 1:n_dim
+            (; δψ, δρ, δoccupation) = solve_ΩplusK_split(ham, ρ, ψ, occupation, εF,
+                                                         eigenvalues, -δHψs[α, s]; q,
+                                                         kwargs...)
+            δoccupations[α, s] = δoccupation
+            δρs[α, s] = δρ
+            δψs[α, s] = δψ
+        end
+    end
```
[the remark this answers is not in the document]
antoine-levitt: I don't understand what's wrong?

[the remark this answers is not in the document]
antoine-levitt: The interaction wrt the terms looks reasonable. It asks all the terms for their contributions to delta H, it computes the delta psi, then it asks all the terms for their contributions to the dynamical matrix.

src/fft.jl
```diff
@@ -46,25 +46,29 @@ Perform an iFFT to obtain the quantity defined by `f_fourier` defined
 on the k-dependent spherical basis set (if `kpt` is given) or the
 k-independent cubic (if it is not) on the real-space grid.
 """
-function ifft(basis::PlaneWaveBasis, f_fourier::AbstractArray)
+function ifft(basis::PlaneWaveBasis{T}, f_fourier::AbstractArray) where {T}
+    # The barrier simplify flow for phonon that often need complex arrays where standard
```
antoine-levitt, line 50: This comment should be in the next function

antoine-levitt: it's also hard to know what you're talking about at this stage

src/postprocess/phonon.jl
```diff
@@ -13,18 +13,88 @@ function dynmat_red_to_cart(model::Model, dynamical_matrix)
...
     cart_mat = zero.(dynamical_matrix)
-    for τ in 1:size(cart_mat, 2), η in 1:size(cart_mat, 4)
-        cart_mat[:, η, :, τ] = inv_lattice' * dynamical_matrix[:, η, :, τ] * inv_lattice
+    for s = 1:size(cart_mat, 2), β = 1:size(cart_mat, 4)
+        cart_mat[:, β, :, s] = inv_lattice' * dynamical_matrix[:, β, :, s] * inv_lattice
     end
     cart_mat
 end
 
+# Create a tensor equivalent to a``(n_{\rm atoms}×n_{\rm dim})^2`` diagonal matrix with the
```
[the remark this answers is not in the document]
antoine-levitt: also this does return the tensor and not the matrix equivalent so it's super misleading to write (Nat Ndim) ^2 in the comment

src/postprocess/phonon.jl
```diff
@@ -13,18 +13,88 @@ function dynmat_red_to_cart(model::Model, dynamical_matrix)
...
+    model = basis.model
+    positions = model.positions
+    n_atoms = length(positions)
+
+    atoms_mass = atomic_mass.(model.atoms)
+    any(iszero.(atoms_mass)) && @error "Some elements have unknown masses"
+    masses = zeros(T, 3, n_atoms, 3, n_atoms)
+    for s in eachindex(atoms_mass)
+        masses_s = @view masses[:, s, :, s]
+        masses_s[diagind(masses_s)] .= atoms_mass[s]
```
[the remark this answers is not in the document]
antoine-levitt: Just write masses[:, s, :, s] = atoms_mass[s] * I

src/postprocess/phonon.jl
```diff
@@ -13,18 +13,88 @@ function dynmat_red_to_cart(model::Model, dynamical_matrix)
...
+function phonon_modes(basis::PlaneWaveBasis{T}, dynamical_matrix) where {T}
+    model = basis.model
+    positions = model.positions
+    n_atoms = length(positions)
+
+    mass_mat = reshape(get_masses(basis), 3*n_atoms, 3*n_atoms)
+    cart_mat = reshape(dynmat_red_to_cart(model, dynamical_matrix), 3*n_atoms, 3*n_atoms)
+
+    res = eigen(cart_mat, mass_mat)
+    norm(imag(res.values)) > sqrt(eps(T)) && @warn "Some large imaginary phonon modes"
```
antoine-levitt, line 52: careful that people use "imaginary phonon" for something else (omega^2 negative, meaning instabilities)

src/postprocess/phonon.jl
```diff
@@ -13,18 +13,88 @@ function dynmat_red_to_cart(model::Model, dynamical_matrix)
...
+Solve the eigenproblem for a dynamical matrix. The phonon frequencies that are returned are
+real numbers in ``{\rm cm}^{-1}``.
+"""
+function phonon_modes(basis::PlaneWaveBasis{T}, dynamical_matrix) where {T}
+    model = basis.model
+    positions = model.positions
+    n_atoms = length(positions)
+
+    mass_mat = reshape(get_masses(basis), 3*n_atoms, 3*n_atoms)
+    cart_mat = reshape(dynmat_red_to_cart(model, dynamical_matrix), 3*n_atoms, 3*n_atoms)
```
antoine-levitt, line 49: weird to call it cart_mat

src/postprocess/phonon.jl
```diff
@@ -13,18 +13,88 @@ function dynmat_red_to_cart(model::Model, dynamical_matrix)
...
+
+    res = eigen(cart_mat, mass_mat)
+    norm(imag(res.values)) > sqrt(eps(T)) && @warn "Some large imaginary phonon modes"
+
+    signs = sign.(real(res.values))
+    values = signs .* hartree_to_cm⁻¹ .* sqrt.(abs.(real(res.values)))
+
+    (; values, res.vectors)
```
antoine-levitt, line 57: these are are in cartesian, but everything else is in reduced by default, so convert to reduced?

src/postprocess/phonon.jl
```diff
@@ -13,18 +13,88 @@ function dynmat_red_to_cart(model::Model, dynamical_matrix)
...
+end
+phonon_frequencies(args...; kwargs...) = phonon_modes(args...; kwargs...).values
```
antoine-levitt, line 59: not sure this is super useful

src/postprocess/phonon.jl
```diff
@@ -13,18 +13,88 @@ function dynmat_red_to_cart(model::Model, dynamical_matrix)
...
 """
 Compute the dynamical matrix in the form of a ``3×n_{\rm atoms}×3×n_{\rm atoms}`` tensor
 in reduced coordinates.
 """
-@timing function compute_dynmat(basis::PlaneWaveBasis, ψ, occupation; kwargs...)
-    dynmats_per_term = [compute_dynmat(term, basis, ψ, occupation; kwargs...)
+@timing function compute_dynmat(basis::PlaneWaveBasis{T}, ψ, occupation; q=zero(Vec3{T}),
+                                ρ=nothing, ham=nothing, εF=nothing, eigenvalues=nothing,
+                                kwargs...) where {T}
+    is_perturbation_needed = any(t -> isa(t, TermAtomicLocal) ||
```
antoine-levitt, line 68: Is this just for perf? I'd remove it

src/postprocess/phonon.jl
```diff
@@ -13,18 +13,88 @@ function dynmat_red_to_cart(model::Model, dynamical_matrix)
...
+                                      isa(t, TermAtomicNonlocal), basis.terms)
+    δψs = nothing
+    δρs = nothing
+    δoccupations = nothing
```
antoine-levitt, line 72: this is just weird

src/postprocess/phonon.jl
```diff
@@ -13,18 +13,88 @@ function dynmat_red_to_cart(model::Model, dynamical_matrix)
...
+        model = basis.model
+        positions = model.positions
+        n_atoms = length(positions)
+        n_dim = model.n_dim
+
+        δHψs = compute_δHψ(basis, ψ; q)
+        δρs = Array{Array{complex(eltype(basis)), 4}, 2}(undef, 3, n_atoms)
+        for s = 1:n_atoms, α = 1:3
+            δρs[α, s] = zeros(T, basis.fft_size..., basis.model.n_spin_components)
+        end
```
antoine-levitt, line 83: see julia> [i for i=1:2, j=1:2]

src/postprocess/phonon.jl
```diff
@@ -13,18 +13,88 @@ function dynmat_red_to_cart(model::Model, dynamical_matrix)
...
+    δψs = nothing
+    δρs = nothing
+    δoccupations = nothing
+    if is_perturbation_needed
+        model = basis.model
+        positions = model.positions
+        n_atoms = length(positions)
+        n_dim = model.n_dim
+
+        δHψs = compute_δHψ(basis, ψ; q)
```
antoine-levitt, line 79: Maybe simpler to just have compute_dHpsi(basis, psi; q, s, alpha)?

src/postprocess/phonon.jl
```diff
@@ -36,3 +106,21 @@ function compute_dynmat_cart(basis::PlaneWaveBasis, ψ, occupation; kwargs...)
     dynmats_reduced = compute_dynmat(basis, ψ, occupation; kwargs...)
     dynmat_red_to_cart(basis.model, dynmats_reduced)
 end
+
+function compute_dynmat(scfres::NamedTuple; kwargs...)
+    compute_dynmat(scfres.basis, scfres.ψ, scfres.occupation; scfres.ρ, scfres.ham,
+                   scfres.occupation_threshold, scfres.εF, scfres.eigenvalues, kwargs...)
+end
+
+function compute_dynmat_cart(scfres; kwargs...)
```
antoine-levitt, line 115: weird, can't you just have a blanket `compute_dynmat_cart(args...; kwargs...) = convert...` ?

src/response/chi0.jl
```diff
@@ -291,49 +292,60 @@ Perform in-place computations of the derivatives of the wave functions by solvin
 a Sternheimer equation for each `k`-points. It is assumed the passed `δψ` are initialised
 to zero.
 """
-function compute_δψ!(δψ, basis, H, ψ, εF, ε, δHψ; ψ_extra=[zeros_like(ψk, size(ψk,1), 0) for ψk in ψ],
-                     kwargs_sternheimer...)
+# For phonon calculations, the equation to be solved is
+#   (H_k - ε_{n,k-q}) δψ_{n,k} = - Q δHψ_{n, k-q},
```
antoine-levitt, line 296: Who is Q?

src/response/chi0.jl
```diff
@@ -291,49 +292,60 @@ Perform in-place computations of the derivatives of the wave functions by solvin
...
+# as the phonon modes couple the k and k-q points.
+# Furthermore, for temperature computations, we also need the eigenvalues at the k point.
```
antoine-levitt, line 298: doesn't really matter why it's needed

src/response/chi0.jl
```diff
@@ -291,49 +292,60 @@ Perform in-place computations of the derivatives of the wave functions by solvin
...
+# Hence the need for the εp and ε variables.
+# In non-phonon calculations, εp ≡ ε, but in phonon calculations, it contains the reordering
+# of the eigenvalues such that εp[ik] ≡ ε_{·,n-k}.
```
antoine-levitt, line 301: I assume it's k-q?

antoine-levitt: Why reordering?

src/response/chi0.jl
antoine-levitt: just document what they are: eps_p is at k-q, eps is at k. Also weird names, eps_k and eps_k_minus_q ?

antoine-levitt: This is super confusing, maybe let's discuss how to document this

src/response/chi0.jl
```diff
@@ -291,49 +292,60 @@ Perform in-place computations of the derivatives of the wave functions by solvin
...
     end
 end
 
 @views @timing function apply_χ0_4P(ham, ψ, occupation, εF, eigenvalues, δHψ;
-                                    occupation_threshold, kwargs_sternheimer...)
-    basis = ham.basis
+                                    occupation_threshold, q=zero(Vec3{eltype(ham.basis)}),
+                                    kwargs_sternheimer...)
+    basis  = ham.basis
+    kpoints_minus_q = k_to_kpq_mapping(basis, -q)
```
antoine-levitt, line 347: _mapping? Also useless temp, so maybe merge with next line?

src/response/chi0.jl
```diff
@@ -347,23 +359,35 @@ end
...
+    else
+        δεF = zero(εF)
+    end
 
     # Then we compute δψ, again in-place into a zero-padded array
-    δψ = zero.(ψ)
-    δψ_occ = [δψ[ik][:, maskk] for (ik, maskk) in enumerate(mask_occ)]
-    compute_δψ!(δψ_occ, basis, ham.blocks, ψ_occ, εF, ε_occ, δHψ_occ; ψ_extra,
-                kwargs_sternheimer...)
+    δψ_shifted = zero.(δHψ)
```
antoine-levitt, line 382: document all of this better

src/response/hessian.jl
```diff
@@ -182,9 +182,11 @@ Solve the problem `(Ω+K) δψ = rhs` using a split algorithm, where `rhs` is ty
...
     end
 
-    δψ, δoccupation, δεF = apply_χ0_4P(ham, ψ, occupation, εF, eigenvalues, δHψ;
-                                       occupation_threshold, tol=tol_sternheimer,
-                                       kwargs...)
+    (; δψ, δoccupation, δεF) = apply_χ0_4P(ham, ψ, occupation, εF, eigenvalues, δHψ;
+                                           occupation_threshold, tol=tol_sternheimer,
+                                           q, kwargs...)
+    # Note that the δρ that is computed above should be equal to the one computed from δψ
+    # with `compute_analytical_δρ`.
```
antoine-levitt, line 189: why is this comment here?

src/terms/local.jl
```diff
@@ -82,78 +82,149 @@ end
...
+            if !haskey(form_factors, (igroup, p))
                 element = model.atoms[first(group)]
-                form_factors[(igroup, q)] = local_potential_fourier(element, q)
+                form_factors[(igroup, p)] = local_potential_fourier(element, p)
             end
         end
     end
 
-    Gs = to_cpu(G_vectors(basis))  # TODO Again for GPU compatibility
+    Gs = to_cpu([G + q for G in G_vectors(basis)])  # TODO Again for GPU compatibility
```
antoine-levitt, line 110: no need for the to_cpu, the [] does it by itself

src/terms/local.jl
```diff
@@ -82,78 +82,149 @@ end
...
+        S = complex(T)
+    end
+    ifft(S, basis, to_device(basis.architecture, pot_fourier))
 end
+(::AtomicLocal)(basis::PlaneWaveBasis) = TermAtomicLocal(compute_local_potential(basis))
 
-@timing "forces: local" function compute_forces(::TermAtomicLocal, basis::PlaneWaveBasis{TT},
-                                                ψ, occupation; ρ, kwargs...) where {TT}
-    T = promote_type(TT, real(eltype(ψ[1])))
+# Standard computation.
```
antoine-levitt, line 130: this comment is more confusing than helpful (also the next)

src/terms/local.jl
```diff
@@ -82,78 +82,149 @@ end
...
+                        for G in G_vectors(basis)]
         for idx in group
             r = model.positions[idx]
-            forces[idx] = _force_local_internal(basis, ρ_fourier, form_factors, r)
+            forces[idx] = -force_type(S, sum(conj(ρ_fourier[iG])
+                                                 .* form_factors[iG]
+                                                 .* cis2pi(-dot(G + q, r))
+                                                 .* (-2T(π)) .* (G + q) .* im
+                                                 ./ sqrt(basis.model.unit_cell_volume)
+                                             for (iG, G) in enumerate(G_vectors(basis))))
```
antoine-levitt, line 157: nice

antoine-levitt: but this is static arrays and scalars, so you might as well not use broadcast and do plain old * and /, shouldn't have a perf impact

src/terms/local.jl
```diff
@@ -82,78 +82,149 @@ end
...
-function _force_local_internal(basis, ρ_fourier, form_factors, r)
-    T = real(eltype(ρ_fourier))
-    f = zero(Vec3{T})
-    for (iG, G) in enumerate(G_vectors(basis))
-        f -= real(conj(ρ_fourier[iG])
-                  .* form_factors[iG]
-                  .* cis2pi(-dot(G, r))
-                  .* (-2T(π)) .* G .* im
-                  ./ sqrt(basis.model.unit_cell_volume))
+# Phonon: Perturbation of the local potential using AD with respect to a displacement on the
```
antoine-levitt, line 163: don't really care that it uses AD

src/terms/local.jl
```diff
@@ -82,78 +82,149 @@ end
...
+    δ²V_fourier = similar(ρ_fourier)
+    for s = 1:n_atoms, α = 1:n_dim
+        for t = 1:n_atoms, β = 1:n_dim
+            fft!(δ²V_fourier, basis, compute_δ²V(term, basis, β, t, α, s))
+            ∫ρδ²V[β, t, α, s] = sum(conj(ρ_fourier) .* δ²V_fourier)
+        end
+    end
+
+
+    ∫δρδV + ∫ρδ²V
```
antoine-levitt, line 212: very nice

src/transfer.jl
```diff
@@ -225,3 +211,53 @@ function k_to_kpq_mapping(basis::PlaneWaveBasis, qcoord)
...
+If ``ΔG ≔ [k+q] - (k+q)``, then we have that
+```math
+    ∑_G \hat{u}_{[k+q]}(G) e^{i(k+q+G)·r} &= ∑_{G'} \hat{u}_{k+q}(G'-ΔG) e^{i(k+q+ΔG+G')·r},
+```
+hence
+```math
+    u_{k+q}(G) = u_{[k+q]}(G + ΔG).
+```
+"""
+function shift_kplusq(basis, kpt, q, ψk)
```
antoine-levitt, line 228: very confusing names, psik is actually at kpt+q?

src/transfer.jl
```diff
@@ -225,3 +211,53 @@ function k_to_kpq_mapping(basis::PlaneWaveBasis, qcoord)
...
+    kcoord_plus_q = kpt.coordinate .+ q
+    index, ΔG = find_equivalent_kpt(basis, kcoord_plus_q, kpt.spin)
+    equiv_kpt_plus_q = basis.kpoints[index]
+
+    kpt_plus_q = build_kpoints(basis, [kcoord_plus_q])[kpt.spin]
+    transfer_blochwave_kpt(ψk, basis, equiv_kpt_plus_q, kpt_plus_q, -ΔG)
+end
```
antoine-levitt, line 235: get_equivalent_blochwave(basis, psik, k) where k=kpt.coordinate+q?

src/transfer.jl
```diff
@@ -225,3 +211,53 @@ function k_to_kpq_mapping(basis::PlaneWaveBasis, qcoord)
...
+
+# Replaces `multiply_by_expiqr`.
+# TODO: Think about a clear and consistent way to define such a function when completing
+# phonon PRs.
+"""
+Return the Fourier coefficients for ``f_{q} ψ_{k-q}`` in the basis of `k`-points, where
+``f`` is a real function.
```
antoine-levitt, line 242: confusing comment, discuss?

src/workarounds/FourierTransforms.jl/ctfft.jl
```diff
@@ -230,15 +230,15 @@ end
 # Like `fftgen`, below, but generates the naïve $\Theta(n^2)$ DFT algorithm:
 function dftgen(T, ω_n, forward::Bool, n::Integer, x, y)
     n == 1 && return :($(y(0)) = $(x(0)))
-    tmpvars = Symbol[ gensym(string("dftgen_", j)) for j in 0:n-1 ]
+    tmpvars = Symbol[ gensym(string("dftgen_", j)) for j = 0:n-1 ]
```
antoine-levitt, line 233: don't touch this file, it's not ours except for bug fixes

mfherbst (commented):
First round of review from me.

In general I have been wondering if it would make sense to introduce a marker struct `Qpoint(q)` that holds the `q`-vector with a definition
```julia
struct QpointZero end
struct Qpoint{T}
   q::Vec3{T}
end
function Qpoint(q)
if iszero(q)
return QpointZero()
else
return Qpoint(q)
end
end
```
This would make it easy to dispatch in a type-stable fashion the special case that the q-point is zero.

antoine-levitt:
> In general I have been wondering if it would make sense to introduce a marker struct Qpoint(q) that holds the q-vector with a definition

That adds complexity just to add type stability that is probably not that important. We might want to factor such a structure (not in this PR), but if we do it let's do it for code reasons, not perf. We meant to refactor the kgrid stuff anyway, so that'd possibly be part of it. If it turns out to be important to optimize the case q=0, we can possibly get constant propagation to do it for us, or do it locally with Val(q), but let's wait to see if/where that's needed.

src/common/threading.jl
```diff
@@ -21,3 +21,22 @@ function disable_threading()
     @assert n_julia == 1  # To exit in non-master MPI nodes
     setup_threading(;n_fft=1, n_blas=1)
 end
+
+# Parallelization loop breaking range into chunks.
+function parallel_loop_over_range(fun, storages::AbstractVector, range)
```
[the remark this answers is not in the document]
antoine-levitt: Yes, I wondered about that too, I don't remember if floops does something like this - in any case this refactor is relatively sane in decoupling the parallelization and the computation and will allow us to experiment with backends easily - if it turns out there's a simple macro for that we can do it later.

src/densities.jl
```diff
@@ -61,14 +50,52 @@ using an optional `occupation_threshold`. By default all occupation numbers are
...
-    ForwardDiff.derivative(zero(T)) do ε
-        ψ_ε   = [ψk   .+ ε .* δψk   for (ψk,   δψk)   in zip(ψ, δψ)]
-        occ_ε = [occk .+ ε .* δocck for (occk, δocck) in zip(occupation, δoccupation)]
-        compute_density(basis, ψ_ε, occ_ε; occupation_threshold)
+@views @timing function compute_δρ(basis::PlaneWaveBasis{T}, ψ, δψ, occupation,
+                                   δoccupation=zero.(occupation);
+                                   occupation_threshold=zero(T), q=zero(Vec3{T})) where {T}
+    S = promote_type(T, eltype(ψ[1]))
+    # δρ is expected to be real when computations are not phonon-related.
+    Tδρ = iszero(q) ? real(S) : S
```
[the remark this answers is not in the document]
antoine-levitt: Type instabilities by themselves are not necessarily bad, so I wouldn't bother unless it actually has a performance impact. That said, fixing accidental instabilities is always good, when it doesn't require convoluted workarounds.

src/postprocess/phonon.jl
```diff
@@ -5,26 +5,89 @@ function dynmat_red_to_cart(model::Model, dynamical_matrix)
...
+    phonon_modes(basis, dynmat_cart)
+end
+
+@doc raw"""
 Compute the dynamical matrix in the form of a ``3×n_{\rm atoms}×3×n_{\rm atoms}`` tensor
 in reduced coordinates.
 """
-@timing function compute_dynmat(basis::PlaneWaveBasis, ψ, occupation; kwargs...)
-    dynmats_per_term = [compute_dynmat(term, basis, ψ, occupation; kwargs...)
+@timing function compute_dynmat(basis::PlaneWaveBasis{T}, ψ, occupation; q=zero(Vec3{T}),
```
mfherbst, line 65: If we call this `compute_dynmat` than we should also call the input variable to `phonon_modes` `dynmat`

src/postprocess/phonon.jl
```diff
@@ -36,3 +99,24 @@ function compute_dynmat_cart(basis::PlaneWaveBasis, ψ, occupation; kwargs...)
...
+                   scfres.occupation_threshold, scfres.εF, scfres.eigenvalues, kwargs...)
+end
+
+function compute_dynmat_cart(scfres; kwargs...)
+    compute_dynmat_cart(scfres.basis, scfres.ψ, scfres.occupation; scfres.ρ, scfres.ham,
+                        scfres.occupation_threshold, scfres.εF, scfres.eigenvalues, kwargs...)
+end
+
+"""
+Assemble the right-hand side term for the Sternheimer equation for all relevant quantities.
```
mfherbst, line 114: All relevant quantities for what ? a bit vague since this is phonon specific, right?

mfherbst: I feel this function as well as `compute_δHψ` need clearer docstrings what the interface is, what is actually computed and returned, what ar the kwargs etc. Otherwise it is a bit magical.

src/terms/hartree.jl
```diff
@@ -26,17 +26,26 @@ struct TermHartree <: TermNonlinear
     # Fourier coefficients of the Green's function of the periodic Poisson equation
     poisson_green_coeffs::AbstractArray
 end
-function TermHartree(basis::PlaneWaveBasis{T}, scaling_factor) where {T}
+function coeffs_poisson_hartree(basis::PlaneWaveBasis{T}, scaling_factor;
```
mfherbst, line 29: `compute_poisson_green_coeffs` to make the connection to `poisson_green_coeffs` clear ?

src/common/types.jl
```diff
@@ -4,6 +4,9 @@ using StaticArrays: setindex
 # Deduce the value type (floating-point type for storing plain an static data
 # in Model and PlaneWaveBasis) from e.g. an interval or a dual type.
 value_type(T) = T
+# Lossy: force interpretation of value as a determined type.
+force_type(::Type{T}, x) where {T <: Real} = real(x)
+force_type(::Type{T}, x) where {T <: Complex} = x
```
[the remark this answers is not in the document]
mfherbst: `convert_enforced(T, stuff ...)` ? (cast is not used for this in the Julia language)

src/terms/terms.jl
```diff
@@ -60,8 +60,10 @@ breaks_symmetries(::Anyonic) = true
 
 # forces computes either nothing or an array forces[at][α] (by default no forces)
 compute_forces(::Term, ::AbstractBasis, ψ, occupation; kwargs...) = nothing
-# dynamical matrix for phonons computations (array dynmat[n_dim, n_atom, n_dim, n_atom])
+# dynamical matrix for phonons computations (array dynmat[3, n_atom, 3, n_atom])
 compute_dynmat(::Term, ::AbstractBasis, ψ, occupation; kwargs...) = nothing
+# variation of the Hamiltonian applied to orbitals for phonons computations
+compute_δHψ_sα(::Term, ::AbstractBasis, ψ, q, s, α; kwargs...) = nothing
```
mfherbst, line 66: Better name, but needs some more details what all these things mean, what is the expected input / output etc. as this function is not as natural as the other ones (at least to me) ... see also my other comments)

src/transfer.jl
```diff
@@ -203,25 +200,65 @@ function find_equivalent_kpt(basis::PlaneWaveBasis{T}, kcoord, spin; tol=sqrt(ep
...
+```math
+    u_{k+q}(G) = u_{[k+q]}(G + ΔG).
+```
+"""
+function kpq_equivalent_blochwave_to_kpq(basis, kpt, q, ψk_plus_q_equivalent)
+    kcoord_plus_q = kpt.coordinate .+ q
+    index, ΔG = find_equivalent_kpt(basis, kcoord_plus_q, kpt.spin)
+    equivalent_kpt_plus_q = basis.kpoints[index]
+
+    kpt_plus_q = build_kpoints(basis, [kcoord_plus_q])[kpt.spin]
```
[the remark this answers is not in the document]
mfherbst: Sorry I don't fully understand at this point what this does and why it's needed. Maybe you can provide some more details ?

src/transfer.jl
```diff
@@ -203,25 +200,65 @@ function find_equivalent_kpt(basis::PlaneWaveBasis{T}, kcoord, spin; tol=sqrt(ep
...
+# Replaces `multiply_by_expiqr`.
+# TODO: Think about a clear and consistent way to define such a function when completing
+# phonon PRs.
+@doc raw"""
+    multiply_ψ_by_blochwave(basis::PlaneWaveBasis, ψ, f_real, q)
+
+Return the Fourier coefficients for the Bloch waves ``f^{\rm real}_{q} ψ_{k-q}`` in an
+element of `basis.kpoints` equivalent to ``k-q``.
+"""
+function multiply_ψ_by_blochwave(basis::PlaneWaveBasis, ψ, f_real, q)
```
mfherbst, line 248: Yeah finding good names for these functions is tricky. I'm not sure we are there yet. So perhaps it's good to document extra clearly why they exist, e.g. I think would be good if you mention the nice explanation δHψ_k = δV_{q} · ψ_{k-q}. here.

src/terms/local.jl
```diff
@@ -82,78 +82,133 @@ end
...
-    T = real(eltype(ρ_fourier))
-    f = zero(Vec3{T})
-    for (iG, G) in enumerate(G_vectors(basis))
-        f -= real(conj(ρ_fourier[iG])
-                  .* form_factors[iG]
-                  .* cis2pi(-dot(G, r))
-                  .* (-2T(π)) .* G .* im
-                  ./ sqrt(basis.model.unit_cell_volume))
+# Phonon: Perturbation of the local potential with respect to a displacement on the
+# direction α of the atom s.
```
mfherbst, line 157: That's a helpful comment for example (along the lines of commenting `compute_δH\psi_sa`)

src/response/chi0.jl
```diff
@@ -291,49 +292,59 @@ Perform in-place computations of the derivatives of the wave functions by solvin
 a Sternheimer equation for each `k`-points. It is assumed the passed `δψ` are initialised
 to zero.
 """
-function compute_δψ!(δψ, basis, H, ψ, εF, ε, δHψ; ψ_extra=[zeros_like(ψk, size(ψk,1), 0) for ψk in ψ],
-                     kwargs_sternheimer...)
+function compute_δψ!(δψ, basis::PlaneWaveBasis{T}, H, ψ, εF, ε, δHψ, ε_minus_q=ε;
+                     ψ_extra=[zeros_like(ψk, size(ψk,1), 0) for ψk in ψ],
+                     q=zero(Vec3{T}), kwargs_sternheimer...) where {T}
+    # We solve the Sternheimer equation
+    #   (H_k - ε_{n,k-q}) δψ_{n,k} = - (1 - P_{k}) δHψ_{n, k-q},
```
mfherbst, line 299: What is `ε_{n,k-q}` ?  Ok, reading on I begin to understand. Maybe have a subparagraph in the general comments around line 194 making clear what is the difference between a q=0 and a q\neq 0 calculation. Otherwise this is the first thing one reads and the jargon and context has not yet been made clear.

src/response/chi0.jl
```diff
@@ -291,49 +292,59 @@ Perform in-place computations of the derivatives of the wave functions by solvin
...
+        for n = 1:length(εk_minus_q)
+            fnk_minus_q = filled_occ * Smearing.occupation(smearing, (εk_minus_q[n]-εF) / temperature)
 
             # Explicit contributions (nonzero only for temperature > 0)
             for m = 1:length(εk)
-                # the n == m contribution in compute_δρ is obtained through
-                # δoccupation, see the explanation above
-                m == n && continue
+                # The n == m contribution in compute_δρ is obtained through δoccupation, see
+                # the explanation above; except if we perform phonon calculations.
```
mfherbst, line 327: Why? What's different ?

src/response/chi0.jl
```diff
@@ -347,33 +358,41 @@ end
...
+    δHψ_minus_q_occ = [δHψ[ik][:, maskk] for (ik, maskk) in enumerate(ordering(mask_occ))]
+    # Only needed for phonon calculations.
+    ε_minus_q_occ  = ordering([eigenvalues[ik][maskk] for (ik, maskk) in enumerate(mask_occ)])
 
     # First we compute δoccupation. We only need to do this for the actually occupied
     # orbitals. So we make a fresh array padded with zeros, but only alter the elements
     # corresponding to the occupied orbitals. (Note both compute_δocc! and compute_δψ!
     # assume that the first array argument has already been initialised to zero).
+    # For phonon calculations when q ≠ 0, we do not use δoccupation, and compute directly
+    # the full perturbation δψ.
```
mfherbst, line 371: Does that mean we use a different convention for representing δρ and δψ ?

src/response/chi0.jl
```diff
@@ -291,49 +292,59 @@ Perform in-place computations of the derivatives of the wave functions by solvin
...
-function compute_δψ!(δψ, basis, H, ψ, εF, ε, δHψ; ψ_extra=[zeros_like(ψk, size(ψk,1), 0) for ψk in ψ],
-                     kwargs_sternheimer...)
+function compute_δψ!(δψ, basis::PlaneWaveBasis{T}, H, ψ, εF, ε, δHψ, ε_minus_q=ε;
+                     ψ_extra=[zeros_like(ψk, size(ψk,1), 0) for ψk in ψ],
+                     q=zero(Vec3{T}), kwargs_sternheimer...) where {T}
+    # We solve the Sternheimer equation
+    #   (H_k - ε_{n,k-q}) δψ_{n,k} = - (1 - P_{k}) δHψ_{n, k-q},
+    # where P_{k} is the projector on ψ_{k} and with the conventions:
+    # - δψ_{k} is the variation of ψ_{k-q}, which implies (for ℬ_{k} the `basis.kpoints`)
+    #     δψ_{k-q} ∈ ℬ_{k-q} and δHψ_{k-q} ∈ ℬ_{k};
```
mfherbst, line 302: What is `ℬ_{k-q}` ? Is k-q not potentially outside of the ks we actually have ? Or do you implicitly apply an operation to find the equivalent point in the irreducible BZ ?

If that's the case than I believe the language in the code is not uniform, e.g. in the comment of the `kpq_equivalent_blochwave_to_kpq` function you also have `[k-q]` or sth, which seems to mean something else ?

src/response/chi0.jl
```diff
@@ -291,49 +292,59 @@ Perform in-place computations of the derivatives of the wave functions by solvin
...
+    # - δHψ[ik] = δHψ_{k-q};
```
mfherbst, line 303: be careful with using `-` for ennumerations here. Could be misinterpreted as a sign.

mfherbst (commented):
Second batch.

src/postprocess/phonon.jl
```diff
@@ -5,26 +5,89 @@ function dynmat_red_to_cart(model::Model, dynamical_matrix)
...
+# Create a ``3×n_{\rm atoms}×3×n_{\rm atoms}`` tensor (for consistency with the format of
+# dynamical matrices) equivalent to a diagonal matrix with the atomic masses of the atoms on
+# the diagonal.
+function get_mass_matrix(basis::PlaneWaveBasis{T}) where {T}
+    model = basis.model
+    positions = model.positions
+    n_atoms = length(positions)
+
+    atoms_mass = atomic_mass.(model.atoms)
+    any(iszero.(atoms_mass)) && @error "Some elements have unknown masses"
```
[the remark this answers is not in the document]
mfherbst: Then `@warn`.

antoine-levitt:
@mfherbst we discussed with Etienne, it's basically done except for comments/docs, so if you have no objection let's merge it and then I'll take a stab at clarifying in a next PR? @epolack will make some perf tests and give us the green light when it's OK.

src/common/types.jl
```diff
@@ -4,6 +4,9 @@ using StaticArrays: setindex
 # Deduce the value type (floating-point type for storing plain an static data
 # in Model and PlaneWaveBasis) from e.g. an interval or a dual type.
 value_type(T) = T
+# Lossy: force interpretation of value as a determined type.
+convert_enforced(::Type{T}, x) where {T <: Real} = real(x)
+convert_enforced(::Type{T}, x) where {T <: Complex} = x
```
mfherbst, line 9: I still wonder about this function and what it is for: Would you not want to do a `convert(T, x)`  as well ? Maybe also stick a `convert_enforced(::Type{T}, x)::T where {T <: xyz} = ... ` to ensure that the type can be propagated by the compiler.

src/densities.jl
```diff
@@ -10,44 +10,39 @@ grid `basis`, where the individual k-points are occupied according to `occupatio
 It is possible to ask only for occupations higher than a certain level to be computed by
 using an optional `occupation_threshold`. By default all occupation numbers are considered.
 """
-@views @timing function compute_density(basis::PlaneWaveBasis{T}, ψ, occupation;
+function compute_density(basis::PlaneWaveBasis{T}, ψ, occupation;
+                         occupation_threshold=zero(T)) where {T}
+    Tρ = promote_type(T, real(eltype(ψ[1])))
+    compute_density(Tρ, basis, ψ, occupation; occupation_threshold)
```
mfherbst, line 16: If we need this for perf, ok, but otherwise I'd prefer the old version without the extra layer for simplicity.

src/external/atomsbase.jl
```diff
@@ -107,6 +107,9 @@ function AtomsBase.atomic_system(lattice::AbstractMatrix{<:Number},
             Atom(atomic_symbol(element), position; kwargs...)
         end
     end
+    # Check consistency of the atomic masses.
+    all(getfield.(atomsbase_atoms, :atomic_mass) .== atomic_mass.(atoms)) ||
```
mfherbst, line 111: This is not where the problem is, actually. What I'm concerned about is `parse_system` above, where an warning like line 56 should be issued. I can take care of this before the merge if you want.

src/fft.jl
```diff
@@ -58,14 +58,17 @@ end
...
     ifft!(similar(f_fourier, basis.fft_size...), basis, kpt, f_fourier; kwargs...)
 end
-
 """
 Perform a real valued iFFT; see [`ifft`](@ref). Note that this function
 silently drops the imaginary part.
 """
 function irfft(basis::PlaneWaveBasis{T}, f_fourier::AbstractArray) where {T}
     real(ifft(basis, f_fourier))
 end
```
mfherbst, line 67: Isn't this just `ifft(real(T), basis, f_fourier)` ?

mfherbst (commented):
Open points for me that need to be clarified before merge:
- [ ] Perf of SCF and response.
- [ ] Subpoint of the above: Is the density function barrier needed?
- [ ] Consistent warning with atomsbase structures of differeing atomic mass

Open points that can be clarified after:
- [ ] Docs of new sternheimer
- [ ] Docs of compute_δHψ_sα

src/common/types.jl
mfherbst: > Not sure I understand. convert(::Real, x::Complex) does not make sense as we lose information.

Yes, but what I meant was the following:
```suggestion
convert_enforced(::Type{T}, x)::T where {T <: Real}    = convert(T, real(x))
convert_enforced(::Type{T}, x)::T where {T <: Complex} = convert(T, x)
```

mfherbst: does this not do the trick ? 

Basically I'm not doubting the function is needed, I'm mostly trying to understand what it is supposed to do and then trying to find a name that is not misleading. `convert` can deal with vectors and arrays, but then the `T` will be an array type as well for example.

mfherbst: Maybe what we want is actually the [Adapt.jl](https://github.com/JuliaGPU/Adapt.jl/) pattern ?

src/external/atomsbase.jl
```diff
@@ -48,6 +48,10 @@ function parse_system(system::AbstractSystem{D}) where {D}
         error("Charged systems not yet supported in DFTK.")
     end
 
+    if !all([atom.atomic_mass for atom in system] .== atomic_mass.(atoms))
```
mfherbst, line 51: This should work, I believe:
```suggestion
    if atomic_mass(system) != atomic_mass.(atoms))
```

src/external/atomsbase.jl
mfherbst: Yes, that's exactly what I meant.

src/common/types.jl
mfherbst: Ok then maybe `convert_enforced` is not the right name. For me it does a `convert`, but is happy to do inexact converts, which would otherwise fail with an Inexact error. Is that the intention ?

mfherbst:
Timings are great, I agree. No need for the wrapper.

src/elements.jl
```diff
@@ -186,11 +201,12 @@ end
...
 
 struct ElementGaussian <: Element
-    α                               # Prefactor
-    L                               # Width of the Gaussian nucleus
+    α               # Prefactor
+    L               # Width of the Gaussian nucleus
     symbol::Symbol  # Element symbol
 end
 AtomsBase.atomic_symbol(el::ElementGaussian) = el.symbol
+AtomsBase.atomic_mass(::ElementGaussian) = nothing  # forces user to chose a mass
```
mfherbst, line 209: You had this for the symbol as well before and it caused more issues than it solved. Let's leave it like it is, but we might need to change this to "zero mass" instead in the future.

src/external/atomsbase.jl
```diff
@@ -19,15 +19,23 @@ function parse_system(system::AbstractSystem{D}) where {D}
     atoms = map(system) do atom
         pseudo = get(atom, :pseudopotential, "")
         if isempty(pseudo)
-            ElementCoulomb(atomic_number(atom))
+            ElementCoulomb(atomic_number(atom); mass=atomic_mass(atom))
         else
-            get!(cached_pspelements, pseudo) do
+            key = pseudo * string(get(atom, :atomic_mass, ""))
```
mfherbst, line 24: Unlike `:pseudopotential`, which is an optional AtomsBase property, all atoms are expected to have a mass, so you can just use `atomic_mass(atom)`.

src/external/atomsbase.jl
```diff
@@ -19,15 +19,23 @@ function parse_system(system::AbstractSystem{D}) where {D}
...
             end
         end
     end
 
+    dftk_atomic_masses = getfield.(element.(getfield.(atoms, :symbol)), :atomic_mass)
+    if (atomic_mass(system) != dftk_atomic_masses)
+        @warn "Some atomic masses differ from the default ones; they may come from \
+               rounding error from the external parser."
+    end
+
```
mfherbst, line 38: Don't need this any more.

src/fft.jl
mfherbst: I'd be curious why @antoine-levitt , but not blocking for the merge.

test/bzmesh_symmetry.jl
```diff
@@ -15,14 +15,14 @@
         basis = PlaneWaveBasis(model_nosym; Ecut=5, case...)
         DFTK.check_group(basis.symmetries)
 
-        scfres = self_consistent_field(basis; is_converged=DFTK.ScfConvergenceDensity(1e-10))
+        scfres = self_consistent_field(basis; tol=1e-10)
```
mfherbst, line 18: Don't remove this everywhere, because if we refactor it later, we don't see things breaking.

src/external/atomsbase.jl
mfherbst: They simply test different things (one is the annotated charge per atom, one the annotated total charge of the system ... sometimes you want one, sometimes the other ... don't ask, it's a mess). We can also have a tolerance in the second test if that puts you more at ease.

test/external/atomsbase.jl
```diff
@@ -214,7 +217,31 @@ end
...
         @test model.atoms[1].psp.identifier == "hgh/lda/c-q4.hgh"
         @test model.atoms[2].psp.identifier == "hgh/lda/c-q4.hgh"
     end
 end
+
+@testitem "AbstractSystem (atomic mass consistency)" begin
+    using DFTK: parse_system
+    using AtomsIO
+    filename = tempname() * ".cif"
+    open(filename, "w") do io
```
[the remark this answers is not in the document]
mfherbst: Yes, but why not use `mktempdir` ?

src/common/types.jl
mfherbst: Ok then why does
```
convert_enforced(::Type{T}, x) where {T <: Real}    = convert(T, real(x))::T
convert_enforced(::Type{T}, x)                      = convert(T, x)::T
```
not do everything you want ? It does a normal convert for complex numbers and other types and first strips the imaginary part when converting to a real number.

mfherbst (commented):
Final nits

mfherbst:
Yeah, but then just add another method on AbstractVector{<:Real} or does that not do the trick ?

src/external/atomsbase.jl
mfherbst: I don't think that's worth it. It will only be noise as this is far too sensitive.

antoine-levitt:
How is real_enforced better than enforce_real? Also beware that lowpass_for_symmetry is possibly slow

mfherbst:
GPU CI did not go through cleanly. Do you guys have access to a CUDA GPU or should I debug ?

mfherbst:
Ok, so I trust the GPU CI goes through. Let's see.

PR #754, opened 2022-10-19
Added reference for the energy cutoff smearing

LaurentVidal95:
I simply added the reference is the arXiv preprint https://arxiv.org/abs/2210.00442 in the example and the kinetic.jl file.
Sorry @jaemolihm we were a bit longer that planned but you can now check the preprint !


mfherbst:
@LaurentVidal95  I added some of this in the meantime. Can you harmonise all links / references to the arxiv version? Just search for your name and you should find all these places :smile:.

PR #758, opened 2022-10-21
Splitting computations of δψ and δocc

epolack:

For phonons, we have to solve
``(Hk - εpn) δψkn = - (1 - Pk) · rhs`` where ``p = k-q``.


src/response/chi0.jl
```diff
@@ -239,94 +241,119 @@ function compute_αmn(fm, fn, ratio)
     ratio * fn / (fn^2 + fm^2)
 end
 
-@views @timing function apply_χ0_4P(ham, ψ, occ, εF, eigenvalues, δHψ;
-                                    occupation_threshold, kwargs_sternheimer...)
-    basis  = ham.basis
+"""
+Compute the derivatives of the Fermi level and of the occupations.
+"""
+function compute_δocc(basis, εF::T; ψ, ε, δHψ, occ) where {T}
```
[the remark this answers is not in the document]
antoine-levitt: Name is fine, I think. Change docstring to `of the occupations (and of the Fermi level)`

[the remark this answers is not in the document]
antoine-levitt: Arguments are actual arguments, no need to make them kwargs

src/response/chi0.jl
```diff
@@ -239,94 +241,120 @@ function compute_αmn(fm, fn, ratio)
...
             D += fpnk * basis.kweights[ik]
         end
-        # compute δεF
+        # Compute δεF
         D = mpi_sum(D, basis.comm_kpts)  # equal to minus the total DOS
         δocc_tot = mpi_sum(sum(basis.kweights .* sum.(δocc)), basis.comm_kpts)
         δεF = !isnothing(model.εF) ? zero(δεF) : δocc_tot / D  # no δεF when Fermi level is fixed
-        # recompute δocc
-        for ik = 1:Nk, (n, εnk) in enumerate(ε_occ[ik])
+        # Recompute δocc
```
antoine-levitt, line 270: , taking into account deps_F

src/response/chi0.jl
```diff
@@ -239,94 +241,120 @@ function compute_αmn(fm, fn, ratio)
...
-        ψk = ψ_occ[ik]
+        Hk  = H[ik]
+        ψk  = ψ[ik]
+        εk  = ε[ik]
         δψk = δψ[ik]
-        Hψk_extra = ham.blocks[ik] * ψ_extra[ik]
 
-        εk = ε_occ[ik]
+        Hψk_extra = Hk * ψ_extra[ik]
+        εk_extra  = diag(real.(ψ_extra[ik]' * (Hk * ψ_extra[ik])))
```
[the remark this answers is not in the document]
antoine-levitt: Reuse Hpsik

antoine-levitt (commented):
Nice refactoring, thanks!

src/response/chi0.jl
```diff
@@ -239,94 +239,119 @@ function compute_αmn(fm, fn, ratio)
...
-            # explicit contributions (nonzero only for temperature > 0)
+            # Explicit contributions (nonzero only for temperature > 0)
             for m = 1:length(εk)
+                εk[m] == εk[n] && continue
                 fmk = filled_occ * Smearing.occupation(smearing, (εk[m]-εF) / temperature)
                 ddiff = Smearing.occupation_divided_difference
                 ratio = filled_occ * ddiff(smearing, εk[m], εk[n], εF, temperature)
                 αmn = compute_αmn(fmk, fnk, ratio)  # fnk * αmn + fmk * αnm = ratio
-                δψk[:, n] .+= ψk[:, m] .* αmn .* (dot(ψk[:, m], δHψ[ik][:, n]) * (n != m))
+                δψk[:, n] .+= ψk[:, m] .* αmn .* dot(ψk[:, m], δHψ[ik][:, n])
```
antoine-levitt, line 312: why?

antoine-levitt: @gkemlin I don't remember why this was, but this commit still passes tests, so possibly our test suite is too coarse?

gkemlin: Hum, that's weird. On master, if I remove `n != m`, the `chi0.jl` test fails as expected. This is because the `n==m` term is taken into account when computing `δρ` through `δoccupation`. Are you sure you didn't forget to specifiy `δoccupation` in `compute_δρ` later on ?

antoine-levitt: OK great, j'ai rien dit!

gkemlin: Can't we have `εk[m] == εk[n]` even though `m!=n`. In theory yes, but in practice hopefully not because of the tolerance on eigensolvers, but I'm worried that this would fail with bad luck ? Also, the condition `m != n` seems more explicit to me.

antoine-levitt: Ah yes! I merged too early... Will make a followup commit

mfherbst:
I see about the failing test, but why did you have to use Applyχ0Model. We don't usually use that because it's pretty expensive.

src/response/chi0.jl
```diff
@@ -146,31 +145,32 @@ function sternheimer_solver(Hk, ψk, εnk, rhs, n;
...
         Rϕ = R(ϕ)
-        # Schur complement of (1-P) (H-εn) (1-P)
+        # Schur complement of (1-P) (H-ε) (1-P)
         # with the splitting Ran(1-P) = Ran(P_extra) ⊕ Ran(R)
         R(H(Rϕ) - Hψk_extra * (ψk_exHψk_ex \ Hψk_extra'Rϕ))
     end
     precon = PreconditionerTPA(basis, kpoint)
-    precondprep!(precon, ψk[:, n])
+    # First column of ψk as there is no natural kinetic energy.
+    precondprep!(precon, ψk[:, 1])
```
mfherbst, line 173: Wouldn't it be more reasonalbe to use the last column where the KE is larger?

mfherbst:
I no it's not used by default. The LdosMixing is. Perhaps we should lower the gradient check tolerance than instead.

src/response/chi0.jl
antoine-levitt: Not sure, I think both are reasonable... The preconditioner is not super sensitive to that anyway. We could just as well set it to 1, but here at least we have more chance to give something sensible in the "weird" cases (scaled kinetic energy, super compressed system...)

PR #760, opened 2022-10-21
Transfer phonon

epolack:

Modification to remove transfer function as argument, and to take care of spins.

### DIFF
--- a/src/transfer.jl
+++ b/src/transfer.jl
@@ -140,7 +140,7 @@ function find_equivalent_kpt(basis::PlaneWaveBasis{T}, kcoord, spin; tol=sqrt(ep
     indices_σ = krange_spin(basis, spin)
     kcoords_σ = getfield.(basis.kpoints[indices_σ], :coordinate)
     # Unique by construction.
-    index::Int = findfirst(isapprox.(kcoords_σ, Ref(kcoord_red); atol=tol)) + indices_σ[1]-1
+    index::Int = findfirst(isapprox(kcoord_red; atol=tol), kcoords_σ) + (indices_σ[1] - 1)
 
     return (; index, ΔG)
 end
--- a/src/transfer.jl
+++ b/src/transfer.jl
@@ -97,38 +97,6 @@ function transfer_blochwave_kpt(ψk_in, basis::PlaneWaveBasis, kpt_in, kpt_out, 
 end
 
 """
-Find the equivalent index of the coordinate `kcoord` ∈ ℝ³ in a list `kcoords` ∈ [-½, ½)³.
-`ΔG` is the vector of ℤ³ such that `kcoords[index] = kcoord + ΔG`.
-"""
-function find_equivalent_kpt(kcoords::Vector{Vec3{T}}, kcoord; tol=100*eps(T)) where {T}
-    kcoord_normalized = mod.(Vector(kcoord), 1)  # coordinate in [0, 1)³
-    kcoord_normalized[kcoord_normalized .≥ 0.5  - tol] .-= 1  # coordinate in [-½, ½)
-
-    indices = findall(e -> is_approx_integer(e - kcoord_normalized; tol), kcoords)
-    index = only(indices)
-
-    ΔG = kcoord_normalized - kcoord
-
-    # ΔG should be an integer.
-    @assert all(is_approx_integer.(ΔG; tol))
-    ΔG = round.(Int, ΔG)
-
-    return (; index, ΔG)
-end
-
-"""
-Return the Fourier coefficients for `ψk · e^{i q·r}` in the basis of `kpt_out`, where `ψk`
-is defined on a basis `kpt_in`.
-"""
-function multiply_by_expiqr(basis, kpt_in, q, ψk; transfer_fn=transfer_blochwave_kpt)
-    kcoords = getfield.(basis.kpoints, :coordinate)
-    shifted_kcoord = kpt_in.coordinate .+ q  # coordinate of ``k``-point in ℝ
-    (index, ΔG) = find_equivalent_kpt(kcoords, shifted_kcoord)
-    kpt_out = basis.kpoints[index]
-    return transfer_fn(ψk, basis, kpt_in, kpt_out, ΔG)
-end
-
-"""
 Transfer Bloch wave between two basis sets. Limited feature set.
 """
 function transfer_blochwave(ψ_in, basis_in::PlaneWaveBasis{T},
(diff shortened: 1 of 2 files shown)


src/transfer.jl
```diff
@@ -120,36 +146,29 @@ end
 Return the Fourier coefficients for `ψk · e^{i q·r}` in the basis of `kpt_out`, where `ψk`
 is defined on a basis `kpt_in`.
 """
-function multiply_by_expiqr(basis, kpt_in, q, ψk; transfer_fn=transfer_blochwave_kpt)
-    kcoords = getfield.(basis.kpoints, :coordinate)
+function multiply_by_expiqr(basis, kpt_in, q, ψk)
+    indices_kpints_σ = findall(e -> e.:spin == kpt_in.spin, basis.kpoints)
+    kcoords = getfield.(basis.kpoints, :coordinate)[indices_kpints_σ]
```
antoine-levitt, line 151: typo

src/transfer.jl
```diff
@@ -120,36 +146,29 @@ end
 Return the Fourier coefficients for `ψk · e^{i q·r}` in the basis of `kpt_out`, where `ψk`
 is defined on a basis `kpt_in`.
 """
-function multiply_by_expiqr(basis, kpt_in, q, ψk; transfer_fn=transfer_blochwave_kpt)
-    kcoords = getfield.(basis.kpoints, :coordinate)
+function multiply_by_expiqr(basis, kpt_in, q, ψk)
+    indices_kpints_σ = findall(e -> e.:spin == kpt_in.spin, basis.kpoints)
```
antoine-levitt, line 150: why the :spin? Just spin?

antoine-levitt: also use k instead of e?

src/transfer.jl
antoine-levitt: `[kpt.coordinate for kpt in kpoints if kpt.spin == kpt_in.spin]`?

src/transfer.jl
```diff
@@ -120,36 +146,29 @@ end
...
+    kcoords = getfield.(basis.kpoints, :coordinate)[indices_kpints_σ]
     shifted_kcoord = kpt_in.coordinate .+ q  # coordinate of ``k``-point in ℝ
-    (index, ΔG) = find_equivalent_kpt(kcoords, shifted_kcoord)
-    kpt_out = basis.kpoints[index]
-    return transfer_fn(ψk, basis, kpt_in, kpt_out, ΔG)
+    index, ΔG = find_equivalent_kpt(kcoords, shifted_kcoord)
```
antoine-levitt, line 153: have this function take kpoints, shifted_kcoord and sigma (both kcoords and sigma are needed to identify kpoints)

src/transfer.jl
```diff
@@ -120,36 +146,29 @@ end
...
+    kpt_out = basis.kpoints[indices_kpints_σ[index]]
+    return transfer_blochwave_kpt(ψk, basis, kpt_in, kpt_out, ΔG)
 end
 
 """
-Transfer Bloch wave between two basis sets. Limited feature set.
+Return an `ordering` function that reorders the `kpoints`. That is for each `kpoint` of the
+`basis`: `kpoints[ik].coordinate + q = ordering(kpoints)[ik].coordinate`.
```
antoine-levitt, line 160: is it really necessary to cache that information (rather than recompute it on the fly)?

antoine-levitt: that logic is a bit convoluted (it's basically a two-stage function)

antoine-levitt: What I meant was that basically it looks like f(x) = x -> x+y, which then you call as fx = f(x); fx(y) but you could just define f(x,y) = x+y and call it like that?

antoine-levitt: Come on, there aren't that many kpoints, it shouldn't be that costly, certainly much less than solving the response equations. Most likely your find_equivalent_kpt function can be sped up significantly.

src/transfer.jl
```diff
@@ -153,3 +121,48 @@ function transfer_blochwave(ψ_in, basis_in::PlaneWaveBasis{T},
...
+
+    # ΔG should be an integer.
+    @assert all(is_approx_integer.(ΔG_approx; tol))
+    ΔG = round.(Int, ΔG_approx)
+
+    indices_σ = findall(kpt -> kpt.spin == spin, kpoints)
+    kpoints_σ = @view kpoints[indices_σ]
+    index_σ = only(findall(kpt -> isapprox(kpt.coordinate, kcoord_normalized; atol=tol),
+                           kpoints_σ))
+    return (; index=indices_σ[index_σ], ΔG)
```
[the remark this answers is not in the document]
antoine-levitt: You can also do a findfirst, kpoint coordinates should of course be unique (you can check that in pwbasis constructor if you're concerned)

src/transfer.jl
```diff
@@ -153,3 +121,47 @@ function transfer_blochwave(ψ_in, basis_in::PlaneWaveBasis{T},
...
+                     k ≥ 0.5 - tol ? k - 1 : k  # coordinate in [-½, ½)³
+                 end
+
+    ΔG = kcoord_red - kcoord
+    # ΔG should be an integer.
+    @assert all(is_approx_integer.(ΔG))
+    ΔG = round.(Int, ΔG)
+
+    index::Int = findfirst(k -> k.spin == spin && isapprox(k.coordinate, kcoord_red; atol=tol),
+                           kpoints)  # unique by construction
```
mfherbst, line 141: Use krange_spin

test/phonon.jl
```diff
@@ -154,15 +154,29 @@ end
...
-                                              transfer_fn=transfer_blochwave_kpt_real)
-        @testset "Testing kpoint $(kpt.coordinate) on kgrid $kgrid" begin
-            @test norm(ψk_out_four - ψk_out_real) < tolerance
+    q = Vec3(q0 .* ishift)
+    @testset "Transfer function" begin
+        for kpt in unique(rand(basis.kpoints, 4))
+            ψk = fft(basis, kpt, ψ)
+
+            ψk_out_four = DFTK.multiply_by_expiqr(basis, kpt, q, ψk)
+            ψk_out_real = begin
```
mfherbst, line 163: I think you want `let`

src/transfer.jl
mfherbst: But on that note you can always define a lambda ;)

test/phonon.jl
mfherbst: let creates a new scope. Begin doesn't.

src/transfer.jl
```diff
@@ -153,3 +121,49 @@ function transfer_blochwave(ψ_in, basis_in::PlaneWaveBasis{T},
...
+
+    ΔG = kcoord_red - kcoord
+    # ΔG should be an integer.
+    @assert all(is_approx_integer.(ΔG))
+    ΔG = round.(Int, ΔG)
+
+    indices_σ = krange_spin(basis, spin)
+    kcoords_σ = getfield.(basis.kpoints[indices_σ], :coordinate)
+    # Unique by construction.
+    index::Int = findfirst(isapprox.(kcoords_σ, Ref(kcoord_red); atol=tol)) + indices_σ[1]-1
```
mfherbst, line 143:
```suggestion
    index::Int = findfirst(isapprox(kcoord_red; atol=tol), kcoords_σ) + indices_σ[1] - 1)
```

src/transfer.jl
```diff
@@ -153,3 +121,49 @@ function transfer_blochwave(ψ_in, basis_in::PlaneWaveBasis{T},
         transfer_blochwave_kpt(ψ_in[ik], basis_in, basis_in.kpoints[ik], basis_out, kpt_out)
     end
 end
+
+"""
+Find the equivalent index of the coordinate `kcoord` ∈ ℝ³ in a list `kcoords` ∈ [-½, ½)³.
+`ΔG` is the vector of ℤ³ such that `kcoords[index] = kcoord + ΔG`.
+"""
+function find_equivalent_kpt(basis::PlaneWaveBasis{T}, kcoord, spin; tol=sqrt(eps(T))) where {T}
```
mfherbst, line 129: Coverage says this function is untested. Can you check you did not mess up the tests accidentially.

src/transfer.jl
```diff
@@ -153,3 +121,49 @@ function transfer_blochwave(ψ_in, basis_in::PlaneWaveBasis{T},
...
+"""
+function multiply_by_expiqr(basis, kpt_in, q, ψk)
+    shifted_kcoord = kpt_in.coordinate .+ q  # coordinate of ``k``-point in ℝ
+    index, ΔG = find_equivalent_kpt(basis, shifted_kcoord, kpt_in.spin)
+    kpt_out = basis.kpoints[index]
+    return transfer_blochwave_kpt(ψk, basis, kpt_in, kpt_out, ΔG)
+end
+
+"""
+Return an `ordering` function that reorders the `kpoints`. That is for each `kpoint` of the
```
antoine-levitt, line 160: update docstring

src/transfer.jl
```diff
@@ -153,3 +121,49 @@ function transfer_blochwave(ψ_in, basis_in::PlaneWaveBasis{T},
...
+`basis`: `kpoints[ik].coordinate + q = ordering(kpoints)[ik].coordinate`.
+"""
+function kpoints_ordering(basis::PlaneWaveBasis, qcoord)
```
antoine-levitt, line 163: I would call it k_to_kpq_mapping maybe?? We used "mapping" for this kind of transfer array in other parts of the code.

antoine-levitt: Otherwise yes I like the array much better, thanks!

PR #762, opened 2022-10-21
Computation of the Fourier transform of the force constant matrix of Ewald

epolack:

Done, unless there is a way compute the reciprocal part of the Hessian with AD.


src/terms/ewald.jl
```diff
@@ -116,41 +129,128 @@ function energy_ewald(lattice, recip_lattice, charges, positions; η=nothing, fo
...
         forces_recip .*= 4T(π) / compute_unit_cell_volume(lattice)
     end
 
     #
     # Real-space sum
     #
+    S = isnothing(ph_disp) ? T : promote_type(complex(T), eltype(ph_disp[1]))
     # Initialize real-space sum with correction term for uniform background
-    sum_real::T = -2η / sqrt(T(π)) * sum(Z -> Z^2, charges)
+    sum_real::S = -2η / sqrt(S(π)) * sum(Z -> Z^2, charges)
```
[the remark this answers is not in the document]
antoine-levitt: Hm, OK, fair point

src/terms/ewald.jl
```diff
@@ -33,31 +33,39 @@ function energy_ewald(model::Model{T}; kwargs...) where {T}
     energy_ewald(model.lattice, charges, model.positions; kwargs...)
 end
 
+# We balance the Ewald spling of ``1/r ≡ erf(η·r)/r + erfc(η·r)/r`` between reciprocal
```
antoine-levitt, line 36: splitting

antoine-levitt: Split that comment in two: first introduce the ewald splitting formula before the function definition (to explain what the function does), then explain the choice before the last line of the function (to explain how it does it)

src/terms/ewald.jl
```diff
@@ -33,31 +33,39 @@ function energy_ewald(model::Model{T}; kwargs...) where {T}
...
+
 # This could be merged with Pairwise, but its use of `atom_types` would slow down this
 # computationally intensive Ewald sums. So we leave it as it for now.
 """
-Compute the electrostatic interaction energy per unit cell between point charges in
-a uniform background of compensating charge to yield net neutrality.`lattice` should
-contain the lattice vectors as columns. `charges` and `positions` are the point charges and
-their positions (as an array of arrays) in fractional coordinates. If `forces` is not
-nothing, minus the derivatives of the energy with respect to `positions` is computed.
+Compute the electrostatic interaction energy per unit cell between point charges at
```
antoine-levitt, line 47: This doc has to be changed. In the phonon case, the energy per unit cell does not make sense, right? Similarly, the forces are the forces on the atoms of the first unit cell?

antoine-levitt: Ah, we converged on a comment in pairwise.jl, so just copy paste it here?

src/terms/ewald.jl
```diff
@@ -140,13 +156,93 @@ function energy_ewald(lattice::AbstractArray{T}, charges, positions;
...
+        dsum += conj(dsum)
+        dynmat_recip -= 0.5 * exp(-Gsq / 4η^2) / Gsq * dsum
+    end
+
+    # Amend `dynmat_recip` by proper scaling factors:
+    dynmat_recip *= 4T(π) / compute_unit_cell_volume(lattice)
+end
+
+# Computes the Fourier transform of the force constant matrix of the Ewald term.
+function dynmat_ewald(model::Model{T}; η=default_η(model.lattice), q=zero(Vec3{T})) where {T}
```
antoine-levitt, line 219: Why do you need this? Don't you have it in the other PR?

src/terms/ewald.jl
```diff
@@ -33,35 +33,52 @@ function energy_ewald(model::Model{T}; kwargs...) where {T}
...
+`q` is the phonon `q`-point (`Vec3`), and `ph_disp` a list of `Vec3` displacements to
+compute the Fourier transform of the force constant matrix.
 
 For now this function returns zero energy and force on non-3D systems. Use a pairwise
 potential term if you want to customise this treatment.
 """
-function energy_ewald(lattice::AbstractArray{T}, charges, positions;
-                      η=nothing, forces=nothing) where {T}
+function energy_ewald(lattice::AbstractArray{T}, charges, positions; η=default_η(lattice),
+                      forces=nothing, q=nothing, ph_disp=nothing) where {T}
```
antoine-levitt, line 67: I'm not sure why we have this API of having the forces being modified... This is very old code from when we did not know julia very well. Can you change the API (here and in pairwise) to just accept a flag, and return the forces if the flag is on? (possibly in a separate PR if that's more convenient)

src/terms/ewald.jl
```diff
@@ -33,35 +33,52 @@ function energy_ewald(model::Model{T}; kwargs...) where {T}
...
     any(iszero.(eachcol(lattice))) && return zero(T)
 
     recip_lattice = compute_recip_lattice(lattice)
     @assert length(charges) == length(positions)
-    if isnothing(η)
-        # Balance between reciprocal summation and real-space summation
-        # with a slight bias towards reciprocal summation
-        η = sqrt(sqrt(T(1.69) * norm(recip_lattice ./ 2T(π)) / norm(lattice))) / 2
+    if !isnothing(ph_disp)
+        @assert !isnothing(q) && !isnothing(forces)
```
antoine-levitt, line 76: it should be possible to have ph_disp non-nothing but not compute the forces (in which case we just compute the local energy)

src/terms/ewald.jl
```diff
@@ -126,27 +144,119 @@ function energy_ewald(lattice::AbstractArray{T}, charges, positions;
...
                 forces_real[i] -= dE_dti
-                forces_real[j] += dE_dti
             end
         end
     end
     energy = (sum_recip + sum_real) / 2  # Divide by 2 (because of double counting)
     if !isnothing(forces)
-        forces .= (forces_recip .+ forces_real) ./ 2
+        forces .= forces_real
+        if isnothing(ph_disp)
```
antoine-levitt, line 177: this logic is just weird?

src/terms/ewald.jl
```diff
@@ -126,27 +144,119 @@ function energy_ewald(lattice::AbstractArray{T}, charges, positions;
...
+            forces .+= forces_recip ./ 2
+        end
     end
     energy
 end
+
+# TODO: See if there is a way to express this with AD.
```
antoine-levitt, line 184: We should discuss at a blackboard

src/terms/ewald.jl
```diff
@@ -126,27 +144,119 @@ function energy_ewald(lattice::AbstractArray{T}, charges, positions;
...
+                hcat(Array.(forces)...)
+            end
+
+            dynmat[:, :, γ, τ] = real_part[1:n_dim, :]
+        end
+    end
+    # Reciprocal part
+    for τ in 1:n_atoms
+        for σ in 1:n_atoms
+            # TODO: why minus?
```
antoine-levitt, line 253: again let's discuss

antoine-levitt (commented):
Very convoluted logic in places. Let's discuss at a blackboard how to compute the forces in reciprocal (in the same way as the real sum)

src/terms/ewald.jl
```diff
@@ -116,23 +136,106 @@ function energy_forces_ewald(lattice::AbstractArray{T}, charges, positions;
...
+    # Reciprocal part
+    for τ in 1:n_atoms
+        for σ in 1:n_atoms
+            dynmat[:, σ, :, τ] += dynmat_ewald_recip(model, σ, τ; η, q)
+        end
+    end
+    dynmat
+end
+
+function compute_dynmat(::TermEwald, scfres::NamedTuple; kwargs...)
```
mfherbst, line 239: Not sure I like the fact, that the `scfres` appears here. For the other computational tasks (Hamiltonian, force) we "unfold" the scfres before passing it down to the terms. I'd suggest we do the same here.

src/terms/ewald.jl
```diff
@@ -43,19 +61,20 @@ For now this function returns zero energy and force on non-3D systems. Use a pai
 potential term if you want to customise this treatment.
 """
 function energy_forces_ewald(lattice::AbstractArray{T}, charges, positions;
-                             η=nothing) where {T}
+                             η=default_η(lattice), q=zero(Vec3{T}), ph_disp=nothing) where {T}
+    S = isnothing(ph_disp) ? T : promote_type(complex(T), eltype(ph_disp[1]))
```
mfherbst, line 65: This feels very hacky to me. Can't we solve this by a function barrier ? I.e. have one function `energy_forces_ewald(lattice, charges, positions, q, ph_disp)` which then calls a function `energy_forces_ewald(T, lattice::AbstractVector{T1}, charges, positions, q, ph_disp)` with `T = promote_type(complex(T1), eltype(ph_disp[1]))` as well as a function  `energy_forces_ewald(T, lattice::AbstractVector{T1}, charges, positions)`  which calls `energy_forces_ewald(T1, lattice::AbstractVector{T1}, charges, positions, zeros(T), nothing)`

test/phonon.jl
```diff
@@ -3,10 +3,11 @@ using Test
 using DFTK
 using DFTK: energy_forces_pairwise
 using LinearAlgebra
-using ForwardDiff
-using StaticArrays
 using Random
```
mfherbst, line 7: I would propose to have this in a file `phonon/main.jl` instead.

test/phonon/ewald.jl
```diff
@@ -0,0 +1,90 @@
...
+    model = Model(silicon.lattice, silicon.atoms, silicon.positions)
+
+    ω_uc = []
+    for q in phonon.qpoints
+        hessian = dynmat_to_cart((; model), DFTK.dynmat_ewald(model; q))
+        push!(ω_uc, compute_ω²(hessian))
+    end
+    ω_uc = sort!(collect(Iterators.flatten(ω_uc)))
+
+    if REFERENCE_PH_CALC
```
mfherbst, line 41: This is convoluted and going to lead to problems down the road. Better make this two `@testset`s where one needs a special argument in the testargs to be activated (maybe reuse the `:FAST_TESTS` thing, see e.g. `silicon_pbe`).

test/phonon/helpers.jl
```diff
@@ -0,0 +1,162 @@
...
+    end |> vec
+
+    (; supercell_size, qpoints)
+end
+
+const REFERENCE_PH_CALC = "REFERENCE PH_CALC" in keys(ENV)
+phonon = (
+    supercell_size = REFERENCE_PH_CALC ? generate_random_supercell() : [2, 1, 3],
+)
+phonon = merge(phonon, (; generate_supercell_qpoints(; phonon.supercell_size).qpoints))
```
mfherbst, line 162: Again solve this in the file above by having two test cases.

test/phonon/helpers.jl
```diff
@@ -0,0 +1,162 @@
...
+    # Thus the transformation between reduced and Cartesian coordinates is not a comatrix.
+    # To transform `dynamical_matrix` from reduced coordinates to `cart_mat` in Cartesian
+    # coordinates, we write
+    #   dr_cart = lattice · dr_red,
+    #   ⇒ dr_redᵀ · D_red · dr_red = dr_cartᵀ · lattice⁻ᵀ · D_red · lattice⁻¹ · dr_cart
+    #                              = dr_cartᵀ · D_cart · dr_cart
+    #   ⇒ D_cart = lattice⁻ᵀ · D_red · lattice⁻¹.
+    for τ in 1:n_atoms
+        for η in 1:n_atoms
+            cart_mat[:, η, :, τ] = inv_lattice' * dynamical_matrix[:, η, :, τ] * inv_lattice
```
mfherbst, line 126: Don't we have a routine for these kind of transformatons ?

test/phonon.jl
mfherbst: Or instead have a `phonon/runtests.jl` which has a similar stucture than the main runtests, e.g. everything is encapsulated by `@testset`s or so ... this also makes sure that things like `unpack` don't bleed out and potentially pollute other tests.

test/phonon/helpers.jl
```diff
@@ -0,0 +1,162 @@
+using Random
+using ForwardDiff
+using LinearAlgebra
+
+# Helpers functions for tests.
+# TODO: Temporary, explanations too scarce. To be changed with proper phonon computations.
+
+# Convert back and forth between Vec3 and columnwise matrix
+fold(x)   = hcat(x...)
+unfold(x) = Vec3.(eachcol(x))
```
mfherbst, line 10: Too generic names. This will lead to name clashes at some point. Maybe put all the phonon test helpers in a submodule or wrap all phonon tests in a common `@testset` ?

mfherbst (commented):
A few smaller nits.

mfherbst:
Needs #876 merged.

PR #770, opened 2022-10-24
More systematic use of named tuples

epolack:
Feel free to discard is if it's too much of a hassle.


examples/error_estimates_forces.jl
```diff
@@ -53,7 +53,7 @@ tol = 1e-8;
 
 # We compute the reference solution ``P_*`` from which we will compute the
 # references forces.
-scfres_ref = self_consistent_field(basis_ref; tol, callback=info->nothing)
+scfres_ref = self_consistent_field(basis_ref; tol, callback=identity)
```
antoine-levitt, line 56: not the same

examples/error_estimates_forces.jl
```diff
@@ -67,8 +67,8 @@ scfres_ref = self_consistent_field(basis_ref; tol, callback=info->nothing)
 #     reference solution is not converged and `Ecut = 15` is such that the
 #     asymptotic regime (crucial to validate the approach) is barely established.
 Ecut = 15
-basis = PlaneWaveBasis(model; Ecut=Ecut, kgrid)
-scfres = self_consistent_field(basis; tol, callback=info->nothing)
+basis = PlaneWaveBasis(model; Ecut, kgrid)
+scfres = self_consistent_field(basis; tol, callback=identity)
```
antoine-levitt, line 71: same as above

antoine-levitt: and a bunch more below

antoine-levitt:
Looks good, thanks. @mfherbst typically on this kind of things I'd be OK with a code formatter

antoine-levitt:
The first error is
The size of the plane wave basis is 230, and you are asking for 4611686018427387904 eigenvalues. Increase Ecut.
which is... concerning. 
```
julia> bitstring(4611686018427387904)
"0100000000000000000000000000000000000000000000000000000000000000"
```
but not sure how to interpret this... uninitialized memory? @mfherbst perhaps something to do with your new band number adaptation?

examples/error_estimates_forces.jl
antoine-levitt: Nothing in particular, but we might want to extend the API in the future (eg so that the callback can request an early stopping)

mfherbst:
Oh I hope not ...

mfherbst:
> typically on this kind of things I'd be OK with a code formatter

Yeah. The annoying thing only is that the tools I know for Julia are not opt in or so fine-grained configurable as the equivalent tools in other languages.

examples/error_estimates_forces.jl
mfherbst: Actually we moved from `info->nothing` to `identity` some time ago to allow chaining of callbacks. Therefore the change is actually reasonable.

src/postprocess/stresses.jl
```diff
@@ -21,8 +21,8 @@ Compute the stresses (= 1/Vol dE/d(M*lattice), taken at M=I) of an obtained SCF
                                    basis.kgrid, basis.kshift, basis.symmetries_respect_rgrid,
                                    basis.comm_kpts)
         ρ = DFTK.compute_density(new_basis, scfres.ψ, scfres.occupation)
-        energies, _ = energy_hamiltonian(new_basis, scfres.ψ, scfres.occupation;
-                                         ρ, scfres.eigenvalues, scfres.εF)
+        energies = energy_hamiltonian(new_basis, scfres.ψ, scfres.occupation;
+                                      ρ, scfres.eigenvalues, scfres.εF).E
```
mfherbst, line 25: This I find confusing. We could rename `E` in `energy_hamiltonian` to `energies` though.

src/scf/potential_mixing.jl
```diff
@@ -340,11 +339,9 @@ trial_damping(damping::FixedDamping, args...) = damping.α
...
 
     ham  = hamiltonian_with_total_potential(ham, info.Vout)
-    info = (ham=ham, basis=basis, energies=info.energies, converged=converged,
-            ρ=info.ρout, eigenvalues=info.eigenvalues, occupation=info.occupation,
-            εF=info.εF, n_iter=n_iter, ψ=info.ψ, info.n_bands_converge,
-            diagonalization=info.diagonalization, stage=:finalize, algorithm="SCF",
-            occupation_threshold=info.occupation_threshold)
+    info = (; ham, basis, info.energies, converged, ρ=info.ρout, info.eigenvalues,
+            info.occupation, info.εF, n_iter, info.ψ, info.n_bands_converge,
+            info.diagonalization, stage=:finalize, algorithm="SCF", info.occupation_threshold)
```
mfherbst, line 344: a bit long

src/terms/Hamiltonian.jl
```diff
@@ -191,11 +191,10 @@ end
     H = Hamiltonian(basis, [HamiltonianBlock(basis, kpt, hks)
                             for (hks, kpt) in zip(hks_per_k, basis.kpoints)])
     E = Energies(basis.model.term_types, energies)
-    (E=E, H=H)
+    (; E, H)
```
mfherbst, line 194: I really wonder if `energies, ham` is not better here for consistency ...

mfherbst: Also to distinguish from the `(E, ops)` return of term instantiation

src/workarounds/forwarddiff_rules.jl
```diff
@@ -208,9 +208,9 @@ function self_consistent_field(basis_dual::PlaneWaveBasis{T};
         ρ_dual = DFTK.compute_density(basis_dual, ψ_dual, occupation_dual)
         εF_dual = T(scfres.εF)  # Only needed for entropy term
         eigenvalues_dual = [T.(εk) for εk in scfres.eigenvalues]
-        _, ham_dual = energy_hamiltonian(basis_dual, ψ_dual, occupation_dual;
-                                         ρ=ρ_dual, eigenvalues=eigenvalues_dual, 
-                                         εF=εF_dual)
+        ham_dual = energy_hamiltonian(basis_dual, ψ_dual, occupation_dual;
+                                      ρ=ρ_dual, eigenvalues=eigenvalues_dual,
+                                      εF=εF_dual).H
```
mfherbst, line 213: Hmm. What I dislike about this rewrite is that I have to go to the very end (and almost miss) that this is only taking the `H` part. We can solve this in Julia 1.7 by using the `(; H) = ` unpacking, but for now maybe we should make an exception for this type of unpacking?

mfherbst (commented):
Thanks for doing this @epolack

src/workarounds/forwarddiff_rules.jl
mfherbst: no you are right. You cannot rename things in that way. So in this case we would need the `.ham` prefix as well.

Ok. To be honest I don't know. The prefix is not so great, but `_, ham = ` is not awesome either. My tendency is to leave it as it was, but I'm happy to be overruled here.

mfherbst:
@antoine-levitt Are you ok with leaving the `_, ham = energy_hamiltonian( ... )`-type unpacking unchanged or any preference towards moving to `ham = energy_hamiltonian( ... ).ham`?

antoine-levitt:
The .ham version is a bit nicer, but I don't really mind either way

mfherbst:
Ok, so then it stays as @epolack has made it now.

PR #771, opened 2022-10-25
update comments and dH->dV

gkemlin:
Following https://github.com/JuliaMolSim/DFTK.jl/pull/758 and https://github.com/JuliaMolSim/DFTK.jl/commit/c799f8a11c241b3585d7f8e1b5e7d38e9f53d7d6 I rewrote the comment describing `apply_chi0_4P` which was out of date, with the new splitting docc / dpsi by Etienne and the reference to our paper. Also, I uniformed dV and dH into the `chi0.jl` file, which was a bit confusing in my opinion.

### DIFF
--- a/src/response/chi0.jl
+++ b/src/response/chi0.jl
@@ -192,16 +192,16 @@ function sternheimer_solver(Hk, ψk, ε, rhs;
 end
 
 # Apply the four-point polarizability operator χ0_4P = -Ω^-1
-# Returns (δψ, δocc, δεF) corresponding to a change in *total* Hamiltonian δV
+# Returns (δψ, δocc, δεF) corresponding to a change in *total* Hamiltonian δH
 # We start from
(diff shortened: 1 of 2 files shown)


antoine-levitt:
I kind of liked dH for a generic hamiltonian perturbation... And then dV for the specific perturbation of a local potential, eg for the Dyson equation. So it's like (H+dH) (psi+dpsi) = (eps+deps) (psi+dpsi), which is pleasantly symmetric. Why do you like dV always?

gkemlin:
I don't have preferences, it's just that is seemed weird to have both in the file :sweat_smile:

mfherbst:
dV always is weird, I agree. Either we go for `dV` and `dH` and then document the difference or just `dH` always, I'd say. Between the last two I don't have a strong preference.

gkemlin:
Ok so I kept it like it was before, that make sense actually, my bad :smile:

mfherbst:
lgtm

PR #772, opened 2022-10-25
Computing density with occupation threshold

epolack:
I saw the `todo` line. Hope this is usefull.

### DIFF
--- a/src/densities.jl
+++ b/src/densities.jl
@@ -4,9 +4,9 @@
 function _check_positive(ρ)
     minimum(ρ) < 0 && @warn("Negative ρ detected", min_ρ=minimum(ρ))
 end
-function _check_total_charge(dvol, ρ::AbstractArray{T}, N) where {T}
+function _check_total_charge(dvol, ρ::AbstractArray{T}, N; tol=T(1e-10)) where {T}
     n_electrons = sum(ρ) * dvol
-    if abs(n_electrons - N) > max(sqrt(eps(T)), T(1e-10))
+    if abs(n_electrons - N) > max(sqrt(eps(T)), tol)
         @warn("Mismatch in number of electrons", sum_ρ=n_electrons, N=N)
     end
 end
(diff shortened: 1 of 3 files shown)


src/densities.jl
```diff
@@ -50,18 +52,21 @@ grid `basis`, where the individual k-points are occupied according to `occupatio
...
     ρ
 end
 
 # Variation in density corresponding to a variation in the orbitals and occupations.
+# The initial value of `occupation_threshold` corresponds to the one for which the δψkn are
+# converged by default.
 @views @timing function compute_δρ(basis::PlaneWaveBasis{T}, ψ, δψ,
-                                   occupation, δoccupation=zero.(occupation)) where {T}
+                                   occupation, δoccupation=zero.(occupation);
+                                   occupation_threshold=default_occupation_threshold()) where {T}
```
mfherbst, line 65: Why not also `zero(T)` here. That was what we had before and it did not harm.

src/densities.jl
```diff
@@ -50,18 +52,21 @@ grid `basis`, where the individual k-points are occupied according to `occupatio
 
     _check_positive(ρ)
     n_elec_check = weighted_ksum(basis, sum.(occupation))
-    _check_total_charge(basis.dvol, ρ, n_elec_check)
+    _check_total_charge(basis.dvol, ρ, n_elec_check; tol=occupation_threshold)
```
mfherbst, line 55: Just out of interest: Is this needed? I mean how large is the effect of using the `occupation_threshold` at all?

mfherbst (commented):
THanks!

src/densities.jl
mfherbst: Actually, looking at the output you have now I would revert my opinion, better to use the `occupation_threshold` for the `tol` and avoid the (spurious) warnings. If a user selects a too high occupation threshold than they cannot expect the density to be correct to 1e-10.

PR #776, opened 2022-11-04
add a keyword argument to plot_dos to support user-defined unit

kunyuan:
The function "plot_bandstructure" supports user-defined units, while "plot_dos" doesn't. The behaviors of these two functions are not consistent in this sense.

This PR adds a keyword argument "unit" to the function "plot_dos" to fix the above consistency problem.

Here is an example of sodium metal dos plot generated by following command: 
```julia
plot_dos(scfres, unit=u"eV", xlims=[-5, 20], ylims=[0, 40])
```

dos_in_eV

### DIFF
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -65,10 +65,13 @@ function plot_band_data(kpath::KPathInterpolant, band_data;
 end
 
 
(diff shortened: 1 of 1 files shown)


mfherbst:
Thanks!

PR #781, opened 2022-11-11
Bugfix for UPF

epolack:
If the repo is subject to change, maybe the more robust way is just to fix the commit to use.
Feel free to discard if the solution must come from fixed path in the pseudo repo.
(https://github.com/JuliaMolSim/PseudoLibrary/issues/7)


mfherbst:
Good idea. Will solve the issue for now :+1:.

PR #786, opened 2022-11-22
Improve DOS plotting default axes ranges

mfherbst:


src/plotting.jl
```diff
@@ -21,9 +21,21 @@ function ScfPlotTrace(plt=Plots.plot(yaxis=:log); kwargs...)
...
 
 
+function default_band_εrange(eigenvalues; εF=nothing)
+    if isnothing(εF)
+        # Can't decide where the interesting region is. Just plot everything
+        (minimum(minimum, eigenvalues), maximum(maximum, eigenvalues))
+    elseif is_metal(eigenvalues, εF)
+        εF .+ (-0.367, 0.367)
+    else
+        εF .+ (-0.147, 0.147)
```
antoine-levitt, line 31: Where do these numbers come from? Are these ev? If so can you say explicitly? Also not 100% sure about the is metal stuff, how about just tracking whether there's a temperature or not?

mfherbst: This is just what pymatgen does and this is just defaults for plotting. You can always change it later. I have no clue why they are doing it like that. Perhaps for metals you have more going on at the Fermi level so you care about a larger range. They are Hartree (in eV its 4 and 10)

antoine-levitt: Just add a comment saying it's 4ev and it's stolen from pymatgen

mfherbst: ok

PR #791, opened 2022-11-23
Switch default scf solver to builtin Anderson

epolack:
Draft to see if some tests fail with this brutal change of the default SCF algorithm.
Will need proper timings also.
(`NLsolve` may be use as an example and not in a test maybe.)

### DIFF
--- a/test/scf_compare.jl
+++ b/test/scf_compare.jl
@@ -12,15 +12,15 @@ include("testcases.jl")
     model = model_LDA(silicon.lattice, silicon.atoms, silicon.positions)
     basis = PlaneWaveBasis(model, Ecut, silicon.kcoords, silicon.kweights; fft_size)
 
-    # Run nlsolve without guess
+    # Run default solver without guess
     ρ0 = zeros(basis.fft_size..., 1)
-    ρ_nl = self_consistent_field(basis; ρ=ρ0, tol).ρ
+    ρ_def = self_consistent_field(basis; ρ=ρ0, tol).ρ
 
     # Run DM
     if mpi_nprocs() == 1  # Distributed implementation not yet available
         @testset "Direct minimization" begin
             ρ_dm = direct_minimization(basis; tol).ρ
-            @test maximum(abs, ρ_dm - ρ_nl) < sqrt(tol) / 10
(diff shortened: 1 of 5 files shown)


mfherbst:
Try a larger problem. Silicon is too small and nice to make a call here.

E.g. try Cr19, WFe, Fe2CrGa , Fe2MnAl from https://github.com/mfherbst/supporting-adaptive-damping/ and AlSiO2H_repeat_20 from https://github.com/mfherbst/supporting-ldos-preconditioning/.

test/scf_compare.jl
```diff
@@ -1,9 +1,22 @@
...
+# NLSolve documentation for details about the other parameters and methods.
+function scf_nlsolve_solver(m=10, method=:anderson; kwargs...)
+    function fp_solver(f, x0, max_iter; tol=1e-6)
+        res = nlsolve(x -> f(x) - x, x0; method, m, xtol=tol,
+                      ftol=0.0, show_trace=false, iterations=max_iter, kwargs...)
+        (; fixpoint=res.zero, converged=converged(res))
+    end
+    fp_solver
+end
+
```
mfherbst, line 19: I think if we really want to dump NLSolve then there is no need to reintroduce it just for the tests.

antoine-levitt:
You're just benchmarking latency here; try running time twice

mfherbst:
supersampling ... I think it has moved into `compute_fft_size`.

But yes, numbers look reasonable, I agree.

mfherbst:
@epolack How far is this from being ready to merge? I mean the timings really don't look all that different.

mfherbst:
What's your damping and mixing?

mfherbst:
> I don't see any damping parameter, so I guess the default one.

Then that's `damping=0.8`. Reduce that a bit.

mfherbst:
> So either we say it's fine as is, or I still need time to fix this last test.

@epolack  I would say it's fine if both have similar trouble.

antoine-levitt:
Yay, thanks for doing this. I tried to play with loading times a few months back but it was a failure, probably mostly due to nlsolve. With this PR and once julia 1.9 is in beta we should revisit this and add precompilation instructions.

PR #792, opened 2022-11-23
Implement more flexible Fermi level algorithm interface

### LINKED ISSUE #787: LDOS mixing for non-monotonic smearings
There is an issue with the LDOS mixing for non-monotonic smearing functions, e.g.

```julia
using DFTK

a = 5.42352  # Bohr
lattice = a / 2 * [[-1  1  1];
                   [ 1 -1  1];
                   [ 1  1 -1]]
atoms     = [ElementPsp(:Fe, psp=load_psp("hgh/lda/Fe-q8.hgh"))]
positions = [zeros(3)];

kgrid = [5, 5, 5]
Ecut = 10
magnetic_moments = [4];

model = model_LDA(lattice, atoms, positions;
                  magnetic_moments,
                  temperature=5e-4,
                  smearing=Smearing.MarzariVanderbilt())
@show model.n_electrons

ρ0 = guess_density(basis, magnetic_moments)
scfres = self_consistent_field(basis, is_converged=DFTK.ScfConvergenceDensity(1e-8);
                               ρ=ρ0, mixing=LdosMixing())
```

Also for Methfessel-Paxton. Might be due to the temperature fiddeling we do in the mixing, which does not play out so well here. Replacing the smearing in ldos by gaussian gives a performance penalty over kerker, but at least the scf converges.

mfherbst:
also implements Gaussian - Newton scheme discussed in #787.

Also fixes #787 (the LDOS issue) which was due to a bad threshold when computing the density.

### DIFF
--- a/src/occupation.jl
+++ b/src/occupation.jl
@@ -12,56 +12,24 @@ import Roots
 # with more advanced methods (e.g. false position), but more care has to
 # be taken with convergence criteria and the like
 
-abstract type FermiLevelAlgorithm end
-struct ZeroTemperature <: FermiLevelAlgorithm end
-struct Bisection       <: FermiLevelAlgorithm end
-struct GaussianNewton  <: FermiLevelAlgorithm end
+abstract type FermiAlgorithm end
 
-function deviation_n_electrons(basis::PlaneWaveBasis, eigenvalues, εF; smearing, temperature)
+"""Default selection of a Fermi level determination algorithm"""
+default_fermialg(::Type{<: Smearing.SmearingFunction}) = FermiTwoStage()
+default_fermialg(::Type{Smearing.Gaussian})   = FermiBisection()  # Monotonic smearing
+default_fermialg(::Type{Smearing.FermiDirac}) = FermiBisection()  # Monotonic smearing
+default_fermialg(::T) where {T <: Smearing.SmearingFunction} = default_fermialg(T)
+default_fermialg(model::Model) = default_fermialg(model.smearing)
+default_fermialg(basis::AbstractBasis) = default_fermialg(basis.model)
+
+function excess_n_electrons(basis::PlaneWaveBasis, eigenvalues, εF; smearing, temperature)
     occupation = compute_occupation(basis, eigenvalues, εF;
                                     smearing, temperature).occupation
     weighted_ksum(basis, sum.(occupation)) - basis.model.n_electrons
 end
 
-
-"""
-Find the occupation and Fermi level.
-"""
-function compute_occupation(basis::PlaneWaveBasis{T}, eigenvalues,
-                            algorithm::FermiLevelAlgorithm=GaussianNewton();
-                            temperature=basis.model.temperature,
-                            smearing=basis.model.smearing) where {T}
-    if !isnothing(basis.model.εF)  # fixed Fermi level
-        εF = basis.model.εF
-    else  # fixed n_electrons
-        # Check there are enough bands
-        max_n_electrons = (filled_occupation(basis.model)
-                           * weighted_ksum(basis, length.(eigenvalues)))
-        if max_n_electrons < n_electrons - sqrt(eps(T))
-            error("Could not obtain required number of electrons by filling every state. " *
-                  "Increase n_bands.")
-        end
-
-        εF = compute_fermi_level(basis, eigenvalues, algorithm; temperature, smearing)
-
-        # Sanity check on Fermi level
-        if deviation_n_electrons(basis, eigenvalues, εF; smearing, temperature) < sqrt(eps(T))
-            if iszero(temperature)
-                error("Unable to find non-fractional occupations that have the " *
-                      "correct number of electrons. You should add a temperature.")
-            else
-                error("This should not happen, debug me!")
-            end
-        end
-    end
-    occupation = compute_occupation(basis, eigenvalues, εF; temperature, smearing)
-
-    (; occupation, εF)
-end
-
-
-"""Compute the occupations, given eigenenergies and a Fermi level"""
-function compute_occupation(basis::PlaneWaveBasis{T}, eigenvalues, εF::Number;
+"""Compute occupation given eigenvalues and Fermi level"""
+function compute_occupation(basis::PlaneWaveBasis{T}, eigenvalues::AbstractVector, εF::Number;
                             temperature=basis.model.temperature,
                             smearing=basis.model.smearing) where {T}
     # This is needed to get the right behaviour for special floating-point types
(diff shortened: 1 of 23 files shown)


src/occupation.jl
```diff
@@ -64,50 +118,24 @@ function compute_fermi_level(basis::PlaneWaveBasis{T}, eigenvalues;
...
-            error("This should not happen, debug me!")
-        end
-    end
-    εF
+    εF_guess = compute_fermi_level(basis, eigenvalues, Bisection();
+                                   temperature, smearing=Smearing.Gaussian())
+
+    n_elec(εF)  = deviation_n_electrons(basis, eigenvalues, εF; smearing, temperature)
+    dn_elec(εF) = ForwardDiff.derivative(objective, εF)
+    Roots.find_zero((n_elec, dn_elec), εF_guess, Newton())
```
azadoks, line 140: See the QE implementation [here](https://github.com/QEF/q-e/blob/9ddde6fa44e0fb1c7debb78532a4b581a6ef82fd/PW/src/efermig.f90#L118).

To me, it seems like they do the Newton's minimization on
```julia
f(eF) = (compute_n_electrons(basis, eigenvalues, eF; smearing, temperature) - n_electrons)^2
df(eF) = ForwardDiff.derivative(f, eF)
ddf(eF) = ForwardDiff.derivative(df, eF)

Roots.find_zero((df, ddf), eF_Gaussian, Newton())
```

This is what I implemented for my project, and it seems to perform better and fail less than Newton's minimization on `(n, dn)`. I haven't given any thought to why this is the case or why Flaviano chose to do it this way, but it would be good to ask.

mfherbst: At least certainly even on simple things the quadratic one performs a lot better.

I guess because the route to the minimum is steeper.

antoine-levitt: What a weird thing to do. Suppose you want to solve for f(x)=0, then the standard newton method requires f' to be nonzero. If instead you set g = f^2 and solve for g' = 2 f' f = 0, that requires g'' = 2f'' f + 2f'^2 = 0. Since f is zero at the solution, that's f'^2. So if f' is small, then f'^2 is even smaller, so that's just worse? So they don't actually solve for f = 0, but rather try to minimize f^2. In your plot in the other thread it looked as if f was small but not exactly zero at the "right" fermi level, correct?

azadoks: Yeah, but I interpreted that as a result of the tolerance they use for MV and MP, which is an absolute error in the number of electrons of 1e-2 (which seems really big to me). For comparison, they use 1e-10 as the tolerance for the bisection method.

src/occupation.jl
```diff
@@ -64,50 +121,24 @@ function compute_fermi_level(basis::PlaneWaveBasis{T}, eigenvalues;
...
-        if iszero(temperature)
-            error("Unable to find non-fractional occupations that have the " *
-                  "correct number of electrons. You should add a temperature.")
-        else
-            error("This should not happen, debug me!")
-        end
-    end
-    εF
+    εF_guess = compute_fermi_level(basis, eigenvalues, Bisection();
+                                   temperature, smearing=Smearing.Gaussian())
```
azadoks, line 139: You could check for an early exit here if eF_guess happens to be right (not impossible).

src/occupation.jl
mfherbst: In my experiments I found that the actual numerical consistency for the cases with multiple Fermi solutions can be pretty bad in any case. I.e. you find a solution and then still DFTK's warning on the electron count for the density flags up.

But I agree a trustworthy approach should find a root for f = 0 and should ensure the electron count is as it should be, certainly to a higher accuracy than 1e-2.

> In your plot in the other thread it looked as if f was small but not exactly zero at the "right" fermi level, correct?

Not sure what you mean here @antoine-levitt. f is zero at the fermi level candidates, but f' can be all over the place ... generally the "best" seems to be the one where it is smallest from my current understanding, though.

src/occupation.jl
mfherbst: Indeed. Actually that's pretty likely even.

mfherbst:
@azadoks I added an interpolation algorithm for the Fermi level of MV and MP starting from a Gaussian fermi level. In my tests it does pretty good. It usually finds a Fermi level with positive density of states. But some cases are really hard to deal with. I guess it's not much the point to go beyond this approach, however, since the issues seem to occur only for really small number of k-points and large temperatures.

src/occupation.jl
mfherbst: > In my experiments I found that the actual numerical consistency for the cases with multiple Fermi solutions can be pretty bad in any case. I.e. you find a solution and then still DFTK's warning on the electron count for the density flags up.

This actually was a bug.

mfherbst:
@antoine-levitt This is ready for a review.

Since this whole business is quite a bit more complicated than I first thought I don't yet feel confident with respect to [Flaviano's idea](https://arxiv.org/abs/2212.07988) or my own algorithm, so this PR is just implementing a more flexible interface for fermi level finding algorithms and a two-stage algorithms for non-monotonic smearings that starts from Gaussian, but still solves the electron count to `eps(T)`.

Still I introduce an `tol_n_elec` argument and default it to the `occupation_threshold`, which serves for more clever Fermi level algorithms as an orientation for how well the electron count needs to be determined. That's currently unused, but we could do so and in [the paper](https://arxiv.org/abs/2212.07988) they mention some arguments why sometimes it makes physical sense to not keep the electron count as it should be.

mfherbst:
@antoine-levitt  ping ;)

antoine-levitt:
Yeah I know, it's on the TODO list :-p

mfherbst:
Ok great, sorry to be a bother ;)

src/occupation.jl
```diff
@@ -1,61 +1,99 @@
...
 import Roots
 
+# Functions for finding the Fermi level and occupation numbers for bands
+# The goal is to find εF so that
+# f_i = filled_occ * f((εi-εF)/θ) with θ = temperature
+# sum_i f_i = n_electrons
+# If temperature is zero, (εi-εF)/θ = ±∞.
+# The occupation function is required to give 1 and 0 respectively in these cases.
+#
+# For finite temperature we right now just use bisection; note that with MP smearing
```
antoine-levitt, line 10: Not true anymore?

src/occupation.jl
```diff
@@ -1,61 +1,99 @@
...
 
-    (; occupation, εF)
-end
+"""Default selection of a Fermi level determination algorithm"""
+default_fermialg(::Type{<: Smearing.SmearingFunction}) = FermiTwoStage()
+default_fermialg(::Type{Smearing.Gaussian})   = FermiBisection()  # Monotonic smearing
+default_fermialg(::Type{Smearing.FermiDirac}) = FermiBisection()  # Monotonic smearing
+default_fermialg(::T) where {T <: Smearing.SmearingFunction} = default_fermialg(T)
+default_fermialg(model::Model) = default_fermialg(model.smearing)
+default_fermialg(basis::AbstractBasis) = default_fermialg(basis.model)
```
antoine-levitt, line 23: that's a lot of methods... Might as well remove the last one (basis). Why don't you also just dispatch on instances instead of types?

src/occupation.jl
```diff
@@ -1,61 +1,99 @@
...
 
+function excess_n_electrons(basis::PlaneWaveBasis, eigenvalues, εF; smearing, temperature)
```
antoine-levitt, line 25: define two functions, one taking occupations, and one computing it?

src/occupation.jl
```diff
@@ -1,61 +1,99 @@
...
+        max_n_electrons = (filled_occupation(basis.model)
+                           * weighted_ksum(basis, length.(eigenvalues)))
+        if max_n_electrons < basis.model.n_electrons - tol_n_elec
+            error("Could not obtain required number of electrons by filling every state. " *
+                  "Increase n_bands.")
+        end
 
-    # Maximum occupation per state
-    filled_occ = filled_occupation(basis.model)
+        @timing "compute_fermi_level" begin
```
antoine-levitt, line 66: really? that should be pretty negligible...

src/occupation.jl
```diff
@@ -1,61 +1,99 @@
...
-    function compute_n_elec(εF)
-        weighted_ksum(basis, sum.(compute_occupation(basis, eigenvalues, εF;
-                                                     smearing, temperature)))
+        if abs(excess) > tol_n_elec
+            @warn("Large deviation of electron count in compute_occupation. (excess=$excess). " *
+                  "This may lead to an unphysical solution. Try decreasing the temperature " *
+                  "or using a different smearing function.")
+        end
+        if dexcess < -sqrt(eps(T))
+            @warn("Negative electron versus Fermi level derivative in compute_occupation. " *
```
antoine-levitt, line 82: that's the density of states, right? Might as well say that

antoine-levitt (commented):
LGTM!

src/scf/potential_mixing.jl
```diff
@@ -258,7 +259,7 @@ trial_damping(damping::FixedDamping, args...) = damping.α
     function EVρ(Vin; diagtol=tol / 10, ψ=nothing, eigenvalues=nothing, occupation=nothing)
         ham_V = hamiltonian_with_total_potential(ham, Vin)
 
-        res_V = next_density(ham_V, nbandsalg; eigensolver, ψ, eigenvalues,
+        res_V = next_density(ham_V, nbandsalg, fermialg; eigensolver, ψ, eigenvalues,
```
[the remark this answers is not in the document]
mfherbst: This pretty much undecided. We also have `nbandsalg`. I opted for that spelling, because this is what SciML people use for these type of options.

src/postprocess/dos.jl
```diff
@@ -50,7 +55,10 @@ function compute_ldos(ε, basis, eigenvalues, ψ; smearing=basis.model.smearing,
     # Use compute_density routine to compute LDOS, using just the modified
     # weights (as "occupations") at each k-point. Note, that this automatically puts in the
     # required symmetrization with respect to kpoints and BZ symmetry
-    compute_density(basis, ψ, weights)
+    compute_density(basis, ψ, weights; occupation_threshold=weight_threshold)
```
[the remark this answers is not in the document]
mfherbst: It's not just a check. It's also doing screening here to speed up the computation.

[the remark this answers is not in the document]
mfherbst: It's not just a check. It's also doing screening here to speed up the computation.

PR #793, opened 2022-12-02
Add developper documentation for GPU computations

GVigne:
This PR aims to add a short developer documentation on GPUs in DFTK. It briefly describes the goals we want to reach, as well as how GPU computations are currently handled in DFTK.

### DIFF
--- a/docs/src/developer/gpu_computations.md
+++ b/docs/src/developer/gpu_computations.md
@@ -1,18 +1,18 @@
 # GPU computations
 
-GPU computations in DFTK is very much a work in progress. The goal is to build on
-Julia's multiple dispatch to have the same code base for CPU and GPU. We hope we
-will be able to reach decent performances without having to write any kernels at all,
-only by relying on the high level functionalities implemented in the GPU packages.
+GPU computations in DFTK is a work in progress. The goal is to build on
+Julia's multiple dispatch to have the same code base for CPU and GPU. Our current
+approach is to aim at decent performance without writing any custom kernels at all,
+relying only on the high level functionalities implemented in the GPU packages.
 
 To go even further with this idea of unified code, we would also like to be able to
-support any type of GPU architecture: unlike other packages, we do not want to
-hard-code the use of a specific architecture, say a NVIDIA CUDA GPU. In order to do
-this, DFTK uses [GPUArrays](https://github.com/JuliaGPU/GPUArrays.jl), which is the
-counterpart of `AbstractArray` but for GPU arrays.
-
-Currently the three main GPU architectures are [CUDA](https://github.com/JuliaGPU/CUDA.jl),
-[ROCm](https://github.com/JuliaGPU/AMDGPU.jl) and [OneAPI](https://github.com/JuliaGPU/oneAPI.jl).
+support any type of GPU architecture: we do not want to hard-code the use of a
+specific architecture, say a NVIDIA CUDA GPU. DFTK does not realy on an
+architecture-specific package ([CUDA](https://github.com/JuliaGPU/CUDA.jl),
+[ROCm](https://github.com/JuliaGPU/AMDGPU.jl),
+[OneAPI](https://github.com/JuliaGPU/oneAPI.jl)...) but rather uses
+[GPUArrays](https://github.com/JuliaGPU/GPUArrays.jl), which is the counterpart of
+`AbstractArray` but for GPU arrays.
 
 ## Current implementation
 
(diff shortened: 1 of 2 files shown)


docs/src/developer/gpu_computations.md
```diff
@@ -0,0 +1,58 @@
+# GPU computations
+
+GPU computations in DFTK is very much a work in progress. The goal is to build on
```
antoine-levitt, line 3: That's a bit pessimistic

docs/src/developer/gpu_computations.md
```diff
@@ -0,0 +1,58 @@
...
+Julia's multiple dispatch to have the same code base for CPU and GPU. We hope we
+will be able to reach decent performances without having to write any kernels at all,
```
antoine-levitt, line 5: performance

docs/src/developer/gpu_computations.md
```diff
@@ -0,0 +1,58 @@
...
+only by relying on the high level functionalities implemented in the GPU packages.
+
+To go even further with this idea of unified code, we would also like to be able to
+support any type of GPU architecture: unlike other packages, we do not want to
```
antoine-levitt, line 9: that's not very nice :-P I think everybody wants that

docs/src/developer/gpu_computations.md
```diff
@@ -0,0 +1,58 @@
...
+PlaneWaveBasis(model; Ecut, kgrid, architecture = DFTK.GPU(CuArray))
+```
+Those `CPU` and `GPU` structures we inspired by the structures of same name in
+[Oceananigans](https://github.com/CliMA/Oceananigans.jl). **However, it is very**
+**likely that this API will change.** We hope that a unified convention will emerge,
+instead of having every package define their own way of dealing with computations on
+multiple architectures.
+
+Not all terms can be used when doing GPU computations: currently, the `Anyonic`,
+`Magnetic`, `TermPairwisePotential` and `Xc` terms are not supported. Symmetries
```
antoine-levitt, line 35: What's wrong with symmetries? They should work, no?

antoine-levitt: Also xc worked I thought?

docs/src/developer/gpu_computations.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Pitfalls
+There are a few things to keep in mind when doing GPU programming in DFTK.
+- Transfers to and from a device are a bit tricky because they are done by
+converting the array to an other type (ex: `CuArray(A)`, `Array(A)`...). This
+means that the array type used for computations should be stored somewhere:
+currently, this is done by the `CPU` and `GPU` structures and the helper functions
+`to_device` and `to_cpu`. `similar` could also be used, but then a reference array
+(one which already lives on the device) needs to be available at any time.
+- Functions' arguments should always be isbits (immutable and contains no
```
antoine-levitt, line 47: functions executed on the GPU you mean

docs/src/developer/gpu_computations.md
GVigne: Haha, I didn't want to be mean. I just wanted to point out that we were being quite ambitious as support for multiple GPU architectures doesn't seem to be the standard for the moment.

docs/src/developer/gpu_computations.md
GVigne: You're right for symmetries: I thought they didn't work, but there is actually not reason they shouldn't.
Xc doesn't work yet. I am building the model using the `model_DFT` function, but by giving an empty array for the functionals, so the Xc term is left out.

docs/src/developer/gpu_computations.md
mfherbst: > doesn't seem to be the standard for the moment.

Well. Everyone wants to, but some people struggle to make it happen. But the major codes (e.g. VASP, SIRIUS-QE) have it now for CUDA, AMD and Intel.

docs/src/developer/gpu_computations.md
```diff
@@ -0,0 +1,58 @@
...
+when creating the basis. `architecture` should be an initialized instance of
+the (non-exported) `CPU` and `GPU` structures. `CPU` does not require any argument,
+but `GPU` requires the type of array which will be used for GPU computations.
+
+```@example gpu_computations
+PlaneWaveBasis(model; Ecut, kgrid, architecture = DFTK.CPU())
+PlaneWaveBasis(model; Ecut, kgrid, architecture = DFTK.GPU(CuArray))
+```
+Those `CPU` and `GPU` structures we inspired by the structures of same name in
+[Oceananigans](https://github.com/CliMA/Oceananigans.jl). **However, it is very**
```
mfherbst, line 29: Well sort of. I would not put Oceananigans here as we got inspired from a couple of directions.

docs/src/developer/gpu_computations.md
```diff
@@ -0,0 +1,58 @@
...
+automatic differentiation.
+
+## Pitfalls
+There are a few things to keep in mind when doing GPU programming in DFTK.
+- Transfers to and from a device are a bit tricky because they are done by
+converting the array to an other type (ex: `CuArray(A)`, `Array(A)`...). This
+means that the array type used for computations should be stored somewhere:
+currently, this is done by the `CPU` and `GPU` structures and the helper functions
+`to_device` and `to_cpu`. `similar` could also be used, but then a reference array
+(one which already lives on the device) needs to be available at any time.
```
mfherbst, line 46: It feels a bit you are explaining the process how we arrived at our current solution. I would not emphasise this so much. I would more focus on how things should now be done. E.g. give examples to explain `to_device` and `to_cpu` and say that they should be preferred over a bare `CuArray(A)`. Then say briefly why. Also keep mentioning `similar` and probably add a reference to `zeros_like`.

docs/src/developer/gpu_computations.md
```diff
@@ -0,0 +1,58 @@
...
+- Functions' arguments should always be isbits (immutable and contains no
+references to other values). This means that the model should never be given
+to a function which will get executed on the GPU or it will fail: instead, only the
+relevant field should be passed.
```
mfherbst, line 50: Pretty hard to follow. Best explain by an example.

docs/src/developer/gpu_computations.md
```diff
@@ -0,0 +1,58 @@
...
+only by relying on the high level functionalities implemented in the GPU packages.
+
+To go even further with this idea of unified code, we would also like to be able to
+support any type of GPU architecture: unlike other packages, we do not want to
+hard-code the use of a specific architecture, say a NVIDIA CUDA GPU. In order to do
+this, DFTK uses [GPUArrays](https://github.com/JuliaGPU/GPUArrays.jl), which is the
+counterpart of `AbstractArray` but for GPU arrays.
+
+Currently the three main GPU architectures are [CUDA](https://github.com/JuliaGPU/CUDA.jl),
+[ROCm](https://github.com/JuliaGPU/AMDGPU.jl) and [OneAPI](https://github.com/JuliaGPU/oneAPI.jl).
```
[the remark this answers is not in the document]
GVigne: Actually, I wanted to remind the reader that there are currently three big GPU architectures, and that they all have their dedicated package. However, we don't want (and don't really need to) use any of those specific packages, we can just rely on GPUArrays.
As you pointed out, it's probably not the best place to put it. I'm going to rephrase this.

docs/src/developer/gpu_computations.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Current implementation
+
+For now, GPU computations are done by specializing the `architecture` keyword argument
+when creating the basis. `architecture` should be an initialized instance of
+the (non-exported) `CPU` and `GPU` structures. `CPU` does not require any argument,
+but `GPU` requires the type of array which will be used for GPU computations.
+
+```@example gpu_computations
```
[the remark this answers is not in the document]
GVigne: I was a bit confused with the `@example` and `@setup` macros. I also didn't knew you could make a ```julia block. Thanks!

docs/src/developer/gpu_computations.md
```diff
@@ -1,18 +1,18 @@
 # GPU computations
 
-GPU computations in DFTK is very much a work in progress. The goal is to build on
-Julia's multiple dispatch to have the same code base for CPU and GPU. We hope we
-will be able to reach decent performances without having to write any kernels at all,
-only by relying on the high level functionalities implemented in the GPU packages.
+GPU computations in DFTK is a work in progress. The goal is to build on
```
mfherbst, line 3:
```suggestion
Performing GPU computations in DFTK is still work in progress. The goal is to build on
```

docs/src/developer/gpu_computations.md
```diff
@@ -21,33 +21,64 @@ when creating the basis. `architecture` should be an initialized instance of
...
 PlaneWaveBasis(model; Ecut, kgrid, architecture = DFTK.CPU())
 PlaneWaveBasis(model; Ecut, kgrid, architecture = DFTK.GPU(CuArray))
 ```
-Those `CPU` and `GPU` structures we inspired by the structures of same name in
-[Oceananigans](https://github.com/CliMA/Oceananigans.jl). **However, it is very**
-**likely that this API will change.** We hope that a unified convention will emerge,
-instead of having every package define their own way of dealing with computations on
-multiple architectures.
+**It is very likely that this API will change.**, based on the evolution of the
+Julia ecosystem concerning distributed architectures.
```
mfherbst, line 29:
```suggestion
!!! note "GPU API is experimental"
    It is very likely that this API will change, based on the evolution of the
    Julia ecosystem concerning distributed architectures.
```

docs/src/developer/gpu_computations.md
```diff
@@ -21,33 +21,64 @@ when creating the basis. `architecture` should be an initialized instance of
...
+provide a level of abstraction while also allowing multiple architectures
+to be used.
+```julia
+cuda_gpu = DFTK.GPU(CuArray)
+cpu_architecture = DFTK.CPU()
+A = rand(10) # A is on the CPU
+B = DFTK.to_device(cuda_gpu, A) # B is a copy of A on the CUDA GPU
+B .+= 1.
+C = DFTK.to_cpu(B) # C is a copy of B on the CPU
+D = DFTK.to_device(cpu_architecture, B) # Equivalent to the previous line, but      should be avoided as it is less clear
```
mfherbst, line 53: keep an eye on code formatting ... usual dftk guidelines apply.

docs/src/developer/gpu_computations.md
```diff
@@ -21,33 +21,64 @@ when creating the basis. `architecture` should be an initialized instance of
...
+```
+*Note:* `similar` could also be used, but then a reference array
 (one which already lives on the device) needs to be available at any time.
```
mfherbst, line 56: at any time. Hmm. Rather at call time (similar allocates a new array).

docs/src/developer/gpu_computations.md
```diff
@@ -21,33 +21,64 @@ when creating the basis. `architecture` should be an initialized instance of
...
-to a function which will get executed on the GPU or it will fail: instead, only the
-relevant field should be passed.
+This was done previously, with helper functions to easily build new arrays
+on a given architecture: see for example
+[`zeros_like`](https://github.com/JuliaMolSim/DFTK.jl/pull/711/commits/ce5da66009440bd8552429eb8cfe96944da16564).
+- Functions which will get executed on the GPU should always have arguments
+which are isbits (immutable and contains no references to other values).
+When using `map`, also make sure that every structure used is also isbits.
+For example, the following map will fail, as `model` contains strings and
+arrays which are not on the GPU.
```
mfherbst, line 64: are not `isbits`.

docs/src/developer/gpu_computations.md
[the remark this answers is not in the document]
mfherbst: sounds much better now!

docs/src/developer/gpu_computations.md
```diff
@@ -0,0 +1,58 @@
...
+Those `CPU` and `GPU` structures we inspired by the structures of same name in
+[Oceananigans](https://github.com/CliMA/Oceananigans.jl). **However, it is very**
+**likely that this API will change.** We hope that a unified convention will emerge,
+instead of having every package define their own way of dealing with computations on
+multiple architectures.
+
+Not all terms can be used when doing GPU computations: currently, the `Anyonic`,
+`Magnetic`, `TermPairwisePotential` and `Xc` terms are not supported. Symmetries
+should also be disabled. Nothing has yet been made to enable GPU support for
+automatic differentiation.
```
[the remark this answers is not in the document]
mfherbst: sounds good now!

docs/src/developer/gpu_computations.md
GVigne: I meant that when you are developing code, you always need to have some sort of reference array at hand: sometimes, this means adding a new argument to a function which will only be used to know what type of device is being used.

mfherbst: Well but in this case you need the `cpu_architecture` at hand. I don't see why one is an advantage over the other.

GVigne: Hum, that's true.

mfherbst:
Looks good enough for a first version. I will merge now, we can follow up with more details later if needed.

GVigne:
Yes, I'll add a small benchmark by the end of the week like Antoine said.

mfherbst:
ok!

PR #795, opened 2022-12-05
Correct PW parameterization comment

epolack:
By default, it is not Teter93 that is done.


antoine-levitt (commented):
Thanks!

antoine-levitt:
Did you mean to mark it WIP? Looks OK to me

PR #797, opened 2022-12-09
Add benchmark in GPU documentation

GVigne:
As we said in https://github.com/JuliaMolSim/DFTK.jl/pull/793, I am now adding a small benchmark in the GPU developer documentation. For now, it's really the bare minimum (code + timers): I think we shouldn't be too verbose for this part, as the benchmarks may change as we implement new features for GPUs. If there is something you think I should add for clarity, just tell me.

### DIFF
--- a/docs/src/developer/gpu_computations.md
+++ b/docs/src/developer/gpu_computations.md
@@ -89,3 +89,91 @@ performance. For example, iterating through the columns of a matrix to compute t
 norms is not efficient, as a new kernel is launched for every column. Instead, it
 is better to build the vector containing these norms, as it is a vectorized
 operation and will be much faster on the GPU.
+
+## A quick benchmark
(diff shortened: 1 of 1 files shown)


mfherbst:
Thanks for this addition. Could you add some details on the DFTK / Julia / CUDA version used and just add a note that there are no guarantees on this and that this only gives a general impression (see the similar remarks on https://docs.dftk.org/stable/tricks/parallelization/)

mfherbst:
Also I now made a change in master, that the GPU CI is only run automatically for
- master
- for branches with a name ending in `gpu` (e.g. `lobpcg-gpu`)
- when runs are manually triggered (e.g. right before merging).

mfherbst:
Closing this as we now should have a benchmark using DFT.

PR #800, opened 2022-12-12
Switch to density tolerance by default

mfherbst:
- [X] Adapt examples
- [X] Adapt docs
- [x] Adapt tests


mfherbst:
@antoine-levitt ping.

antoine-levitt:
Oh I'm all in favor of course. Maybe can you add a bit to the self_consistent_field docstring? We currently don't have much information there, but the tol is probably the most important to document

mfherbst:
Agreed. Will do.

PR #802, opened 2022-12-18
Rely on AtomsIO for structure parsing

mfherbst:
Requires fixes in https://github.com/JuliaMolSim/AtomsBase.jl/pull/58.


Project.toml
```diff
@@ -33,7 +34,6 @@ Primes = "27ebfcd6-29c5-5fa9-bf4b-fb8fc14df3ae"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 ProgressMeter = "92933f4c-e287-5a05-a399-4b506db050ca"
 PseudoPotentialIO = "cb339c56-07fa-4cb2-923a-142469552264"
-PyCall = "438e738f-606a-5dbb-bf0a-cddfbfd45ab0"
```
antoine-levitt, line 36: farewell python, we're not particularly going to miss you

docs/src/guide/installation.md
```diff
@@ -26,29 +26,23 @@ That's it. With this you are all set to
...
-The installation of the ASE python package is therefore strongly recommended.
-To do so first find out which version of python is used by the
-[PyCall.jl](https://github.com/JuliaPy/PyCall.jl) package.
-You can do this for example with the Julia commands
+## Recommended optional packages
+While not strictly speaking required to use DFTK it is usually convenient to install
+a couple of standard packages from the [AtomsBase](https://github.com/JuliaMolSim/AtomsBase.jl)
+ecosystem, such as [AtomsIO.jl](https://github.com/mfherbst/AtomsIO.jl)
+or [ASEconvert.jl](https://github.com/mfherbst/ASEconvert.jl).
+You can install these two packages using
```
antoine-levitt, line 34: Fine but won't the user still need to install python ase? This is just shifting the problem to another package, right?

examples/atomsbase.jl
```diff
@@ -5,6 +5,49 @@
...
+
+# To use an AbstractSystem in DFTK, we attach pseudopotentials, construct a DFT model,
+# discretise and solve:
+system = attach_psp(system; Si="hgh/lda/si-q4")
+
+model  = model_LDA(system; temperature=1e-3)
+basis  = PlaneWaveBasis(model; Ecut=15, kgrid=[4, 4, 4])
+scfres = self_consistent_field(basis, tol=1e-8);
+
+# If we did not want to use ASE we could of course use any other package,
```
antoine-levitt, line 28: Using commas like this ("any other package, which yields an AS object, ...") is a germanism: remove the first comma

examples/atomsbase.jl
```diff
@@ -5,6 +5,49 @@
...
+# which yields an AbstractSystem object, this includes:
```
antoine-levitt, line 29: . This includes

examples/gaas_surface.jl
```diff
@@ -3,10 +3,15 @@
 # This example shows how to use the
 # [atomistic simulation environment](https://wiki.fysik.dtu.dk/ase/index.html),
 # or ASE for short,
-# to set up a particular gallium arsenide surface
-# and run the resulting calculation in DFTK.
-# The particular example we consider the (1, 1, 0) GaAs surface separated by vacuum.
+# to set up and run a particular calculation of a gallium arsenide surface.
+# ASE is a Python package which intents to simplify the process of setting up,
```
antoine-levitt, line 7: intends

antoine-levitt: which intents to simplify -> to simplify

examples/gaas_surface.jl
```diff
@@ -3,10 +3,15 @@
...
+# running and analysing results from atomistic simulations across different simulation codes.
+# By means of [ASEconvert](https://github.com/mfherbst/ASEconvert.jl) it is seamlessly
+# integrated with the AtomsBase ecosystem and thus available to DFTK via our own
+# [AtomsBase integration](@ref).
 #
+# In this example we will consider modelling the (1, 1, 0) GaAs surface separated by vacuum
+# using density-functional theory.
```
antoine-levitt, line 14: Pretty clear that we're using DFT :-p

examples/input_output.jl
```diff
@@ -3,51 +3,40 @@
...
-# [atomistic simulation environment](https://wiki.fysik.dtu.dk/ase/index.html),
-# a Python package to simplify the process of setting up, running and
-# analysing results from atomistic simulations across different programs.
-# If it is installed it is automatically used by DFTK in order to read a wide range
-# of input files, such as xyz, CIF, input files of various other codes
-# (e.g. Quantum Espresso, VASP, ABINIT, CASTEP, …).
-# The full list of formats
-# can be found in the [ASE IO documentation](https://wiki.fysik.dtu.dk/ase/ase/io/io.html).
+# ## Reading / writing files supported by AtomsIO
+# [AtomsIO](https://github.com/mfherbst/AtomsIO.jl) is a Julia package, which allows
```
antoine-levitt, line 7: same as above, comma is a germanism

antoine-levitt: which allows to -> to

examples/silicon.jl
```diff
@@ -9,6 +11,8 @@ Si = ElementPsp(:Si, psp=load_psp("hgh/lda/Si-q4"))
 atoms     = [Si, Si]
 positions = [ones(3)/8, -ones(3)/8]
 
-model = model_LDA(lattice, atoms, positions)
+# model = model_LDA(lattice, atoms, positions, temperature=1e-2, smearing=Smearing.MethfesselPaxton(2))
+model = model_LDA(lattice, atoms, positions, temperature=1e-2u"Ry", smearing=Smearing.MarzariVanderbilt())
```
antoine-levitt, line 15: What? Don't use smearing for silicon. Also silicon is our hello world, so I would keep this as basic as possible and stick to the defaults

src/Model.jl
```diff
@@ -190,6 +190,15 @@ function Model(lattice::AbstractMatrix{<:Quantity}, atoms::Vector{<:Element},
...
 end
 
+
+"""
+    Model(system::AbstractSystem; kwargs...)
+
+AtomsBase-compatible Model constructor. Sets structural information (`atoms`, `positions`,
+`lattice`, `n_electrons` etc.) from the passed `system`.
+"""
+Model(system::AbstractSystem; kwargs...) = _call_with_system(Model, system; kwargs...)
```
antoine-levitt, line 200: This _call_with_system API is super weird

antoine-levitt (commented):
Yay for removing stuff!

docs/src/guide/installation.md
mfherbst: No as ASEconvert does this automatically.

examples/silicon.jl
mfherbst: Oh that is an error. I was playing around with this file.

docs/src/guide/installation.md
antoine-levitt: OK but automatically in python means "it fails only 50% of the time", no?

mfherbst: no this is pretty reliable and independent of your system python (makes a custom conda environment and such ... try it, it's pretty cool). Has just one issue: You cannot use PyCall-based packages and PythonCall-based packages at the same time.

antoine-levitt: That means no DFTK+PyPlot? That's a pretty huge caveat! I don't think we should do it until that's fixed... What's the issue number?

mfherbst: You can use [PythonPlot](https://github.com/stevengj/PythonPlot.jl) instead. Also it's not an issue more a fundamental thing. The caveat is down to race conditions due to the way the libpython and down-stream dependencies are dynamically loaded into the Julia process. If you want to avoid that you can always opt-out of the automatic python installation stuff in PythonCall and manage everything yourself (using PyCall and conda packages as before). Point is just that novice users don't need to bother with this.

mfherbst: Ah turns out this is only an issue if the python interpreters are actually different: https://cjdoris.github.io/PythonCall.jl/stable/pycall/

antoine-levitt: I mean, novice users are just going to install this, then pyplot, then see crashes. Probably we should add a note somewhere, maybe including the error message so that people can find it when googling.

mfherbst: I should probably add some details in the docs about this.

PR #804, opened 2022-12-20
Remove deprecated features

mfherbst:


mfherbst:
People can always pin "0.5" if they really need this. That's the point about semantic versioning.

PR #806, opened 2022-12-21
Add simplistic time per SCF step printing

mfherbst:


mfherbst:
@antoine-levitt ok?

antoine-levitt:
Hm, maybe put it at the end instead?

mfherbst:
Ah next to the diag, that makes sense.

PR #811, opened 2023-01-12
Change internal fixed-point API to (x, info) = f(x, info) to simplify SCF metadata tracking

epolack:
PR from @niklasschmitz. Will fix the tests when CI completes.


mfherbst:
Since there are conflicts against master the CI will not run. @epolack @niklasschmitz You must first resolve the conflicts and/or rebase.

mfherbst:
> Should I close it and open a new one?

If you want to continue the development I'd say that's the simplest.

You can already get an idea what's the issue right now from https://git.uni-paderborn.de/herbstm/DFTK.jl/-/jobs/123457 by the way.

src/scf/self_consistent_field.jl
```diff
@@ -137,37 +129,43 @@ Overview of parameters:
...
     end
 
+    info_init = (; ρin=ρ, ψ=ψ, occupation=nothing, eigenvalues=nothing, εF=nothing, 
+                   n_iter=0, converged=false)
+
     # Tolerance and maxiter are only dummy here: Convergence is flagged by is_converged
     # inside the fixpoint_map.
-    solver(fixpoint_map, ρout, maxiter; tol=eps(T))
-
+    _, info = solver(fixpoint_map, ρ, info_init, maxiter; tol=eps(T)) # TODO ?? why dummy?
```
niklasschmitz, line 152: On that TODO: I think I understand now the comment above for custom convergence flagging - I wonder whether we could move this to the fixed-point solvers API though? instead of specifying the tolerance, one could do something like `solver(fixpoint_map, ρ, info_init, maxiter; is_converged_fn=is_converged_fn)`?

mfherbst (commented):
Very nice. Has the potential for quite some simplifications (e.g. unifying SCF routines in density and potential).

src/scf/scf_solvers.jl
```diff
@@ -57,7 +60,7 @@ CROP-accelerated root-finding iteration for `f`, starting from `x0` and keeping
 a history of `m` steps. Optionally `warming` specifies the number of non-accelerated
 steps to perform for warming up the history.
 """
-function CROP(f, x0, m::Int, max_iter::Int, tol::Real, warming=0)
+function CROP(f, x0, info0, m::Int, max_iter::Int, tol::Real, warming=0)
```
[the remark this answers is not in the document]
antoine-levitt: It's useful to have since it parallels the anderson implementation, and we might want to look at it again in 5 years. Why can't `info` be an optional arg (or kwarg) so that solvers don't have to bother with it?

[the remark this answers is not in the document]
mfherbst: If we want the solvers to be able to use the isconverged callback it somehow needs an info. But perhaps we can have a default-scfres function, which would be the default for the arg.

src/scf/self_consistent_field.jl
```diff
@@ -181,42 +170,49 @@ Overview of parameters:
...
+                       ρin, α=damping, n_iter, nbandsalg.occupation_threshold,
+                       runtime_ns=time_ns() - start_ns, nextstate...,
+                       diagonalization=[nextstate.diagonalization])
 
         # Compute the energy of the new state
         if compute_consistent_energies
             (; energies) = energy(basis, ψ, occupation; ρ=ρout, eigenvalues, εF)
         end
+        history_Etot = vcat(info.history_Etot, energies.total)
+        history_Δρ = vcat(info.history_Δρ, norm(Δρ) * sqrt(basis.dvol))
```
mfherbst, line 183: I think it's ok to just push and risk that the `info.history_*` things are globally updated.

src/scf/self_consistent_field.jl
```diff
@@ -181,42 +170,49 @@ Overview of parameters:
...
+        # Apply mixing and pass it the full info as kwargs
+        ρnext = ρin .+ T(damping) .* mix_density(mixing, basis, Δρ; info_next...)
 
-        converged = is_converged(info)
-        converged = MPI.bcast(converged, 0, MPI.COMM_WORLD)  # Ensure same converged
-        callback(merge(info, (; converged)))
+        callback(info_next)
+        if is_converged(info_next)
+            info_next = merge(info_next, (; converged=true))
+        end
```
mfherbst, line 192: Be careful here, this is not doing the same thing (order of callback and is_converged swapped) and also you removed the MPI broadcast (which could explain the failing MPI tests)

src/scf/self_consistent_field.jl
mfherbst: We can very much remove this dummy stuff now !

mfherbst (commented):
Main blocker for me is the thing with the MPI tests. Might be due to the removed broadcast.

src/scf/self_consistent_field.jl
niklasschmitz: Thanks! Added back the MPI bcast, indeed MPI tests are passing now.

src/scf/scf_solvers.jl
```diff
@@ -55,7 +58,7 @@ CROP-accelerated root-finding iteration for `f`, starting from `x0` and keeping
 a history of `m` steps. Optionally `warming` specifies the number of non-accelerated
 steps to perform for warming up the history.
 """
-function CROP(f, x0, m::Int, maxiter::Int, tol::Real, warming=0)
+function CROP(f, x0, info0, m::Int, maxiter::Int, tol::Real, warming=0)
```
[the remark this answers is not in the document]
niklasschmitz: Removed CROP in [e9a3394](https://github.com/JuliaMolSim/DFTK.jl/pull/811/commits/e9a3394c76f79e338fc4f14b642112f9fa3afaff).

src/scf/self_consistent_field.jl
niklasschmitz: Done in [caee55e](https://github.com/JuliaMolSim/DFTK.jl/pull/811/commits/caee55e87e9eb0a1c9b487d3c25daca60a312b10)

src/scf/self_consistent_field.jl
niklasschmitz: Agree mostly, but I think avoiding this risk makes the state more transparent in distinguishing `info` from `info_next`.

examples/custom_solvers.jl
```diff
@@ -16,19 +16,18 @@ model = model_LDA(lattice, atoms, positions)
...
+        if info.converged
             break
         end
-        x = x + mixing_factor * inc
-        fx = f(x)
+        x = x + mixing_factor * (fx - x)
     end
-    (; fixpoint=x, converged=norm(fx-x) < tol)
+    (; fixpoint=x, info)
 end;
```
mfherbst, line 31: In line with the first example here I would probably show the "proper way" to supply new solvers, but also show with a custom `is_converged` callback how to customise the convergence check.

docs/src/guide/self_consistent_field.jl
```diff
@@ -123,25 +123,25 @@
...
-        ## If change less than tolerance, break iterations:
-        if norm(Fρ - ρ) < tol
+        Fρ, info = F(ρ, info)
+        if info.converged
             break
         end
-        ρ  = Fρ
-        Fρ = F(ρ)
+        ρ = Fρ
     end
```
mfherbst, line 141: I agree this would be the "right" way to modify this function, but I'm wondering whether for demonstration purposes (this is supposed to be a low-entrance barrier example) it would be better not to "hide away" the convergence check. Don't you think @niklasschmitz ?

mfherbst (commented):
Nits on the docs.

docs/src/guide/self_consistent_field.jl
niklasschmitz: Yes, that sounds better for pedagogical purposes. I will just add a note to link to the "custom solvers" example, where it should be explained how `info.converged` works.

examples/custom_solvers.jl
niklasschmitz: I added a short note on this one the custom solvers docs, and appended a customizing example with stresses in analogy to ScfConvergenceForce in [3a37b2a](https://github.com/JuliaMolSim/DFTK.jl/pull/811/commits/3a37b2a99f81d5be3151d7c88d96033777857c9a). Though I wonder whether this is too complicated.

src/scf/self_consistent_field.jl
```diff
@@ -144,29 +144,18 @@ Overview of parameters:
...
 
     # We do density mixing in the real representation
     # TODO support other mixing types
-    function fixpoint_map(ρin)
-        converged && return ρin  # No more iterations if convergence flagged
-        MPI.bcast(Dates.now() ≥ end_time, MPI.COMM_WORLD) && return ρin
+    function fixpoint_map(ρin, info)
+        (; ψ, occupation, eigenvalues, εF, n_iter, converged, timedout) = info
+        converged && return ρin, info  # No more iterations if convergence flagged
+        timedout && return ρin, info
```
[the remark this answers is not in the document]
niklasschmitz: I agree on removing them. I would even leave out the assert, leaving just flagging to the inner function and the decision-making to the solver, if that makes sense?

examples/custom_solvers.jl
```diff
@@ -69,8 +73,31 @@ scfres = self_consistent_field(basis;
...
+    error = isnothing(conv.previous_stress) ? NaN : norm(conv.previous_stress - stress)
+    conv.previous_stress = stress
+    error < conv.tolerance
+end
+
+scfres2 = self_consistent_field(basis;
+                                solver=my_fp_solver,
+                                is_converged=ScfConvergenceStress(1e-8),
+                                eigensolver=my_eig_solver,
+                                mixing=MyMixing());
```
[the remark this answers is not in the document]
niklasschmitz: It works, but looks too complicated indeed, I replaced it by energy convergence.

src/scf/self_consistent_field.jl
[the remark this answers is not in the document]
mfherbst: I thought that too, but could it not lead to hard-to-debug errors if a solver just forgets to check this ? But then if you write your own solver you mess with pretty low-level stuff anyway, so you should know what you're doing.

examples/custom_solvers.jl
```diff
@@ -79,25 +79,20 @@ scfres = self_consistent_field(basis;
...
-    (; basis, ψ, occupation, eigenvalues, εF, ρout) = info
-    stress = compute_stresses_cart((; basis, ψ, occupation, eigenvalues, εF, ρ=ρout))
-    error = isnothing(conv.previous_stress) ? NaN : norm(conv.previous_stress - stress)
-    conv.previous_stress = stress
-    error < conv.tolerance
+function my_convergence_criterion(info)
+    tol = 1e-10
+    if last(info.history_Δρ) > 10sqrt(conv.tolerance)
+        return false  # The ρ change should also be small to avoid the SCF being just stuck
+    end
```
mfherbst, line 88: Not even sure I would do this in this simple example.

niklasschmitz: Removed in [0a89022](https://github.com/JuliaMolSim/DFTK.jl/pull/811/commits/0a8902294752f0f345849f08b3b448a395c9eecd)

src/scf/scf_solvers.jl
```diff
@@ -1,30 +1,27 @@
-# these provide fixed-point solvers that can be passed to scf()
+# these provide fixed-point solvers that can be passed to self_consistent_field()
 
-# the fp_solver function must accept being called like fp_solver(f, x0, tol,
-# maxiter), where f(x) is the fixed-point map. It must return an
-# object supporting res.sol and res.converged
+# the fp_solver function must accept being called like
+# fp_solver(f, x0, info0; maxiter), where f(x, info) is the fixed-point map. 
+# It must return an object supporting res.fixpoint and res.info
```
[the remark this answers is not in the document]
niklasschmitz: Added a comment in [1d405e9](https://github.com/JuliaMolSim/DFTK.jl/pull/811/commits/1d405e98d26dc7c8ec24e189d2aa3d4908309c9c)

PR #816, opened 2023-02-08
Model update constructor

mfherbst:
Some of the improvements suggested in #675.

### DIFF
--- a/src/Model.jl
+++ b/src/Model.jl
@@ -237,7 +237,7 @@ function Model(model::Model{T};
                positions::Vector{<:AbstractVector}=model.positions,
                kwargs...) where {T, U}
     TT = promote_type(T, U, eltype(positions[1]))
-    Model{TT}(model; lattice, positions, atoms, kwargs...)
+    Model{TT}(model; lattice, positions, kwargs...)
 end
 
 Base.convert(::Type{Model{T}}, model::Model{T}) where {T}    = model
--- a/src/Model.jl
+++ b/src/Model.jl
@@ -203,6 +203,42 @@ function Model(system::AbstractSystem; kwargs...)
     Model(parsed.lattice, parsed.atoms, parsed.positions; parsed.magnetic_moments, kwargs...)
 end
 
+"""
+    Model(model;
+          lattice=model.lattice,
+          positions=model.positions,
+          atoms=model.atoms, kwargs...)
+
+Construct an identical model to `model` with the option to change some of the contained
+parameters. Especially changing `lattice` or `positions` is useful for geometry or
+lattice optimisations.
+"""
+function Model(model::Model{T};
+               lattice::AbstractMatrix{U}=model.lattice,
+               positions::Vector{<:AbstractVector}=model.positions,
+               atoms=model.atoms,
+               kwargs...) where {T, U}
+    TT = promote_type(T, U, eltype(positions[1]))
+    Model(TT.(lattice), atoms, positions;
+          model.model_name,
+          model.n_electrons,
+          magnetic_moments=[],  # not used because symmetries explicitly given
+          terms=model.term_types,
+          model.temperature,
+          model.smearing,
+          model.εF,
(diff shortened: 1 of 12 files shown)


src/Model.jl
```diff
@@ -203,6 +203,46 @@ function Model(system::AbstractSystem; kwargs...)
...
+          magnetic_moments=[],  # not used because symmetries explicitly given
+          terms=model.term_types,
+          model.temperature,
+          model.smearing,
+          model.εF,
+          model.spin_polarization,
+          model.symmetries,
+          # Can be safely disabled: this has been checked for model
+          disable_electrostatics_check=true,
+          kwargs...
```
antoine-levitt, line 232: This means the kwargs can't override the other settings (eg temperature etc). Maybe do a merge here?

src/Model.jl
```diff
@@ -203,6 +203,46 @@ function Model(system::AbstractSystem; kwargs...)
     Model(parsed.lattice, parsed.atoms, parsed.positions; parsed.magnetic_moments, kwargs...)
 end
 
+"""
+    Model(model; [lattice, positions, atoms, kwargs...])
+    Model{T}(model; [lattice, positions, atoms, kwargs...])
+
+Construct an identical model to `model` with the option to change some of the contained
+parameters. If one of `lattice`, `positions`, `atoms` is missing, the respective value
+from `model` is used. This constructor is useful for changing the data type in the model
```
antoine-levitt, line 212: That sentence is kind of obvious no?

antoine-levitt: (I mean the "if one of ...")

src/fft.jl
```diff
@@ -60,7 +60,7 @@ Perform a real valued iFFT; see [`ifft`](@ref).
 """
 function irfft(basis::PlaneWaveBasis{T}, f_fourier::AbstractArray; check=Val(true)) where {T}
     f_real = ifft(basis, f_fourier)
-    (check == Val(true)) && @assert norm(imag(f_real)) < sqrt(eps(T))
+    (check == Val(true)) && @assert norm(imag(f_real)) < max(1e-12, sqrt(eps(T)))
```
antoine-levitt, line 63: I hate these checks so much. Let's just kill them all...

src/postprocess/stresses.jl
```diff
@@ -6,18 +6,7 @@ Compute the stresses (= 1/Vol dE/d(M*lattice), taken at M=I) of an obtained SCF
...
-                          magnetic_moments=[], # not used because symmetries explicitly given
-                          terms=model.term_types,
-                          model.temperature,
-                          model.smearing,
-                          model.εF,
-                          model.spin_polarization,
-                          model.symmetries,
-                          # Can be safely disabled: this has been checked for basis.model
-                          disable_electrostatics_check=true)
+        new_model = Model(basis.model; lattice)
```
antoine-levitt, line 9: nice

antoine-levitt (commented):
Cool!

src/Model.jl
mfherbst: Interestingly not. The `kwargs` take preference.

src/fft.jl
mfherbst: Not sure. They do help during development for sure.

src/Model.jl
antoine-levitt: Really?
```
julia> sum([]; dims=1, dims=2)
ERROR: syntax: keyword argument "dims" repeated in call to "sum" around REPL[1]:1
```

mfherbst: Yeah but this does not use a `kwargs...` construct. That makes a difference.

antoine-levitt:
```
julia> kwargs = (; dims=1)
(dims = 1,)

julia> sum(; kwargs..., dims=2)
ERROR: MethodError: no method matching sum(; dims=2)
```
Woah this is wild. Is this documented anywhere?

PR #822, opened 2023-02-10
Factorisation of phonon test cases

epolack:
Phonon PR: implement the _true solutions_ (AD is the reference) for phonon computations  & tests simplification.

### DIFF
--- a/src/response/cg.jl
+++ b/src/response/cg.jl
@@ -35,11 +35,10 @@ function cg!(x::AbstractVector{T}, A::LinearMap{T}, b::AbstractVector{T};
 
     # preconditioned conjugate gradient
     while n_iter < maxiter
-        n_iter += 1
-
         # output
         info = (; A, b, n_iter, x, r, residual_norm, converged, stage=:iterate)
         callback(info)
+        n_iter += 1
         if (n_iter ≥ miniter) && residual_norm ≤ tol
             converged = true
             break
--- a/src/response/cg.jl
+++ b/src/response/cg.jl
@@ -28,7 +28,7 @@ function cg!(x::AbstractVector{T}, A::LinearMap{T}, b::AbstractVector{T};
     # p is the descent direction
     p = copy(c)
     n_iter = 0
-    residual_norm = zero(real(T))
+    residual_norm = norm(r)
 
     # convergence history
     converged = false
@@ -36,6 +36,14 @@ function cg!(x::AbstractVector{T}, A::LinearMap{T}, b::AbstractVector{T};
     # preconditioned conjugate gradient
     while n_iter < maxiter
         n_iter += 1
+
+        # output
+        info = (; A, b, n_iter, x, r, residual_norm, converged, stage=:iterate)
+        callback(info)
+        if (n_iter ≥ miniter) && residual_norm ≤ tol
+            converged = true
+            break
+        end
         mul!(c, A, p)
         α = γ / dot(p, c)
 
@@ -43,14 +51,6 @@ function cg!(x::AbstractVector{T}, A::LinearMap{T}, b::AbstractVector{T};
         x .= proj(x .+ α .* p)
         r .= proj(r .- α .* c)
         residual_norm = norm(r)
-
-        # output
-        info = (; A, b, n_iter, x, r, residual_norm, converged, stage=:iterate)
-        callback(info)
-        if (n_iter > miniter) && residual_norm <= tol
-            converged = true
-            break
-        end
 
         # apply preconditioner and prepare next iteration
         ldiv!(c, precon, r)
--- a/src/response/chi0.jl
+++ b/src/response/chi0.jl
@@ -110,6 +110,7 @@ function sternheimer_solver(Hk, ψk, ε, rhs;
                             callback=identity, cg_callback=identity,
                             ψk_extra=zeros_like(ψk, size(ψk, 1), 0), εk_extra=zeros(0),
                             Hψk_extra=zeros_like(ψk, size(ψk, 1), 0), tol=1e-9)
+    iszero(rhs) && return(zero(rhs))
     basis = Hk.basis
     kpoint = Hk.kpoint
 
--- a/src/response/chi0.jl
+++ b/src/response/chi0.jl
@@ -110,7 +110,6 @@ function sternheimer_solver(Hk, ψk, ε, rhs;
                             callback=identity, cg_callback=identity,
                             ψk_extra=zeros_like(ψk, size(ψk, 1), 0), εk_extra=zeros(0),
                             Hψk_extra=zeros_like(ψk, size(ψk, 1), 0), tol=1e-9)
-    iszero(rhs) && return zero(rhs)
     basis = Hk.basis
     kpoint = Hk.kpoint
 
--- a/src/response/chi0.jl
+++ b/src/response/chi0.jl
@@ -350,7 +350,7 @@ end
     # occupation_threshold for which we compute the associated response δψn,
     # the others being discarded to ψ_extra.
     # We then use the extra information we have from these additional bands,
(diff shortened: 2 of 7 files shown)


src/PlaneWaveBasis.jl
```diff
@@ -346,6 +346,17 @@ Creates a new basis identical to `basis`, but with a custom set of kpoints
                    basis.symmetries_respect_rgrid, basis.comm_kpts, basis.architecture)
 end
 
+"""
+Creates a new basis identical to `basis`, but with an updated model.
+"""
+@timing function PlaneWaveBasis(basis::PlaneWaveBasis, new_model::Model)
+    PlaneWaveBasis(new_model, basis.Ecut, basis.fft_size,
+                   basis.variational, basis.kcoords_global,
+                   basis.kweights, basis.kgrid, basis.kshift,
```
antoine-levitt, line 355: kweights_global probably

src/response/phonon/ad.jl
```diff
@@ -0,0 +1,32 @@
...
+function compute_dynmat_ad(basis::PlaneWaveBasis{T}; scf_kwargs...) where {T}
+    # TODO: Cannot use symmetries at all, see https://github.com/JuliaMolSim/DFTK.jl/issues/817
+    @assert isone(only(basis.model.symmetries))
+
+    model = basis.model
+    cell = (; model.lattice, model.atoms, model.positions)
+    n_atoms = length(model.positions)
+    n_dim = model.n_dim
+
+    dynamical_matrix = zeros(ComplexF64, (n_dim, n_atoms, n_dim, n_atoms))
```
antoine-levitt, line 14: use eltype from basis

src/response/phonon/ad.jl
```diff
@@ -0,0 +1,32 @@
...
+    for τ in 1:n_atoms
+        for γ in 1:n_dim
+            displacement = zero.(model.positions)
+            displacement[τ] = StaticArrays.setindex(displacement[τ], one(T), γ)
```
antoine-levitt, line 18: Should we define our own setindex, and alias it to staticarrays'? I think there are other places in the code where `setindex` would be nice

src/response/phonon/commons.jl
```diff
@@ -0,0 +1,60 @@
+dalton_to_amu   = austrip(1u"u")
+hartree_to_cm⁻¹ = ustrip(u"cm^-1", 1u"bohr^-1") ./ austrip(1u"c0") / 2π
```
antoine-levitt, line 2: move this to common/?

src/response/phonon/commons.jl
```diff
@@ -0,0 +1,60 @@
...
+    mass_matrix = zeros(ComplexF64, n_dim*n_atoms, n_dim*n_atoms)
+    if !iszero(minimum(charges))
+        for iτ in eachindex(charges)
+            for γ in 1:n_dim
+                mass_τγ = ustrip(periodic_table[charges[iτ]].atomic_mass) * dalton_to_amu
+                mass_matrix[γ + (iτ - 1)*n_dim, γ + (iτ - 1)*n_dim] = mass_τγ
+            end
+        end
+    else
+        @warn "Some elements have unknown masses"
```
antoine-levitt, line 23: Make this an error

antoine-levitt: Make this part of the API for Element: atomic_mass(::Element)

src/response/phonon/commons.jl
```diff
@@ -0,0 +1,60 @@
...
+    n_dim = model.n_dim
+
+    charges = charge_nuclear.(model.atoms)
+    # Some elements may be unknown (for example when using ElementGaussian).
+    mass_matrix = zeros(ComplexF64, n_dim*n_atoms, n_dim*n_atoms)
+    if !iszero(minimum(charges))
+        for iτ in eachindex(charges)
+            for γ in 1:n_dim
+                mass_τγ = ustrip(periodic_table[charges[iτ]].atomic_mass) * dalton_to_amu
+                mass_matrix[γ + (iτ - 1)*n_dim, γ + (iτ - 1)*n_dim] = mass_τγ
```
antoine-levitt, line 19: Should be a `Diagonal` matrix

src/response/phonon/commons.jl
```diff
@@ -0,0 +1,60 @@
...
+    n_atoms = length(positions)
+    lattice = model.lattice
+    n_dim = model.n_dim
+    inv_lattice = compute_inverse_lattice(lattice)[1:n_dim, 1:n_dim]
+
+    cart_mat = zero.(dynamical_matrix)
+    for i in 1:n_dim:n_dim*n_atoms
+        for j in 1:n_dim:n_dim*n_atoms
+            sub_matrix = dynamical_matrix[i:i+(n_dim-1), j:j+(n_dim-1)]
+            cart_mat[i:i+(n_dim-1), j:j+(n_dim-1)] = inv_lattice' * sub_matrix * inv_lattice
```
antoine-levitt, line 43: we have a function for this (comatrix_red_to_cart). Your conversion looks fishy btw, should be lattice' on the right, no?

antoine-levitt (commented):
We should be able to simplify this code a bit, but looks reasonable

src/response/phonon/ad.jl
antoine-levitt: Sure

src/response/chi0.jl
```diff
@@ -110,6 +110,7 @@ function sternheimer_solver(Hk, ψk, ε, rhs;
                             callback=identity, cg_callback=identity,
                             ψk_extra=zeros_like(ψk, size(ψk, 1), 0), εk_extra=zeros(0),
                             Hψk_extra=zeros_like(ψk, size(ψk, 1), 0), tol=1e-9)
+    iszero(rhs) && return(zero(rhs))
```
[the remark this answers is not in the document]
mfherbst: Also not sure. Comparisons against exact zero are always fishy, so it should probably be some tolerance. But which ?

Also don't braket around the `return` ;)

test/phonon/helpers.jl
```diff
@@ -57,47 +26,97 @@ function generate_supercell_qpoints(; supercell_size=generate_random_supercell()
     (; supercell_size, qpoints)
 end
 
-# Test against a reference array.
-function test_frequencies(testcase, terms, ω_ref; tol=1e-9, supercell_size=[2, 1, 3])
-    model = Model(testcase.lattice, testcase.atoms, testcase.positions; terms)
-    basis_bs = PlaneWaveBasis(model; Ecut=5)
+function are_approx_frequencies(ω_uc, ω_ref; tol=1e-10)
```
mfherbst, line 29: I'd keep the test because you run the `@test` macro inside.

src/response/chi0.jl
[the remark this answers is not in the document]
mfherbst: `Tolerance of Sterneimer/10` is tricky, because what is a good cutoff depends on the conditioning of `H-ε`. I mean what you do now fixes the immediate issue, but it does not guarantee we don't fall into this trap again when the `rhs` is small.

But yes the cg should have some protection to detect cases where we solve A*x = ε with ε small. I bet the other CG implementations might have this. I suppose you can get some ideas there.

If it turns out to be too tricky, then let's just leave the quick fix you propose.

src/response/cg.jl
```diff
@@ -28,29 +28,29 @@ function cg!(x::AbstractVector{T}, A::LinearMap{T}, b::AbstractVector{T};
...
     n_iter = 0
-    residual_norm = zero(real(T))
+    residual_norm = norm(r)
 
     # convergence history
     converged = false
 
     # preconditioned conjugate gradient
     while n_iter < maxiter
         n_iter += 1
```
mfherbst, line 38: Good idea. I think I would move the `n_iter += 1` also below the callback.

PR #824, opened 2023-02-16
Bugfix for ScfDiagtol

epolack:
Not sure this is a bug, but just in case.


mfherbst:
Not a bug. This ensures the actually returned value can never grow as it changes `diagtol_max`.

PR #828, opened 2023-03-02
Remove nonnegative density check

antoine-levitt:
It's just obnoxious. We build a density that is positive by construction. Then we symmetrize it, which can make it numerically negative. That is fine and expected behavior. Either we enforce it to be positive by clipping to zero, or we don't, but there's no point in warning about it.

### DIFF
--- a/src/densities.jl
+++ b/src/densities.jl
@@ -1,10 +1,6 @@
 # Densities (and potentials) are represented by arrays
 # ρ[ix,iy,iz,iσ] in real space, where iσ ∈ [1:n_spin_components]
 
-function _check_nonnegative(ρ::AbstractArray{T}; tol=eps(T)) where {T}
-    minimum(ρ) < -tol && @warn("Negative ρ detected", min_ρ=minimum(ρ))
-end
-
 """
     compute_density(basis::PlaneWaveBasis, ψ::AbstractVector, occupation::AbstractVector)
 
@@ -55,7 +51,6 @@ using an optional `occupation_threshold`. By default all occupation numbers are 
 
     mpi_sum!(ρ, basis.comm_kpts)
     ρ = symmetrize_ρ(basis, ρ; do_lowpass=false)
-    _check_nonnegative(ρ; tol=5occupation_threshold)
     ρ
 end
 


mfherbst:
Not sure. The problem is that the density can be negative e.g. for negative occupation numbers. Might be good to warn about it in that case, no?

mfherbst:
We could go with the clipping though, but we should still warn in case the negativities get too large.

antoine-levitt:
I mean, why? If you're using a method with negative occupation numbers, you know you're doing it, so what good is a warning?

mfherbst:
The point is if you know what you are doing the density is not negative even if you can have negative occupation numbers. It's only negative if you happen to have hit the wrong Fermi level and thus e.g. use too few k-Points / too low temperature or sth like that. In that case it's helpful to get a warning that points you in the right direction.

antoine-levitt:
I think that depends on the system : if you model a molecule you're likely going to have very small densities at the edge of the system, and any small numerical issue will push you to negative values

mfherbst:
Sure can happen, but then the density will be super small-negative and hardly < `-5occupation_threshold` or not?

I'm not necessarily against removing the check, I just can't think of another way to give users a fair warning for the case, e.g. when using MP and when settings are messed up. Because if you silently drop it in that case, likely the SCF will at best not converge or at worst converge to complete garbage.

The key problem here is that selecting a reasonable k-Point mesh and temperature for MP is not always trivial. So I think one should not just argue: Oh if you choose an unreasonable k-Point setup, you cannot expect a reasonable answer.

antoine-levitt:
> Sure can happen, but then the density will be super small-negative and hardly < -5occupation_threshold or not?

I've definitely had it happen for the anyon stuff, without any temperature at all, just because the density at the boundary of the box is already of the order of the machine precision. So it's not particularly indicative of defective usage...

> I'm not necessarily against removing the check, I just can't think of another way to give users a fair warning for the case, e.g. when using MP and when settings are messed up. Because if you silently drop it in that case, likely the SCF will at best not converge or at worst converge to complete garbage.

Could we warn on very negative occupations instead?

mfherbst:
> Could we warn on very negative occupations instead?

We could, but I'm not sure that's more reliable as a single large negative occupation (-0.2) can be ok if it's compensated by many many positive ones, giving overall a positive density.

mfherbst:
I think I understand why you got this check for the anyons. You used direct minimisation, where `compute_density` is called without the occupation_threshold parameter, such that this defaults to `0`. Therefore there was no leverage freedom for the density to become slightly negative due to numerical noise. I'll fix this case in #839.

PR #829, opened 2023-03-03
Fix Hamiltonian application for nonstandard models

antoine-levitt:
Has been broken since https://github.com/JuliaMolSim/DFTK.jl/commit/530548459ce95dd6bb9dceb0748f7c3838b52eba (and was only accidentally working before). Good thing nobody uses magnetic fields, huh :-)

### DIFF
--- a/src/terms/Hamiltonian.jl
+++ b/src/terms/Hamiltonian.jl
@@ -36,16 +36,18 @@ function HamiltonianBlock(basis, kpoint, operators, scratch=ham_allocate_scratch
     fourier_ops  = filter(o -> o isa FourierMultiplication,   optimized_operators)
     real_ops     = filter(o -> o isa RealSpaceMultiplication, optimized_operators)
     nonlocal_ops = filter(o -> o isa NonlocalOperator,        optimized_operators)
-    divAgrid_ops = filter(o -> o isa DivAgradOperator,        optimized_operators)
+    divAgrad_ops = filter(o -> o isa DivAgradOperator,        optimized_operators)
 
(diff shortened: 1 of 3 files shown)


test/anyons.jl
```diff
@@ -18,3 +18,35 @@ using Test
     @test norm(curlA - 2π*DFTK.ρref_real_2D(x, y, M, σ)) < 1e-4
     @test abs(divA) < 1e-6
 end
+
+@testset "Anyons: check E11" begin
```
mfherbst, line 22: What is E11 ?

src/terms/Hamiltonian.jl
```diff
@@ -229,5 +231,5 @@ end
 function hamiltonian_with_total_potential(Hk::HamiltonianBlock, V)
     operators = [op for op in Hk.operators if !(op isa RealSpaceMultiplication)]
     push!(operators, RealSpaceMultiplication(Hk.basis, Hk.kpoint, V))
-    HamiltonianBlock(Hk.basis, Hk.kpoint, operators, Hk.scratch)
+    HamiltonianBlock(Hk.basis, Hk.kpoint, operators)
```
mfherbst, line 234: Why remove this? Saves quite a bit of allocation for potential-based SCFs

test/anyons.jl
antoine-levitt: Some quantity defined in the paper, doesn't really matter what it is, it's just a test we reproduce the results in the paper

src/terms/Hamiltonian.jl
antoine-levitt: OK, I added it back. It's a bit ugly, but this is a pretty internal thing anyway

PR #835, opened 2023-03-20
PseudoPotentialIO

azadoks:
I released PseudoPotentialIO v0.2.0 this morning, so while everything is fresh, I started to work on a fewest-changes integration of it into DFTK.

I'm still considering a possible interface change on the PseudoPotentialIO side that could help with code duplication and would require a bit of a rework here, but doing this should help shake out any remaining bugs / rough edges in PseudoPotentialIO.

Most everything is working well with relatively minor changes, but there are a couple things that might need to be changed in PseudoPotentialIO to get e.g. ForwardDiff working as expected.

- [ ] Change some typing in PPIO HghPsP for forward diff?
- [ ] Re-export `load_psp` in DFTK so that `PseudoPotentialIO` doesn't have to be `using`ed every time you want to load a pseudo?


azadoks:
See https://github.com/azadoks/PseudoPotentialIO.jl/pull/11 for the alternative interface in PseudoPotentialIO

PR #839, opened 2023-03-30
More graceful numerical checks

mfherbst:
I'm running some Float32 calculations for the paper and I'm running into some problems with the assertions in `irfft`, even for the cases where an `enforce_real!` just before the `irfft` should make sure that the array is truly real.

This PR does two things:
   - Increase occupation threshold for low-precision floats (is probably reasonable ... density error of 1e-6 is unrealistic with Float32 anyway)
   - Fudge in `irfft` (which I don't like)

In light of the discussion in #828 I wonder if we should maybe have some form of a "production" mode, which disables the irfft and density checks (and similar) and a "testing" mode, which keeps them? Certainly in the unit tests it would be nice to keep them on, just to stumble over accidental errors.


antoine-levitt:
You know my position: we should just remove this check, it's super annoying for no good reason. Otherwise we'll just play tolerance chasing for ages.

mfherbst:
Ok, agreed. I'll drop the check.

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -454,6 +454,10 @@ end
         # Quick sanity check
         for i = 1:size(X, 2)
             @views if abs(BX[:, i]'X[:, i] - 1) >= sqrt(eps(real(eltype(X))))
+                # TODO error-ing is too harsh here. Better throw some exception (e.g. in
+                # reduced precision this can be interpreted as an indicator to re-try the
+                # orthogonalisation in elevated precision or to stop the reduced precision
+                # iterations as a whole.
                 error("LOBPCG is badly failing to keep the vectors normalized; this should never happen")
```
antoine-levitt, line 461: Did this really fail in your tests? X here should be the result of orthogonal operations of matrices returned by ortho, so they really should be orthogonal to some small multiple of machine precision

mfherbst: It fails for large-ish systems in Float32. Maybe there is sth to debug, but I could not be bothered.

mfherbst: Sent you something.

PR #850, opened 2023-04-16
Fix initial density guess generation for GPU runs

### LINKED ISSUE #841: Initial guess density currently broken on GPU
Regression introduced in #766. We should see if there is at least a quick fix ...

zhubonan:
Fixes #841 and make the GPU test work in `test/gpu.jl` by ensuring the initial density calculation is always done on CPU.  Also fixed an issue in Xc calculation with hard-coded `Array` in `TermXc`. 

Also added some notes on debugging GPU code and anti-patterns in the documentation. 

Replaced some `zeros` with `zeros_like`, should reduce potential issues. These changes should have no effect for running the code with CPU.


docs/src/developer/gpu_computations.md
```diff
@@ -88,3 +88,62 @@ performance. For example, iterating through the columns of a matrix to compute t
...
+
+```julia
+rho_temp = similar(basis.G_vectors, T, basis.fft_size...)
+fill!(rho_temp, 0)
+```
+
+
+## Debugging
+
+First, make sure the scalar indexing is disabled as these calculations are not computed on GPU but instead on the CPU.
```
mfherbst, line 119: I'd rather write `... as otherwise fallback routines on the CPU are allowed. While these are helpful for interactive debugging, these always imply a synchronisation on the device and a data transfer to main memory. In practice runs this then results in an unnecessary overhead and the calculation appearing to be struck.`

Point is that computation is not necessarily happening on the CPU (depends on what you do with the scalar data), but the sync+transfer can be a bad idea.

docs/src/developer/gpu_computations.md
```diff
@@ -88,3 +88,62 @@ performance. For example, iterating through the columns of a matrix to compute t
...
+Data must be transferred for each call so the speed can be extremely slow, resulting the calculation appears to be *stuck*.
+
+For CUDA GPU, scalar indexing is allowed on REPL, and a warning will be print. 
+However, this does not tell you where the offending lines are.
```
mfherbst, line 123: Maybe better: `Note that by default scalar indexing is allowed on REPL and results in a warning being printed. However, ... ` (because scalar indexing is also the default for other GPU backends).

docs/src/developer/gpu_computations.md
```diff
@@ -88,3 +88,62 @@ performance. For example, iterating through the columns of a matrix to compute t
...
+For CUDA GPU, scalar indexing is allowed on REPL, and a warning will be print. 
+However, this does not tell you where the offending lines are. 
+
+To completely disable scalar indexing, just run
+
+```julia
+CUDA.allowscalar(false)
+```
+
+which will allow the program to error when scalar indexing is used.
```
mfherbst, line 131: typo: which will cause the program to error

docs/src/developer/gpu_computations.md
```diff
@@ -88,3 +88,62 @@ performance. For example, iterating through the columns of a matrix to compute t
...
+The offending code should be rewritten using array operations. 
+If it is not possible or non-trivial to do so, one make the calculation to be run on CPU with,
```
mfherbst, line 133: maybe clearer: `to do so, it is usually best to have the calculation to run on the CPU, `

src/density_methods.jl
```diff
@@ -143,20 +143,24 @@ end
 function atomic_density_superposition(basis::PlaneWaveBasis{T},
                                       form_factors::IdDict{Tuple{Int,T},T};
                                       coefficients=ones(T, length(basis.model.atoms))
-                                      )::Array{T,3} where {T}
```
mfherbst, line 146: If I recall correctly we did this to avoid a type stablity / help the compiler to infer types. Best to check this in case you have not yet done so (e.g. with Cthulu.jl)
@azadoks do you remember?

src/density_methods.jl
```diff
@@ -143,20 +143,24 @@ end
...
+                                      ) where {T}
     model = basis.model
-    G_cart = G_vectors_cart(basis)
+    G_cart = to_cpu(G_vectors_cart(basis))
+    Gv = to_cpu(G_vectors(basis))
+    atom_groups = to_cpu(model.atom_groups)
+    positions = to_cpu(model.positions)
```
mfherbst, line 151: These should not be on the GPU. Are you sure this is needed?

src/density_methods.jl
```diff
@@ -143,20 +143,24 @@ end
 function atomic_density_superposition(basis::PlaneWaveBasis{T},
                                       form_factors::IdDict{Tuple{Int,T},T};
                                       coefficients=ones(T, length(basis.model.atoms))
-                                      )::Array{T,3} where {T}
+                                      ) where {T}
     model = basis.model
-    G_cart = G_vectors_cart(basis)
+    G_cart = to_cpu(G_vectors_cart(basis))
+    Gv = to_cpu(G_vectors(basis))
```
mfherbst, line 149: I'm tempted to save the line here inline this into the map(enumerate( )), but @antoine-levitt might disagree here.

src/density_methods.jl
```diff
@@ -166,9 +170,11 @@ function atomic_density_form_factors(basis::PlaneWaveBasis{T},
                                      )::IdDict{Tuple{Int,T},T} where {T<:Real}
     model = basis.model
     form_factors = IdDict{Tuple{Int,T},T}()  # IdDict for Dual compatability
-    for G in G_vectors_cart(basis)
+    G_cart = to_cpu(G_vectors_cart(basis))
+    atom_groups = to_cpu(model.atom_groups)
```
mfherbst, line 174: see comment above

src/density_methods.jl
```diff
@@ -166,9 +170,11 @@ function atomic_density_form_factors(basis::PlaneWaveBasis{T},
...
+    for G in G_cart
```
mfherbst, line 175: Inline?

src/terms/magnetic.jl
```diff
@@ -27,7 +27,7 @@ function TermMagnetic(basis::PlaneWaveBasis{T}, Afunction::Function) where {T}
             end
         end
     end
-    TermMagnetic(Apotential)
+    TermMagnetic(to_device.(Ref(basis.architecture), Apotential))
```
mfherbst, line 30: Did you check that the Magnetic term now works on the GPU. If yes, we maybe can remove the respective comment in docs/src/developer/gpu_computations.md

src/terms/terms.jl
```diff
@@ -80,7 +80,7 @@ In this case the matrix has effectively 4 blocks
 """
 @timing function compute_kernel(basis::PlaneWaveBasis{T}; kwargs...) where {T}
     n_spin = basis.model.n_spin_components
-    kernel = zeros(T, n_spin * prod(basis.fft_size), n_spin * prod(basis.fft_size))
+    kernel = zeros_like(basis.G_vectors, T, n_spin * prod(basis.fft_size), n_spin * prod(basis.fft_size))
```
mfherbst, line 83: The line now becomes a bit long. Maybe cache the `prod(basis.fft_size)` or the `n_spin * prod(basis.fft_size)` in a variable and use it in the line thereafter for clarity.

mfherbst (commented):
Awesome! Thanks very much @zhubonan for your PR. Here are some comments for discussion.

mfherbst:
Also while you are at it, could you remove the workarounds I introduced in 

https://github.com/JuliaMolSim/DFTK.jl/commit/6a8c25c22999d6153b98574ebd99211779c8dbae#diff-21d2f263953fee8479af0575d1c861d2659db2fb962f8a3d94c26dc263004223

src/density_methods.jl
azadoks: I vaguely remember having some type stability issues here, but explicitly checking it would be good.

zhubonan: OK I will try to investigate it. But have `Array{T,3}` would force the returned density to be on CPU I think.

mfherbst: > But have Array{T,3} would force the returned density to be on CPU I think.

Absolutely. If we really need the annotation, in that case we would have to be a bit smarter (e.g. by introducing some type assert or annotation where the type is derived from the array type stored in the PlaneWaveBasis).

docs/src/developer/gpu_computations.md
```diff
@@ -88,3 +88,62 @@ performance. For example, iterating through the columns of a matrix to compute t
...
+```julia
+rho_temp = zeros_like(basis.G_vectors, T, basis.fft_size...)
+```
+
+or 
+
+```julia
+rho_temp = similar(basis.G_vectors, T, basis.fft_size...)
+fill!(rho_temp, 0)
+```
```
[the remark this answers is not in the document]
zhubonan: Thanks for the suggestion, I have change it now!

src/density_methods.jl
zhubonan: Indeed, I have deleted those. The model is always on CPU.

src/density_methods.jl
zhubonan: Ah OK have changed it now.

src/terms/terms.jl
zhubonan: Good point, updated now.

zhubonan:
@mfherbst Thanks for all the comments. I have made some extra changes. 
The `Magnetic`, `ExternalFromReal` and `LocalNonlinearity` terms should all work now. 

Still need to wait for CI tests to finish, though. Do you mind taking another look? Thanks.

docs/src/developer/gpu_computations.md
```diff
@@ -88,3 +88,56 @@ performance. For example, iterating through the columns of a matrix to compute t
...
+should be written as:
+
+```julia
+rho_temp = zeros_like(basis.G_vectors, T, basis.fft_size...)
+```
+
+
+## Debugging
+
+First, make sure the scalar indexing is disabled as these calculations are not computed on GPU but instead on the CPU,
```
mfherbst, line 112: Sorry, I think I was unclear here:
```suggestion
First, make sure the scalar indexing is disabled
```

src/terms/anyonic.jl
```diff
@@ -82,7 +82,9 @@ struct TermAnyonic{T <: Real, Tρ, TA} <: Term
     ρref::Tρ
     Aref::TA
 end
+
 function TermAnyonic(basis::PlaneWaveBasis{T}, hbar, β) where {T}
+    @assert isa(basis.architecture, CPU) "Anyonic term is not implement for GPU yet."
```
mfherbst, line 87: We don't really do asserts with a message. In such a case we would just have the assertion and a comment in the code. Alternatively use the `error( )` function if you think that's cleaner.

The argument for a simple `@assert` here is (a) Anyonic models are a pretty specific use case and (b) when this  happens, the user would anyway need to open the code to fix it.

src/terms/operators.jl
```diff
@@ -114,16 +114,23 @@ struct MagneticFieldOperator{T <: Real, AT} <: RealFourierOperator
...
-        ∂αψ_fourier = pα .* ψ.fourier
+        ∂αψ_fourier = p[α, :] .* ψ.fourier
         ∂αψ_real = ifft(op.basis, op.kpoint, ∂αψ_fourier)
         Hψ.real .+= op.Apot[α] .* ∂αψ_real
     end
 end
+
+function Gv_to_array(Gv::AbstractArray{SVector{3, T}}) where {T}
+    reinterpret(reshape, T, Gv)
+end
```
mfherbst, line 132: Why do you need this as a separate function?

src/terms/operators.jl
```diff
@@ -114,16 +114,23 @@ struct MagneticFieldOperator{T <: Real, AT} <: RealFourierOperator
     kpoint::Kpoint{T}
     Apot::AT  # Apot[α][i,j,k] is the A field in (cartesian) direction α
 end
+
 function apply!(Hψ, op::MagneticFieldOperator, ψ)
     # TODO this could probably be better optimized
+    # The cartesian coordinates should be pre-allocated....
+    p = Gv_to_array(Gplusk_vectors_cart(op.basis, op.kpoint))
```
mfherbst, line 121: This looks very unusual to me. Why do you need this in the first place?

src/terms/magnetic.jl
```diff
@@ -50,3 +50,30 @@ function ene_ops(term::TermMagnetic, basis::PlaneWaveBasis{T}, ψ, occupation;
 
     (; E, ops)
 end
+
+"""
+Specialised for AbstractGPUArray
+"""
+function ene_ops(term::TermMagnetic, basis::PlaneWaveBasis{T}, ψ::Vector{A}, occupation;
```
mfherbst, line 57: That's a lot of code duplication. I'm sure there is a better way to do this. Let's not bother for now. It's really not all that important.

src/terms/magnetic.jl
mfherbst: Ah, thanks for your efforts, but I was mostly wondering if they now happen to work by accident ...

src/workarounds/cuda_arrays.jl
```diff
@@ -1,16 +1,5 @@
...
-function LinearAlgebra.eigen(A::Hermitian{<:Complex,<:CUDA.CuArray})
-    values, vectors = CUDA.CUSOLVER.heevd!('V', 'U', A.data)
-    (; values, vectors)
-end
-
-function LinearAlgebra.eigen(A::Hermitian{<:Real,<:CUDA.CuArray})
-    values, vectors = CUDA.CUSOLVER.syevd!('V', 'U', A.data)
-    (; values, vectors)
-end
-
```
mfherbst, line 13: Are you sure this works? The issue is still open.

test/gpu.jl
```diff
@@ -49,7 +50,63 @@ end
     @test norm(scfres_cpu.ρ - Array(scfres_gpu.ρ)) < 1e-6
 end
 
+@testset "CUDA Magnetic+other terms" begin
```
mfherbst, line 53: That's a lot of code for a test of a rather obscure feature. Can you not just test the hamiltonian apply of the terms or something like that? See the hamiltonian_consistency.jl test if you're unsure what I mean. 

Maybe you can easily add a gpu branch to it ... if you come up with some reasonable base functions and stick them into some common file it should not imply too much extra code.

src/terms/magnetic.jl
mfherbst: Also we put GPU workarounds in src/workarounds/gpu_arrays.jl

src/terms/magnetic.jl
mfherbst: BTW our main paradigm in DFTK is to keep the code base manageable. That means we really want to avoid duplicating code to the largest extend possible ... and with Julia this very often works. So if you find yourself copying large chunks of code, better first ask for some advice before going through the hassle ;)

src/terms/local.jl
```diff
@@ -36,7 +36,9 @@ struct ExternalFromReal
 end
 
 function (external::ExternalFromReal)(basis::PlaneWaveBasis{T}) where {T}
-    pot_real = external.potential.(r_vectors_cart(basis))
+    # Move this operation on CPU as GPU broadcast is very strict on type stability
+    rv = to_cpu(r_vectors_cart(basis))
+    pot_real = to_device(basis.architecture, external.potential.(rv))
```
mfherbst, line 41: I don't think this is a good idea. It's better to keep thinks on the GPU and force users to code up their `external.potential` function properly. Otherwise you are making it impossible to do this sort of evaluation *only* on the GPU, which IMO is a very bad design decision.

mfherbst (commented):
Thanks for the update. Some more comments.

Don't feel obliged to deal with putting the extra terms on the GPU right now (in fact this would be very much a follow-up PR in my opinion).

src/workarounds/cuda_arrays.jl
zhubonan: > Also while you are at it, could you remove the workarounds I introduced in

> https://github.com/JuliaMolSim/DFTK.jl/commit/6a8c25c22999d6153b98574ebd99211779c8dbae#diff-21d2f263953fee8479af0575d1c861d2659db2fb962f8a3d94c26dc263004223

I thought you meant to remove it? Seems to work for my local tests. I think the methods are now in CUDA.jl via this PR: https://github.com/JuliaGPU/CUDA.jl/pull/1603/files#diff-e1fa24914853993ba813e2b2ee908a06f18ebcf7282c9dc9fc3e47842e91aac5?

test/gpu.jl
zhubonan: Good point, I have now added dedicated "hamiltonian_consistency" tests for GPUs, although this requires lifting the common functions to a different source file.

src/terms/local.jl
zhubonan: Thanks. I have removed it.

src/terms/magnetic.jl
zhubonan: The problem here is that the GPU version works for CPU, but it is much slower using CPU. The CPU version does not work for GPU... I have moved it to the `workarounds` now.

src/terms/operators.jl
zhubonan: This is because the G vectors are stored as a vector of `SVector{3, T}`. This makes it impossible to use broadcast with slicing that is needed for GPU. On CPU one can do `[elem[1] for elem in Gplusk_vectors_cart(op.basis, op.kpoint)]` to get say the first element of each vector  just fine, but this results in scalar indexing that does not work for GPU. 

The solution here is to reinterpret a vector of `SVector{3,T}` as a dense array, as the underlying memory layout is the same, I think. Then we can use slicing with the dot syntax.

src/terms/operators.jl
zhubonan: I put it as a separate function because it may be useful for other things as well. Although at the moment it is only used in `apply!(::MagneticFileOperator`. Putting it as a function also makes the type interface possible as the `T` needs to be inferred from the input.

src/terms/anyonic.jl
zhubonan: Makes sense, I have changed it to a simple comment.

zhubonan:
> Thanks for the update. Some more comments.

> Don't feel obliged to deal with putting the extra terms on the GPU right now (in fact this would be very much a follow-up PR in my opinion)

Thanks! I have just updated it with the comments addressed. Let me know if you more.

src/workarounds/cuda_arrays.jl
mfherbst: Indeed, your're right this now works. Great to see that :+1:.

But the workarounds I was talking about was in the test/gpu.jl file.

src/terms/operators.jl
mfherbst: I see, but in any case  this is a bit of a hack and I would not endorse this with a function on the level of the `G_vectors(basis)`-family of functions. Also how brittle is this? Reinterpret can (and has lead us) to very subtle and annoying bugs, so I'm a bit hesitant. For sure you should add a comment explaining the precise assumptions about memory layout etc. to the place you are using this code.

mfherbst (commented):
Hmm. I think the initial density stuff is good to go, modulo removing the workaround code in the tests.

Regarding the Magnetic term and the term tests, I'm a bit unhappy with the code duplication still, but I lack the time to take a look at this in detail now.

I'd suggest to split up the PR, get the density thing merged quickly and then deal with the other changes in a follow-up.

src/density_methods.jl
zhubonan: No, I have not checked, but there is indeed a type instability introduce by this PR due to

```
ρ = to_device(basis.architecture, ρ_cpu)
```

Because it is not possible to infer the `architecture` based on `basis` alone. I think we should include `architecture` in the type parameter of `PlanewaveBasis`. This does not affect the run time of the function itself, though. Is this function a performance critical part and get called lots of times? Otherwise, I think instability does not harm too much.

I checked that the original version did not to have type instability for the output type in the first place anyway, so there is no need to have the `Array` (@azadoks do you remember any detail?) However, there are lots of allocations when called - I think the function itself can be optimised a bit.

zhubonan: Maybe it can introduce issue with types other than Float64? I am not sure...

mfherbst: Sorry it took me a bit to follow up with this. So it actually was two type instabilities. The one you found @zhubonan, but also one we missed earlier and which in fact caused us to put the explicit return type. With that fixed the return type can also be removed.

mfherbst:
Thanks @zhubonan for your help on getting this fixed!

zhubonan:
Thanks for the update, glad to see the type stability getting enhanced!

PR #851, opened 2023-04-20
Multicomponents

epolack:
### API question to settle
- [ ] Possibility to have components for `TermKinetic.kinetic_energies`, `RealSpaceMultiplication.potential` to prevent loops on the components; but maybe in a second time (i.e., future PR). Implementing one at a time needed generalisations?
  It was what was done for the preconditioner before (e.g., hash 9869eb59137997162f9660167db14177c14dbeb0)

### Remaining points

∅

### Future PR
- use a [BlochWaves structure](https://github.com/JuliaMolSim/DFTK.jl/pull/931) and BlochWave!!
-  add occupation to `BlochWaves`
- not sure what to print by default (`Base.show` methods); maybe report to `ψk::BlochWave`.
- see for how to do in a memory friendly-way operations such that
  ```julia
  for (σ, n) in eachindex(ψk::BlochWave; region=[:components, :bands])
  ```
- Tullio.jl for when doing tricky stuff?

### DIFF
--- a/benchmark/regression/testcases.jl
+++ b/benchmark/regression/testcases.jl
@@ -39,8 +39,7 @@ end
 
         n_electrons = testcase.n_electrons
         n_bands = div(n_electrons, 2, RoundUp)
-        ψ = [Matrix(qr(randn(ComplexF64, length(G_vectors(basis, kpt)), n_bands)).Q)
-             for kpt in basis.kpoints]
+        ψ = [DFTK.random_orbitals(basis, kpt, n_bands) for kpt in basis.kpoints]
         filled_occ = DFTK.filled_occupation(model)
         occupation = [filled_occ * rand(n_bands) for _ = 1:length(basis.kpoints)]
         occ_scaling = n_electrons / sum(sum(occupation))


antoine-levitt:
We discussed the following: by default things like psi (to be discussed what rho would be: do we want the full rho[alpha, beta, i, j, k] ?) would be arrays of arrays static arrays (yes...), ie psi[k][G,n][alpha]. This is the same in memory as psi[k][alpha,G,n]. There would be convenience functions nest(psi) : psi[alpha, G, n] -> psi[G, n][alpha], denest(psi) : psi[alpha, G, n] -> psi[G, n][alpha], and denest(psi, alpha) which would select only component alpha. The main objection to this is that it's pretty annoying to work with such a nested structure: eg printing by default is atrocious...

antoine-levitt:
Ping @mfherbst does the above seem reasonable? We'd like to go ahead soon

mfherbst:
Thanks for the reminder. I'll take a look!

mfherbst:
Ah now I get why componentarrays don't work: They encapsulate the wrong way round.

But in that sense your basic "element" becomes not `T`, but some StaticArray type, right? So maybe it could help to make things simpler if this is exposed more. Like right now the `T` of PlaneWaveBasis is used to deduce a lot of types. But now it should probably be the respective `SArray{T}` right?

What I really would like to avoid is to add a lot of additional complexity to the code, which then gets combinatorially multiplied with AD, GPU and other things, such that we end up writing a lot of  workaround code to get things working with our highly specific data layout.

mfherbst:
BTW. I guess it might be good to schedule a call to discuss this after the long weekend.

src/Psi.jl
```diff
@@ -0,0 +1,57 @@
+struct PlaneWave#{T <: Real}
```
LaurentVidal95, line 1: I would have put "BlochState" here or something not used for `e_G` in the doc.

src/Psi.jl
```diff
@@ -0,0 +1,28 @@
+struct BlockWaves{T, Tψ, Basis <: AbstractBasis{T}}
+    basis::Basis
+    # TODO: Find better name, and better way to type it
+    value::Vector{VT} where {VT <: AbstractArray3{Tψ}}
```
antoine-levitt, line 4: just call it psi?

src/Psi.jl
```diff
@@ -0,0 +1,28 @@
+struct BlockWaves{T, Tψ, Basis <: AbstractBasis{T}}
```
antoine-levitt, line 1: BlochWaves more than BlockWaves, definitely, the latter just looks like a typo

src/Psi.jl
```diff
@@ -0,0 +1,28 @@
...
+    basis::Basis
+    # TODO: Find better name, and better way to type it
+    value::Vector{VT} where {VT <: AbstractArray3{Tψ}}
+    n_components::Int
+    n_bands::Int
+end
+function BlockWaves(basis::PlaneWaveBasis, ψ::Vector{VT}; do_nimp=false) where {VT <: AbstractArray3}
+    n_bands = minimum(size.(ψ, 3))
+    # Remove do_nimp, maybe n_bands == nothing if not same number of bands.
```
antoine-levitt, line 10: why do you need the field n_bands? Just don't assume the number of bands is constant?

src/Psi.jl
```diff
@@ -0,0 +1,28 @@
...
+    do_nimp || @assert n_bands == maximum(size.(ψ, 3))
+    BlockWaves(basis, ψ, basis.model.n_components, n_bands)
+end
+BlockWaves(basis) = BlockWaves(basis, [Array{eltype(basis), 3}(undef, 0, 0, 0)], 0, 0)
```
antoine-levitt, line 14: Not sure this should exist

src/Psi.jl
```diff
@@ -0,0 +1,28 @@
...
+Base.length(ψ::BlockWaves) = length(ψ.value)
+Base.iterate(ψ::BlockWaves, args...) = Base.iterate(ψ.value, args...)
+Base.getindex(ψ::BlockWaves, indices...) = ψ.value[indices...]
+Base.isnothing(ψ::BlockWaves) = iszero(ψ.n_bands)
+Base.similar(ψ::BlockWaves, args...) = similar(ψ.value, args...)
+
+# TODO: Have efficient copy function with consistency checks when only the basis is changed.
+# TODO: Do not change flatten the size of array by default (1:1)
+viewcomp(ψk::AbstractArray3, σ) = ψk[σ, :, :]
```
antoine-levitt, line 23: view_component is not much longer

antoine-levitt: Should it not be a `@view`?

src/Psi.jl
```diff
@@ -0,0 +1,28 @@
...
+viewcomp(ψ::BlockWaves, σ) = [viewcomp(ψk, σ) for ψk in ψ]
+denest(ψ::BlockWaves; σ=1:ψ.basis.model.n_components) = [viewcomp(ψk, σ) for ψk in ψ]
```
antoine-levitt, line 25: add doc of what this does

src/eigen/preconditioners.jl
```diff
@@ -56,6 +56,18 @@ end
     end
     Y
 end
+@views function ldiv!(Y, P::PreconditionerTPA, R::AbstractArray3)
```
antoine-levitt, line 59: merge with previous function

src/fft.jl
```diff
@@ -256,25 +301,32 @@ end
 Plan a FFT of type `T` and size `fft_size`, spending some time on finding an
 optimal algorithm. (Inplace, out-of-place) x (forward, backward) FFT plans are returned.
 """
-function build_fft_plans!(tmp::Array{Complex{Float64}})
-    ipFFT = FFTW.plan_fft!(tmp; flags=FFTW.MEASURE)
-    opFFT = FFTW.plan_fft(tmp;  flags=FFTW.MEASURE)
+function build_fft_plans!(tmp::Array{Complex{Float64}}, dims::AbstractVector{Int}=1:ndims(tmp))
```
antoine-levitt, line 304: doc, and don't type

antoine-levitt: also kwarg

src/terms/Hamiltonian.jl
```diff
@@ -168,14 +173,26 @@ end
 
     Hψ
 end
+@views @timing "DftHamiltonian multiplication" function LinearAlgebra.mul!(Hψ::AbstractVecOrMat,
```
antoine-levitt, line 176: Comment with why this is needed

src/terms/kinetic.jl
```diff
@@ -45,10 +46,18 @@ end
 
     E = zero(T)
     for (ik, ψk) in enumerate(ψ)
-        for iband = 1:size(ψk, 2)
-            ψnk = @views ψk[:, iband]
-            E += (basis.kweights[ik] * occupation[ik][iband]
-                  * real(dot(ψnk, Diagonal(term.kinetic_energies[ik]), ψnk)))
+        @static if VERSION < v"1.9"  # eachslice(ψk; dims) not yet available
```
antoine-levitt, line 49: Put more standard code path first, and workaround in else

antoine-levitt: Also can't we just use Compat?

antoine-levitt: But I have to say, I find the pre-1.9 version much more readable...

src/terms/operators.jl
```diff
@@ -22,13 +22,33 @@ function LinearAlgebra.mul!(Hψ::AbstractVector, op::RealFourierOperator, ψ::Ab
...
+        @views for σ in 1:size(ψ, 1)
+            mul!(Hψ[σ, :, i], op, ψ[σ, :, i])
+        end
     end
     Hψ
 end
 Base.:*(op::RealFourierOperator, ψ) = mul!(similar(ψ), op, ψ)
+# Default transformation: from two components tensors to diagonal matrices for each
+# reciprocal vector.
+# (Re)-naming is ***. Matrix → tensor.
```
antoine-levitt, line 37: Rename tensor to Array

src/terms/operators.jl
```diff
@@ -39,9 +59,10 @@ struct NoopOperator{T <: Real} <: RealFourierOperator
     kpoint::Kpoint{T}
 end
 apply!(Hψ, op::NoopOperator, ψ) = nothing
-function Matrix(op::NoopOperator)
+function tensor(op::NoopOperator)
     n_Gk = length(G_vectors(op.basis, op.kpoint))
-    zeros_like(op.basis.G_vectors, eltype(op.basis), n_Gk, n_Gk)
+    n_comps = op.basis.model.n_components
+    zeros_like(op.basis.G_vectors, eltype(op.basis), n_comps, n_Gk, n_Gk)
```
antoine-levitt, line 65: Why is this not n_comps x n_G x n_comps x n_G ?

src/terms/operators.jl
```diff
@@ -22,13 +22,33 @@ function LinearAlgebra.mul!(Hψ::AbstractVector, op::RealFourierOperator, ψ::Ab
...
+# (Re)-naming is ***. Matrix → tensor.
+function Matrix(op::RealFourierOperator)
+    tensor_op = tensor(op)
+    n_Gk = length(G_vectors(op.basis, op.kpoint))
+    n_comps = op.basis.model.n_components
+    H = zeros(complex(eltype(op.basis)), n_comps*n_Gk, n_comps*n_Gk)
+    for j in eachindex(G_vectors(op.basis, op.kpoint))
+        for i in eachindex(G_vectors(op.basis, op.kpoint))
+            for σ in 1:n_comps
+                H[σ + (i-1)*n_comps, σ + (j-1)*n_comps] = tensor_op[σ, i, j]
```
antoine-levitt, line 46: Just use reshape?

src/terms/operators.jl
```diff
@@ -87,7 +111,16 @@ end
 function apply!(Hψ, op::FourierMultiplication, ψ)
     Hψ.fourier .+= op.multiplier .* ψ.fourier
 end
-Matrix(op::FourierMultiplication) = Array(Diagonal(op.multiplier))
+function tensor(op::FourierMultiplication)
```
antoine-levitt, line 114: Use `repeat` maybe?

antoine-levitt (commented):
Some misc things but in general I think this is coming along nicely! Very cool that not that many new stuff is needed.

src/fft.jl
antoine-levitt: no I mean dims is usually a kwarg rather than an optional arg in julia

src/terms/operators.jl
antoine-levitt: Yeah sorry repeat is no good. But this looks OK:
```
julia> a = [1 2;3 4]
2×2 Matrix{Int64}:
 1  2
 3  4

julia> kron(a, I(2))
4×4 Matrix{Int64}:
 1  0  2  0
 0  1  0  2
 3  0  4  0
 0  3  0  4
```

antoine-levitt:
```
to_composite_σG
- to_composite_σG(basis, kpt, ψk) : ψk[σ,G] -> ψk[σG]
- to_composite_σG(basis, kpt, ψk) : ψk[σ,G,n] -> ψk[σG,n]
- to_composite_σG(basis, ψ) : ψ[k][σ,G] -> ψ[k][σG]
- to_composite_σG(basis, ψ) : ψ[k][σ,G,n] -> ψ[k][σG,n]

from_composite_σG
```

src/common/ortho.jl
```diff
@@ -1,9 +1,5 @@
-@timing function ortho_qr(φk::ArrayType) where {ArrayType <: AbstractArray}
-    x = convert(ArrayType, qr(φk).Q)
-    if size(x) == size(φk)
-        return x
-    else
-        # Sometimes QR (but funnily not always) CUDA messes up the size here
```
antoine-levitt, line 6: @mfherbst do you remember what was going on there? The API for qr is really annoying, because some people use the "full" QR decomposition (with Q being a NxN matrix if A was Nxn), so that this code is possibly a massive hidden performance trap which we should report to CUDA

mfherbst: If I recall correctly the sizes it reported (i.e. `size(qr(X)`) were fine, just when you did the `Matrix(qr(X).Q)` then you got a full NxN array back.

src/Psi.jl
```diff
@@ -0,0 +1,57 @@
+struct PlaneWave#{T <: Real}
+    kpoint::Kpoint#{T}
+    # No type because of AD
```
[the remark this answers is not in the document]
mfherbst: sorry not sure where these came from ...

antoine-levitt:
> If I recall correctly the sizes it reported (i.e. size(qr(X)) were fine, just when you did the Matrix(qr(X).Q) then you got a full NxN array back.

OK but then this is massively expensive, and by just taking the [:, 1:n] we are hiding a massive perf issue. Do you remember how to trigger this?

mfherbst:
It was really random. Like I had it once, but then removed it later and it was ok, but had to reintroduce the workaround at some other point to make it work. It could have been to do with versions of CUDA.jl the underlying cuda library etc. Main point was to get it to work. The CUDA code is horribly slow anyway, so I have not yet bothered to investigate.

mfherbst:
> Could you give me some feedback on this PR? It takes quite some time to keep it up to date with the master branch.

Understood. I'm pretty busy right now, but I'll do it as soon as I can.

PR #853, opened 2023-05-03
API change for Ewald and pairwise computations

epolack:


src/terms/ewald.jl
```diff
@@ -21,16 +23,13 @@ end
 
 @timing "forces: Ewald" function compute_forces(term::TermEwald, basis::PlaneWaveBasis{T},
```
antoine-levitt, line 24: move the timing, since this is now trivial

src/terms/ewald.jl
```diff
@@ -9,10 +9,12 @@ struct Ewald end
...
 
 struct TermEwald{T} <: Term
-    energy::T  # precomputed energy
+    energy::T                # precomputed energy
+    forces::Vector{Vec3{T}}  # and forces
 end
 function TermEwald(basis::PlaneWaveBasis{T}) where {T}
-    TermEwald(T(energy_ewald(basis.model)))
+    energy, forces = energy_forces_ewald(basis.model; compute_forces=true)
+    TermEwald(T(energy), forces)
```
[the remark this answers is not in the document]
antoine-levitt: It was probably for performance reasons, you can remove it and see if it breaks

antoine-levitt:
Maybe open an issue at interval arithmetic if you're able to reproduce it outside of DFTK?

antoine-levitt:
Then maybe just post it without a MWE and see if the error message rings a bell for them

dsestil:
It is issue of interval arithmetic. `IrrationalConstants.invsqrtπ * @interval(1., 10.)` raises error in windows only. Looks like this issue is solved but not published. (https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/556)

antoine-levitt:
That's reasonable but wow, 100s precompile? I get much less than this I think

antoine-levitt:
On 1.9 ?

antoine-levitt:
```
julia> @time Pkg.precompile("DFTK")
Precompiling DFTK
  1 dependency successfully precompiled in 49 seconds. 119 already precompiled.
 49.997652 seconds (543.25 k allocations: 59.760 MiB, 0.02% gc time, 0.04% compilation time)

julia> @time using DFTK
  2.474813 seconds (5.53 M allocations: 370.662 MiB, 4.88% gc time, 11.11% compilation time: 44% of which was recompilation)
```
This is on 1.9.0. Weird...

src/terms/ewald.jl
```diff
@@ -9,28 +9,23 @@ struct Ewald end
...
 
 struct TermEwald{T} <: Term
-    energy::T  # precomputed energy
+    energy::T                # precomputed energy
+    forces::Vector{Vec3{T}}  # and forces
 end
-function TermEwald(basis::PlaneWaveBasis{T}) where {T}
-    TermEwald(T(energy_ewald(basis.model)))
+@timing "precomp: Ewald" function TermEwald(basis::PlaneWaveBasis{T}) where {T}
+    energy, forces = energy_forces_ewald(basis.model; compute_forces=true)
```
mfherbst, line 16: If we anyway always precompute them. Why don't we always compute them? Would make things even simpler.

mfherbst: Also you can now use
```
(; energy, forces) = 
```
as we moved to at least Julia 1.7.

src/terms/pairwise.jl
```diff
@@ -17,40 +17,34 @@ function PairwisePotential(V, params; max_radius=100)
     params = Dict(minmax(key[1], key[2]) => value for (key, value) in params)
     PairwisePotential(V, params, max_radius)
 end
-function (P::PairwisePotential)(basis::PlaneWaveBasis{T}) where {T}
-    E = energy_pairwise(basis.model, P.V, P.params; P.max_radius)
-    TermPairwisePotential(P.V, P.params, T(P.max_radius), E)
+@timing "precomp: Pairwise" function (P::PairwisePotential)(basis::PlaneWaveBasis{T}) where {T}
+    E, forces = energy_forces_pairwise(basis.model, P.V, P.params; P.max_radius,
+                                       compute_forces=true)
```
mfherbst, line 22: Same argument as for ewald.

src/terms/pairwise.jl
```diff
@@ -111,18 +104,21 @@ function energy_pairwise(lattice, symbols, positions, V, params;
...
-    if !isnothing(forces)
-        forces .= forces_pairwise
+    res = (; energy)
+
+    if compute_forces
+        res = merge(res, (; forces))
     end
-    energy
+
+    res
```
mfherbst, line 123: see ewald

mfherbst (commented):
Very nice, sorry for the delay. I would just give up on the flag completely and always compute the forces.

PR #854, opened 2023-05-11
Rewrite Wannier interface

qiaojunfeng:
Finally, it's here! 🚀 
Now I have finished the rewriting of Wannier function interface, including
- use [WannierIO.jl](https://github.com/qiaojunfeng/WannierIO.jl) for reading/writing Wannier90 files
- In addition to the current `wannier90_jll` example, added an example of using [`Wannier.jl`](https://github.com/qiaojunfeng/Wannier.jl) to construct Wannier functions
- For initial projections, in addition to the current random Gaussain initial guess, added
    - Hydrogenic orbital (i.e., analytic solution of radial function) projection with different angular momentum
    - Pseudized-atomic orbitals (from pseudopotentials) projection, currently only for `PspUpf` since it contains the `pswfcs` field
        - this should also enable us to compute PDOS, projected bands, and DFT+U in the future
    - SCDM projection (QR factorization of wavefunctions)
- support `:collinear` calculation

Still have to add some more tests, but maybe now we can discuss the file/function namings, design, etc.


qiaojunfeng:
Oh it miserably failed since my packages are not registered yet, it's ongoing https://github.com/qiaojunfeng/WannierIO.jl/issues/1

antoine-levitt:
Oh wow that's awesome, thanks so much for doing this! Give me a couple days and I'll review. In the meantime @LaurentVidal95 can you maybe take a look?

LaurentVidal95:
Super nice ! I'll give it a look tomorrow with pleasure :+1:

mfherbst:
Very nice. Thanks for the PR @qiaojunfeng. I'll make a pass as well within the next few days.

examples/wannier.jl
```diff
@@ -0,0 +1,92 @@
...
+exclude_bands = DFTK._default_exclude_bands(scfres)
+basis, ψ, eigenvalues = DFTK.unfold_scfres_wannier(scfres, exclude_bands)
+
+# Then compute the SCDM initial guess with [`compute_amn_scdm`](@ref)
+# Since this is an entangled case, we need a weight factor, using `erfc` function.
+# Note that the unit of `μ` and `σ` are `Ha`, as in the DFTK convention.
+# Here we choose these numbers by inspecting the band structure, you can also
+# test with different values to see how it affects the result.
+μ, σ = 0.0, 0.01
+f = DFTK.scdm_f_erfc(basis, eigenvalues, μ, σ)
```
mfherbst, line 68: If you make `\mu` and `\sigma` kwargs, then this is obvious without the extra line assigning `\mu` and `sigma`

examples/wannier.jl
```diff
@@ -0,0 +1,92 @@
...
+# and construct 5 MLWFs
+n_wann = 5
+A = DFTK.compute_amn_scdm(basis, ψ, n_wann, f)
```
mfherbst, line 71: same here.

mfherbst: also better spell it out `n_wannier`

examples/wannier.jl
```diff
@@ -0,0 +1,92 @@
...
+
+# The Wannier.jl interface is very similar to the Wannier90 example,
+# except that the function `run_wannier` is used instead of `run_wannier90`.
+# To further explore the functionalities of the MLWF interface, in this example
+# we use SCDM to generate a better initial guess for the MLWFs
+# (by default, `run_wannier` will use random initial guess which is not good).
+# We need to first unfold the `scfres` to a MP kgrid for Wannierization,
+# and remove the possibly unconverged bands (bands above `scfres.n_bands_converge`)
+exclude_bands = DFTK._default_exclude_bands(scfres)
+basis, ψ, eigenvalues = DFTK.unfold_scfres_wannier(scfres, exclude_bands)
```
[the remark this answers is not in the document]
mfherbst: (BTW I'm reading this top to bottom, so some of my comments might better apply to the wannierio.jl files)

examples/wannier.jl
```diff
@@ -0,0 +1,92 @@
...
+# and construct 5 MLWFs
+n_wann = 5
+A = DFTK.compute_amn_scdm(basis, ψ, n_wann, f)
+# we pass the `A` matrix to `run_wannier`, so it will skip the `compute_amn` step
+wann_model, = run_wannier(
+    scfres;
+    fileprefix="wannier/graphene",
+    n_wann,
+    A,
+    dis_froz_max=0.1,
```
mfherbst, line 78: I guess you use these cryptic keyword arguments for compatability with wannier90, but I'm really not a fan of this. Can we not avoid that somehow and give it more descriptive names?

If not: at least add a comment here.

examples/wannier.jl
mfherbst: We should bikeshed these names ... `scdm_f_erfc` is quite cryptic to someone who does not already know wannier90.

examples/wannier.jl
```diff
@@ -0,0 +1,92 @@
...
+# Note that the unit of `μ` and `σ` are `Ha`, as in the DFTK convention.
+# Here we choose these numbers by inspecting the band structure, you can also
+# test with different values to see how it affects the result.
+μ, σ = 0.0, 0.01
+f = DFTK.scdm_f_erfc(basis, eigenvalues, μ, σ)
+# and construct 5 MLWFs
+n_wann = 5
+A = DFTK.compute_amn_scdm(basis, ψ, n_wann, f)
+# we pass the `A` matrix to `run_wannier`, so it will skip the `compute_amn` step
+wann_model, = run_wannier(
```
mfherbst, line 73: Does it need to be called `run_wannier` ? I think `wannierize` or sth might be much better. We only used this name because wannier90 has such a bad interface and there is very little you can do other than just run the whole black box.

mfherbst: Also why `wann_model,` instead of just `wann_model` ?

src/external/stubs.jl
```diff
@@ -1,17 +1 @@
...
-Wannier90 for the disentanglement may be added as keyword arguments.
-The function returns the `fileprefix`.
-
-!!! warning "Experimental feature"
-    Currently this is an experimental feature, which has not yet been tested
-    to full depth. The interface is considered unstable and may change
-    incompatibly in the future. Use at your own risk and please report bugs
-    in case you encounter any.
-"""
-function run_wannier90 end
```
mfherbst, line 17: Why did you remove that?

The idea is that the function is *always* available in DFTK, but the *methods* are only added once wannier90 and wannier are loaded (using the requires magic). I would keep that also for `run_wannier` and `run_wannier90` (or whatever the methods will end up being called).

test/external/wannier.jl
```diff
@@ -0,0 +1,28 @@
+using Test
+using DFTK
+include("../testcases.jl")
+
+if !Sys.iswindows() && mpi_nprocs() == 1
```
mfherbst, line 5: This gate was only needed because wannier90 is not available in Yggdrasil on windows. But I hope your code works there?

mfherbst: I suppose one could save some code duplication by merging this with the wannier90 test.

src/terms/nonlocal.jl
```diff
@@ -227,3 +227,85 @@ function build_form_factors(psp, qs::Array)
     end
     form_factors
 end
+
```
mfherbst, line 230: These changes are again a lot of code duplication, which can be avoided with the tricks mentioned above.

src/pseudo/PspUpf.jl
```diff
@@ -181,6 +182,17 @@ function eval_psp_projector_fourier(psp::PspUpf, i, l, q::T)::T where {T <: Real
     4T(π) * s
 end
 
+# For UPFs, the integral is transformed to the following sum:
+# 4π Σ{k} j_l(q r[k]) (r[k]^2 p_{il}(r[k]) dr[k])
+function eval_psp_pswfc_fourier(psp::PspUpf, i, l, q::T)::T where {T <: Real}
```
mfherbst, line 187: Ok so the pswfc essentially have the same structure as the projectors. Therefore we should not use prefixes but marker structs for the dispatch (e.g. `struct Proj end; struct PsWfc end`). @azadoks might have some plans for this already in PspIO, so ideally we use something here that can be easily integrated with his ideas.

src/external/wannierio.jl
```diff
@@ -0,0 +1,693 @@
+using Dates
+using Printf: @sprintf
```
mfherbst, line 2: That should just be available actually

src/external/wannierio.jl
```diff
@@ -0,0 +1,693 @@
...
+using SpecialFunctions: erfc
```
mfherbst, line 3: same

src/external/wannierio.jl
```diff
@@ -0,0 +1,693 @@
...
+
+"""
+Return the kpoint index shift for spin-polarized case.
+
+For spin-polarized calculation, first half of kpoints are for spin up,
+and the second half are for spin down.
+"""
+function _get_ik_shift(basis::PlaneWaveBasis, spin::Int)
```
mfherbst, line 11: we have a function for that kspin_range or similar

src/external/wannierio.jl
```diff
@@ -0,0 +1,693 @@
...
+    α = [q[3] for q in lnα]
+    guess = guess_amn_hydrogenic(basis, centers, l; n, α)
+    guess_filtered(kpt) = guess(kpt)[:, indices]
+    guess_filtered
+end
+
+"""
+Random Gaussian initial guess for Wannier functions.
+"""
+@timing function guess_amn_random(basis::PlaneWaveBasis, n_wann::Integer)
```
mfherbst, line 265: in general things like `_random`, `_hydrogenic` should be done by dispatch on a marker struct in julia.

src/external/wannierio.jl
```diff
@@ -0,0 +1,693 @@
+using Dates
+using Printf: @sprintf
+using SpecialFunctions: erfc
+
```
mfherbst, line 4: You generally use far too restrictive type bounds. Let's discuss this in person @qiaojunfeng.

src/external/wannierio.jl
```diff
@@ -0,0 +1,693 @@
...
+- `kpb_G`: the shifting vectors of k+b, size: `(3, n_bvecs, n_kpts)`.
+    1st index: x,y,z component of b-vector,
+    2nd index: b-vector,
+    3rd index: shifting vector to get the shifted k-point `kpb_k[ib,ik]` to
+    the actual k+b point, thus the actual k+b is `kpb_k[ib,ik] + kpb_G[:,ib,ik]`.
+
+# Keyword arguments
+- `spin`: the spin channel to be used, either `1` or `2`.
+"""
+@timing function compute_mmn(
```
mfherbst, line 53: If this computes the overlaps, than that's what it should be called. Julia allows more than 8 character function names ;).

src/external/wannierio.jl
```diff
@@ -0,0 +1,693 @@
...
+    @assert size(kpb_G) == (3, n_bvecs, n_kpts)
+
+    kpts = krange_spin(basis, spin)
+    @assert length(kpts) == n_kpts
+    ik_shift = _get_ik_shift(basis, spin)
+
+    M = zeros(eltype(ψ[1]), n_bands, n_bands, n_bvecs, n_kpts)
+
+    for ik in kpts
+        k = basis.kpoints[ik]
```
mfherbst, line 69: Just use `enumerate(basis.kpoints)` or the krange spin thing we have

src/external/wannierio.jl
```diff
@@ -0,0 +1,693 @@
...
+            equivalent_G_vectors = [
+                (iGk, index_G_vectors(basis, kpb, Gk + G_shift))
+                for (iGk, Gk) in enumerate(G_vectors(basis, k))
+            ]
+            iGk = [eqG[1] for eqG in equivalent_G_vectors if !isnothing(eqG[2])]
+            iGkpb = [eqG[2] for eqG in equivalent_G_vectors if !isnothing(eqG[2])]
+
+            # Compute overlaps
+            for n in 1:n_bands
+                for m in 1:n_bands
```
mfherbst, line 88: merge these loops

src/external/wannierio.jl
```diff
@@ -0,0 +1,693 @@
...
+                    M[m, n, ib, ik0] = dot(ψ[ik][iGk, m], ψ[ikpb][iGkpb, n])
+                end
+            end
+        end
+    end
```
mfherbst, line 93: this many nestings point that there is something better that can be done.

src/external/wannierio.jl
```diff
@@ -0,0 +1,693 @@
...
+
+Same as Wannier90 user guide Table 3.3.
+
+# Arguments
+- `r`: radial grid
+- `n`: principal quantum number
+- `α`: diffusivity, ``\frac{Z}/{a}`` where ``Z`` is the atomic number and
+    ``a`` is the Bohr radius.
+"""
+function radial_hydrogenic(r::AbstractVector{T}, n::Integer, α::Real=1.0) where {T<:Real}
```
mfherbst, line 108: these hydrogenic things should live in common or maybe even in pspio. They sound like the could be useful for other things, too.

src/external/wannierio.jl
```diff
@@ -0,0 +1,693 @@
...
+    kpts = krange_spin(basis, spin)
+    eigs = eigenvalues[kpts]
+
+    n_kpts = length(kpts)
+    n_bands = length(eigs[1])
+    E = zeros(eltype(eigs[1]), n_bands, n_kpts)
+
+    for (ik, εk) in enumerate(eigs)
+        for (n, εnk) in enumerate(εk)
+            E[n, ik] = auconvert(Unitful.eV, εnk).val
```
mfherbst, line 470: better use `ustrip`.

src/external/wannierio.jl
```diff
@@ -0,0 +1,693 @@
...
+Unfold symmetries, exclude bands.
+
+# Arguments
+- `exclude_bands`: the band indices to be excluded from the Wannierization,
+    usually the semicore states.
+"""
+function unfold_scfres_wannier(
+    scfres::NamedTuple,
+    exclude_bands::Union{AbstractArray{<:Integer},Nothing}=nothing,
+)
```
mfherbst, line 566: @LaurentVidal95 Didn't you write a function at some point that does more or less this?

src/external/wannierio.jl
```diff
@@ -0,0 +1,693 @@
...
+- remaining keyword arguments are passed to `get_wannier90_win` and written
+    to the `win` file.
+
+!!! warning
+
+    For simplicity, here only use random initial guess for Wannier functions.
+    In principle, one should use better initial guesses, e.g.,
+    [`guess_amn_hydrogenic`](@ref), [`guess_amn_psp`](@ref).
+"""
+function save_wannier(
```
mfherbst, line 627: Is there a way to run Wannier.jl without saving these?

src/external/wannierio.jl
```diff
@@ -0,0 +1,693 @@
...
+    @info "Wannier matrices written to prefix=$(fileprefix)"
+
+    (; win, M, A, E)
+end
+
+function _default_exclude_bands(scfres::NamedTuple)
+    n_bands = length(scfres.eigenvalues[1])
+    exclude_bands = (scfres.n_bands_converge+1):n_bands
+    exclude_bands
+end
```
mfherbst, line 693: why do you need this?

src/external/wannierio.jl
mfherbst: Also I don't like this monolithic function. I think it's much better to have individual functions to write these files (e.g. by providing DFTK-specific methods to `WannierIO.write_amn` etc.) and then the user just has to call them in the right order and supplying the right guess methods etc.

antoine-levitt:
Thanks for the pass @mfherbst , I'll take the next one!

PR #858, opened 2023-05-17
Experimental AMDGPU support

mfherbst:
Works now with a pre-installed AMDGPU stack, todo:

- [x] Fix unit test issues (maybe due to the changes in fft.jl in combination with generic floats ?)
- [x] Some unit tests for amdgpu
- [x] See if Anderson acceleration can be enabled. (Let's see about this later)


src/workarounds/roc_arrays.jl
```diff
@@ -0,0 +1 @@
+synchronize_device(::GPU{<:AMDGPU.ROCArray}) = AMDGPU.Device.sync_workgroup()
```
pxl-th, line 1: `sync_workgroup` is meant to be called from a kernel.
AMDGPU has equivalent to CUDA `AMDGPU.synchronize` method.

```suggestion
synchronize_device(::GPU{<:AMDGPU.ROCArray}) = AMDGPU.synchronize()
```

pxl-th (commented):
FFT seems to be working fine.
Segfalts come from missing rocSOLVER `qr` used in [`ortho_qr`](https://github.com/JuliaMolSim/DFTK.jl/blob/bfbe7afb89c1d0cbfd51adfa2cb7b931ab31e477/src/common/ortho.jl#L2).

pxl-th:
Now `amgpu_simple.jl` fails with:
```
ERROR: LoadError: Scalar indexing is disallowed.
Invocation of getindex resulted in scalar indexing of a GPU array.
This is typically caused by calling an iterating implementation of a method.
Such implementations *do not* execute on the GPU, but very slowly on the CPU,
and therefore are only permitted from the REPL for prototyping purposes.
If you did intend to index this array, annotate the caller with @allowscalar.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] assertscalar(op::String)
   @ GPUArraysCore ~/.julia/packages/GPUArraysCore/uOYfN/src/GPUArraysCore.jl:103
 [3] getindex
   @ ~/.julia/packages/GPUArrays/5XhED/src/host/indexing.jl:9 [inlined]
 [4] mul!(Hψ::ROCMatrix{ComplexF64, AMDGPU.Runtime.Mem.HIPBuffer}, H::Hamiltonian, ψ::ROCMatrix{ComplexF64, AMDGPU.Runtime.Mem.HIPBuffer})
   @ DFTK ~/code/DFTK.jl/src/terms/Hamiltonian.jl:83
 [5] *(H::Hamiltonian, ψ::ROCMatrix{ComplexF64, AMDGPU.Runtime.Mem.HIPBuffer})
   @ DFTK ~/code/DFTK.jl/src/terms/Hamiltonian.jl:88
 [6] top-level scope
   @ ~/code/DFTK.jl/examples/amdgpu_simple.jl:23
in expression starting at /home/pxl-th/code/DFTK.jl/examples/amdgpu_simple.jl:23
```

pxl-th:
Same issue is with CUDA though

src/workarounds/roc_arrays.jl
mfherbst: Thanks!

mfherbst:
Hi @pxl-th Thanks for your help. Yes, that probably should not happen. I'll take a look.

mfherbst:
I tried to use the code with AMDGPU.jl master, but I get some issues, see https://github.com/JuliaGPU/AMDGPU.jl/commits/master#461.

mfherbst:
Getting reasonably far now. Current breaking point is `cholesky(::Hermitian{<:RocArray})` which leads to scalar indexing (i.e. probably some fallback conversion to CPU arrays is invoked).

mfherbst:
@abussy DFTK works now on AMDGPUs (tested on bulk silicon so far). I might try to merge at least a preliminary version soon-ish to allow us to experiment.

PR #860, opened 2023-05-20
Fix various smaller bugs

mfherbst:
Fixes for smaller bugs I triggered while playing around today.

### DIFF
--- a/src/eigen/lobpcg_hyper_impl.jl
+++ b/src/eigen/lobpcg_hyper_impl.jl
@@ -118,7 +118,7 @@ end
 # Perform a Rayleigh-Ritz for the N first eigenvectors.
 @timing function rayleigh_ritz(X, AX, N)
     XAX = X' * AX
-    @assert all(!isnan, XAX)
+    @assert !any(isnan, XAX)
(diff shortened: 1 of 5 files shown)


src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -118,7 +118,7 @@ end
 # Perform a Rayleigh-Ritz for the N first eigenvectors.
 @timing function rayleigh_ritz(X, AX, N)
     XAX = X' * AX
-    @assert all(!isnan, XAX)
+    @assert !any(isnan, XAX)
```
antoine-levitt, line 121: any particular reason for this?

mfherbst: Yes. On cuda funnily the two statements are not equal. I had the original assertion fail even without nans in the array.

antoine-levitt: What? Sounds like a bug that should be reported?

mfherbst: Yeah, probably. In any case we very easily get NaNs when using CUDA, so we need to give this a closer look anyway.

PR #861, opened 2023-06-20
Reallow to skip precompilation

epolack:
Not sure why the possibility was removed. It is useful for API-breaking changes.


mfherbst:
Also see the info box at the bottom of https://docs.dftk.org/stable/guide/installation/#Developer-setup

PR #868, opened 2023-07-24
AD for forces

epolack:
Precompiling is the same (but slow ~100s, I do not understand why; I had this also in https://github.com/JuliaMolSim/DFTK.jl/pull/853, then it reverse to normal timings without any changes from my part…). 

On a simple silicon supercell of size 2×2×2 the cold and hot timings are similar:
master: 22s then 6.29s
branch: 24.2s then 6.57s

So if the change seems fine for you, I will test this on more robust test cases (ideas are welcome).

### DIFF
--- a/src/terms/local.jl
+++ b/src/terms/local.jl
@@ -68,7 +68,7 @@ end
 Atomic local potential defined by `model.atoms`.
 """
 struct AtomicLocal end
-function (::AtomicLocal)(basis::PlaneWaveBasis{T}) where {T}
+function compute_local_potential(basis::PlaneWaveBasis{T}; positions=basis.model.positions) where {T}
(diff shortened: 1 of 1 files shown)


antoine-levitt:
Not too sure about this. Generally, trying to add more AD introduces more headaches down the line. Esp. for forces it's probably possible to use adjoint tricks, which I don't think we do currently but would be probably easier to do with explicit derivatives. Also, here you have to do a lot more FFTs. Is there a motivation to this?

(as an aside, we have the pattern "build a superposition of radial things" quite a lot, we should factor it out eventually)

mfherbst:
I'm also not too sure about this as we don't actually loose that much code. Also not sure how well this plays with GPU. I think we should have a proper think about how to do this, if we really want to.

PR #869, opened 2023-07-27
Fix for Ewald

epolack:
Sorry, I missed a change in Ewald API. I have noticed two other small bugs in `diag.jl` and `cg.jl`.

### DIFF
--- a/src/response/cg.jl
+++ b/src/response/cg.jl
@@ -1,7 +1,7 @@
 using LinearMaps
 
 function default_cg_print(info)
-    @printf("%3d\t%1.2e\n", info.iterations, info.residual_norm)
+    @printf("%3d\t%1.2e\n", info.n_iter, info.residual_norm)
 end
 
 """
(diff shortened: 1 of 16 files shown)


src/terms/ewald.jl
```diff
@@ -47,7 +47,7 @@ function energy_forces_ewald(lattice::AbstractArray{T}, charges, positions;
     # TODO should something more clever be done here? For now
     # we assume that we are not interested in the Ewald
     # energy of non-3D systems
-    any(iszero.(eachcol(lattice))) && return zero(T)
+    any(iszero.(eachcol(lattice))) && (; energy=zero(T), forces=zero(positions))
```
mfherbst, line 50: Missing `return`

src/response/cg.jl
```diff
@@ -1,7 +1,7 @@
 using LinearMaps
 
 function default_cg_print(info)
-    @printf("%3d\t%1.2e\n", info.n_iter, info.residual_norm)
+    @printf("%3d\t%1.2e\n", info.iterations, info.residual_norm)
```
mfherbst, line 4: `n_iter` is for consistency with SCF data structures. Why change it?

PR #872, opened 2023-08-15
Add versioninfo function

mfherbst:
Useful for debugging: All info in one central place. Might wait for https://github.com/JuliaParallel/MPI.jl/pull/760.


mfherbst:
Merging this as we need it for the Aiida interface.

PR #875, opened 2023-08-16
Initial support for serialisation to JSON

mfherbst:
- [x] More tests
- [x] Documentation


mfherbst:
Merging this as we need it for the Aiida interface. We can fix up smaller things in a follow up.

PR #877, opened 2023-08-29
New AtomicPotentials.jl + PseudoPotentialIO.jl backend

azadoks:
This **draft** replaces the existing pseudopotential framework in `pseudos` with the packages
- `AtomicPotentials.jl` which implements numerical pseudopotentials and the existing analytical `Element` types under a common interface.
- `PseudoPotentialIOExperimental.jl` which removes all pseudopotential evaluation code from `PseudoPotentialIO.jl` and instead focuses _only_ on parsing pseudopotential file formats.

Ideally, this refactor aims to
- Homogenize use of atomic potentials to subsequently
  - increase code sharing among Hamiltonian terms which are derived from the atomic potentials (`AtomicLocal`, `AtomicNonlocal`, `PspCorrection`, and `Xc`) which will
  - facilitate code readability and performance optimization.
- Facilitate support for evaluating atomic potentials on accelerators (GPU)

There's a lot to this PR, so it will take a while to write up, but the key progress points now are:

- All `Element` subtypes are gone, replaced by `Element(symbol, potential)`
- All `pseudos` code is gone, replaced by `AtomicPotentials.jl`
- Loading pseudopotentials is done by `load_psp`, which supports the DFTK `"hgh/.../..."` syntax as well as `"/path/to/file"` and the PseudoLibrary ("family", "filename")

- Form factor calculation is centralized in `form_factors.jl`: angular + radial parts
  - Radial form factors in Fourier space are now interpolated using either Linear or CubicSplines interpolation from `Interpolations.jl` (GPU-supported)
- Structure factor and structure factor gradients are centralized in `structure_factors.jl`

- Projection vector construction is centralized in `projection_vectors.jl`
  - Projection coefficients are now stored in sparse block-diagonal arrays
- Local atomic superposition (i.e. valence density, core density, and local potential) construction and force calculations (for `AtomicLocal` and `Xc` terms) are centralized in `atomic_superposition.jl`

- Density guess code is refactored and much much shorter (using `atomic_superposition.jl`)
- New orbital guess code is in `orbitals.jl`, using UPF's `PSWFC`
  - TODO: need to parse the pseudo-energy of the orbitals so that they can be sorted by lowest energy if fewer orbitals than available are requested for the guess

- Construction of all the affected Hamiltonian terms with HGH, UPF, and PSP8 pseudopotentials is working
- Total energies, forces, and stresses (with AD) are working and agree for at least `silicon.jl`

- A non-zero amount of effort has been put in to get pseudos-on-GPU working: radial Fourier transforms, interpolation, etc. can all be done on GPU. Some deep sleuthing will need to be done to make the code compile with CUDA though (I get all sorts of invalid IR problems, mostly to do with `compute_angular_form_factors` and `compute_radial_form_factors` for the moment).


simonganne01:
Are you still working on this @azadoks ? I would like to build upon this in order to implement ultrasoft psps and PAW in the future. @antoine-levitt or @mfherbst is this the way forward for DFTK?

mfherbst:
See discussion in #987 ... longterm this would be refactored and included, but so far no one is working on this.

PR #878, opened 2023-08-31
Implemented density transfer, added timings to hessian

a-azzali:
Added `transfer_density`
Added a non-k-point-specific `transfer_mapping`
Added timings to hessian functions

### DIFF
--- a/src/transfer.jl
+++ b/src/transfer.jl
@@ -2,32 +2,32 @@ using SparseArrays
 
 """
 Compute the index mapping between the global grids of two bases.
-Returns two matrices `idcs_in` and `idcs_out` whose 8 elements
-are of type `CartesianIndices` (multidimensional rectangular blocks of
-indices). Iterated over the blocks,
-`x_out_fourier[block_out] = x_in_fourier[block_in]` does the transfer from the
-Fourier coefficients `x_in_fourier` (defined on `basis_in`) to `x_out_fourier`
-(defined on `basis_out`, equally provided as Fourier coefficients).
+Returns an iterator of 8 pairs `(block_in, block_out)`. Iterated over these pairs
+`x_out_fourier[block_out, :] = x_in_fourier[block_in, :]` does the transfer from
+the Fourier coefficients `x_in_fourier` (defined on `basis_in`) to
+`x_out_fourier` (defined on `basis_out`, equally provided as Fourier coefficients).
 """
 function transfer_mapping(basis_in::PlaneWaveBasis, basis_out::PlaneWaveBasis)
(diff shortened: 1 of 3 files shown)


src/transfer.jl
```diff
@@ -122,6 +150,32 @@ function transfer_blochwave(ψ_in, basis_in::PlaneWaveBasis{T},
...
+"""
+function transfer_density(ρ_in, basis_in::PlaneWaveBasis{T},
+		          basis_out::PlaneWaveBasis{T}) where {T}
+
+    @assert basis_in.model.lattice == basis_out.model.lattice
+    @assert length(size(ρ_in)) ∈ (3, 4)
+
+    ρ_freq_in = fft(basis_in, ρ_in)
+    ρ_freq_out = similar(ρ_freq_in, basis_out.fft_size..., size(ρ_in, 4))
+    ρ_freq_out .= 0
```
antoine-levitt, line 168: We have a `zeros_like` for this

src/transfer.jl
```diff
@@ -122,6 +150,32 @@ function transfer_blochwave(ψ_in, basis_in::PlaneWaveBasis{T},
...
+
+    idcs_in, idcs_out = transfer_mapping(basis_in, basis_out)
+    map(zip(idcs_in, idcs_out)) do (block_in,block_out)
```
antoine-levitt, line 171: why not do this with a plain old loop?

src/transfer.jl
```diff
@@ -122,6 +150,32 @@ function transfer_blochwave(ψ_in, basis_in::PlaneWaveBasis{T},
...
+        ρ_freq_out[block_out,:] .= ρ_freq_in[block_in,:]
+    end
+    
+    ρ_real_out = ifft(basis_out, ρ_freq_out)
+    ρ_real_out
```
antoine-levitt, line 176: no need for that last line

antoine-levitt (commented):
Hi and thanks for this, looks useful! Generally the code looks a bit complicated because of all the cartesian indices. Would it not be simpler to vectorize the arrays (ie rho_vec = reshape(rho, prod(fft_size), 2) and have transfer_mapping work with just integers?

a-azzali:
Thank you for suggesting those 3 changes to `transfer_density`, I hope it's ok now.

About the array vectorization: We're using the CartesianIndices because one has 8 parallelepiped-shaped blocks of indices to be returned (we wanted to make use of this). I haven't found an easy way of doing this in 1d...

Btw I should have implemented the tests for the density transfer and fixed some bugs.

PR #881, opened 2023-09-13
Refinement of density and forces

a-azzali:
DRAFT

Added postprocess/refine.jl

New functions `refine_scfres`, `refine_forces` and `refine_densities` should be the blackbox version of what described in the _Practical error bounds for the forces_ example.

MPI:
Temporarily replaced `solve_ΩplusK` with `solve_ΩplusK_split`, that is MPI-compatible.
TODO check that the `pre_refinement` key of `scfres` is saved correctly when using MPI.


mfherbst:
Closed in favour of #1016

PR #882, opened 2023-09-15
fix for documenter 1.1

epolack:


mfherbst:
Yes removing it works. I tried that with other projects.

mfherbst:
But it seems that one still needs to keep it in `deploy`, but it's fine to remove in `makedocs`.

PR #885, opened 2023-09-22
Correct typo in preconditioners.jl

gkemlin:
`default_shift` is not defined in the `mul!` function for preconditioners, it should be `P.default_shift`.


antoine-levitt:
Thanks!

PR #886, opened 2023-09-22
Updating SCF doc

epolack:
Really simple fix for a "strange" behavior: If the computations are asked with temperature, then actually the tolerance of the SCF solver will be that of the maximum between the asked `scf_tol` and the `default_occupation_threshold` of 10⁻⁶.
I think this is not intuitive.
I would have changed `default_occupation_threshold` to take into account the `scf_tol`, but this seems to change quiet a bit the structures, so I think this PR is a sane middle ground @mfherbst.


mfherbst:
Sorry I don't get it. Could you post an example script triggering the problem?

Issue #866, opened 2023-07-18
Use TestItemRunner.jl for managing tests


mfherbst:
https://github.com/julia-vscode/TestItemRunner.jl

PR #888, opened 2023-10-02
Switching to TestItemRunner

epolack:
Closes https://github.com/JuliaMolSim/DFTK.jl/issues/866

By default, former `all` and `fast` tests are run (i.e., no `tags` ⇒ test will be run in those case).

TODOs:
- [x] New phonon tests need to be adapted.

To be settled:
- [x] `serialisation.jl` by mfherbst
- [x] `hessian.jl` by mfherbst
- [x] `using ..TestCases` or passing functions along.
- [x] renamed `quick` → `minimal`, `functionality` → `core`

Done:
- [x] There is something fishy with MPI:
  - [x] For serialisation, pass in https://github.com/JuliaMolSim/DFTK.jl/actions/runs/6394141485?pr=888 but not in https://github.com/JuliaMolSim/DFTK.jl/actions/runs/6394940927 which changes nothing, for machine-epsilons where strict equality is checked.
  - [x] Same with AtomsBase https://github.com/JuliaMolSim/DFTK.jl/actions/runs/6398765207/job/17369599039?pr=888, but it should be alright to relax the tolerance to 1e-13.
- [x] The `@testitem` lines can be really long. I do not not how to cut them.
- [x] Some tests are both `:quick` and `:slow` (inverse of `fast`); this is a bit strange and should maybe be changed. But it was to have the same behavior as before.
- [x] I used `@testsetup` in some places; can be removed if useless.
- [x] When `scf_compare.jl` fails in mixing, the output is really difficult to parse because of the really long lines (e.g., https://github.com/JuliaMolSim/DFTK.jl/actions/runs/6406092720/job/17390018132?pr=851). I am using `Meta.parse |> eval`. Not pretty, but does the job. Is it ok?
- [x] Update the developer doc w.r.t. this change?

### DIFF
--- a/test/pairwise.jl
+++ b/test/pairwise.jl
@@ -21,7 +21,7 @@
 
     model = Model(lattice, atoms, positions; terms=[term])
     basis = PlaneWaveBasis(model; Ecut=20, kgrid=(1, 1, 1))
-    forces = compute_forces(basis.terms[1], basis, nothing, nothing)
+    forces = compute_forces(only(basis.terms), basis, nothing, nothing)
 
     # Compare forces to finite differences
     ε=1e-8
--- a/test/phonon/ewald.jl
+++ b/test/phonon/ewald.jl
@@ -1,28 +1,9 @@
-# TODO Far too much code duplication with pairwise tests here
-#      Needs badly refactoring
 @testitem "Phonon: Ewald: comparison to ref testcase" #=
     =#    tags=[:phonon, :dont_test_mpi] setup=[Phonon, TestCases] begin
     using DFTK
-    using DFTK: compute_dynmat_cart
-    using .Phonon: generate_supercell_qpoints, compute_squared_frequencies
-    using LinearAlgebra
-    silicon = TestCases.silicon
 
-    tol = 1e-9
+    testcase = TestCases.silicon
     terms = [Ewald()]
-    model = Model(silicon.lattice, silicon.atoms, silicon.positions; terms)
-    basis_bs = PlaneWaveBasis(model; Ecut=5)
-
-    supercell_size = [2, 1, 3]
-    phonon = (; supercell_size, generate_supercell_qpoints(; supercell_size).qpoints)
-
-    ω_uc = []
-    for q in phonon.qpoints
-        hessian = compute_dynmat_cart(basis_bs, nothing, nothing; q)
-        push!(ω_uc, compute_squared_frequencies(hessian))
-    end
-    ω_uc = sort!(collect(Iterators.flatten(ω_uc)))
-
     ω_ref = [ -0.2442311083805831
               -0.24423110838058237
               -0.23442208238107232
@@ -59,47 +40,14 @@
               1.7010950724885228
               1.7010950724885254
               1.752506588801463]
-    @test norm(ω_uc - ω_ref) < tol
+    Phonon.test_frequencies(testcase, terms, ω_ref)
 end
 
 @testitem "Phonon: Ewald: comparison to automatic differentiation" #=
     =#    tags=[:phonon, :slow, :dont_test_mpi] setup=[Phonon, TestCases] begin
     using DFTK
-    using DFTK: compute_dynmat_cart, energy_forces_ewald
-    using .Phonon: generate_random_supercell, generate_supercell_qpoints
-    using .Phonon: compute_squared_frequencies, ph_compute_reference
-    using LinearAlgebra
-    using Random
-    silicon = TestCases.silicon
+    testcase = TestCases.silicon
 
-    Random.seed!()
-    tol = 1e-9
     terms = [Ewald()]
-    model = Model(silicon.lattice, silicon.atoms, silicon.positions; terms)
-    basis_bs = PlaneWaveBasis(model; Ecut=5)
-
-    supercell_size = supercell_size=generate_random_supercell()
-    phonon = (; supercell_size, generate_supercell_qpoints(; supercell_size).qpoints)
-
-    ω_uc = []
-    for q in phonon.qpoints
-        hessian = compute_dynmat_cart(basis_bs, nothing, nothing; q)
-        push!(ω_uc, compute_squared_frequencies(hessian))
-    end
-    ω_uc = sort!(collect(Iterators.flatten(ω_uc)))
-
-    supercell = create_supercell(silicon.lattice, silicon.atoms, silicon.positions,
-                                    phonon.supercell_size)
-    model_supercell = Model(supercell.lattice, supercell.atoms, supercell.positions;
-                            terms)
-    basis_supercell_bs = PlaneWaveBasis(model_supercell; Ecut=5)
-    hessian_supercell = compute_dynmat_cart(basis_supercell_bs, nothing, nothing)
-    ω_supercell = sort(compute_squared_frequencies(hessian_supercell))
-    @test norm(ω_uc - ω_supercell) < tol
-
-    ω_ad = ph_compute_reference(model_supercell) do term, lattice, atoms, positions
-        charges = charge_ionic.(atoms)
-        energy_forces_ewald(lattice, charges, positions)
-    end
-    @test norm(ω_ad - ω_supercell) < tol
+    Phonon.test_rand_frequencies(testcase, terms)
 end
--- a/test/phonon/helpers.jl
+++ b/test/phonon/helpers.jl
@@ -1,6 +1,8 @@
 # Helpers functions for tests.
 @testsetup module Phonon
-using DFTK: setindex, dynmat_red_to_cart, normalize_kpoint_coordinate
+using Test
+using DFTK
+using DFTK: compute_dynmat_cart, setindex, dynmat_red_to_cart, normalize_kpoint_coordinate
 using Random
 using LinearAlgebra
 using ForwardDiff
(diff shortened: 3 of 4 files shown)


mfherbst:
> For serialisation, 

yes this can actually happen in MPI because different processors might not compute numerically exactly the same stuff. We should not check strict equality there, but allow a few ulp of change (maybe 10eps) ... but add a comment somewhere (maybe above the test cases in a prominent place) why this has been done.

> Same with AtomsBase 

I don't think this is an MPI thing, but I think it's fair to allow a tolerance for lattice here as well.

> When scf_compare.jl fails in mixing

I could not find the exact line where this is happening. Could you pinpoint that ?

>  Update the developer doc w.r.t. this change?

yes please

test/PspHgh.jl
```diff
@@ -82,7 +81,13 @@ end
 
 end
 
-@testset "Check qcut routines" begin
+@testitem "Check qcut routines" tags=[:psp] begin
+    using LinearAlgebra
+    using DFTK: load_psp, eval_psp_projector_fourier, eval_psp_local_fourier
+    using DFTK: eval_psp_projector_real, psp_local_polynomial, eval_psp_local_real
+    using DFTK: psp_projector_polynomial, qcut_psp_projector, qcut_psp_local
+    using DFTK: eval_psp_energy_correction, count_n_proj_radial
```
mfherbst, line 89: Would be nice to be more specific here and not just duplicate the same blob a couple of times ...

test/PspHgh.jl
```diff
@@ -91,18 +96,24 @@ end
     @test (res[1] < res[2]) == (res[3] < res[2])
 
     for i in 1:2, l in 0:2
-        qcut = qcut_psp_projector(psp, i, l)
-        res = eval_psp_projector_fourier.(psp, i, l, [qcut - ε, qcut, qcut + ε])
-        @test (res[1] < res[2]) == (res[3] < res[2])
+        qcut1 = qcut_psp_projector(psp, i, l)
```
mfherbst, line 99: Why the rename ?

test/PspHgh.jl
```diff
@@ -91,18 +96,24 @@ end
...
-    Qloc = psp_local_polynomial(Float64, psp)
-    evalQloc(q) = let t = q * psp.rloc; Qloc(t) * exp(-t^2 / 2) / t^2; end
+@testitem "Agreement of polynomial implementation and eval functions" tags=[:psp] begin
+    using LinearAlgebra
+    using DFTK: load_psp, eval_psp_projector_fourier, eval_psp_local_fourier
+    using DFTK: eval_psp_projector_real, psp_local_polynomial, eval_psp_local_real
+    using DFTK: psp_projector_polynomial, qcut_psp_projector, qcut_psp_local
+    using DFTK: eval_psp_energy_correction, count_n_proj_radial
+
+    psp0 = load_psp("hgh/lda/Si-q4")
```
mfherbst, line 112: Why the rename ?

test/bzmesh.jl
```diff
@@ -30,9 +26,11 @@ include("testcases.jl")
     test_against_spglib([ 9, 11, 13])
 end
 
-if mpi_nprocs() == 1  # PythonCall does not play nicely with MPI
-@testset "bzmesh_ir_wedge is correct reduction" begin
+# PythonCall does not play nicely with MPI.
+@testitem "bzmesh_ir_wedge is correct reduction" tags=[:nompi] begin
```
mfherbst, line 30: Maybe use a more descriptive tag like `dont_test_mpi`

test/serialisation.jl
```diff
@@ -1,44 +1,111 @@
...
-    # Note: Apart from the energy (which is recomputed on loading) the other quantities
-    #       should be exactly as stored. For MPI runs it is possible that the density differs
-    #       slightly for non-master processes as the version from the master process is stored.
-    @test tested.n_iter         == ref.n_iter
-    @test tested.energies.total ≈  ref.energies.total atol=1e-13
-    @test tested.eigenvalues    == ref.eigenvalues
-    @test tested.occupation     == ref.occupation
-    @test tested.ψ              == ref.ψ
-    @test tested.ρ              ≈  ref.ρ rtol=1e-14
+@testsetup module Serialisation
```
mfherbst, line 1: Can we do two Testsetup modules here ?

test/hessian.jl
```diff
@@ -70,13 +68,13 @@ include("testcases.jl")
         @testset "solve_ΩplusK_split <=> solve_ΩplusK" begin
             scfres = self_consistent_field(basis; tol=1e-10)
             δψ1 = solve_ΩplusK_split(scfres, rhs).δψ
+            # TODO: No idea why Hessian mandatory. Same issue even without using setup…
```
mfherbst, line 71: I don't get this comment ?

test/hessian.jl
```diff
@@ -70,13 +68,13 @@ include("testcases.jl")
...
             δψ1, _ = select_occupied_orbitals(basis, δψ1, occupation)
-            ψ, occupation = select_occupied_orbitals(basis, scfres.ψ, scfres.occupation)
-            rhs, _ = select_occupied_orbitals(basis, rhs, occupation)
-            δψ2 = solve_ΩplusK(basis, ψ, rhs, occupation).δψ
+            ψ1, occupation1 = select_occupied_orbitals(basis, scfres.ψ, scfres.occupation)
```
mfherbst, line 73: Why the rename ?

test/hamiltonian_consistency.jl
```diff
@@ -1,109 +1,112 @@
...
-    test_consistency_term(Xc(:lda_xc_teter93), spin_polarization=:collinear)
-    test_consistency_term(Xc(:gga_x_pbe), spin_polarization=:collinear)
-    test_consistency_term(Xc(:mgga_x_tpss))
-    test_consistency_term(Xc(:mgga_x_scan))
-    test_consistency_term(Xc(:mgga_c_scan), spin_polarization=:collinear)
-    test_consistency_term(Xc(:mgga_x_b00))
-    test_consistency_term(Xc(:mgga_c_b94), spin_polarization=:collinear)
+@testitem "Hamiltonian consistency" setup=[HamConsistency] begin
+    using DFTK
+    using LinearAlgebra
```
mfherbst, line 81: Can't you just do a
```
using HamConsistency: test_consistency_term
```
instead ?

test/hamiltonian_consistency.jl
```diff
@@ -1,109 +1,112 @@
-using Test
-using DFTK
-using Logging
-import DFTK: mpi_sum
+@testsetup module HamConsistency
```
mfherbst, line 1: I would vote for *not* indenting the `@testsetup` blocks. That probably makes the code less deeply nested, which overall leads to better readability. What do you think ?

test/forwarddiff.jl
```diff
@@ -1,12 +1,9 @@
-using DFTK
-using ForwardDiff
-using Test
-using ComponentArrays
+@testitem "Force derivatives using ForwardDiff" tags=[:nompi] begin
+    using DFTK
+    using ForwardDiff
+		using LinearAlgebra
```
mfherbst, line 4: Weird ...

mfherbst (commented):
Generally looks good, thanks for doing this. I think a few stylistic things we still need to clarify ... and there is some copy n' paste that you should clean up ;)

test/PspHgh.jl
mfherbst: Strange, I thought the whole point of this testrunner busyness was to avoid putting things in global scope. What does it clash with ? Maybe we should rename that instead ?

test/PspHgh.jl
mfherbst: Ah now I get why these things are needed. Here I would actually just put the code block above inside a `let ... end` then the `qcut` is not on the global scope.

mfherbst:
>It is at the summary of the test.

Ah yeah, well I'd say that's ok.

test/PspHgh.jl
mfherbst: Yeah as I mentioned elsewhere. My strategy would be to wrap the `psp` in the code directly above the loop in a `let` block.

test/serialisation.jl
mfherbst: Well, if it leads to code duplication, than better have one module.

test/hessian.jl
mfherbst: Ok, let's finish up the rest and than I'll take a look at that and see if I can figure it out.

test/hamiltonian_consistency.jl
mfherbst: ok.

mfherbst:
> It's not super pretty, but I guess better.

good for me

mfherbst:
> Some tests are both :quick and :slow (inverse of fast); this is a bit strange and should maybe be changed. But it was to have the same behavior as before.

Yeah this fast / slow business is not ideal. I think it's fine for this PR.

mfherbst:
Oh and another way to break the `@testitem` lines should be to use brackets, e.g.
```julia
@testitem("Silicon without XC (large)", tags=[:slow, :functionality], setup=[SiliconRedHF, TestCases, RunSCF], begin
   ...
end)
```

test/silicon_nlcc.jl
```diff
@@ -1,7 +1,9 @@
-include("run_scf_and_compare.jl")
-include("testcases.jl")
+@testsetup module SiliconNLCC
+using DFTK
+using LazyArtifacts
 
-function run_silicon_nlcc(T; Ecut=5, grid_size=15, spin_polarization=:none, kwargs...)
+function run_silicon_nlcc(T, silicon, run_scf_and_compare; Ecut=5, grid_size=15,
```
mfherbst, line 5: Maybe rename to `testcase` for consistency with other functions with a variable carrying an unspecified testcase.

test/serialisation.jl
mfherbst: Yeah that's really not pretty either.

mfherbst:
Ok, I let you finish up the `@testitem` formatting and then I'll take a look at that `serialisation.jl` and `hessian.jl` stuff.

mfherbst:
But I think it overall turned out really nicely!

test/serialisation.jl
mfherbst: Fair enough. Maybe we'll have to bite some bullets in the end.

mfherbst: Could you make a PR at TestItemRunner (e.g. in their docs) discussing this feature. Because it might be that this just "works" but is not intended behaviour. In that case I'm afraid we might have our tests randomly break in the future.

If they sort of give it a green light by admitting it to their public documentation, then I'm ok with using this here.

mfherbst:
Needs #876 and #762 merged.

mfherbst:
@epolack The new pohnon tests need adapting, but I don't think I will get there today. If you want to quickly take a look at that, that would be great!

mfherbst:
Let me know once this is good to go.

PR #899, opened 2023-10-30
Add basic Wannier.jl integration

Technici4n:
This PR adds basic `Wannier.jl` integration. The communication happens via the Wannier90 files at the moment; this is not the fastest, but it required the least amount of code. I split the Wannier code across `wannier_shared.jl`, `wannier90.jl` and `wannier.jl`, the last 2 requiring the relevant package.

I also refactored the initial projection code, to make the choice of the initial projections more flexible. Hopefully that can be extended in the future to project onto atomic orbitals without changing the API again.

Following the recommendation in one of the #854 comments, I updated the example to mention `Wannier.jl` first.


Technici4n:
Junfeng pushed a 0.3.2 Wannier update, unfortunately this requires Julia 1.8.4 at least. Is it acceptable to update Julia to 1.8, at least for testing? (I have 1.9 on my machine and did not catch this issue...)

src/DFTK.jl
```diff
@@ -39,6 +39,7 @@ include("common/zeros_like.jl")
 include("common/norm.jl")
 include("common/quadrature.jl")
 include("common/hankel.jl")
+include("common/hydrogenic.jl")
```
Technici4n, line 42: (It would be nice if these lines were sorted to reduce merge conflicts between PRs that append files.)

Technici4n:
@mfherbst CI is passing (except for nightly due to a `@main` macro conflict inside of Wannier).

Technici4n:
Merge conflicts addressed. :)

src/DFTK.jl
mfherbst: You mean within each block. That might not be a bad idea.

src/common/hydrogenic.jl
```diff
@@ -0,0 +1,22 @@
+@doc raw"""
+Radial functions from solutions of Hydrogenic Schrödinger equation.
+Same as Wannier90 user guide Table 3.3.
+# Arguments
+- `r`: radial grid
+- `n`: principal quantum number
+- `α`: diffusivity, ``\frac{Z}/{a}`` where ``Z`` is the atomic number and
+    ``a`` is the Bohr radius.
+"""
+function radial_hydrogenic(r::AbstractVector{T}, n::Integer, α::Real=1.0) where {T<:Real}
```
mfherbst, line 10: This function is not type stable. Even if `T` is `Float32`, this function will return a `Float64` because `(3/2)` automatically gives a Float64. One way to solve this is to use the pattern `(3/T(2))` for example.

src/external/stubs.jl
```diff
@@ -1,13 +1,27 @@
...
+    Currently this is an experimental feature, which has not yet been tested
+    to full depth. The interface is considered unstable and may change
+    incompatibly in the future. Use at your own risk and please report bugs
+    in case you encounter any.
+"""
+function get_wannier_model end
+
+"""
+Wannerize the obtained bands using wannier90.
+Same arguments as `get_wannier_model`.
```
mfherbst, line 22:
```suggestion
Same arguments as [`get_wannier_model`](@ref).
```

src/external/wannier.jl
```diff
@@ -0,0 +1,36 @@
...
+            end
+        end)
+
+    (nntot=model.n_bvecs, nnkpts)
+end
+
+"""
+Build a Wannier.jl model that can be used for Wannierization.
+"""
+@timing function get_wannier_model(scfres;
```
mfherbst, line 22: I think I would drop the `get_` and just call it `wannier_model` or if the type in Wannier.jl is always the same and Wannier.jl uses the convention of having constructors that "do things (instead of just storing data), it can even be an additional constructor method (e.g. if the type is `WannierModel` define a method `Wannier.WannierModel( ... )` with the arguments you pass.

src/external/wannier_shared.jl
```diff
@@ -0,0 +1,407 @@
...
+    model = basis.model
+    [begin
+        q_cart = recip_vector_red_to_cart(model, q)
+        exp( 2π*(-im*dot(q, proj.center) - dot(q_cart, q_cart) / 4) )
+    end for q in qs]
+end
+
+function random_gaussian_projection()
+    GaussianWannierProjection(rand(3))
+end
```
mfherbst, line 41: Why does this function even exist? Can't you just use it inline?

src/external/wannier_shared.jl
```diff
@@ -0,0 +1,407 @@
...
+As a starting point for the search, we can provide an initial guess function ``g``
+for the shape of the Wannier functions, based on what we expect from knowledge of the problem or physical intuition.
+This starting matrix is called ``[A_k]_{m,n}``, and is computed as follows:
+``[A_k]_{m,n} = \langle ψ_m^k | g^{\text{per}}_n \rangle``.
+The matrix will be orthonormalized by the chosen Wannier program, we don't need to do so ourselves.
+"""
+abstract type WannierProjection end
+
+# Dispatch function for WannierProjection implementations, evaluating g in Fourier space.
+function get_fourier_projection_coefficients end
```
mfherbst, line 19: Could simply be the call operator, i.e.
```
function (proj::GaussianWannierProjection)(basis::PlaneWaveBasis, qs)
   # stuff
end
```
That makes it even more flexible as one could pass arbitrary functions instead of these predefined functors when doing the wannierisation.

src/external/wannier_shared.jl
```diff
@@ -0,0 +1,407 @@
...
+Base type for Wannier projections.
+
+Wannierization searchs for a unitary matrix ``U_{m_n}``.
+As a starting point for the search, we can provide an initial guess function ``g``
+for the shape of the Wannier functions, based on what we expect from knowledge of the problem or physical intuition.
+This starting matrix is called ``[A_k]_{m,n}``, and is computed as follows:
+``[A_k]_{m,n} = \langle ψ_m^k | g^{\text{per}}_n \rangle``.
+The matrix will be orthonormalized by the chosen Wannier program, we don't need to do so ourselves.
+"""
+abstract type WannierProjection end
```
mfherbst, line 16: I don't think this abstract type is needed. Let me know if you have reason to believe otherwise.

src/external/wannier_shared.jl
```diff
@@ -0,0 +1,407 @@
...
+
+function get_fourier_projection_coefficients(proj::GaussianWannierProjection, basis::PlaneWaveBasis, qs)
+    # associate a center with the fourier transform of the corresponding gaussian
+    # TODO: what is the normalization here?
+
+    model = basis.model
+    [begin
+        q_cart = recip_vector_red_to_cart(model, q)
+        exp( 2π*(-im*dot(q, proj.center) - dot(q_cart, q_cart) / 4) )
+    end for q in qs]
```
mfherbst, line 36: we don't use begin and end for things like this. If you need this use a 
```
map(qs) do q
  # stuff
  # more stuff
end
```

src/external/wannier_shared.jl
```diff
@@ -0,0 +1,407 @@
...
+end
+
+function random_gaussian_projection()
+    GaussianWannierProjection(rand(3))
+end
+
+"""
+A sum of two Gaussians. Can be used as an approximation of a p-like orbital.
+"""
+struct DoubleGaussianWannierProjection <: WannierProjection
```
mfherbst, line 46: And here is a prime example why being able to just use functions is great. Then the 40 lines of code to make the rather restrictive type `DoubleGaussianWannierProjection` can be reduced to just 5:

```
function plike_gaussian_projection(center1::AbstractVector, center2::AbstractVector)
    function inner(basis::PlaneWaveBasis, qs)
       GaussianWannierProjection(center1)(basis, qs) - GaussianWannierProjection(center1)(basis, qs)
    end
end
```

src/external/wannier_shared.jl
```diff
@@ -0,0 +1,407 @@
...
+    # rgrid depends on α
+    r = exp.(x) ./ proj.α
+    dr = r .* dx
+    R = radial_hydrogenic(r, proj.n, proj.α)
+    r2_R_dr = r.^2 .* R .* dr
+
+    model = basis.model
+
+    [begin
+        q_cart = recip_vector_red_to_cart(model, q)
```
mfherbst, line 104: same comment as above.

src/external/wannier_shared.jl
```diff
@@ -0,0 +1,407 @@
...
+
+Centers are to be given in lattice coordinates and G_vectors in reduced coordinates.
+The dot product is computed in the Fourier space.
+
+Given an orbital ``g_n``, the periodized orbital is defined by :
+ ``g^{per}_n =  \sum\limits_{R \in {\rm lattice}} g_n( \cdot - R)``.
+The  Fourier coefficient of ``g^{per}_n`` at any G
+is given by the value of the Fourier transform of ``g_n`` in G.
+"""
+function compute_amn_kpoint(basis::PlaneWaveBasis, ψk, kpt, projections::AbstractVector{<:WannierProjection}, n_bands)
```
mfherbst, line 318: in internal function names we typically abbreviate `kpoint` as `kpt`
Typically we have the `kpt` argument follow the `basis`
Don't annotate the type of the projections (only `AbstractVector`)
Make `n_bands` a kwarg like in the other functions.

Side note: If we do a restriction of the ψk arrays inside the ψ vector itself using a view we could rid of passing both the ψk and the n_bands. Here is a code example that does this using masks:
https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/response/chi0.jl#L348

src/external/wannier90.jl
```diff
@@ -1,288 +1,24 @@
...
-    mkpath(dir)
-
-    # Write input file and launch Wannier90 preprocessing
-    write_w90_win(fileprefix, basis;
-                  num_wann=length(centers), num_bands=n_bands, wannier_plot, kwargs...)
-    wannier90_jll.wannier90(exe -> run(Cmd(`$exe -pp $prefix`; dir)))
-    nnkp = read_w90_nnkp(fileprefix)
+        n_bands=scfres.n_bands_converge,
+        n_wannier=n_bands,
+        projections::AbstractVector{<:WannierProjection}=default_wannier_centres(n_wannier),
```
mfherbst, line 8: no type annotation ;)

mfherbst (commented):
Pretty nice and sorry that it took me so long to reply.

src/external/wannier.jl
Technici4n: I went for `wannier_model` because the type is unfortunately `Wannier.Model` and that might conflict with DFTK's own `Model` type.

mfherbst: Ah yes, that's probably true. In that case `wannier_model` is probably a good name.

examples/wannier.jl
```diff
@@ -0,0 +1,144 @@
...
+# - 3 bond-centered 2s hydrogenic orbitals for the expected σ bonds
+# - 2 atom-centered 2pz hydrogenic orbitals for the expected π bands
+
+using Wannier # Needed to make wannier_model available
+
+## Helper functions to produce hydrogenic starting guesses
+s_guess(center) = DFTK.HydrogenicWannierProjection(center, 2, 0, 0, C.Z)
+pz_guess(center) = DFTK.HydrogenicWannierProjection(center, 2, 1, 0, C.Z)
+
+wannier_model = wannier_model(scfres;
```
mfherbst, line 65: this won't work as you cannot use the name of a function as a variable

src/common/hydrogenic.jl
```diff
@@ -0,0 +1,23 @@
+@doc raw"""
+Radial functions from solutions of Hydrogenic Schrödinger equation.
+Same as Wannier90 user guide Table 3.3.
+# Arguments
+- `r`: radial grid
+- `n`: principal quantum number
+- `α`: diffusivity, ``\frac{Z}/{a}`` where ``Z`` is the atomic number and
+    ``a`` is the Bohr radius.
+"""
+function radial_hydrogenic(r::AbstractVector{T}, n::Integer, α::Real=one(T)) where {T<:Real}
```
mfherbst, line 10: Have you checked the type stability eplicitly. This is easy to miss. Just include a test with `@test_inferred` or something like (julia docs on Unit Tests) and a call based on a float32 number.

Technici4n: I did. Added a test as well (the vscode integration is very nice...)

examples/wannier.jl
Technici4n: `wannier_model` is a very natural name for the variable, so I chose to go for a `Wannier.Model` method to fix this issue.

src/external/wannier_shared.jl
Technici4n: I just moved `kpt` to the second position for now. (We could move to an array of views but it makes the code less readable IMO and would need to be applied to the other functions as well.)

Technici4n:
All comments addressed, and I moved the Wannier.jl and Wannier90 integrations to extensions as discussed earlier. We have to wait for another Wannier bump due to the Spglib bump (https://github.com/qiaojunfeng/Wannier.jl/pull/35).

src/external/wannier_shared.jl
```diff
@@ -48,7 +128,7 @@ function write_w90_win(fileprefix::String, basis::PlaneWaveBasis;
             path = kpath.paths[1]
 
             println(fp, "begin kpoint_path")
-            for i = 1:length(path)-1
+            for i in 1:length(path)-1
```
mfherbst, line 131: be careful. Etienne changed this on purpose to be =

src/external/wannier_shared.jl
```diff
@@ -1,5 +1,85 @@
...
+    n_r = round(Int, (log(rmax) - xmin) / dx) + 1
+    x = range(; start=xmin, length=n_r, step=dx)
+    # rgrid depends on α
+    r = exp.(x) ./ proj.α
+    dr = r .* dx
+    R = radial_hydrogenic(r, proj.n, proj.α)
+    r2_R_dr = r.^2 .* R .* dr
+
+    model = basis.model
+
```
mfherbst, line 64: newline not needed

src/external/wannier_shared.jl
```diff
@@ -1,5 +1,85 @@
...
+end
+
+@doc raw"""
+A hydrogenic initial guess.
+
+`α` is the diffusivity, ``\frac{Z}/{a}`` where ``Z`` is the atomic number and
+    ``a`` is the Bohr radius.
+"""
+struct HydrogenicWannierProjection
+    center::AbstractVector{Float64}
```
mfherbst, line 40: The `Float64` is weird here as it's an abstract type. I'd leave it off.

src/external/wannier_shared.jl
```diff
@@ -1,5 +1,85 @@
...
+    R = radial_hydrogenic(r, proj.n, proj.α)
+    r2_R_dr = r.^2 .* R .* dr
+
+    model = basis.model
+
+    map(qs) do q
+        q_cart = recip_vector_red_to_cart(model, q)
+
+        qnorm = norm(q_cart)
+        radial_part = 0
```
mfherbst, line 69: Not that this function is type sable at all, but since this is a common antipattern I wanted to point it out:

This is a type instability as `radial_part` is an integer, but later floats are added, initialise as `= 0.0`.

mfherbst (commented):
Small nits, otherwise ok. But definitely not the fastest code ... there are a few type instabilities and some smaller things where performance could be gained, but probably ok for now.

src/external/wannier_shared.jl
mfherbst: ok

Project.toml
```diff
@@ -50,10 +50,14 @@ UnitfulAtomic = "a7773ee8-282e-5fa2-be4e-bd808c38a91a"
...
 [weakdeps]
 JSON3 = "0f8b85d8-7281-11e9-16c2-39a750bddbf1"
+Wannier = "2b19380a-1f7e-4d7d-b1b8-8aa60b3321c9"
+wannier90_jll = "c5400fa0-8d08-52c2-913f-1e3f656c1ce9"
 WriteVTK = "64499a7a-5c06-52f2-abe2-ccb03c286192"
 
 [extensions]
 DFTKJSON3Ext = "JSON3"
+DFTKWannierExt = "Wannier"
+DFTKWannier90Ext = "wannier90_jll"
```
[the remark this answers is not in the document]
Technici4n: As you prefer, I'm happy to change the name. The two existing extensions are already inconsistent, btw. :smile:

examples/wannier.jl
```diff
@@ -0,0 +1,144 @@
...
+              projections=[
+                  ## Note: fractional coordinates for the centers!
+                  ## 3 bond-centered 2s hydrogenic orbitals to imitate σ bonds
+                  s_guess((positions[1] + positions[2]) / 2),
+                  s_guess((positions[1] + positions[2] + [0, -1, 0]) / 2),
+                  s_guess((positions[1] + positions[2] + [-1, -1, 0]) / 2),
+                  ## 2 atom-centered 2pz hydrogenic orbitals
+                  pz_guess(positions[1]),
+                  pz_guess(positions[2]),
+              ],
```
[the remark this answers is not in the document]
Technici4n: I added a small explanation, hopefully it helps. (It's hard to explain how exactly this works.)

src/DFTK.jl
```diff
@@ -184,9 +185,11 @@ include("pseudo/list_psp.jl")
 include("pseudo/attach_psp.jl")
 
 export atomic_system, periodic_system  # Reexport from AtomsBase
+export wannier_model
```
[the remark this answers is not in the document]
Technici4n: nope, how embarrassing...

ext/DFTKWannier90Ext.jl
```diff
@@ -0,0 +1,30 @@
+module DFTKWannier90Ext
+
+using DFTK
+import wannier90_jll
```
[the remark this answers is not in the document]
Technici4n: Right, I didn't think of that. The `DFTK.` prefix is overall shorter though, so I'll keep it.

test/external/wannier.jl
```diff
@@ -0,0 +1,68 @@
...
+    using DFTK
+    using Wannier
+    silicon = TestCases.silicon
+
+    model  = model_LDA(silicon.lattice, silicon.atoms, silicon.positions)
+    basis  = PlaneWaveBasis(model; Ecut=5, kgrid=[4, 4, 4], kshift=[1, 1, 1]/2)
+    nbandsalg = AdaptiveBands(model; n_bands_converge=12)
+    scfres = self_consistent_field(basis; nbandsalg, tol=1e-12)
+
+    fileprefix = "wannierjl_outputs/Si"
```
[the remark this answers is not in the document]
Technici4n: Changed it to use a tempdir. I like having the tests split so I can run the Wannierjl ones on my Windows without needing to use WSL, but arguably saving an SCF is worth it.

src/external/wannier_shared.jl
```diff
@@ -1,5 +1,85 @@
...
+    map(qs) do q
+        q_cart = recip_vector_red_to_cart(model, q)
+        exp( 2π*(-im*dot(q, proj.center) - dot(q_cart, q_cart) / 4) )
+    end
+end
+
+"""
+A p-like initial guess, using the difference of two Gaussians with different centers.
+"""
+function opposite_gaussians_projection(center1::AbstractVector, center2::AbstractVector)
```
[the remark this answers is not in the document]
Technici4n: I'm not too sure myself. If you already know that you want a plike orbital you probably want to use `HydrogenicWannierProjection` anyway (or even the pswfc from the PSP in the future).

Project.toml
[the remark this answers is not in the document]
mfherbst: ok, whatever.

test/external/wannier.jl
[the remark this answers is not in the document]
mfherbst: Sorry no special casing for windows support. :smile: I'd like to save the SCF if we can and here it's easy ... plus it enables the comparison now or later.

src/external/wannier_shared.jl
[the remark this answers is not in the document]
mfherbst: yes, that's true. Good point. OTOH this makes it questionable why this function even exists in the main code. Might be good to include in the example to show how custom projectors could be made. Maybe it's better put there.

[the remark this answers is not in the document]
Technici4n: Just removed it. I don't think that encouraging custom projectors is particularly a good idea, in principle it shouldn't be needed and the interface is still experimental. (I was using `opposite_gaussians_projection` before the hydrogenic guess was implemented, now it can be useful for comparison against the hydrogenic guess but that doesn't need to be in the mainline code).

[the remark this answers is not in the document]
mfherbst: >  (I was using opposite_gaussians_projection before the hydrogenic guess was implemented, now it can be useful for comparison against the hydrogenic guess but that doesn't need to be in the mainline code).

Yes, but it would be good to keep the idea documented in the wannier example. That's what I meant. Sorry if I was not clear.

examples/wannier.jl
```diff
@@ -0,0 +1,160 @@
...
+# ### Custom initial guesses
+#
+# We can also provide custom initial guesses for Wannierization,
+# by passing a callable function in the `projections` array.
+# The function receives the basis and a list of points (fractional coordinates in reciprocal space),
+# and returns the Fourier transform of the initial guess function evaluated at each point.
+#
+# For example, we could use Gaussians for the σ and pz guesses with the following code:
+s_guess(center) = DFTK.GaussianWannierProjection(center)
+pz_guess(center) = begin
```
mfherbst, line 116: Dont use this pattern. Use function name(args)
end

examples/wannier.jl
```diff
@@ -0,0 +1,161 @@
...
+    ## Approximate with two Gaussians offset by 0.5 Å from the center of the atom
+    offset = model.inv_lattice * [0, 0, austrip(0.5u"Å")]
+    center1 = center + offset
+    center2 = center - offset
+    ## Build the custom projector
+    (basis, qs) -> DFTK.GaussianWannierProjection(center1)(basis, qs) - DFTK.GaussianWannierProjection(center2)(basis, qs)
+end
```
mfherbst, line 123: Actually code up thd call. At least in a comment

Technici4n: not sure I understand

mfherbst: That's fine no worries, not too important anyways.

PR #906, opened 2023-11-15
Add cutoff parameter to PspUpf

azadoks:
Quick and dirty


src/pseudo/load_psp.jl
```diff
@@ -9,7 +9,7 @@ The file is searched in the directory `datadir_psp()` and by the `key`.
 If the `key` is a path to a valid file, the extension is used to determine
 the type of the pseudopotential file format and a respective class is returned.
 """
-function load_psp(key::AbstractString)
+function load_psp(key::AbstractString; rcut=nothing)
```
mfherbst, line 12: kwargs...

src/pseudo/load_psp.jl
```diff
@@ -34,7 +34,7 @@ function load_psp(key::AbstractString)
     end
 
     if isfile(fullpath)
-        return pseudo_type(fullpath; identifier)
+        return pseudo_type(fullpath; identifier, rcut)
```
mfherbst, line 37: kwargs...

src/pseudo/PspUpf.jl
```diff
@@ -173,24 +176,26 @@ end
...
-    end
-    4T(π) * (s - psp.Zion / q) / q
+    # ABINIT STYLE
+    # f = psp.rgrid .* (psp.rgrid .* psp.vloc .- -psp.Zion) .* sphericalbesselj_fast.(0, q .* psp.rgrid)
+    # I = NumericalIntegration.integrate(psp.rgrid, f, Trapezoidal())
+    # 4T(π) * (I + -psp.Zion / q^2)
+
+    # QE STYLE
+    ir_cut = length(psp.rgrid)
+    f = psp.rgrid .* (psp.rgrid .* psp.vloc .- -psp.Zion * erf.(psp.rgrid)) .* sphericalbesselj_fast.(0, q .* psp.rgrid)
```
mfherbst, line 196: line too long

src/pseudo/PspUpf.jl
```diff
@@ -173,24 +176,26 @@ end
...
-    s = zero(T)
-    @inbounds for ir = eachindex(psp.r_vloc_corr_dr)
-        s += sin(q * psp.rgrid[ir]) * psp.r_vloc_corr_dr[ir]
-    end
-    4T(π) * (s - psp.Zion / q) / q
+    # ABINIT STYLE
+    # f = psp.rgrid .* (psp.rgrid .* psp.vloc .- -psp.Zion) .* sphericalbesselj_fast.(0, q .* psp.rgrid)
+    # I = NumericalIntegration.integrate(psp.rgrid, f, Trapezoidal())
+    # 4T(π) * (I + -psp.Zion / q^2)
+
```
mfherbst, line 193: Remove ?

src/pseudo/PspUpf.jl
```diff
@@ -128,12 +121,33 @@ function PspUpf(path; identifier=path)
...
+        ρcore = ρcore[mask]
+    end
+
     return PspUpf(Zion, lmax, rgrid, drgrid, vloc, r_projs, h, pswfcs, pswfc_occs,
                   r2_4π_ρion, ρcore; identifier, description)
 end
 
 function PspUpf(Zion, lmax, rgrid::Vector{T}, drgrid, vloc, r_projs, h, pswfcs, pswfc_occs,
                 r2_4π_ρion, ρcore; identifier="", description="") where {T <: Real}
+
```
mfherbst, line 150: no need.

src/pseudo/PspUpf.jl
```diff
@@ -128,12 +121,33 @@ function PspUpf(path; identifier=path)
         ρcore = zeros(Float64, length(rgrid))
     end
 
+    #* HACK!
+    if !isnothing(rcut)
```
mfherbst, line 125: Not really a hack. I think removing the rcut is fine.

Also at least put one test somewhere to ensure the code is run, so that we don't accidentially break it.

antoine-levitt (commented):
I like the cleanup, but weren't those fields precomputed for performance reasons?

azadoks:
> I like the cleanup, but weren't those fields precomputed for performance reasons?

I can reoptimize in the morning, in general 'rcut' _should_ be set, and as a side effect, things are then a bit quicker. I'll give a full explanation tomorrow for historical documentation.

src/common/trapezoidal.jl
```diff
@@ -0,0 +1,16 @@
+# NumericalIntegration.jl also implements this (in a slightly different way)
+# however, it is unmaintained and has stale, conflicting version requirements
+# for Interpolations.jl
+@inbounds function trapezoidal(x::AbstractVector, y::AbstractVector)
+    n = length(x)
+    n == length(y) || error("vectors `x` and `y` must have the same number of elements")
```
mfherbst, line 6: I think I would even make this an assert.

src/pseudo/PspUpf.jl
```diff
@@ -128,101 +120,104 @@ function PspUpf(path; identifier=path)
...
-    s = zero(T)
-    @inbounds for ir = eachindex(r2_proj_dr)
-        s += sphericalbesselj_fast(l, q * psp.rgrid[ir]) * r2_proj_dr[ir]
-    end
-    4T(π) * s
+function eval_psp_projector_fourier(psp::PspUpf, i, l, q::T)::T where {T<:Real}
+    ir_cut = length(psp.r_projs[l+1][i])
+    f = (
+        psp.rgrid[1:ir_cut] .* psp.r_projs[l+1][i] .*
+        sphericalbesselj_fast.(l, q .* psp.rgrid[1:ir_cut])
```
mfherbst, line 173: For these things I like putting the `.*` in front to make the operation more visible

antoine-levitt: You also probably want `@views`. I sometimes wonder if we should not enclose the whole of DFTK in an `@views`

src/common/trapezoidal.jl
azadoks: I may be misremembering this, but I think `@assert` can be dropped by the compiler.

src/pseudo/PspUpf.jl
```diff
@@ -128,101 +120,104 @@ function PspUpf(path; identifier=path)
         ρcore = zeros(Float64, length(rgrid))
     end
 
-    return PspUpf(Zion, lmax, rgrid, drgrid, vloc, r_projs, h, pswfcs, pswfc_occs,
-                  r2_4π_ρion, ρcore; identifier, description)
-end
+    if !isnothing(rcut)
+        mask = rgrid .<= rcut
```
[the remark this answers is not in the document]
azadoks: I'm not at all worried about the time it takes to read and parse the UPF files; I also remembered that I shouldn't cut off the pseudo-atomic wavefunctions, so this isn't being done anymore anyway.

src/common/trapezoidal.jl
```diff
@@ -0,0 +1,16 @@
+# NumericalIntegration.jl also implements this (in a slightly different way)
+# however, it is unmaintained and has stale, conflicting version requirements
+# for Interpolations.jl
+@inbounds function trapezoidal(x::AbstractVector, y::AbstractVector)
```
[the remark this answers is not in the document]
azadoks: I've added vectorization for `trapezoidal` and `_simpson_uniform`, but `@turbo` breaks `_simpson_nonuniform`, so I've left it with (maybe useless) `@simd`.

azadoks:
Note that the `psp` tests run ~5% _slower_ with `@turbo` vs `@simd`.

azadoks:
The radial cutoff and integration methods are added to improve the agreement between DFTK and QE/ABINIT for simulations involving some particular pseudopotentials which are not well behaved under the current implementation. Specifically, the minimum and curvature of equations of state computed by DFTK are _very_ different from those computed by QE/ABINIT and from an all-electron reference EOS.

After some digging, I discovered that one contributing factor to the disagreement in results is that `PspUpf` performs Fourier/Hankel transforms by integrating over the entire radial mesh provided by the UPF file, while QE integrates select quantities only up to 10 Bohr (a magic constant), and my ABINIT reference calculations were performed with psp8 format pseudos which have radial meshes up to 5.99 Bohr.

For some reason, the pseudopotential data at high radial distance is significantly lower quality for some pseudos compared to that at lower radial distance (maybe to do with the logarithmic mesh used by ONCVPSP).

This can be seen by computing `Vloc(q)` for various real-space radial cutoff values
```
Vloc(q)[rcut] = 4pi Int{0,rcut}[r^2 (Vloc(r) - C(r)) j0(qr) dr] + H[C(r)]
where
C(r) = -Zion erf(r) / r; H[C(r)] = -Zion/q^2 exp(-q^2 / 4)
```

Then plotting the finite-difference sensitivity against rcut
```
Int{0,15}[|Vloc(q)[ircut + 1] - Vloc(q)[ircut]|] / (rcut[ircut + 1] - rcut[ircut])
```

crappy_finite_diff_vloc_int

Particular elements, e.g. Co, Cr, Cu, etc., have very high sensitivity around specific values (~11, ~13.5, ~9.8 Bohr). These pseudos should be cut off around these values to get accurate results for quantities like the equation of state.

Additionally, I found that we really _do_ need to use trapezoidal or Simpson's method integration to agree more closely with existing codes.

src/common/trapezoidal.jl
mfherbst: You can actually mark such checks as "the bounds checks" using a macro. I forgot how it works, but I suppose that's the way to go here.

src/pseudo/PspUpf.jl
```diff
@@ -65,164 +65,169 @@ Does not support:
...
-    rgrid       = pseudo["radial_grid"]
-    drgrid      = pseudo["radial_grid_derivative"]
-    lmax        = pseudo["header"]["l_max"]
-    vloc        = pseudo["local_potential"] ./ 2  # (Ry -> Ha)
+    Zion = Int(pseudo["header"]["z_valence"])
+    rgrid = pseudo["radial_grid"]
+    drgrid = pseudo["radial_grid_derivative"]
+    lmax = pseudo["header"]["l_max"]
+    vloc = pseudo["local_potential"] ./ 2  # (Ry -> Ha)
     description = get(pseudo["header"], "comment", "")
```
mfherbst, line 86: This was formatted so nicely ... don't break that for no reason.

src/pseudo/PspUpf.jl
```diff
@@ -65,164 +65,169 @@ Does not support:
...
-        s += sin(q * psp.rgrid[ir]) * psp.r_vloc_corr_dr[ir]
-    end
-    4T(π) * (s - psp.Zion / q) / q
+function eval_psp_pswfc_fourier(psp::PspUpf, i, l, q::T)::T where {T<:Real}
+    # Pseudo-atomic wavefunctions are _not_ currently cut off like the other
+    # quantities. They are the reason that PseudoDojo UPF files have a much
+    # larger radial grid than their psp8 counterparts.
+    # If issues arise, try cutting them off too.
+    ircut = length(psp.r2_pswfcs[l+1][i])
+    rgrid = @view psp.rgrid[1:ircut]
```
mfherbst, line 185: maybe then don't use the variable name `ircut` to avoid someone missing this point when glossing over the code.

mfherbst (commented):
Nice.

src/pseudo/PspUpf.jl
azadoks: image
Sorry!

mfherbst (commented):
lgtm

antoine-levitt:
This is very annoying, why do people publish pseudos if they know they're rubbish above some cutoff radius? Or are the pseudos good and is it just that the integration error blows up for large rcut because we integrate against a r^l or something like this? But surely the pseudos should decay fast enough for this not to be an issue?

WuyihanVictor:
1. **In `PspUpf.jl`, Line 88**: 
   - to ensure `rcut <= rgrid`.
   - change the code as follows:
     ```julia
     rcut = isnothing(rcut) ? last(rgrid) : min(rcut, last(rgrid))
     ```

2. **In `atomsbase.jl`, Line 19**:
   -  add following modifications to facilitate the AiidaDFTK.jl building system from atoms.
   - 
     ```julia
     atoms = map(system) do atom
         pseudo = get(atom, :pseudopotential, "")
         rcut_val = get(atom, :rcut, nothing)
         if isempty(pseudo)
             ElementCoulomb(atomic_number(atom))
         else
             get!(cached_pspelements, pseudo) do
                 ElementPsp(atomic_number(atom); psp=load_psp(pseudo; rcut=rcut_val))
             end
         end
     end
     ```
   - read 'rcut' from each atom, load_psp with 'rcut'

mfherbst:
> This is very annoying, why do people publish pseudos if they know they're rubbish above some cutoff radius?

Unclear to me, but on my top list to ask a pseudo expert next time I meet one ;).

> But surely the pseudos should decay fast enough for this not to be an issue? 

Well, if you have a logarithmic grid and don't put in proper weights during the fit of the potential, then I suppose it could  happen that the values far away from zero are only defined to a very low accuracy. 

The issue we observed also does not occur for all elements. For most (where the decay is faster it seems), things are fine.

@WuyihanVictor runs some tests with variable `rcut` against all electron and QE. Let's see what we learn.

mfherbst:
> add following modifications to facilitate the AiidaDFTK.jl building system from atoms.

This feels very much like a hack. In an ideal world, this should just be part of the pseudo definition ...

I would call this `pseudopotential_rcut` to avoid any clash with other potential keywords in AtomsBase.

mfherbst:
> This feels very much like a hack. In an ideal world, this should just be part of the pseudo definition ...

BTW with our "PseudoLibrary" we could easily do that, by supplying an extra file that has the cutoff data, once we have settled on good values.

azadoks:
> Or are the pseudos good and is it just that the integration error blows up for large rcut because we integrate against a r^l or something like this?

The pseudos are [good](https://doi.org/10.1038/s42254-023-00655-3).
For 'Vloc', we integrate with 'r^2 j_0(qr)', which should also decay with large 'r'.

The UPF pseudos are simply on a larger domain than [requested in the generation input](https://github.com/PseudoDojo/ONCVPSP-PBE-PDv0.4/blob/134d757178552b8228a43a848b0e53f232cdda84/Cu/Cu-sp.in) because of the [inclusion of pseudo atomic wavefunctions](https://github.com/gmatteo/oncvpsp/blob/30f79147b4904f12200b8c7abecb421c153083b3/src/upfout.f90#L95).

azadoks:
> BTW with our "PseudoLibrary" we could easily do that, by supplying an extra file that has the cutoff data, once we have settled on good values.

This would be super nice to have in general, particularly also for the energy cutoffs recommended by PseudoDojo in their .djrepo files.

PR #909, opened 2023-11-21
Fix bug with spin and local_nonlinearity

antoine-levitt:


antoine-levitt:
We should make sure this is tested

antoine-levitt:
Closing because clémentine is going to supersede this in another PR

PR #910, opened 2023-11-22
Misc fixes

epolack:
Tivial PR (I had this in another PR) but should make tests pass. At least two of them.
I updated the affiliation, but did not update @antoine-levitt mail in case he wants to keep the old one (which also appears in `Project.toml`).


antoine-levitt:
Thanks for this! Both email are fine.

PR #916, opened 2023-11-30
Suggestion for CI simplification

mfherbst:
@epolack This is my suggestion to simplify the CI handling. Does this satisfy your needs ?


mfherbst:
Not sure. I think long term the way it should work is that everything is run single-threaded using CPUs and then we additionally tag some important tests, that are *also* being run by `:gpu`, `:mpi`, `:windows`, `:macos` or a respective subset. But let's first settle on something that works for all of us and then refactor.

PR #921, opened 2023-12-08
Atoms calculators compatible Calculator

CedricTravelletti:
This PR implements an AtomsCalculator calculator for DFTK, with the goal of using it for geometry optimization (see [https://github.com/CedricTravelletti/GeomOpt.jl](https://github.com/CedricTravelletti/GeomOpt.jl)).


src/DFTK.jl
```diff
@@ -18,6 +18,9 @@ using Random
 using ChainRulesCore
 using PrecompileTools
 
+using AtomsBase
+using AtomsCalculators
```
mfherbst, line 22: `using AtomsBase` should already be done somewhere, so not needed. The `using AtomsCalculators` you best do just in the file implementing the interface.

src/atoms_calculators.jl
```diff
@@ -0,0 +1,106 @@
...
+# TODO: Find out if symmetries can be re-enabled (see issue on GH).
+
+struct DFTKState
+    scfres::NamedTuple
+end
+function construct_dummy_state(basis)
+    ρ = guess_density(basis)
+    ψ = nothing # Will get initialized by SCF.
+    return DFTKState((;ρ, ψ, basis))
+end
```
mfherbst, line 18: Actually all we need should be `DFTKState(system::AbstractSystem, params::DFTKParameters)`. From this you should be able to get a decent dummy state and do the `prepare_basis`

src/atoms_calculators.jl
```diff
@@ -0,0 +1,106 @@
...
+    
+    # Note that we use the state's densities and orbitals, but change the basis 
+    # to reflect system changes.
+    scfres = self_consistent_field(basis, ρ=state.scfres.ρ, ψ=state.scfres.ψ, tol=calculator.params.tol, callback=calculator.scf_callback)
+    calculator.state = DFTKState(scfres)
+    return calculator.state.scfres.energies.total
+end
+    
+AtomsCalculators.@generate_interface function AtomsCalculators.forces(
+        system::AbstractSystem, calculator::DFTKCalculator; cartesian=false)
```
mfherbst, line 80: Again, just always compute cartesian. Also I believe the interface needs to have units (depends what the convention is, to be checked and synchronised with the others).

src/atoms_calculators.jl
```diff
@@ -0,0 +1,106 @@
...
+        _compute_forces = compute_forces
+    else
+        _compute_forces = compute_forces_cart
+    end
+    scfres = self_consistent_field(basis, tol=calculator.params.tol, callback=calculator.scf_callback)
+    calculator.state = DFTKState(scfres)
+    return _compute_forces(calculator.state.scfres)
+end
+AtomsCalculators.@generate_interface function AtomsCalculators.forces(
+        system::AbstractSystem, calculator::DFTKCalculator, state::DFTKState; cartesian=false)
```
mfherbst, line 92: merge into the above. Just have the argument be `state=calculator.state`

src/atoms_calculators.jl
```diff
@@ -0,0 +1,106 @@
...
+    ψ = nothing # Will get initialized by SCF.
+    return DFTKState((;ρ, ψ, basis))
+end
+
+struct DFTKParameters
+    Ecut::Real
+    kgrid::Union{Nothing,<:AbstractVector{Int}}
+    tol::Real
+    temperature::Real
+end
```
mfherbst, line 25: I would do as in https://github.com/JuliaMolSim/ElectronicStructure.jl/blob/master/src/dftk.jl, i.e. just have the generic dicts here.

src/atoms_calculators.jl
```diff
@@ -0,0 +1,106 @@
...
+
+function prepare_basis(system::AbstractSystem, params::DFTKParameters)
+    model = model_LDA(system; temperature=params.temperature, symmetries=false)
+    basis = PlaneWaveBasis(model; params.Ecut, params.kgrid)
+    return basis
+end
+
+function DFTKCalculator(system; Ecut::Real, kgrid::Union{Nothing,<:AbstractVector{Int}}, tol=1e-6,
+        temperature=zero(Real), verbose_scf=false, state=nothing
+    )
```
mfherbst, line 41: Here have only a low-level constructor for now taking the dicts (basically can be done using the `Base.@kwdef`). I agree it's not a bad idea to set `callback=identity` by default in the `scf_kwargs`. If you want you can also make a high-level constructor a bit like the one you have here that maps down to the low-level one.

CedricTravelletti:
Fixes are implemented now. Ready for another round of review.

src/atoms_calculators.jl
```diff
@@ -0,0 +1,76 @@
...
+    basis_kwargs = (; )
+    scf_kwargs = (; )
+end
+
+struct DFTKState
+    scfres::NamedTuple
+end
+function DFTKState(system::AbstractSystem, params::DFTKParameters)
+    basis = prepare_basis(system, params)
+    ρ = guess_density(basis)
```
mfherbst, line 25:
```suggestion
    ρ = guess_density(basis, system)
```
as systems can be magnetic (and this has an influence on the basis)

src/atoms_calculators.jl
```diff
@@ -0,0 +1,76 @@
...
+    ψ = nothing # Will get initialized by SCF.
+    return DFTKState((;ρ, ψ, basis))
```
mfherbst, line 27:
```suggestion
    return DFTKState((; ρ, ψ, basis))
```

src/atoms_calculators.jl
```diff
@@ -0,0 +1,76 @@
...
+end
+
+function prepare_basis(system::AbstractSystem, params::DFTKParameters)
+    # By default create an LDA model.
+    model = model_LDA(system; symmetries=false, params.model_kwargs...)
```
mfherbst, line 32: This should be a little more flexible to be useful. Use `model_DFT` here, which takes the XC functionals as an argument after the system, a bit like this:
https://github.com/mfherbst/AiidaDFTK.jl/blob/master/src/AiidaDFTK.jl#L52
These you can also store in DFTKParameters inside the `model_kwargs` maybe and then use a `pop!` or similar to single the `xc` arguments out. Alternatively you can also make an extra method of `model_DFT` that takes the `xc` as a kwarg (that might actually be the better solution here)

src/atoms_calculators.jl
```diff
@@ -0,0 +1,76 @@
...
+    basis = PlaneWaveBasis(model; params.basis_kwargs...)
+    return basis
```
mfherbst, line 34:
```suggestion
    PlaneWaveBasis(model; params.basis_kwargs...)
```
No explicit return if not needed, no need to introduce a variable.

src/atoms_calculators.jl
```diff
@@ -0,0 +1,76 @@
...
+end
+
+Base.@kwdef mutable struct DFTKCalculator
+    params::DFTKParameters
+    state::DFTKState = nothing
+end
+
+function DFTKCalculator(system; state=nothing, verbose_scf=false, model_kwargs, basis_kwargs, scf_kwargs)
+    if !verbose_scf
+        scf_kwargs = (; scf_kwargs..., callback = (x) -> nothing)
```
mfherbst, line 44: There is the function `merge` to do this.

src/atoms_calculators.jl
```diff
@@ -0,0 +1,76 @@
...
+    end
+    params = DFTKParameters(;model_kwargs, basis_kwargs, scf_kwargs)
+
+    # Create dummy state if not given.
+    if isnothing(state)
+        state = DFTKState(system, params)
+    end
+
+    return DFTKCalculator(params, state)
```
mfherbst, line 53: no return

src/atoms_calculators.jl
```diff
@@ -0,0 +1,76 @@
...
+end
+
+AtomsCalculators.@generate_interface function AtomsCalculators.potential_energy(system::AbstractSystem, calculator::DFTKCalculator; state = calculator.state)
```
mfherbst, line 56: Line too long (we keep it around 90--92)

src/atoms_calculators.jl
```diff
@@ -0,0 +1,76 @@
...
+    # Update basis (and the enregy terms within).
+    basis = prepare_basis(system, calculator.params)
+    
+    # Note that we use the state's densities and orbitals, but change the basis 
+    # to reflect system changes.
+    scfres = self_consistent_field(basis; ρ=state.scfres.ρ, ψ=state.scfres.ψ, calculator.params.scf_kwargs...)
```
mfherbst, line 62: can skip the `ρ=` and `\psi=`

src/atoms_calculators.jl
```diff
@@ -0,0 +1,76 @@
...
+    calculator.state = DFTKState(scfres)
+    return calculator.state.scfres.energies.total
+end
+    
+AtomsCalculators.@generate_interface function AtomsCalculators.forces(system::AbstractSystem, calculator::DFTKCalculator; state = calculator.state)
+    # Update basis (and the enregy terms within).
+    basis = prepare_basis(system, calculator.params)
+    
+    # Note that we use the state's densities and orbitals, but change the basis 
+    # to reflect system changes.
```
mfherbst, line 72: Copying the code is ok for these small functions, but certainly no need to copy the comments word by word. Just don't have them.

CedricTravelletti:
Fixes implemented. Also added test for energy and force computations.

test/atoms_calculators.jl
```diff
@@ -0,0 +1,23 @@
+@testitem "Test AtomsCalculators energy and forces interface" tags=[:dont_test_mpi, :dont_test_windows] setup=[TestCases] begin
```
mfherbst, line 1: Why not test MPI or windows ?

test/atoms_calculators.jl
```diff
@@ -0,0 +1,23 @@
...
+    using AtomsBase
+    using AtomsCalculators
+    silicon = TestCases.silicon
+
+    # Converto to AtomsBase system
+    silicon = periodic_system(silicon.lattice, silicon.atoms, silicon.positions)
+    
+    model_kwargs = (; functionals = [:lda_x, :lda_c_pw])
+    basis_kwargs = (; kgrid = [4, 4, 4], Ecut = 5.0)
+    scf_kwargs = (; tol = 1e-7)
```
mfherbst, line 12: We don't put spaces after `=` in kwargs and named tuples.

test/atoms_calculators.jl
```diff
@@ -0,0 +1,23 @@
...
+    calculator = DFTKCalculator(silicon; model_kwargs, basis_kwargs, scf_kwargs)
+
+    energy = AtomsCalculators.potential_energy(silicon, calculator)
+    ref_energy = -7.8691334531
+    @test isapprox(energy, ref_energy; rtol=1e-3)
```
mfherbst, line 17:
```suggestion
    @test energy ≈ ref_energy rtol=1e-3
```

test/atoms_calculators.jl
```diff
@@ -0,0 +1,23 @@
+@testitem "Test AtomsCalculators energy and forces interface" tags=[:dont_test_mpi, :dont_test_windows] setup=[TestCases] begin
+    using DFTK
+    using AtomsBase
+    using AtomsCalculators
+    silicon = TestCases.silicon
+
+    # Converto to AtomsBase system
+    silicon = periodic_system(silicon.lattice, silicon.atoms, silicon.positions)
```
mfherbst, line 8: Maybe add some noise to the positions to ensure the forces are not zero.

test/atoms_calculators.jl
```diff
@@ -0,0 +1,23 @@
...
+    calculator = DFTKCalculator(silicon; model_kwargs, basis_kwargs, scf_kwargs)
+
+    energy = AtomsCalculators.potential_energy(silicon, calculator)
+    ref_energy = -7.8691334531
+    @test isapprox(energy, ref_energy; rtol=1e-3)
+
+    forces = AtomsCalculators.forces(silicon, calculator)
+    ref_forces = [[6.071013955320855e-10, -4.690795889974796e-10, -6.299121885664237e-10],
+                  [4.750013095864486e-10, -3.6292771194166813e-10, 3.5873004768426427e-10]]
+    @test isapprox(forces, ref_forces; atol=1e-5)
```
mfherbst, line 22: same as above

src/atoms_calculators.jl
```diff
@@ -0,0 +1,76 @@
...
+end
+function DFTKState(system::AbstractSystem, params::DFTKParameters)
+    basis = prepare_basis(system, params)
+    ρ = guess_density(basis, system)
+    ψ = nothing # Will get initialized by SCF.
+    DFTKState((; ρ, ψ, basis))
+end
+
+function prepare_basis(system::AbstractSystem, params::DFTKParameters)
+    # By default create an LDA model.
```
mfherbst, line 31: Obsolete comment.

src/atoms_calculators.jl
```diff
@@ -0,0 +1,76 @@
...
+# see https://github.com/JuliaMolSim/Molly.jl/blob/master/src/types.jl
+#
+# TODO: Decide how Hamiltonian terms should be updated upon change of the positions.
+# These terms are held inside the `basis`. Currently, the whole basis is rebuilt.
+# TODO: Find out if symmetries can be re-enabled (see issue on GH).
+
+using AtomsBase
+using AtomsCalculators
+
+Base.@kwdef struct DFTKParameters
```
mfherbst, line 14: Maybe add a super brief docstring, documenting the minimal requirements:
```suggestion
"""
Construct a [AtomsCalculators](https://github.com/JuliaMolSim/AtomsCalculators.jl) compatible calculator
for DFTK. The `model_kwargs` are passed onto the [`Model`](@ref) constructor, the `basis_kwargs` to the [`PlaneWaveBasis`](@ref) constructor, the `scf_kwargs` to [`self_consistent_field`](@ref). At the very least the DFT `functionals` and the `Ecut` needs to be specified.

## Example
```julia-repl
julia> DFTKParameters(; model_kwargs=(; functionals=[:lda_x, :lda_c_vwn]),
                       basis_kwargs=(; Ecut=10, kgrid=(2, 2, 2)),
                       scf_kwargs=(; tol=1e-4))
```
"""
```


(the formatting is messed up ... look at the code and copy n' paste what is between the `suggestion` codeblock.)

File: src/atoms_calculators.jl [hunk 23]
```diff
@@ -0,0 +1,76 @@
...
+end
+    
+AtomsCalculators.@generate_interface function AtomsCalculators.forces(
+        system::AbstractSystem, calculator::DFTKCalculator; state = calculator.state)
+    # Update basis (and the enregy terms within).
+    basis = prepare_basis(system, calculator.params)
+    
+    scfres = self_consistent_field(basis;
+                    state.scfres.ρ, state.scfres.ψ, calculator.params.scf_kwargs...)
+    calculator.state = DFTKState(scfres)
```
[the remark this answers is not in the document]
CedricTravelletti: I do not agree on this one. It seems to me that what is being done here is just to compute forces, there is not "true" updating of the geometry going on.

CedricTravelletti:
All the above changes are implemented, except one (see last comment). 
Ready for another round of review.

src/atoms_calculators.jl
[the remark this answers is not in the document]
mfherbst: Yes, I agree the suggested name is bad, but my point was to somehow avoid the 4 repetitive lines in energy and forces. Because as we add more properties (likely going to happen) we will need to repeat them over and over again.

mfherbst:
Note that there are still some unaddressed comments in `src/atoms_calculators.jl`. But I agree with all changes you made so far.

test/atoms_calculators.jl
mfherbst: You did not address this.

src/atoms_calculators.jl
```diff
@@ -0,0 +1,75 @@
+#
+# Define AtomsCalculators interface for DFTK.
+#
+# This interface is inspired by the one used in Molly.jl, 
+# see https://github.com/JuliaMolSim/Molly.jl/blob/master/src/types.jl
+#
+# Important: By default, when the calculator is called with a `state`, the
```
mfherbst, line 7: Sure it's important, that's why it is there ;).

src/atoms_calculators.jl
```diff
@@ -0,0 +1,75 @@
...
+    if isnothing(state)
+        state = DFTKState(system, params)
+    end
+    DFTKCalculator(params, state)
+end
+
+function compute_scf!(system::AbstractSystem, calculator::DFTKCalculator, state::DFTKState)
+    model = model_DFT(system; calculator.params.model_kwargs...,
+                      symmetries=state.scfres.basis.symmetries)
+    # Note that we re-use the symmetries from the state, to avoid degenerate cases.
```
mfherbst, line 54: This comment looks out of place.

src/atoms_calculators.jl
```diff
@@ -0,0 +1,75 @@
...
+end
+
+struct DFTKState
+    scfres::NamedTuple
+end
+function DFTKState(system::AbstractSystem, params::DFTKParameters)
+    model = model_DFT(system; params.model_kwargs...)
+    basis = PlaneWaveBasis(model; params.basis_kwargs...)
+    ρ = guess_density(basis, system)
+    ψ = nothing # Will get initialized by SCF.
```
mfherbst, line 28:
```suggestion
    ψ = nothing  # Will get initialized by SCF.
```

src/atoms_calculators.jl
```diff
@@ -0,0 +1,75 @@
...
+end
+
+function compute_scf!(system::AbstractSystem, calculator::DFTKCalculator, state::DFTKState)
+    model = model_DFT(system; calculator.params.model_kwargs...,
+                      symmetries=state.scfres.basis.symmetries)
+    # Note that we re-use the symmetries from the state, to avoid degenerate cases.
+    basis = PlaneWaveBasis(model; calculator.params.basis_kwargs...)
+    
+    # Note that we use the state's densities and orbitals, but change the basis 
+    # to reflect system changes.
```
mfherbst, line 58: This is obvious, because usually one would not specify the `ρ` and `ψ` kwargs.
```suggestion
```

src/atoms_calculators.jl
mfherbst, line 54: Ah now I understand. Don't use comments to refer to the line *above* (unless absolutely necessary and then with clear explicit indication).
```suggestion
    # Note that we re-use the symmetries from the state, to avoid degenerate cases.
    model = model_DFT(system; calculator.params.model_kwargs...,
                      symmetries=state.scfres.basis.symmetries)
```

src/atoms_calculators.jl
```diff
@@ -0,0 +1,76 @@
...
+Base.@kwdef mutable struct DFTKCalculator
+    params::DFTKParameters
+    state::DFTKState = nothing
+end
+
+function DFTKCalculator(system; state=nothing, verbose_scf=false, model_kwargs, basis_kwargs, scf_kwargs)
+    if !verbose_scf
+        scf_kwargs = merge(scf_kwargs, (;callback = identity))
+    end
+    params = DFTKParameters(;model_kwargs, basis_kwargs, scf_kwargs)
```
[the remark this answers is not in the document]
mfherbst:
```suggestion
    params = DFTKParameters(; model_kwargs, basis_kwargs, scf_kwargs)
```

src/atoms_calculators.jl
```diff
@@ -0,0 +1,75 @@
...
+
+Base.@kwdef mutable struct DFTKCalculator
+    params::DFTKParameters
+    state::DFTKState
+end
+
+function DFTKCalculator(system; state=nothing, verbose=false,
+                        model_kwargs, basis_kwargs, scf_kwargs)
+    if !verbose
+        scf_kwargs = merge(scf_kwargs, (;callback=identity))
```
mfherbst, line 40:
```suggestion
        scf_kwargs = merge(scf_kwargs, (; callback=identity))
```

src/atoms_calculators.jl
mfherbst:
```suggestion
# By default, when the calculator is called with a `state`, the 
```

mfherbst:
@CedricTravelletti Also take a look why the tests fail.

CedricTravelletti:
> @CedricTravelletti Also take a look why the tests fail.

On my local machine, the `test/silicon_lda.jl` fails because it is unable to find non-fractional occupations. 
This does not seem related to this PR (can be fixed by adding temperature).

mfherbst:
>  This does not seem related to this PR (can be fixed by adding temperature).

Indeed, but suspicious it pops up.

mfherbst:
The problem is you do
```
silicon.positions[1] += [0.05, 0, 0]
```
which modifies the testcase.

mfherbst:
@

mfherbst:
I tried to refactor a little to avoid the need to pass a system at all at Calculator construction.

@CedricTravelletti What do you think about this idea ?

Oh and you made a small mistake in passing the `basis.symmetries` to the Model. These symmetries are *not* necessarily the same as the basis symmetries also take the discretisation into account (which can break some symmetries itself).

PR #924, opened 2023-12-09
Remove collinear spin c-interface, which is now native in spglib

mfherbst:
Needs Spglib main to be released as 0.8.6.


singularitti:
Actually, I was going to release it as v0.9.0 according to semantic versioning since this is a new feature.

mfherbst:
> since this is a new feature.

I'm not sure why you would want to do that. If it's just a new feature it's not breaking, so no need to increment the minor in pre-1.0 versions.

Needs https://github.com/thchr/Brillouin.jl/pull/28

singularitti:
Hi Michael, I checked the semantic versioning rules:

> - Patch version Z (x.y.Z | x > 0) MUST be incremented if only backward compatible bug fixes are introduced. A bug fix is defined as an internal change that fixes incorrect behavior.

> - Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backward compatible functionality is introduced to the public API. It MUST be incremented if any public API functionality is marked as deprecated. It MAY be incremented if substantial new functionality or improvements are introduced within the private code. It MAY include patch level changes. Patch version MUST be reset to 0 when minor version is incremented.

According to semantic versioning rules (more explicitly defined for versions greater than 0.x.y), the increment of the patch version is intended for bug fixes, and the increment of the minor version is for adding new functionality. While there aren't clear-cut rules for versions below 1.0.0, the principle remains. The focus of minor version increments is on introducing 'new functionality,' and for patch versions, it's on 'bug fixes.' I think it's better to keep this consistency.

Are there specific norms for versioning in pre-1.0.0 releases that you're aware of? I'm open to insights on this.

Besides, there is a possible breaking changes for the users (https://github.com/singularitti/Spglib.jl/pull/174), where `MVector ` is replaced by `ReducedCoordinates` (an immutable vector type defined [here](https://github.com/MineralsCloud/CrystallographyCore.jl/blob/v0.3.3/src/cell.jl#L8-L13)). Although I don't think it should cause some big surprises for most people.

mfherbst:
I think that's a convention in the Julia community that I read somewhere. For sure the PRs in the General registry are marked as BREAKING for major bumps post-1.0 and for minor bumps pre-1.0.

singularitti:
It seems that the Julia community is particularly cautious about introducing breaking changes, which explains their focus on highlighting potential disruptions.

With this in mind, I'm ready to launch version 1.0, as it encompasses all the major features. What are your thoughts on this milestone?

Looking ahead, I'm considering some significant modifications for version 2.0. One such change involves [removing `Lattice` as a subtype of `AbstractMatrix`](https://github.com/thchr/Brillouin.jl/issues/26#issuecomment-1761252330), along with a few other adjustments I'd like to explore. However, I'm committed to maintaining the functionality of your existing code, so I'm planning for version 1.0+ to be part of a Long-Term Support (LTS) series. I will inform all `Spglib.jl` users about these developments in a forthcoming issue.

mfherbst:
> With this in mind, I'm ready to launch version 1.0, as it encompasses all the major features. With 1.0+, there will be no ambiguity. What are your thoughts on this milestone?

Sure, if you think it's ready, go for it.

> One such change involves removing Lattice as a subtype of AbstractMatrix, along with a few other adjustments I'd like to explore.

I see. Interesting design choice. In my opinion a lattice can just be a matrix, exactly because it is after all a linear mapping between fractional and cartesian space, so sometimes you do want to use standard linear algebra tools on it. But ok, YMMV.

singularitti:
> it is after all a linear mapping between fractional and cartesian space

I agree. But being a matrix is more than that: you will have all operations defined in `LinearAlgebra` and related packages, e.g., `eigvals` and `eigvecs`, `qr` and `lu`, `tr`, etc. I don't know what they mean on a crystallographic lattice. For the conversion between fractional and cartesian spaces, we could just define `*` and `\` on lattices, or use [functors](https://github.com/MineralsCloud/CrystallographyBase.jl/blob/37ed51c8c02b6e81eeef07332bdbc1065ea148b5/test/transform.jl#L13-L17).

antoine-levitt:
I've not followed the discussion too closely so I might be off here, but generally I find that being too strict about allowed operations is more annoying than actively useful. You're basically trading off convenience against a form of defensive programming. That's sometimes a good tradeoff, but not always. In general julia tends to favor "composition over inheritance", ie if something has some characteristics, just have it be a field of the structure instead of having the structure itself behave with the specified characteristic. I also personally tend to prefer designs with functions and plain data types like arrays instead of pushing control flow into clever structures, but that might be just me.

singularitti:
@antoine-levitt You're probably right. But even if I choose to subtype `AbstractLattice` from `AbstractMatrix`, the method is still not unique.

For example, we know `Lattice` is a 3×3 matrix, then how do I define `Base.similar` for `Lattice? Should I subtype it from `StaticMatrix{3,3,T}` defined in `StaticArrays.jl`? I did [try this once](https://github.com/MineralsCloud/CrystallographyCore.jl/pull/9), but it turned out more troublesome than I thought (even though it is more performant), so I finally [changed it back](https://github.com/MineralsCloud/CrystallographyCore.jl/pull/11).

I also found out it is confusing to [define `+`, `-`, `*` & `/` for `Lattice`s](https://github.com/MineralsCloud/CrystallographyCore.jl/pull/11/files#diff-94ffd526188fb689992655802a9d798f4fcf28e0d16b68e6a5a41cba9f18321eL176-R175). What does it mean for a `Lattice` plus a column `Vector`? How about a row `Vector`? Shifting? It's actually not that easy since the conversion between fractional and Cartesian spaces is by convention for lattices whose 3 basis vectors join at the origin in the Cartesian space. That's why I defined [`ShiftedCell`s](https://github.com/MineralsCloud/CrystallographyBase.jl/blob/main/src/shifted.jl#L37-L40) since I want to plot origin-shifted cells in space.

I wanted to tread carefully before confusing myself. I wish I could have explored more on how to use Julia to do this.

antoine-levitt:
That's where the whole paradigm of composition over inheritance: instead of
```
struct Lattice <: AbstractMatrix
    __my_private_data_do_not_touch_or_I_will_kill_you
end
*(x::Lattice, y) = x.__my_private_data_do_not_touch_or_I_will_kill_you * y
... basically reimplement the array API
```
you simply do
```
struct Lattice
    basis
end
reciprocal(x::Lattice) = Lattice(inv(basis'))
```
and you let the users of Lattice use basis when they want "low-level access". We've followed this in DFTK and I feel it has stopped us from going overboard with needless abstractions (we've possibly even gone too far in that direction, but it's simpler to add abstraction after the fact than it is to remove it). This is the kind of design that is super frowned upon in OOP (although "composition over inheritance" is also a thing in OOP), where accessing fields directly is basically a cardinal sin, but it makes less sense in Julia (to be clear: there are valid reasons for wanting to hide state from the outside world, eg if you want to preserve invariants or decouple implementation from interface, but it's just not the case for this kind of simple wrapper structure).

antoine-levitt:
Basically, to be a bit provocative: what do the 200 lines of https://github.com/MineralsCloud/CrystallographyCore.jl/blob/main/src/lattice.jl add to the world?

singularitti:
I'm not sure that I follow this discussion. I have never doubted "composition over inheritance", I was talking about what operations (API) are **well-defined** on lattices.

First, I want to distinguish `Lattice` & `ReciprocalLattice`, since in my opinion users should not directly generate `ReciprocalLattice` from constructors, they should always use `reciprocal(lattice)` to get `ReciprocalLattice`s. Also, sometimes reciprocal lattices could have different coordinate conversion rules than lattices, if their basis vectors are stacked horizontally (although in the current implementation they are not).

Second, there are some operations that I don't think are **well-defined** on lattices, but are automatically valid for matrix-based lattices:

```julia
julia> using CrystallographyCore  # v0.3.3

julia> lattice = Lattice([1, 0, 0], [0, 1, 0], [0, 0, 1])
3×3 Lattice{Int64}
 1  0  0
 0  1  0
 0  0  1


julia> lattice + rand(3, 3)  # What does it mean?
3×3 Lattice{Float64}
 1.5014863697678984  0.7312723496109128  0.09387166521207868
 0.03245372982602057  1.9990831059837737  0.3028171310629082
 0.4638982244970916  0.5334845994633506  1.0331766280721062

julia> lattice .+ rand(3)  # What does it mean? An expansion of shifting of the origins?
3×3 Lattice{Float64}
 1.643023714353042  0.643023714353042  0.643023714353042
 0.12836310515285643  1.1283631051528564  0.12836310515285643
 0.03587227261008108  0.03587227261008108  1.035872272610081


julia> lattice .+ rand(3)'  # What does it mean? An expansion of shifting of the origins?
3×3 Lattice{Float64}
 1.371864256792086  0.31784154655251984  0.018266877709759544
 0.371864256792086  1.3178415465525197  0.018266877709759544
 0.371864256792086  0.31784154655251984  1.0182668777097597

julia> similar(lattice, 3, 3)
3×3 Lattice{Int64}
 1  4  -1
 2  5  -1
 3  6  0


julia> similar(lattice, 3, 4)  # Should I allow `similar` for size 3×4 matrix?
ERROR: ArgumentError: invalid dims `(3, 4)` for `Lattice`!
Stacktrace:
 [1] similar(lattice::Lattice{Int64}, ::Type{Int64}, dims::Tuple{Int64, Int64})
   @ CrystallographyCore ~/.julia/dev/CrystallographyCore/src/lattice.jl:163
 [2] similar(::Lattice{Int64}, ::Int64, ::Int64)
   @ Base ./abstractarray.jl:829

julia> qr(lattice)  # What does it mean?
QR{Float64, Lattice{Float64}, Vector{Float64}}
Q factor: 3×3 LinearAlgebra.QRPackedQ{Float64, Lattice{Float64}, Vector{Float64}}
R factor:
3×3 Matrix{Float64}:
 -1.0   0.0  0.0
  0.0  -1.0  0.0
  0.0   0.0  1.0

julia> lattice == I  # A lattice is equivalent to an identity matrix? They are of different types!
true
```

Making a lattice a matrix automatically transferring so many features that bond with matrices, which I'm not sure whether they are applicable on the physical entity.

singularitti:
> what do the 200 lines of https://github.com/MineralsCloud/CrystallographyCore.jl/blob/main/src/lattice.jl add to the world?

I'm only implementing the iterable interface, the indexing interface, and the broadcasting interface, no array interface as suggested by Julia. Those interface are well-defined and usually of no ambiguity.

singularitti:
Another substantial problem is in my other package [LinearElasticityBase.jl](https://github.com/MineralsCloud/LinearElasticityBase.jl):

```julia
julia> using LinearElasticityBase

julia> EngineeringStress(0, 0, 0, 1, 0, 0)
6-element EngineeringStress{Int64}:
 0
 0
 0
 1
 0
 0

julia> b = EngineeringStress(0, 0, 0, 1, 0, 0)
6-element EngineeringStress{Int64}:
 0
 0
 0
 1
 0
 0

julia> a = EngineeringStrain(0, 0, 0, 1, 0, 0)
6-element EngineeringStrain{Int64}:
 0
 0
 0
 1
 0
 0

julia> a == b
true
```

Obviously, `a` and `b` completely different things and should have different methods to dispatch on, but they are "equal" since `==` just compare their elements.

antoine-levitt:
What I meant is that when I design things I'm less focused on types and more on functions. Types I see just as glorified containers. Eg in DFTK's immutable Model we simply have a lattice and a reciprocal lattice as a 3x3 matrix. Then we have functions like vector_red_to_cart(model, vector) or covector_red_to_cart(model, covector) to do conversions, and vector and covectors are plain vectors. An alternative design would be to have separate types for lattices and reciprocal lattices, vectors and covectors, etc. But then you get a lot of headaches code in getting those types right (and being compatible with other types of arrays, like GPU & co), and I'm not sure there's a lot of benefit; certainly not in terms of lines of code, which is the primary metric I'm designing towards. The lack of abstraction also proves useful when debugging performance issues, or when getting new users up to speed. Again that's personal preference and tastes differ, but being basic has served us well in DFTK.

PR #926, opened 2023-12-11
Fix a bug with scaling factors in the Hartree term

ClementineBarat:
Correction of a small bug when using a scaling factor in the Hartree term.


mfherbst:
Thanks !

antoine-levitt:
Congrats on your first PR 🎉

PR #928, opened 2023-12-12
Correction of compute_fermi_level for collinear spins.

ClementineBarat:
This PR changes compute_fermi_level to have the two spin components of a same kpoint be treated together in the case of collinear spins. The eigenvalues of the two spin components are merged before searching for the HOMO and LUMO.

### DIFF
--- a/src/occupation.jl
+++ b/src/occupation.jl
@@ -146,23 +146,33 @@ function compute_fermi_level(basis::PlaneWaveBasis{T}, eigenvalues, ::FermiZeroT
               "occupation $filled_occ. Typically this indicates that you need to put " *
               "a temperature or switch to a calculation with collinear spin polarization.")
     end
-    n_fill = div(n_electrons, n_spin * filled_occ, RoundUp)
+    n_fill = div(n_electrons, filled_occ, RoundUp)
+
+    # For a given kpoint, all the eigenvalues corresponding to different spins must be treated together.
+    kpt_coordinates = unique([basis.kpoints[i].coordinate for i in 1:length(basis.kpoints)])
+    if length(kpt_coordinates) < length(basis.kpoints)
+        merged_spin_eigenvalues = [sort(vcat(eigenvalues[findall(j->basis.kpoints[j].coordinate==coordinate, 1:length(basis.kpoints))]...))
+                for coordinate in kpt_coordinates]
+    else
+        merged_spin_eigenvalues = eigenvalues
+    end
 
     # For zero temperature, two cases arise: either there are as many bands
     # as electrons, in which case we set εF to the highest energy level
     # reached, or there are unoccupied conduction bands and we take
     # εF as the midpoint between valence and conduction bands.
+
     if n_fill == length(eigenvalues[1])
-        εF = maximum(maximum, eigenvalues) + 1
+        εF = maximum(maximum, merged_spin_eigenvalues) + 1
         εF = mpi_max(εF, basis.comm_kpts)
     else
         # highest occupied energy level
-        HOMO = maximum([εk[n_fill] for εk in eigenvalues])
+        HOMO = maximum([εk[n_fill] for εk in merged_spin_eigenvalues])
         HOMO = mpi_max(HOMO, basis.comm_kpts)
         # lowest unoccupied energy level, be careful that not all k-points
         # might have at least n_fill+1 energy levels so we have to take care
         # of that by specifying init to minimum
-        LUMO = minimum(minimum.([εk[n_fill+1:end] for εk in eigenvalues]; init=T(Inf)))
+        LUMO = minimum(minimum.([εk[n_fill+1:end] for εk in merged_spin_eigenvalues]; init=T(Inf)))
         LUMO = mpi_min(LUMO, basis.comm_kpts)
         εF = (HOMO + LUMO) / 2
     end


antoine-levitt:
I think you can just brutally concatenate everything and get rid of n_spin, no? Also should be able to get rid of some of the rest of that function? I think that function should just look like the bisection in FermiBisection but doing the bisection on the index of the eigenvalues rather than on the eigenvalues themselves, ie find the index i such that excess(all_eigs[i]) == 0 and set the fermi level to all_eigs[i] + all_eigs[i+1]. Doesn't look like there's a integer bisection already coded, so just do it by hand I guess.

@gkemlin this came up because we had trouble with SCF for magnetic systems (without temperature), there's no reason to want to have the same number of occupied states in every kpoint, is there?

mfherbst:
The krange_spin function might also be helpful.

For zero temperature there might be hidden assumptions about how many bands are actually converged in the diagonalisation or sth. like that.

antoine-levitt:
We discussed the semantics here, and it seems reasonable to allow a variable number of occupied states per kpoint, but emit a warning. It's helpful when eg something is an insulator but the SCF goes through a phase where it thinks it's a metal. Also doing a metal at zero temperature is relatively sensible too. I'm not sure what the other codes do here.

mfherbst:
I agree, sounds reasonable and useful to have.

gkemlin:
Hi, that's a good question. From my very sparse experience with spins (I think it only concerns what we did with Fe2MnAl), I remember that
- for a fixed spin channel, there is no reason to assume that all kpts have the same number of occupied states
- for a fixed kpt, there is no reason to assume that both spin channels have the same number of occupied states
However, this is was with positive temperature but I guess that at zero temperature there is still no reason to enforce same occupancy for every kpt and spin so that's something reasonable to have, yes. 

Still, just to make sure I get it, at zero temperature, if you look at every kpts with both spin channels taken together, the number of occupied states should be the same everywhere right ? Otherwise I can't see how the Fermi level can en up in a spectral gap.

antoine-levitt:
> Still, just to make sure I get it, at zero temperature (e.g. for ferromagnetic insulators), if you look at every kpts with both spin channels taken together, the number of occupied states should be the same everywhere right ? Otherwise I can't see how the Fermi level can en up in a spectral gap.

Yes, if the thing is an insulator. However, it can possibly happen that during SCF iterations the system gets lost and temporarily believes it's a metal (I think I encountered this issue before). Also, there's no particular reason we shouldn't be able to do metals at zero temperature (it might blow up, but it might not, and we should be able to try).

src/occupation.jl
```diff
@@ -137,34 +137,39 @@ function compute_fermi_level(basis::PlaneWaveBasis{T}, eigenvalues, ::FermiZeroT
...
-    # as electrons, in which case we set εF to the highest energy level
-    # reached, or there are unoccupied conduction bands and we take
-    # εF as the midpoint between valence and conduction bands.
-    if n_fill == length(eigenvalues[1])
-        εF = maximum(maximum, eigenvalues) + 1
-        εF = mpi_max(εF, basis.comm_kpts)
+
+    all_eigenvalues = sort(vcat(eigenvalues...))
+
+    # Bissection method to find  the index of the eigenvalue such that excess_n_electrons = 0
```
antoine-levitt, line 151: One s at bisection, and there's too much space between find and the

src/occupation.jl
```diff
@@ -137,34 +137,39 @@ function compute_fermi_level(basis::PlaneWaveBasis{T}, eigenvalues, ::FermiZeroT
...
+    i_min = 1
+    i_max = length(all_eigenvalues)
+
+    if excess_n_electrons(basis, eigenvalues, all_eigenvalues[i_max]; temperature, smearing) == 0
+        εF = all_eigenvalues[i_max]
```
antoine-levitt, line 156: We used +1 before but I'm not sure why, this is fine.

src/occupation.jl
```diff
@@ -137,34 +137,39 @@ function compute_fermi_level(basis::PlaneWaveBasis{T}, eigenvalues, ::FermiZeroT
...
-
-    # For zero temperature, two cases arise: either there are as many bands
-    # as electrons, in which case we set εF to the highest energy level
-    # reached, or there are unoccupied conduction bands and we take
-    # εF as the midpoint between valence and conduction bands.
-    if n_fill == length(eigenvalues[1])
-        εF = maximum(maximum, eigenvalues) + 1
-        εF = mpi_max(εF, basis.comm_kpts)
+
+    all_eigenvalues = sort(vcat(eigenvalues...))
```
antoine-levitt, line 149: Careful that this will not work with MPI (hopefully the tests should catch this)

src/occupation.jl
```diff
@@ -137,34 +137,39 @@ function compute_fermi_level(basis::PlaneWaveBasis{T}, eigenvalues, ::FermiZeroT
...
+            else 
+                i_max = i
+            end
+        end
+        εF = 1/2*(all_eigenvalues[i_min]+all_eigenvalues[i_max])
+    end
+
+    if !allequal(compute_occupation(basis, eigenvalues, εF; temperature, smearing).occupation)
+        @warn("Not all kpoints have the same number of occupied states, which could mean "*
+              "that a metallic system is treated at zero temperature.")
```
antoine-levitt, line 172: We want this to be fine for an insulating spin-polarized system (ie where occupations are different between different spins, but not different kpoints)

src/occupation.jl
antoine-levitt: You need to wrap and call MPI.Allgatherv here, so that each process does a bisection on the full array. (You cannot do a Gatherv because excess_n_electrons calls mpi_sum and will hang if not all processes participate.)

src/occupation.jl
```diff
@@ -137,41 +138,78 @@ function compute_fermi_level(basis::PlaneWaveBasis{T}, eigenvalues, ::FermiZeroT
...
+            εF = all_eigenvalues[i] + T(1/2)*(all_eigenvalues[i+1]-all_eigenvalues[i])
+            excess = excess_n_electrons(basis, eigenvalues, εF; temperature, smearing)
+            if excess < -tol_n_elec
+                i_min = i
+            elseif excess > tol_n_elec
+                i_max = i
+            else 
+                i_min = i
+                i_max = i
+            end
```
antoine-levitt, line 188: searchsortedfirst

antoine-levitt: searchsorted qui renvoie tous les indices

PR #929, opened 2023-12-12
regression tests & benchmark

epolack:
* [x] If the action is to merge on master, maybe compare to previous known release? Otherwise, makes mostly sense only for PR.
* [x] Locally, the results are consistent. On the CI the target is way to fast, maybe because for the action to make sense, I have to run against the exact same version. I tested with `retune`, and the results are more logical; so I am keeping it that way for now:
  [no-retune](https://github.com/JuliaMolSim/DFTK.jl/actions/runs/7180407054/job/19552599295?pr=929) v.s. [retune](https://github.com/JuliaMolSim/DFTK.jl/actions/runs/7180599170/job/19553213320?pr=929)
* [x] Separate load testing from regression?
* [x] For the humongous tests it still need some work for it to be flexible and easily usable; e.g., allow user to chose baseline, allow to use retune option.
* [x] **Silence tests (https://github.com/JuliaMolSim/DFTK.jl/pull/957)**
* [ ] Add regression tests for Julia Nightly?
* [ ] Document for the developer.
* [ ] For humongous tests, retune not useful
* [ ] Have `@benchmarkable` directly in code?
* [ ] Density and ad tests are too inconsistent between run. Increase load.
* [ ] use DFTK-testproblems as submodule? Modify CI with
  ```
  with:
    submodules: true
  ```
* [ ] The github-bot does not comment on the PR, even if indicated as so. May be a permission issue, but do we really need this feature? Also, will be messed up if two concurrent jobs.

Also: added `@testset` to continue even if some tests fails…

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,11 +21,11 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - {mode: stable,  os: ubuntu-latest,  payload: noslow-example }
-          - {mode: stable,  os: macOS-latest,   payload: noslow         }
-          - {mode: stable,  os: windows-latest, payload: noslow         }
-          - {mode: stable,  os: ubuntu-latest,  payload: noslow-mpi     }
-          - {mode: nightly, os: ubuntu-latest,  payload: noslow         }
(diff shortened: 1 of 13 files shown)


mfherbst:
Superseeded by #1093.

PR #933, opened 2023-12-21
Fixing gpu

epolack:
Ok, maybe I should not have removed `to_cpu`. The GPU CI failed on the exact line.

(Trying to trigger the Gitlab GPU CI with the name of the branch to investigate.)


mfherbst:
>  I had the same issue locally, and it disappeared after I did MPI.install_mpiexecjl() and restarted Julia…

This sounds unrelated. I'm not sure. We do it in the same way other MPI tests seem to do it.

PR #936, opened 2023-12-26
Uniform interface for direct minimisation and density interpolation

mfherbst:
todo

- [x] More tests for density interpolation

### DIFF
--- a/src/scf/direct_minimization.jl
+++ b/src/scf/direct_minimization.jl
@@ -60,7 +60,7 @@ end
 
 """
 Computes the ground state by direct minimization. `kwargs...` are
-passed to `Optim.Options()` and `optim_method` selects the optim approach,
+passed to `Optim.Options()` and `optim_method` selects the optim approach
 which is employed.
 """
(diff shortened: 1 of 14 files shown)


src/scf/direct_minimization.jl
```diff
@@ -60,13 +60,19 @@ end
 
 """
 Computes the ground state by direct minimization. `kwargs...` are
-passed to `Optim.Options()`. Note that the resulting ψ are not
-necessarily eigenvectors of the Hamiltonian.
+passed to `Optim.Options()` and `optim_method` selects the optim approach,
+which is employed.
```
antoine-levitt, line 64: I've noticed this pattern quite a bit in your writing : "the X, which is Y" means that all X are automatically Y, and is used for additional info on the X; "the X which is Y" means that some of the X are Y, and is used to specify which one you're talking about. My german is extremely rusty but I think the mistake comes from german, where there's automatically a comma before dass?

mfherbst: Oh yes this is indeed a German error and one of my frequent ones. Thanks for pointing out.

PR #941, opened 2023-12-29
Crude import of some material from DFTK school

mfherbst:
Because the stuff becomes too outdated otherwise to be useful.


mfherbst:
@antoine-levitt Probably can be improved at many places, but any objections in merging the current state to have it display ?

My idea was to then get students to use this to learn about DFTK and submit PRs with their fixes when needed.

antoine-levitt:
Sure, but can you split categories for DFTK tutorial (what we have now, which basically assumes people have a reasonable idea of DFT/condensed matter) and general (CM/DFT) tutorials?

mfherbst:
Good idea. Maybe like a getting started and a background section ?

antoine-levitt:
Yup!

PR #942, opened 2024-01-04
Add initial exact exchange implementation

mfherbst:
Small overhowl of #717.


mfherbst:
@antoine-levitt Thoughts ?

src/standard_models.jl
```diff
@@ -44,6 +50,20 @@ function model_DFT(lattice::AbstractMatrix,
     model_DFT(lattice, atoms, positions, Xc(functionals); kwargs...)
 end
 
+"""
+Build an Hartree-Fock model from the specified atoms.
+"""
+function model_HF(lattice::AbstractMatrix, atoms::Vector{<:Element},
+                  positions::Vector{<:AbstractVector}; extra_terms=[], kwargs...)
+    @warn "Exact exchange in DFTK is hardly optimised and not yet production-ready."
```
antoine-levitt, line 58: put that check in the instantiation of the exchange term?

src/terms/exact_exchange.jl
```diff
@@ -0,0 +1,76 @@
+@doc raw"""
+Exact exchange term: the Hartree-Exact exchange energy of the orbitals
```
antoine-levitt, line 2: calling it hartree is confusing?

src/terms/exact_exchange.jl
```diff
@@ -0,0 +1,76 @@
...
+end
+
+# Note: Implementing exact exchange in a scalable and numerically stable way, such that it
+# rapidly converges with k-points is tricky. This implementation here is far too simple and
+# slow to be useful.
+#
+# ABINIT/src/66_wfs/m_getghc.F90
+# ABINIT/src/66_wfs/m_fock_getghc.F90
+# ABINIT/src/66_wfs/m_prep_kgb.F90
+# ABINIT/src/66_wfs/m_bandfft_kpt.F90
```
antoine-levitt, line 34: I'd rather not give explicit references to code (as that might give the impression that we reproduced the implementation instead of a clean-room reimplementation)

src/terms/exact_exchange.jl
```diff
@@ -0,0 +1,76 @@
...
+#
+# For further information (in particular on regularising the Coulomb), consider the following
+#      https://www.vasp.at/wiki/index.php/Coulomb_singularity
+#      https://journals.aps.org/prb/pdf/10.1103/PhysRevB.34.4405   (QE default)
+#      https://journals.aps.org/prb/pdf/10.1103/PhysRevB.73.205119
+#      https://journals.aps.org/prb/pdf/10.1103/PhysRevB.77.193110
+#      https://docs.abinit.org/topics/documents/hybrids-2017.pdf (Abinit apparently
+#           uses a short-ranged Coulomb)
```
antoine-levitt, line 42: Yeah that we should meet and think about, also with @ELallinec who is thinking about the Coulomb singularity

antoine-levitt: Definitely as a first attempt it's reasonable to have a short-ranged Coulomb.

antoine-levitt:
Have you tested this in any way? I'm reluctant to merge something that might just be flat out wrong, esp. with the Coulomb singularity. Maybe H2?

antoine-levitt:
Is this even converging at all? There's basically no damping on the Fock exchange by doing it this way. Direct minimization OTOH should work outside the box

antoine-levitt:
Roadmap after discussion:

- be careful about Coulomb singularity (eg implement Gygi-Baldereschi correction as a first step)
- make sure HF postprocessing (ie computing the exchange energy from an already converged LDA calculation) works
- compare it to another code
- implement self-consistency

simonganne01:
Hello, I am a doctoral researcher at the university of Ghent. I am considering DFTK as a main software to run my simulations. But I need the coulomb and exchange interaction between orbitals. I see that you were already succesfull in implementing the coulomb interaction for gamma-point calculations. I made a somewhat adapted version for my specific needs (coulomb interactions between Wannier orbitals). For now i got it to work for gamma point calculations as well. But i struggle to get realistic results for simulations with multiple kpoints. I get relatively large imaginary parts and other inconsistancies. I was wondering if there was any other internal progress made, not yet publicly available. I will share my code shortly as well, but for now i just wanted signal that i am working an a similar problem and open for conversation.

src/terms/exact_exchange.jl
```diff
@@ -0,0 +1,76 @@
...
+
+    for (n, ψn) in enumerate(eachcol(ψk))
+        ψn_real = ifft(basis, kpt, ψn)
+        for (m, ψm) in enumerate(eachcol(ψk))
+            ψm_real  = ifft(basis, kpt, ψm)
+            ρnm_real = conj(ψn_real) .* ψm_real
+            ρnm_fourier = fft(basis, ρnm_real)
+
+            fac_mn = occk[n] * occk[m] / T(2)
+            E -= fac_mn * real(dot(ρnm_fourier .* term.poisson_green_coeffs, ρnm_fourier))
```
simonganne01, line 70: What is the justification for taking the real part here? shouldn't it automatically be real? (except for small numerical errors of course)

antoine-levitt:
Hi and thanks for your interest! I haven't worked on this at all, but I'm still interested so let me know if you want help!

src/terms/exact_exchange.jl
antoine-levitt: You have to take real parts at some point to get a real number out. If you take it out at the end you have to declare E = zero(Complex(T)) or your code is unstable.

mfherbst:
Hi @simonganne01

thanks for your interest and intent to use DFTK !

For the moment nobody is actively working on this.

Effectively the missing piece is to implement an appropriate regularisation for the coulomb operator following the roadmap mentioned in [above reply](https://github.com/JuliaMolSim/DFTK.jl/pull/942?notification_referrer_id=NT_kwDOAKpBhbQxNDU0MDU0NjA0MzoxMTE1Nzg5Mw#issuecomment-1994316568).

If you happen to make any progress in this direction, we would be very happy to integrate it!. We are certainly open for conversation and happy to discuss a PR to DFTK.

antoine-levitt:
After discussion with Lin, the simplest thing to do here is truncated Coulomb

PR #945, opened 2024-01-10
Formatting / code improvement

epolack:


antoine-levitt:
There's also Iterators.flatten to replace reduce(vcat) (might need a `collect` if an array output is requested, but otherwise it should be fine)

src/Smearing.jl
```diff
@@ -43,10 +43,10 @@ function occupation_divided_difference(S::SmearingFunction, x, y, εF, temp)
     else
         f(z) = occupation(S, (z-εF) / temp)
         fder(z) = occupation_derivative(S, (z-εF)/temp) / temp
-        divided_difference_(f, fder, x, y)
+        _divided_difference(f, fder, x, y)
```
antoine-levitt, line 46: Here and at other places: why? _ are for very internal "do not touch outside of immediate context" functions, these look generally useful

src/terms/nonlocal.jl
```diff
@@ -60,28 +60,28 @@ end
...
+                    dPdR = [-2T(π)*im*Gk[α] for Gk in G_plus_k] .* P
                     ψk = ψ[ik]
                     dHψk = P * (C * (dPdR' * ψk))
                     -sum(occupation[ik][iband] * basis.kweights[ik] *
                          2real(dot(ψk[:, iband], dHψk[:, iband]))
                          for iband=1:size(ψk, 2))
-                end  # α
-            end  # r
-        end  # kpt
-    end  # group
```
antoine-levitt, line 84: these are quite neat, don't remove them

src/terms/Hamiltonian.jl
```diff
@@ -155,8 +155,8 @@ end
             @timeit to "divAgrad" begin
                 apply!((; fourier=Hψ[:, iband], real=nothing),
                        H.divAgrad_op,
-                       (; fourier=ψ[:, iband], real=nothing),
-                       ψ_real)  # ψ_real used as scratch
+                       (; fourier=ψ[:, iband], real=nothing);
+                       ψ_scratch=ψ_real)
```
antoine-levitt, line 159: nice

antoine-levitt (commented):
Apart from these, LGTM!

src/Smearing.jl
mfherbst: Don't care too much in this case. I think it was even @antoine-levitt who chose the name :smile:.

mfherbst: Also we should really start a document where we list these conventions (i.e. in this case: "Prefix function name by `_` if it is an internal helper function, which is not of general use and should thus be kept close to the vicinity of the calling function")

mfherbst (commented):
LGTM

mfherbst:
> Note: Renaming q to Gk because of phonons.

ok for me, but did we not settle on `p` for this ?

src/external/wannier_shared.jl
```diff
@@ -209,31 +209,31 @@ end
...
-    kpb = basis.kpoints[ikpb]
-    equivalent_G_vectors = [(iGk, index_G_vectors(basis, kpb, Gk + G_shift))
-                            for (iGk, Gk) in enumerate(G_vectors(basis, k))]
-    iGk   = [eqG[1] for eqG in equivalent_G_vectors if !isnothing(eqG[2])]
-    iGkpb = [eqG[2] for eqG in equivalent_G_vectors if !isnothing(eqG[2])]
+    k        = basis.kpoints[ik]
+    k_plus_b = basis.kpoints[ik_plus_b]
+    equivalent_G_vectors = [(ip, index_G_vectors(basis, k_plus_b, p + G_shift))
+                            for (ip, p) in enumerate(G_vectors(basis, k))]
+    ip        = [eqG[1] for eqG in equivalent_G_vectors if !isnothing(eqG[1])]
```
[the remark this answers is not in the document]
mfherbst: I don't think that is a mistake, because the first entry in a tuple is just an index that is never nothing, but `index_G_vectors` can be nothing (if that `G` vector is not used in the `k`-point specific basis.

src/bzmesh.jl
```diff
@@ -57,8 +57,8 @@ function irreducible_kcoords(kgrid::MonkhorstPack, symmetries::AbstractVector{<:
         ks == 1//2
     end
     rotations = [symop.W for symop in symmetries]
-    qpoints = [Vec3(0, 0, 0)]
-    spg_mesh = Spglib.get_stabilized_reciprocal_mesh(rotations, kgrid.kgrid_size, qpoints;
+    ppoints = [Vec3(0, 0, 0)]
+    spg_mesh = Spglib.get_stabilized_reciprocal_mesh(rotations, kgrid.kgrid_size, ppoints;
```
mfherbst, line 61: No that's really `q` (i.e. phonon stuff)

docs/src/developer/conventions.md
```diff
@@ -10,7 +10,8 @@ to simplify typing them.
 
 - **Reciprocal-space vectors:** ``k`` for vectors in the Brillouin zone,
   ``G`` for vectors of the reciprocal lattice,
-  ``q`` for general vectors
+  ``q`` for phonon vectors,
+  ``p`` for general vectors.
```
mfherbst, line 14: ok, but you should mention what the difference is (one potentially goes out of the irreducible BZ mesh, right and is like an extra grid to deal with non-crystal-conserving perturabations)

docs/src/developer/conventions.md
```diff
@@ -116,3 +117,17 @@ normalized as ``\frac{|\Omega|}{N} \sum_{r} |\psi(r)|^{2} = 1`` where
...
+
+- line lengths should be 92 characters.
+- avoid shortening variable names only for its own sake.
+- named tuples should be explicit, i.e., `(; var=val)` over `(var=val)`.
+- empty callbacks should use `identity`.
+- use `=` to loop over a range but `in` to loop over elements.
+- always format the `where` keyword with explicit braces: `where {T <: Any}`.
+- do not use implicit arguments in functions but explicit keyword.
+- prefix function name by `_` if it is an internal helper function, which is not of general
+  use and should thus be kept close to the vicinity of the calling function.
```
mfherbst, line 133: Thanks for getting this started.

I think this deserves an extra page. We have many more, but they are just not documented yet.

PR #947, opened 2024-01-19
Inline equations fix

epolack:
PR #946 introduced a bug where inline math was not rendered properly, as modifying `:tex` overwrote the default configuration. There were also minor style issues that I have corrected.

However, I have noticed signatures issues that I do not understand. Examples [here](https://docs.dftk.org/stable/api/#DFTK.sphericalbesselj_fast-Union{Tuple{T},%20Tuple{Integer,%20T}}%20where%20T) and [there](https://docs.dftk.org/stable/api/#DFTK.compute_%CE%B4%CF%88!-Union{Tuple{T},%20Tuple{Any,%20PlaneWaveBasis{T,%20VT}%20where%20VT%3C:Real,%20Vararg{Any,%205}},%20Tuple{Any,%20PlaneWaveBasis{T,%20VT}%20where%20VT%3C:Real,%20Vararg{Any,%206}}}%20where%20T).

Signatures done by hand are not taken into accounts, and some are duplicated.


mfherbst:
> However, I have noticed signatures issues that I do not understand. 

Have not looked at the details, but as far as I understand it with [DocStringExtensions](https://github.com/JuliaDocs/DocStringExtensions.jl) we don't need to do the signatures by hand any more, but maybe we use it wrongly.

mfherbst:
> Oh, that's it! I did not noticed it was imported; thanks! I'll do a pass to remove the signatures we have.

ok, but don't do it blindly before checking that DocStringExtensions does not accidentially loose information.

mfherbst:
@epolack Great catch ! Thanks very much !

PR #950, opened 2024-01-25
Update GTH pseudopotential database

mkrack:
- Script efficiency (speed) improved
- Full sync with current GTH-PP database performed


mfherbst:
Thanks @mkrack !

mfherbst:
By the way we are not hung up on the naming convention. In particular long-term the plan is to move to a central pseudopotential registry in the Julia ecosystem (https://github.com/JuliaMolSim/PseudoLibrary). So if there are more/other GTH-type potentials worth having available, over there would be a good place to put them.

mkrack:
@mfherbst Thanks for merging.

The GTH pseudopotentials in https://github.com/cp2k/cp2k-data/tree/master/potentials/Goedecker/cp2k have been generated in continuation of the work of [C. Hartwigsen et al.](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.58.3641) on the LDA parameter sets based on the same recipes and codes for other exchange-correlation functionals, especially GGA (see [Krack 2005](https://doi.org/10.1007/s00214-005-0655-y)). The corresponding [GTH parameter files](https://github.com/cp2k/cp2k-data/tree/master/potentials/Goedecker/cp2k_soc) including the spin-orbit coupling (SOC) parameters have been release only recently, but, if I am not mistaken, SOC is not yet a feature supported by DFTK, but might be of relevance in the future. 

Alternative versions of GTH PP parameter sets are indeed available which have been optimized using the atomic code `ATOM` integrated as a module in CP2K.

It makes sense to keep the parameter files in a separate registry. I suggest just to mirror the relevant xcf sub-folders of the [cp2k-data](https://github.com/cp2k/cp2k-data/tree/master/potentials/Goedecker/cp2k) repository in the case of the GTH (HGH) potentials.

PR #954, opened 2024-02-02
Lattice symmetries

epolack:
PR to work on being able to reproduce Abinit default symmetries behaviour.
Should not be much to do code-wise, but lots of work to check that this is indeed what is done in Abinit + update the doc to explain.

### DIFF
--- a/src/PlaneWaveBasis.jl
+++ b/src/PlaneWaveBasis.jl
@@ -106,10 +106,13 @@ struct PlaneWaveBasis{T,
     ## Symmetry operations that leave the discretized model (k and r grids) invariant.
     # Subset of model.symmetries.
     symmetries::Vector{SymOp{VT}}
-    # Whether the symmetry operations leave the rgrid invariant
+    # Whether the symmetry operations leave the rgrid invariant.
     # If this is true, the symmetries are a property of the complete discretized model.
     # Therefore, all quantities should be symmetric to machine precision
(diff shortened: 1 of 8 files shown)


src/PlaneWaveBasis.jl
```diff
@@ -106,10 +106,13 @@ struct PlaneWaveBasis{T,
...
     # Subset of model.symmetries.
     symmetries::Vector{SymOp{VT}}
-    # Whether the symmetry operations leave the rgrid invariant
+    # Whether the symmetry operations leave the rgrid invariant.
     # If this is true, the symmetries are a property of the complete discretized model.
     # Therefore, all quantities should be symmetric to machine precision
     symmetries_respect_rgrid::Bool
+    # Whether the symmetry operations leave the basis invariant. Otherwise the symmetry
+    # operations will leave the lattice invariant.
+    symmetries_respect_basis::Bool
```
mfherbst, line 115: This is just a very confusing name. Should rather be respect_BZ or respect_kgrid, no ?

Also the boolean flags maybe go out of hand here. Maybe it's better to have a separate struct that collect the symmetry properties (preserves lattice, preserves kgrid, preserves structure, etc.). That might also make it easier for users to specify cases where exactly one symmetry is broken (e.g. along a displacement vector). Just a thought. Feel free to discard it if this is too much work.

Issue #56, opened 2019-11-12
Use Logging


antoine-levitt:
The Logging stdlib is very nicely done. The main design point is separating production of log messages from processing : we can define several log levels, and then decide what we want to do with them - ignore some, print some, write some to an output file, etc. In particular we can just pepper the code liberally with `@info` and `@debug` (or similar) and worry later about if we want to display it or not. Added benefit is we can disable some and the corresponding macro will not get evaluated, which might be a way to do expensive checks.

mfherbst:
On that note I also thought about integrating with [TimerOutputs.jl](https://github.com/KristofferC/TimerOutputs.jl), which is kind of an optional light-weight profiling.

PR #957, opened 2024-02-08
Using `Suppressor` to silence tests

epolack:
I find it always bothering to have to scroll to look where tests fail. This is a simple change that will only output failing tests.
It's a quick alternative to more fine-grained outputs (https://github.com/JuliaMolSim/DFTK.jl/issues/56).

~(I have checked locally that the full backtrace is indeed printed in different failing test cases.)~
Lots of edge cases, and it feels too hacky. Not sure if there's a way for a simpler alternative…

~Let's see if passed tests are printed at the end.~ (see https://github.com/JuliaMolSim/DFTK.jl/actions/runs/7832171350/job/21370246591?pr=851 for example. Strangely error is not printed, but was not printed either in https://github.com/JuliaMolSim/DFTK.jl/actions/runs/7831316138/job/21367378204?pr=851)

Maybe actually better to take the time to use Logging, and rewrite println... Should not be that big à work and much more flexible and clean.


mfherbst:
The test ecosystem has evolved quite a bit and this would need to be restarted from scratch if considered further. Thus I'm closing this for now.

PR #959, opened 2024-03-05
Virial interface for AtomsCalculator

CedricTravelletti:
This PR adds virial calculation to the DFTK AtomsCalculators inteface.

It also fixes the units in the AtomsCalculators interface. 

Indeed, AtomsCalculators.jl specifies that we should return Unitful results.


src/external/atoms_calculators.jl
```diff
@@ -69,12 +71,20 @@ AtomsCalculators.@generate_interface function AtomsCalculators.potential_energy(
...
-    compute_forces_cart(calculator.state.scfres)
+    compute_forces_cart(calculator.state.scfres) * u"hartree/bohr"
+end
+
+AtomsCalculators.@generate_interface function AtomsCalculators.virial(
+        system::AbstractSystem, calculator::DFTKCalculator; state = DFTKState(),
+        kwargs...)
+    compute_scf!(system, calculator, state)
+    - (compute_stresses_cart(calculator.state.scfres)
+     * calculator. state.scfres.basis.model.unit_cell_volume) * u"hartree"
```
mfherbst, line 89: There is a space too much after the `calculator.`

Plus are you sure this is the right unit ? I had in mind it was something of the dimensions energy / volume, but I could be wrong.

test/external/atoms_calculators.jl
```diff
@@ -15,10 +17,52 @@
 
     energy = AtomsCalculators.potential_energy(silicon, calculator)
     ref_energy = -7.86059
-    @test energy ≈ ref_energy rtol=1e-3
+    @test austrip.(energy) ≈ ref_energy rtol=1e-3
```
mfherbst, line 20: no `austrip.`, just `austrip` (this is a scalar, no vector)

src/external/atoms_calculators.jl
CedricTravelletti: According to [AtomsCalculators](https://juliamolsim.github.io/AtomsCalculators.jl/dev/interface-definition/) interface definition: 
`Virial is a square matrix (3x3 in 3D) that has units of force times length or energy`, so here hartrees are appropriate.

CedricTravelletti:
Thank your for the review. 
Comments have been adressed. Ready for another round of review.

test/external/atoms_calculators.jl
```diff
@@ -15,10 +17,52 @@
...
+    silicon = TestCases.silicon
+
+    # Perturb from equilibrium so forces are not 0.
+    positions = [silicon.positions[1] + [0.05, 0, 0], silicon.positions[2]]
+    silicon = periodic_system(silicon.lattice, silicon.atoms, positions)
+
+    model_kwargs = (; temperature=1e-3, functionals=[:lda_x, :lda_c_pw])
+    basis_kwargs = (; kgrid=[4, 4, 4], Ecut=5.0)
+    scf_kwargs = (; tol=1e-7)
+    calculator = DFTKCalculator(; model_kwargs, basis_kwargs, scf_kwargs, verbose=true)
```
mfherbst, line 38: This is now essentially 3 times the same code. Can you make a function singling this out (see other files how this is done).

src/external/atoms_calculators.jl
```diff
@@ -69,12 +71,20 @@ AtomsCalculators.@generate_interface function AtomsCalculators.potential_energy(
...
 AtomsCalculators.@generate_interface function AtomsCalculators.forces(
         system::AbstractSystem, calculator::DFTKCalculator; state = DFTKState(),
         kwargs...)
     compute_scf!(system, calculator, state)
-    compute_forces_cart(calculator.state.scfres)
+    compute_forces_cart(calculator.state.scfres) * u"hartree/bohr"
+end
+
+AtomsCalculators.@generate_interface function AtomsCalculators.virial(
+        system::AbstractSystem, calculator::DFTKCalculator; state = DFTKState(),
```
mfherbst, line 85: We don't put spaces around `=` in Keyword arguments (same above, actually, which I did not catch last time)

mfherbst (commented):
Rest looks good to me.

test/external/atoms_calculators.jl
CedricTravelletti: fixed

CedricTravelletti:
@mfherbst Do you think there are other modifications to be done or we can merge?

PR #960, opened 2024-03-05
Fix normalization of wavefunctions

antoine-levitt:
@azadoks Do you agree? The wavefunctions are not in units of energy, so there should not be a factor of 2 here.


mfherbst:
@azadoks  ping :smile:

azadoks:
Ah yes, my bad!

antoine-levitt:
I don't understand your comment, is the current version (with fix in) correct?

azadoks:
I think the factor of two needs to go, but we should also probably be normalizing the wavefunctions by their norm. Abinit does `chi ./= sqrt.(integrate(r, chi.^2))` where `chi` is the data from the file (i.e. with the factor of `r` included). QE does the same thing, but it also sets the occupations for any wavefunctions which have a norm `< 1e-8` to a small negative number so that they are not used in the initial guess wavefunction.

antoine-levitt:
OK, I'll open an issue

PR #972, opened 2024-05-15
Add a preference option for DFTK threading

antoine-levitt:
Draft because untested so there's probably typos, I'll put it through its paces sometime later, but should be complete.

### DIFF
--- a/docs/src/tricks/parallelization.md
+++ b/docs/src/tricks/parallelization.md
@@ -165,8 +165,8 @@ More than two FFT threads is rarely useful.
 
 ## Advanced threading tweaks
 The default threading setup done by `setup_threading` is to select
-one FFT thread and the same number of BLAS and Julia threads.
-This section provides some info in case you want to change these defaults.
+one FFT thread, and use `Threads.nthreads()` to determine the number of DFTK
+and BLAS threads. This section provides some info in case you want to change these defaults.
 
 ### BLAS threads
 All BLAS calls in Julia go through a parallelized OpenBlas
@@ -191,12 +191,17 @@ where `N` is the number of threads you desire.
 To **check the number of BLAS threads** currently used, you can use `BLAS.get_num_threads()`.
 
 ### DFTK threads
-On top of BLAS threading DFTK uses Julia threads
-in a couple of places to parallelize over ``k``-points (density computation)
-or bands (Hamiltonian application).
-The number of threads used for these aspects is controlled by default by the
-flag `-t` passed to Julia or the *environment variable* `JULIA_NUM_THREADS`.
-It can also be set through `setup_threading(; n_DFTK)`.
+On top of BLAS threading DFTK uses Julia threads in a couple of places
+to parallelize over ``k``-points (density computation) or bands
+(Hamiltonian application). The number of threads used for these
+aspects is controlled by default by `Threads.nthreads()`, which can be
+set using the flag `-t` passed to Julia or the *environment variable*
+`JULIA_NUM_THREADS`. Optionally, you can use `setup_threading(; n_DFTK)`
+to set an upper bound to the number of threads used by DFTK,
+regardless of `Threads.nthreads()` (for instance, if you want to
+thread several DFTK runs and don't want DFTK's threading to
+interfere with that). This is done through `Preferences.jl` and
+requires a restart of Julia.
 
 ### FFT threads
 Since FFT threading is only used in DFTK inside the regions already parallelized
--- a/docs/src/tricks/parallelization.md
+++ b/docs/src/tricks/parallelization.md
@@ -190,13 +190,13 @@ BLAS.set_num_threads(N)
 where `N` is the number of threads you desire.
 To **check the number of BLAS threads** currently used, you can use `BLAS.get_num_threads()`.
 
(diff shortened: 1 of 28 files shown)


src/common/threading.jl
```diff
@@ -1,42 +1,77 @@
 import FFTW
 using LinearAlgebra
 
-function setup_threading(; n_fft=1, n_blas=Threads.nthreads())
-    n_julia = Threads.nthreads()
+function setup_threading(; n_fft=1, n_blas=Threads.nthreads(), n_DFTK=nothing)
```
mfherbst, line 4: Is `n_DFTK` a good name ? I think this is all potentially very confusing on the relationship between all these different threading parameters. Don't you think it would be clearer to just have a boolean flag to indicate whether we use the available Julia threads inside DFTK or not ?

My main grips here is that `n_DFTK` is a compile-time thing while all the other options here are runtime things. So using this function can easily lead to rather unexpected behaviour (one sometimes needs to exit the shell and recompile DFTK, sometimes not).

With a boolean flag and a separate function for that you avoid that issue.

antoine-levitt: I think it makes sense to be able to set the number of threads that DFTK uses rather than just a flag. In some sense the new names are less confusing because they more accurately reflect what's going on: there are exactly three sources of threading, DFTK, BLAS and FFTW, and now we control the number of threads of all these three in this function.

The compile-time thing is confusing, but with the info printing it should be OK? I also added some safeguards.

antoine-levitt:
This works fine locally, so let's merge?

antoine-levitt:
Nope, I figured let's do the simple thing and require a reboot...

src/densities.jl
```diff
@@ -76,7 +76,7 @@ end
     #   |ψ_{n,k}|² is 2 ψ_{n,k} * δψ_{n,k+q}.
     # Hence, we first get the δψ_{[k+q]} as δψ_{k+q}…
     δψ_plus_k = transfer_blochwave_equivalent_to_actual(basis, δψ, q)
-    storages = parallel_loop_over_range(allocate_local_storage, range) do storage, kn
+    storages = parallel_loop_over_range(range; allocate_local_storage) do kn, storage
```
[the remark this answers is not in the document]
antoine-levitt: I don't get what you mean?

antoine-levitt:
Thanks! @mfherbst OK to merge?

src/common/threading.jl
```diff
@@ -1,42 +1,92 @@
...
+    if n_DFTK != nothing
+        set_DFTK_threads!(n_DFTK)
+    end
+    n_DFTK = @load_preference("DFTK_threads", Threads.nthreads())
     FFTW.set_num_threads(n_fft)
     BLAS.set_num_threads(n_blas)
-    mpi_master() && @info "Threading setup:" n_fft n_blas n_julia
+    mpi_master() && @info "Threading setup: $(Threads.nthreads()) Julia threads, $n_DFTK DFTK threads, $n_fft FFT threads, $n_blas BLAS threads"
```
mfherbst, line 11: Why? The other option gave a nice table?
Also line too long :-P

src/common/threading.jl
```diff
@@ -1,42 +1,92 @@
...
+function set_DFTK_threads!()
+    @delete_preferences!("DFTK_threads")
+end
+# If unset, use nthreads() threads
+function get_DFTK_threads()
+    nthreads = @load_preference("DFTK_threads", nothing)
+    if isnothing(nthreads)
+        nthreads = Threads.nthreads()
+    elseif nthreads > Threads.nthreads()
+        error("You set your preference for $n_DFTK DFTK threads, but only ran julia with $(Threads.nthreads()) threads.")
```
mfherbst, line 41: Does such a crash also happen in case the `get_DFTK_threads()` function is not called ? Calling that function might not always be the case.

src/common/threading.jl
mfherbst: Add some docs to this function. Otherwise it's quite confusing now. (In particular mention that Julia threads also exist and are considered, so essentially n_DFTK and Julia threads configure the same thing: One per runtime and one per compile time.

docs/src/tricks/parallelization.md
```diff
@@ -190,13 +190,13 @@ BLAS.set_num_threads(N)
...
-On top of BLAS threading DFTK uses Julia threads (`Thread.@threads`)
+### DFTK threads
+On top of BLAS threading DFTK uses Julia threads
 in a couple of places to parallelize over ``k``-points (density computation)
 or bands (Hamiltonian application).
-The number of threads used for these aspects is controlled by the
+The number of threads used for these aspects is controlled by default by the
 flag `-t` passed to Julia or the *environment variable* `JULIA_NUM_THREADS`.
-To **check the number of Julia threads** use `Threads.nthreads()`.
+It can also be set through `setup_threading(; n_DFTK)`.
```
mfherbst, line 200: I think one should clearly describe the different purpose between Julia threads and DFTK threads:
- Julia threads is a runtime thing
- DFTK threads is a compile time thing (preference)
So it gives two different ways to configure threading.

antoine-levitt:
Review done

PR #975, opened 2024-05-23
Disable timings by default

antoine-levitt:


mfherbst:
I strongly disagree with this. Getting the timing output by default on bigger calculations has more than once helped me retroactively to find issues and track issues in DFTK. If for a project or your workflow timings cause issues, it's not all that hard to disable them locally or globally using the preference system.

Doing it the other way round (enable using preferences) is much harder in automated cluster-based calculations.

antoine-levitt:
Hm, my thought was that timings are actually pretty disruptive (stacktraces, threading) and are of interest mostly to developers who can turn them on easily, so it'd make more sense to turn them off by default. But OK sure if you find it useful let's keep it.

PR #993, opened 2024-08-09
Solutions to discretisation exercises

dianetambey:
finite differences: boundary conditions and formula fix
ex 2: * missing in formula
ex 4: put in square box
sol 1: formula fix

PR #995, opened 2024-08-13
Discretisation-exercises

dianetambey:
ex 6 problem with time plot


mfherbst:
Isn't this a duplicate of #993

PR #997, opened 2024-08-21
Enhance user experience with MPI

abussy:
This PR aims at enhancing the user experience when using MPI parallelization by not stopping when `n_ranks > n_kpt`. 

The current solution of stopping execution with an error message is not optimal. Indeed, for an arbitrary system, it is not trivial to know the number of irreducible K-points in advance. This is particularly annoying when calculations are launched in an automated fashion, where the only safe bet is to not use MPI at all.

The solution proposed here is simple. When a calculation is started with more MPI ranks than K-points, a new MPI communicator is created. This communicator has as many ranks as there are K-points, while the remaining processes exit the program.

While this may lead to idling CPU time, I believe that not crashing improves the experience. Moreover, a warning is printed describing the situation to the user, so that they can optimize their next run. I would also argue that this is not worse than under-parallelizing in some instances where the number of K-point is not a multiple of the number of MPI ranks (maybe we should also issue a warning in such a case?).

I also took the opportunity to fix testing with the `:mpi` tag. The current test on `parse(Bool, get(ENV, "CI", "false"))` in the `PlaneWaveBasis` creation, which essentially disables MPI testing on the CI, has been removed. Instead, all tests involving kgrids with a single k-point have received the `:dont_test_mpi` tag. The number of MPI ranks for MPI testing is also hardcoded to 2: because all tests are run as a single calculation, killing processes when `n_ranks > n_kpt` would make the tests hang. This also fixes local testing via `Pkg.test("DFTK"; test_args = ["mpi"])`.

### DIFF
--- a/src/PlaneWaveBasis.jl
+++ b/src/PlaneWaveBasis.jl
@@ -176,6 +176,35 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Real, fft_size::Tuple{Int, Int, I
     # Right now we split only the kcoords: both spin channels have to be handled
     # by the same process
     n_kpt   = length(kcoords_global)
+
+    # K-points are distributed over MPI ranks. If there are more ranks than k-points,
+    # the code crashes, and very specific fixes of "reducing over empty collections"
+    # would need to be implemented. In order to avoid that, while letting the code run
+    # without crashing, a new MPI communicator is created. This communicator has the same
+    # amount of ranks as there are k-point. All processors not in the communicator exit
+    # at this stage. While this may lead to idle CPU time, the user experience is enhanced.
+    if mpi_nprocs(comm_kpts) > n_kpt
+
+        # Create a new MPI communicator with a maximum of n_kpt procs
+        color = nothing
+        rank_id = mpi_rankid(comm_kpts)
+        nprocs_init = mpi_nprocs(comm_kpts)
+        if rank_id < n_kpt
+            color = 1
+        end
+        comm_kpts = MPI.Comm_split(comm_kpts, color, rank_id)
+
+        # Processes not in the communicator exit the program here
+        if comm_kpts == MPI.COMM_NULL
+            MPI.Finalize()
+            exit()
+        end
+
+        @warn("Attempting to parallelize a calculation of $n_kpt K-points  over $nprocs_init " *
+              "MPI ranks. DFTK cannot handle having more MPI ranks than K-points. The calculation " *
+              "continues with $n_kpt MPI ranks instead.")
+
+    end
     n_procs = mpi_nprocs(comm_kpts)
 
     if n_procs > n_kpt
--- a/src/PlaneWaveBasis.jl
+++ b/src/PlaneWaveBasis.jl
@@ -441,9 +441,12 @@ function irreducible_kcoords_global(basis::PlaneWaveBasis)
 end
 
 function irreducible_kweights_global(basis::PlaneWaveBasis{T}) where {T}
-    @inline function same_kpoint(i_irr, i_dupl)
+    function same_kpoint(i_irr, i_dupl)
         maximum(abs, basis.kcoords_global[i_dupl]-basis.kcoords_global[i_irr]) < eps(T)
     end
+
+    # Check that weights add up to 1 on entry (non spin doubled k-points)
+    @assert sum(basis.kweights_global) ≈ 1
 
     # Assume that duplicated k-points are appended at the end of the kcoords array
     irr_kweights = basis.kweights_global[1:basis.n_irreducible_kpoints]
@@ -451,6 +454,7 @@ function irreducible_kweights_global(basis::PlaneWaveBasis{T}) where {T}
         for i_irr = 1:basis.n_irreducible_kpoints
             if same_kpoint(i_irr, i_dupl)
                 irr_kweights[i_irr] += basis.kweights_global[i_dupl]
+                break
             end
         end
     end
--- a/src/PlaneWaveBasis.jl
+++ b/src/PlaneWaveBasis.jl
@@ -175,37 +175,24 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Real, fft_size::Tuple{Int, Int, I
     # Compute k-point information and spread them across processors
     # Right now we split only the kcoords: both spin channels have to be handled
     # by the same process
+    n_procs = mpi_nprocs(comm_kpts)
     n_kpt   = length(kcoords_global)
 
-    # K-points are distributed over MPI ranks. If there are more ranks than k-points,
-    # the code crashes, and very specific fixes of "reducing over empty collections"
-    # would need to be implemented. In order to avoid that, while letting the code run
-    # without crashing, a new MPI communicator is created. This communicator has the same
-    # amount of ranks as there are k-point. All processors not in the communicator exit
-    # at this stage. While this may lead to idle CPU time, the user experience is enhanced.
-    if mpi_nprocs(comm_kpts) > n_kpt
-
-        # Create a new MPI communicator with a maximum of n_kpt procs
-        color = nothing
-        rank_id = mpi_rankid(comm_kpts)
-        nprocs_init = mpi_nprocs(comm_kpts)
-        if rank_id < n_kpt
-            color = 1
+    # The code cannot handle MPI ranks without k-points. If there are more prcocessas
+    # than k-points, we duplicate k-points with the highest weight on the empty MPI
+    # ranks (and scale the weight accordingly)
+    if n_procs > n_kpt
+        for i in n_kpt+1:n_procs
+            idx = argmax(kweights_global)
+            kweights_global[idx] *= 0.5
+            push!(kweights_global, kweights_global[idx])
+            push!(kcoords_global, kcoords_global[idx])
         end
-        comm_kpts = MPI.Comm_split(comm_kpts, color, rank_id)
-
-        # Processes not in the communicator exit the program here
-        if comm_kpts == MPI.COMM_NULL
-            MPI.Finalize()
-            exit()
-        end
-
-        @warn("Attempting to parallelize a calculation of $n_kpt K-points  over $nprocs_init " *
-              "MPI ranks. DFTK cannot handle having more MPI ranks than K-points. The calculation " *
-              "continues with $n_kpt MPI ranks instead.")
-
+        @warn("Attempting to parallelize $n_kpt k-points over $n_procs MPI ranks. DFTK does " *
+              "not support processes empty of k-point. Some k-points were duplicated over the " *
+              "extra ranks with scaled weights.")
     end
-    n_procs = mpi_nprocs(comm_kpts)
+    n_kpt = length(kcoords_global)
 
     if n_procs > n_kpt
         # XXX Supporting more processors than kpoints would require
--- a/src/PlaneWaveBasis.jl
+++ b/src/PlaneWaveBasis.jl
@@ -58,7 +58,9 @@ struct PlaneWaveBasis{T,
     ## These fields are not actually used in computation, but can be used to reconstruct a basis
     # Monkhorst-Pack grid used to generate the k-points, or nothing for custom k-points
     kgrid::AbstractKgrid
-    # full list of (non spin doubled) k-point coordinates in the irreducible BZ (duplicates possible)
+    # Full list of (non spin doubled) k-point coordinates in the irreducible BZ (duplicates possible)
+    # Best to use the irreducible_kcoords_global() and irreducible_kweights_global() functions 
+    # to insure none of the k-points are duplicated
     kcoords_global::Vector{Vec3{T}}
     kweights_global::Vector{T}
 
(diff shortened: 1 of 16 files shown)


mfherbst (commented):
Great idea, thanks for the PR.

Two comments:
- Calling exit from within PlaneWaveBasis is super unexpected and can easily lead to spurious bugs. I'm thinking of a case where one uses MPI for DFTK *and* to run other things and all of a sudden part of the processes in the global communicator are just gone and the program hangs or does undefined stuff.
- Instead I propose to actually provide a mechanism to do the appropriate splitting of the communicator outside of a `PlaneWaveBasis` call. What I'm thinking is to essentially internally call `build_kpoints` to determine the `length(kcoords_global)` plus a helper function to do the "shrinking" of the communicator plus potentially calling exit from there. The user would then explicitly call 3 functions. The one to determine the actual numer of k-points, then the split and exit function and then PlaneWaveBasis( ) with the communicator of the appropriate length.

Regarding the tests: I had in mind for a long time to actually switch the logic. Instead of "disabling" MPI tests where needed I think it's better to annotate those, which do support MPI instead. (I.e. instead of a `:dont_test_mpi` have a simple `:mpi` on the counter-set). If it's too much work, leave it as is and we do a follow-up ...

mfherbst:
BTW the test failures in the "normal" tests seem fake (because of a too tightly chosen test tolerance, I think), but the cancelled MPI test points to an issue in your implementation.

abussy:
> Calling exit from within PlaneWaveBasis is super unexpected and can easily lead to spurious bugs. I'm thinking of a case where one uses MPI for DFTK and to run other things and all of a sudden part of the processes in the global communicator are just gone and the program hangs or does undefined stuff.

Yes, I agree actually. In fact, I was facing this exact problem when trying to run the tests with more than 2 MPI ranks.

Ideally, for a robust behavior, the extra MPI ranks should not exit the program, but wait at a synchronization barrier at the end of execution. I am not quite sure how to implement that in a light way, without an arbitrarily large `if` block after the creation of the sub-communicator. I will give it some thoughts.

There are 2 important questions to be addressed on the form I think:
- Should an input script look any different when a parallel run is intended?
- Should the user have to think about creating a sub-communicator or setup parallelization when writing their input?

I would tend to say no to both of the above. I'll try to come up with something along these lines.

abussy:
The latest commit brings major simplifications to this MPI issue. Instead of building sub-communicators and killing processes in a dangerous fashion, I propose to duplicate some k-points over the empty MPI ranks. The weights are adjusted in order to keep exactness of the results. A warning is issued as well.

This is the lightest fix I could think of. It is also safe and robust, and the user does not have to think about parallelism in their script. The only potential issue I see is: if the user queries for the k-point mesh after the calculation, they could get a non irreducible k-mesh with duplicates. One possible solution would be to add an extra field to the basis with the original mesh, or write a function that always returns the irreducible mesh.

> Regarding the tests: I had in mind for a long time to actually switch the logic. Instead of "disabling" MPI tests where needed I think it's better to annotate those, which do support MPI instead. (I.e. instead of a :dont_test_mpi have a simple :mpi on the counter-set). If it's too much work, leave it as is and we do a follow-up ...

I don't mind changing from `:dont_test_mpi` to `:mpi` tags in the tests. However, if this PR is accepted, I believe all tests could actually be run in parallel (to be checked by me). In such a case, we could get rid of this tag completely.

> BTW the test failures in the "normal" tests seem fake (because of a too tightly chosen test tolerance, I think), but the cancelled MPI test points to an issue in your implementation (could be related to the exit() btw.)

In my experience, it seems that the `anderson.jl` test is particularly unstable.

mfherbst:
I'm in favour of this. @abussy Please update and then we can merge this !

@antoine-levitt Objections ?

mfherbst:
@abussy  The only thing that we have to be careful with is unfolding the k-point mesh when not using symmetries and during IO operations.

antoine-levitt:
I haven't looked in details but it looks reasonable, go ahead!

abussy:
Notes:
- Added a `n_irreducible_kpoints` filed to the `PlaneWaveBasis` --> used in I/O so that only irreducible k-points are considered
- Added `irreducible_kcoords()` and `irreducible_kweights()` functions for the same reasons
- BZ unfolding currently not implemented with MPI, so duplicated k-point won't affect it. Added an extra error message in case MPI support gets implemented
- Rebased for compatibility with master branch

src/PlaneWaveBasis.jl
```diff
@@ -264,8 +269,25 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Real, fft_size::Tuple{Int, Int, I
...
+    # ranks (and scale the weight accordingly)
+    if n_procs > n_kpt
+        for i in n_kpt+1:n_procs
+            idx = argmax(kweights_global)
+            kweights_global[idx] *= 0.5
+            push!(kweights_global, kweights_global[idx])
+            push!(kcoords_global, kcoords_global[idx])
+        end
+        @warn("Attempting to parallelize $n_kpt k-points over $n_procs MPI ranks. DFTK does " *
+              "not support processes empty of k-point. Some k-points were duplicated over the " *
```
mfherbst, line 287: That line is a bit long.

src/PlaneWaveBasis.jl
```diff
@@ -543,6 +565,29 @@ function weighted_ksum(basis::PlaneWaveBasis, array)
...
+    # Assume that duplicated k-points are appended at the end of the kcoords array
+    basis.kcoords_global[1:basis.n_irreducible_kpoints]
+end
+
+function irreducible_kweights(basis::PlaneWaveBasis{T}) where {T}
+    irr_kweights = basis.kweights_global[1:basis.n_irreducible_kpoints]
+
+    # Assume that duplicated k-points are appended at the end of the kcoords/kweights array
+    for i_dupl in basis.n_irreducible_kpoints+1:length(basis.kweights_global)
+        for i_irr in 1:basis.n_irreducible_kpoints
```
mfherbst, line 582: We use `=` when loops are over integers.

src/common/mpi.jl
```diff
@@ -5,6 +5,7 @@ import MPI
 Number of processors used in MPI. Can be called without ensuring initialization.
 """
 mpi_nprocs(comm=MPI.COMM_WORLD) = (MPI.Init(); MPI.Comm_size(comm))
+mpi_rankid(comm=MPI.COMM_WORLD) = (MPI.Init(); MPI.Comm_rank(comm))
```
mfherbst, line 8: Unused function ... remove or test

src/symmetry.jl
```diff
@@ -414,6 +414,10 @@ function unfold_array(basis_irred, basis_unfolded, data, is_ψ)
     if !(basis_irred.comm_kpts == basis_irred.comm_kpts == MPI.COMM_WORLD)
         error("Brillouin zone symmetry unfolding not supported with MPI yet")
     end
+    if basis_irred.n_irreducible_kpoints < mpi_nprocs(basis_irred.comm_kpts)
+        #Note: if this routine is ever generalised for MPI, need special care for duplicated KP
```
mfherbst, line 418: Space after `#`. May also be a bit too long.

test/PlaneWaveBasis.jl
```diff
@@ -58,7 +58,7 @@ end
     test_pw_cutoffs(silicon, 4.0, [11, 13, 11])
 end
 
-@testitem "PlaneWaveBasis: Check cubic basis and cubic index" setup=[TestCases] begin
+@testitem "PlaneWaveBasis: Check cubic basis and cubic index" setup=[TestCases] tags=[:dont_test_mpi] begin
```
mfherbst, line 61: Just to check: Make sure that at least one test tests the code path with k-point duplication, because missing errors in this from refactors can be super annoying.

src/input_output.jl
```diff
@@ -224,11 +224,21 @@ function band_data_to_dict!(dict, band_data::NamedTuple; save_ψ=false, save_ρ=
...
-            dict[key] = reshape(gathered, (size(data[1])..., n_kpoints, n_spin))
+            n_kpt_tot = length(basis.kcoords_global)
+
+            dict[key] = reshape(gathered, (size(data[1])..., n_kpt_tot, n_spin))
+
+            # Only store irreducible k-points (assumed to be first in an array)
+            if n_kpt_tot > n_kpoints
+                index = ntuple(_ -> Colon(), ndims(dict[key]))
+                index = Base.setindex(index, 1:n_kpoints, ndims(dict[key]) - 1)
+                dict[key] = dict[key][index...]
```
mfherbst, line 240: I don't like that you change the thing in the dict after writing. For me once things are put into the dict they are "final". I'd suggest to introduce a temporary variable to hold the reshaped array and apply the additional masking afterwards.

src/PlaneWaveBasis.jl
```diff
@@ -543,6 +565,29 @@ function weighted_ksum(basis::PlaneWaveBasis, array)
...
+            if maximum(abs.(basis.kcoords_global[i_dupl]-basis.kcoords_global[i_irr])) < eps(T)
```
mfherbst, line 583: A few comments:
- `maximum(abs, ... )` is better (no allocation).
- Hmm. This mechanics may be a little brittle. I'd be much more defensive here, e.g. make sure each `i_irr` only matches a single kpoint, e.g. by using `only` in combination with `findall` or so. Similarly I'd put an assertion that the weights add up to the total electron count or so (see how it's done in other places, the correct number depends on the computational setup).

Also too long a line :smile:.

src/PlaneWaveBasis.jl
```diff
@@ -264,8 +269,25 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Real, fft_size::Tuple{Int, Int, I
...
+            push!(kweights_global, kweights_global[idx])
+            push!(kcoords_global, kcoords_global[idx])
+        end
+        @warn("Attempting to parallelize $n_kpt k-points over $n_procs MPI ranks. DFTK does " *
+              "not support processes empty of k-point. Some k-points were duplicated over the " *
+              "extra ranks with scaled weights.")
+    end
+    n_kpt = length(kcoords_global)
 
     if n_procs > n_kpt
```
mfherbst, line 292: Can't this stuff now be removed ?

test/PlaneWaveBasis.jl
mfherbst: Also in a follow-up we should really invert from `:dont_test_mpi` to `:mpi` on all mpi-enabled tests.

src/PlaneWaveBasis.jl
```diff
@@ -89,10 +89,15 @@ struct PlaneWaveBasis{T,
     ## These fields are not actually used in computation, but can be used to reconstruct a basis
     # Monkhorst-Pack grid used to generate the k-points, or nothing for custom k-points
     kgrid::AbstractKgrid
-    # full list of (non spin doubled) k-point coordinates in the irreducible BZ
+    # full list of (non spin doubled) k-point coordinates in the irreducible BZ (duplicates possible)
```
mfherbst, line 92: Can you also add a comment that these are better accessed through the functions you introduce below
to take care of the duplication.

src/PlaneWaveBasis.jl
```diff
@@ -543,6 +565,29 @@ function weighted_ksum(basis::PlaneWaveBasis, array)
     mpi_sum(res, basis.comm_kpts)
 end
 
+"""
+Utilities to get information about the irreducible k-point mesh (in case of duplication)
+Useful for I/O, where k-point information should not be duplicated
+"""
+function irreducible_kcoords(basis::PlaneWaveBasis)
```
mfherbst, line 572: Add the `_global` postfix

src/PlaneWaveBasis.jl
```diff
@@ -543,6 +565,29 @@ function weighted_ksum(basis::PlaneWaveBasis, array)
...
+    # Assume that duplicated k-points are appended at the end of the kcoords array
+    basis.kcoords_global[1:basis.n_irreducible_kpoints]
+end
+
+function irreducible_kweights(basis::PlaneWaveBasis{T}) where {T}
```
mfherbst, line 577: Add the `_global` postfix

test/PlaneWaveBasis.jl
abussy: I don't mind doing this as part of this PR

src/PlaneWaveBasis.jl
abussy: You mean the warning message? I think it has its use, as over-parallelizing leads to waste of resources.

test/PlaneWaveBasis.jl
mfherbst: This was not a comment specifically to this test, but more a general remark.

src/PlaneWaveBasis.jl
mfherbst: No I mean everything after line 292 in the new line count.

abussy: Oh yes, for sure!

abussy:
Implemented most concerns raised by @mfherbst during his review. I'd like to raise two points for discussion:

1) I gave some thoughts to changing the `:dont_test_mpi` test tag to `:mpi`. I think that, by default, DFTK should be MPI compatible. Any method of function not implemented for MPI is then an exception, and using the `:dont_test_mpi` test flag makes that clear. I am afraid that changing to `:mpi` instead would  lead to a lower impact of the tag.

2) When reworking the `irreducible_kweights_global()` function in `PlaneWaveBasis.jl`, I realised that in some cases, duplicated k-points already exist (even in non-parallel runs), and I refrained from using functions such as `findall()`. In particular, when calculating band structures with the `compute_bands()` function, the K-point path might go through special points multiple times, on purpose. As a result, I kept the logic of looking for replicated k-points in `basis.kcoords_global[basis.n_irreducible_kpoints+1:end]` explicitly. I might possibly change nomenclature though, as the original set of K-points is not necessarily the irreducible one.

mfherbst:
> I gave some thoughts to changing the :dont_test_mpi test tag to :mpi. I think that, by default, DFTK should be MPI compatible. Any method of function not implemented for MPI is then an exception, and using the :dont_test_mpi test flag makes that clear. I am afraid that changing to :mpi instead would lead to a lower impact of the tag.

Hmm. Yes, but I think that's ok, because it's fine if only the core is tested with MPI and not the rest. But let's think about this some more and leave as is for now.


> , I realised that in some cases, duplicated k-points already exist (even in non-parallel runs),
> ...
> In particular, when calculating band structures with the compute_bands() function,

True, good point.

But still there is a problem. The coordinates can be the same if the spin is different. I'm not sure you handle this proplerly right now ?

src/PlaneWaveBasis.jl
```diff
@@ -422,6 +431,34 @@ function weighted_ksum(basis::PlaneWaveBasis, array)
...
+Utilities to get information about the irreducible k-point mesh (in case of duplication)
+Useful for I/O, where k-point information should not be duplicated
+"""
+function irreducible_kcoords_global(basis::PlaneWaveBasis)
+    # Assume that duplicated k-points are appended at the end of the kcoords array
+    basis.kcoords_global[1:basis.n_irreducible_kpoints]
+end
+
+function irreducible_kweights_global(basis::PlaneWaveBasis{T}) where {T}
+    @inline function same_kpoint(i_irr, i_dupl)
```
mfherbst, line 444: You need the `@inline` ? Is the compiler not smart enough to do this ?
Also anyway not performance crtical ...

src/PlaneWaveBasis.jl
```diff
@@ -422,6 +431,34 @@ function weighted_ksum(basis::PlaneWaveBasis, array)
...
+        maximum(abs, basis.kcoords_global[i_dupl]-basis.kcoords_global[i_irr]) < eps(T)
+    end
+
+    # Assume that duplicated k-points are appended at the end of the kcoords array
+    irr_kweights = basis.kweights_global[1:basis.n_irreducible_kpoints]
+    for i_dupl = basis.n_irreducible_kpoints+1:length(basis.kweights_global)
+        for i_irr = 1:basis.n_irreducible_kpoints
+            if same_kpoint(i_irr, i_dupl)
```
mfherbst, line 452: Still some of my defensiveness comments apply, don't they ? Like check the weight before and after is the same and add a `continue` ?

src/PlaneWaveBasis.jl
abussy: Probably overkill, I agree

src/PlaneWaveBasis.jl
abussy: I don't think a `continue` fits here: a given k-point could be duplicated multiple times, and all the weights must be added. Extreme example: a single k-point calculation on any number of MPI ranks.

mfherbst: just reverse the order of the loops ?

PR #999, opened 2024-08-30
MPI workarounds for aarch64

abussy:
This PR contains MPI workarounds for `aarch64`. As a result, DFTK can now be run on ARM architecture with full MPI support. These workarounds are meant to be temporary, until there is an upstream fix (whenever that happens, issue https://github.com/JuliaParallel/MPI.jl/issues/404 was already opened 4 years ago). 

The procedure was the following: `:mpi` tagged tests were run on an ARM machine, until a crash would occur. A set of specific MPI reduction methods were then written for the custom type involved. Some other custom types might pop up in the future. Adding a new method for them should be trivial.

I know this is rather clunky, and it inflates the code size and complexity. However, I think it is crucial to fully support DFTK on ARM architecture, especially now that many modern HPC clusters run on it (e.g. Alps at CSCS).


mfherbst:
Thanks. Two remarks:

- We used to have a workarounds folder. I suggest to put the code there.
- Did you report these fixes upstream or is this not easy to do ?

abussy:
I moved the fixes to the `workarounds` directory for better consistency.

I did not explicitly report these fixes upstream, but my mention of the original issue in my initial message ensures that this PR appears there for reference.

src/workarounds/aarch64_mpi.jl
```diff
@@ -0,0 +1,90 @@
...
+end
+
+function mpi_max(bool::Bool, comm::MPI.Comm)
+    int = Int(bool)
+    Bool(mpi_max(int, comm))
+end
+
+# Vec3{T} must be cast to Vector{T} before MPI reduction
+function mpi_sum!(arr::Vector{Vec3{T}}, comm::MPI.Comm) where{T}
+    n = length(arr)
```
mfherbst, line 22: Can this not be solved by a reinterpret(reshape, ... )? Then there is no copy

src/workarounds/aarch64_mpi.jl
```diff
@@ -0,0 +1,90 @@
...
+	    arr[i] = Vec3{T}(new_arr[3(i-1)+1:3(i-1)+3])
+    end
+    arr
+end
+
+# ForwardDiff.Dual{T, U, V} and arrays of it must be cast to Vector{U} as well
+# utility function to cast a Dual type to an array containing a value and the partial diffs
+function dual_array(dual::ForwardDiff.Dual{T, U, V}) where{T, U, V}
+    dual_array = [ForwardDiff.value(dual)]
+    append!(dual_array, collect(ForwardDiff.partials(dual)))
```
mfherbst, line 38: I vaguely recall there are again ways to do thiy without copy. Check the forwarddiff rules for array operations (multiplication neefs to be done equally to all partials)

src/workarounds/aarch64_mpi.jl
```diff
@@ -0,0 +1,90 @@
...
+    dual_array
+end
+
+# utility function that casts back an array to a Dual type, based on a template Dual
+function new_dual(dual_array, template)
```
mfherbst, line 43: Again Forwarddifc may have sth here

src/workarounds/aarch64_mpi.jl
```diff
@@ -0,0 +1,90 @@
...
+    lengths = []
+    for i in 1:length(dual)
+        tmp = dual_array(dual[i])
+        append!(array, tmp)
+        append!(lengths, length(tmp))
+    end
+    mpi_sum!(array, comm)
+    offset = 0
+    for i in 1:length(dual)
+        dual[i] = new_dual(array[offset+1:offset+lengths[i]], dual[i])
```
mfherbst, line 86: Index operation does a copy by default. Use views (@views)

src/workarounds/aarch64_mpi.jl
abussy: I think this one is tricky to get without copies, because `SVector` are immutable. So even if I use something like `new_aarr = reshape(reinterpret(T, arr), 3, :)`, I still could not call `mpi_sum!()`. And assuming I call `mpi_sum()` instead, I also end up with a copy.

abussy:
In the last commit, the amount of copies and/or temporary arrays in the MPI workarounds for `aarch64` was minimized. There are unfortunately a few things that cannot be avoided, e.g. the creation of a standard type arrays for MPI communication, and some copies when accessing elements of a `Vector{ForwardDiff.Dual}`.

mfherbst:
@abussy There seems to be some update in the issue on the MPI.jl side as well. Maybe that helps.

abussy:
Indeed, since MPI.jl v0.20.22, it is possible to manually register custom operations for MPI on `aarch64`. This PR now uses this feature, and is much cleaner as a result. Thanks for pointing that out!

mfherbst:
Awesome ! Thanks @abussy lets merge this.

PR #1005, opened 2024-10-08
Make `solve_ΩplusK` MPI-compatible

Technici4n:
Only the `dot` and `norm` in the CG solver had to be made MPI-compatible.

Some quick numbers from benchmarking an operation dominated by `solve_ΩplusK` on my computer:

| MPI processes | running time |
|---|---|
|1|97.7s|
|2|49.4s|
|4|27.5s|

### DIFF
--- a/src/response/cg.jl
+++ b/src/response/cg.jl
@@ -7,10 +7,18 @@ end
 """
 Implementation of the conjugate gradient method which allows for preconditioning
 and projection operations along iterations.
+
+To use this function with MPI:
+- Make `A`, `precon` and `proj` MPI-aware.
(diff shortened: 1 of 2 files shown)


antoine-levitt:
Nice! Can you move mpi_dot and mpi_norm to mpi.jl? (potentially we should audit the code base to see if there are other places that could use it as well) Also not worth documenting the mpi in that file (things are already complicated enough and the mpi is pretty simple here).

Technici4n:
Done.

> potentially we should audit the code base to see if there are other places that could use it as well

I just did a quick search for `mpi_sum` and did not find other places :)

mfherbst (approved):

PR #1006, opened 2024-10-09
Optimize `apply_K`

Technici4n:
A few low-hanging fruits to optimize `solve_ΩplusK`, which is currently dominated by `apply_K`.

### 1. Optimize allocations in `compute_δρ`
A couple of `.` were missing to fuse array operations together.

Before:
```julia
julia> @time δρ = DFTK.compute_δρ(basis_ref, ψ_upscaled, δψ.δψ, occupation);
  0.422897 seconds (5.26 k allocations: 538.114 MiB, 4.43% gc time)
```
After:
```julia
julia> @time δρ = DFTK.compute_δρ(basis_ref, ψ_upscaled, δψ.δψ, occupation);
  0.328420 seconds (4.61 k allocations: 23.110 MiB)
```

### 2. Optimize allocations in `apply_K`
For each k-point, `apply_K` was allocating three arrays per band: one to hold the IFFT, one to hold `δV .* ψnk_real`, and one to hold the FFT.

The change is to now preallocate a single array to hold the result of the IFFT, and of the multiplication by `δV`.

Before:
```julia
julia> @time x = DFTK.apply_K(basis_ref, δψ.δψ, ψ_upscaled, ρ, occupation);
  0.870054 seconds (6.35 k allocations: 689.433 MiB, 5.23% gc time)
```

After:
```julia
julia> @time x = DFTK.apply_K(basis_ref, δψ.δψ, ψ_upscaled, ρ, occupation);
  0.644315 seconds (5.49 k allocations: 65.835 MiB)
```

### 3. Reuse IFFT of `ψ`
~~Every time $\pmb{K}$ is applied, the IFFT of `ψ` was computed **twice**: once in `compute_δρ` and once in `apply_K` itself. It does not change over the entire `solve_ΩplusK` call, so we can compute it once and reuse it.~~

Reverted because it increases memory requirements by an order of magnitude.

### DIFF
--- a/src/densities.jl
+++ b/src/densities.jl
@@ -56,7 +56,7 @@ end
 # Variation in density corresponding to a variation in the orbitals and occupations.
 @views @timing function compute_δρ(basis::PlaneWaveBasis{T}, ψ, δψ, occupation,
                                    δoccupation=zero.(occupation);
-                                   occupation_threshold=zero(T), q=zero(Vec3{T})) where {T}
+                                   occupation_threshold=zero(T), q=zero(Vec3{T}), ψ_real=nothing) where {T}
     Tψ = promote_type(T, eltype(ψ[1]))
     # δρ is expected to be real when computations are not phonon-related.
     Tδρ = iszero(q) ? real(Tψ) : Tψ
@@ -84,7 +84,11 @@ end
         (ik, n) = kn
 
         kpt = basis.kpoints[ik]
-        ifft!(storage.ψnk_real, basis, kpt, ψ[ik][:, n])
+        if isnothing(ψ_real)
+            ifft!(storage.ψnk_real, basis, kpt, ψ[ik][:, n])
+        else
+            storage.ψnk_real .= ψ_real[ik][:, :, :, n]
+        end
         # … and then we compute the real Fourier transform in the adequate basis.
         ifft!(storage.δψnk_real, basis, δψ_plus_k[ik].kpt, δψ_plus_k[ik].ψk[:, n])
 
--- a/src/densities.jl
+++ b/src/densities.jl
@@ -56,7 +56,7 @@ end
 # Variation in density corresponding to a variation in the orbitals and occupations.
 @views @timing function compute_δρ(basis::PlaneWaveBasis{T}, ψ, δψ, occupation,
                                    δoccupation=zero.(occupation);
-                                   occupation_threshold=zero(T), q=zero(Vec3{T}), ψ_real=nothing) where {T}
+                                   occupation_threshold=zero(T), q=zero(Vec3{T})) where {T}
     Tψ = promote_type(T, eltype(ψ[1]))
     # δρ is expected to be real when computations are not phonon-related.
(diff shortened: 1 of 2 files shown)


antoine-levitt:
Nice! I assume this is for a very small test case? For bigger ones everything to do with K should be pretty much negligible.

antoine-levitt:
The point is that omega scales like the number of bands and kpoints, and K doesn't, so for larger unit cells and/or kpoint sampling you'd get rid of K. But still good to have anyway!

src/densities.jl
```diff
@@ -84,12 +84,16 @@ end
         (ik, n) = kn
 
         kpt = basis.kpoints[ik]
-        ifft!(storage.ψnk_real, basis, kpt, ψ[ik][:, n])
+        if isnothing(ψ_real)
+            ifft!(storage.ψnk_real, basis, kpt, ψ[ik][:, n])
+        else
+            storage.ψnk_real .= ψ_real[ik][:, :, :, n]
```
antoine-levitt, line 90: you can just avoid this copy by havnig a psink_real which points either to storage.psink_real or to psi_real

src/response/hessian.jl
```diff
@@ -99,13 +104,20 @@ Return δψ where (Ω+K) δψ = rhs
         x .= pack(Pδψ)
     end
 
+    # Precompute IFFT of ψ
+    ψ_real = map(enumerate(basis.kpoints)) do (ik, kpt)
+        stack(map(eachcol(ψ[ik])) do ψnk
+            ifft(basis, kpt, ψnk)
+        end)
+    end
+
```
antoine-levitt, line 113: the reason we didn't do it this way was because it kills the RAM for larger systems. So I'd rather not do this unless there's a very good reason to do it...

antoine-levitt: If the time saving is significant we might want to add it as an option, but I'm not convinced it's worth the complexity. Your testcase is silicon unit cell? What kpoint sampling? What are the timings for a reasonably converged calculation (eg kpoint grid 8x8x8)?

Technici4n: Here is a Silicon profile with a kgrid of 8x8x8 and refined Ecut of 120. The starting point was obtained from Ecut=90. Only entries with >= 1000 samples are included. Only the first two commits of this PR have been applied, **not** the IFFT caching.

https://gist.github.com/Technici4n/faf102bf9aa1f368ed859557a4f97254

You can see that `apply_K` (`100902` samples) is roughly 2x slower than `apply_Ω` (`53891` samples), and that roughly half of `apply_K` is the IFFTs of `ψ` (`25305` samples + `20430` samples).

In this example however, the memory usage of `ψ` is roughly 0.5 GB, and the memory usage of its IFFT is roughly 13 GB. So we should probably not go for this optimization. Overall, it can only save 20-25 % of the total running time.

antoine-levitt: OK, can you remove the last optimization and add a comment detailing the tradeoff, and then we merge this?

PR #1010, opened 2024-10-22
Use latest pre-release version instead of nightly for CI against upcoming versions

Technici4n:
It should be a lot more stable. This is setup to test against the latest version, pre-releases included. Right now this checks against 1.11.1, but I expect it to test against 1.12 pre-releases once they get released.

Thanks to @gdalle for the idea.


mfherbst:
Funny how this increases coverage by 1% ...

PR #1013, opened 2024-11-01
ForwardDiff for response derivatives with PspUpf pseudopotentials

niklasschmitz:
Currently, one cannot use UPF pseudopotentials in ForwardDiff response computations of non-Hellman-Feynman derivatives. The dual version of `self_consistent_field` raises a MethodError on `construct_value(::PspUpf)` being not implemented.

This PR includes a simple fix by adding this method as an identity, under the restriction that the PspUpf contains only AbstractFloat types, in particular it does not (yet) allow ForwardDiff.Dual inside the PspUpf.

I added a test case for force derivatives using PspUpf, which is however failing with a large numerical error in comparison to finite difference. I'm opening this as a draft PR for now, until we track down the numerical problem.


test/forwarddiff.jl
```diff
@@ -54,6 +54,17 @@
...
+    @testset "Using PspUpf" begin
+        Si = ElementPsp(:Si; psp=load_psp(silicon.psp_upf))
+        atoms = [Si, Si]
+
+        derivative_ε1_fd = let ε1 = 1e-5
+            (compute_force(ε1, 0.0; atoms) - compute_force(-ε1, 0.0; atoms)) / 2ε1
+        end
+        derivative_ε1 = ForwardDiff.derivative(ε1 -> compute_force(ε1, 0.0; atoms), 0.0)
+        @test norm(derivative_ε1 - derivative_ε1_fd) < 1e-4
+    end
```
niklasschmitz, line 67: This test case is currently failing with a large error of ~0.07 against finite differences:
```
Test Failed at ~/git/DFTK.jl/test/forwarddiff.jl:66

  Expression: norm(derivative_ε1 - derivative_ε1_fd) < 0.0001

   Evaluated: 0.07632223701033393 < 0.0001
```

niklasschmitz:
I also temporarily include a more in-depth example in `sandbox/upf_forwarddiff.jl`. Interestingly, the suggested fix works correctly with the HGH-PspUpf, and is also consistent with HGH-PspHgh results. The numerical inconsistency appears only after switching to the Pseudodojo-PspUpf potential...

test/forwarddiff.jl
niklasschmitz: This now works, thanks to #1014! (Error down to ~1e-7)

mfherbst:
The MacOS failure is unrelated. Should we just merge this @niklasschmitz ?

niklasschmitz:
I added an extra test case for `compute_kernel` when using NLCC, to test it independently from ForwardDiff. It indeed was failing on master, and is fixed by this PR. Ready from my side

test/kernel.jl
```diff
@@ -97,4 +100,11 @@
     test_kernel(:collinear, Xc([:gga_c_pbe]), test_compute=false)
     test_kernel(:collinear, Xc([:gga_x_pbe]), test_compute=false)
     test_kernel(:collinear, Xc([:gga_x_pbe, :gga_c_pbe]), test_compute=false)
+
+    @testset "Non-linear core correction (NLCC)" begin
+        psp = TestCases.silicon.psp_upf  # PseudoDojo v0.4.1 Si includes NLCC
```
mfherbst, line 105: @test psp.has_nlcc (or whatever it's called) just to make sure this does not break randomly in the future ?

Also do we need all three of these tests ? Should not one be sufficient ?

niklasschmitz: Makes sense, I added a test for `has_core_density(psp)` here, and just kept just LDA.

PR #1014, opened 2024-11-01
Add core density to XC kernel

mfherbst:
The core density is currently missing in the kernel computation.

This is me looking at this for 5 minutes and  doing no testing whatsoever.

Does this help @niklasschmitz  for #1013 ?


mfherbst:
Sounds good. Closing this here, #1013 can take over.

PR #920, opened 2023-12-07
Misc changes

epolack:
You can cherry pick from those commits (I can revert anything left unchecked)
- [x] [Some fixes for the documentation](https://github.com/JuliaMolSim/DFTK.jl/commit/44782bea97146eda6b05392c36c5dc77216f5965)
- [x] [Some more consistency in the use of named tuples](https://github.com/JuliaMolSim/DFTK.jl/commit/5969b9f33c085a385a10d0ca4406b4162b46f47b)
- [x] @antoine-levitt [likes idiom `for i = 1:n`](https://github.com/JuliaMolSim/DFTK.jl/commit/22268a5f7a2656bb816f33c4cdb15661be9fda7a) 
  ```console
  find ./ -type f -name '*.jl' -not -path "*/FourierTransforms.jl/*" -exec sed -E -i 's/in ([0-9]+):/= \1:/g' {} \;
  ```
- [x] [Fix as docs could not be built locally anymore](https://github.com/JuliaMolSim/DFTK.jl/commit/33613e66742d01515e5f2320c142afb983ba8fa9)
- [x] [Using artifacts when building doc](https://github.com/JuliaMolSim/DFTK.jl/commit/bae4edba9b799128ba619bfac43607051ea7dd1f): I find it really useful to debug the doc as it is always built anyway (e.g., [here](https://github.com/epolack/DFTK.jl/actions/runs/7119002924))

PR #1016, opened 2024-11-12
Refinement of density and forces

Technici4n:
Continuation of #881.

### Concept
The new functions `refine_scfres`, `refine_forces` and `refine_density` should be the blackbox version of what is described in the _Practical error bounds for the forces_ example.

### Details
- Usage looks like this:
```julia
# Here basis_ref has a larger Ecut than scfres.basis
refinement = refine_scfres(scfres, basis_ref; ΩpK_tol);
refined_density = refine_density(refinement);
refined_forces = refine_forces(refinement; scfres);
```
- The example was updated to use these new functions, but keep the comparisons against other estimates like in the paper.

TODO:
- [x] In-code TODOs
- [x] Add test?
- [x] Find better name for `ΩpK_tol`.
- [x] ~~Occupation handling looks messy.~~ Virtual orbitals are not supported, so it looks like `occupation` is only there to be passed to functions that require it.


src/postprocess/refine.jl
```diff
@@ -0,0 +1,144 @@
...
+Result of calling the [`refine_scfres`](@ref) function.
+- `basis`: Refinement basis, larger than the basis used to
+           run a first [`self_consistent_field`](@ref) computation.
+- `ψ`, `ρ`, `occupation`: Quantities from the scfres, transferred to the refinement basis
+                          and with virtual orbitals removed.
+- `δψ`, `δρ`: First order corrections to the wavefunctions and density.
+              The sign is such that the refined quantities are ψ - δψ and ρ - δρ.
+"""
+struct RefinementResult
+    basis :: PlaneWaveBasis
```
mfherbst, line 19: No spaces around `::`

src/postprocess/refine.jl
```diff
@@ -0,0 +1,144 @@
...
+through [`refine_density`](@ref) and [`refine_forces`](@ref).
+"""
+function refine_scfres(scfres, basis_ref::PlaneWaveBasis{T}; ΩpK_tol,
+                       occ_threshold=default_occupation_threshold(T), kwargs...) where {T}
+    basis = scfres.basis
+
+    @assert basis.model.lattice == basis_ref.model.lattice
+    @assert length(basis.kpoints) == length(basis_ref.kpoints)
+    @assert all(basis.kpoints[ik].coordinate == basis_ref.kpoints[ik].coordinate
+                for ik in 1:length(basis.kpoints))
```
mfherbst, line 41: This needs an assertion about occupation / temperature somewhere (maybe taken care of elsewhere), just to make sure that for metals this fails.

src/postprocess/refine.jl
```diff
@@ -0,0 +1,144 @@
...
+        proj_tangent_kpt!(δφnk, φk)
+        op(x) = apply_M(φk, Pk, x, n)
+        function f_ldiv!(x, y)
+            x .= proj_tangent_kpt(y, φk)
+            x ./= (Pk.mean_kin[n] .+ Pk.kin)
+            proj_tangent_kpt!(x, φk)
+        end
+        J = LinearMap{eltype(φk)}(op, size(δφnk, 1))
+        δφnk = IterativeSolvers.cg(J, δφnk, Pl=FunctionPreconditioner(f_ldiv!),
+                  verbose=false, reltol=0, abstol=1e-15)
```
mfherbst, line 77: Here are also some tolerances we may want to tweak, respectively some output / convergence statistics we may want to access (let's discuss ...)

src/postprocess/refine.jl
```diff
@@ -0,0 +1,144 @@
...
+        proj_tangent_kpt!(δφnk, φk)
+    end
+
+    function apply_metric(φ, P, δφ, A::Function)
```
mfherbst, line 81: This name is a little unexpected. Because of the `A` argument, which differs depending on the purpose of this function.

src/postprocess/refine.jl
```diff
@@ -0,0 +1,144 @@
...
+        Hψk = Hk * ψk
+        ψk'Hψk
+    end # Rayleigh coefficients
+    ΩpKe2 = apply_Ω(e2, ψr, hamr, Λ) .+ apply_K(basis_ref, e2, ψr, ρr, occ)
+    ΩpKe2 = transfer_blochwave(ΩpKe2, basis_ref, basis)
+
+    rhs = resLF - ΩpKe2
+
+    # Invert Ω+K on the small space
+    e1 = solve_ΩplusK(basis, ψ, rhs, occ; tol=ΩpK_tol).δψ
```
mfherbst, line 111: Again this throws away potentially useful details (convergence, history, verbosity). Also one may want to switch algorithms etc with kwargs.

src/postprocess/refine.jl
```diff
@@ -0,0 +1,144 @@
...
+    # the density.
+    δρ = compute_δρ(basis_ref, ψr, schur_residual, occ)
+
+    RefinementResult(basis_ref, ψr, ρr, occ, schur_residual, δρ)
+end
+
+"""
+Retrieve the refined density from a [`RefinementResult`](@ref).
+"""
+function refine_density(refinement::RefinementResult)
```
mfherbst, line 126: I'm wondering if it may be better to target `-δρ` (i.e. the correction) instead of `ρ - δρ` as the quantity to return. Pros is that you will need `δρ` as an error estimate and that doing the final manipulation of adding the correction and `ρ` is trivial and obvious.

mfherbst: But forcing to compute a quantity and its correction at the same time has advantages too as one can use tricks such as `DiffResults` to avoid duplicate computations, see below.

src/postprocess/refine.jl
```diff
@@ -0,0 +1,144 @@
...
+Either the unrefined forces must be provided, or an `scfres` to compute them.
+"""
+function refine_forces(refinement::RefinementResult, forces::AbstractArray)
+    dF = ForwardDiff.derivative(ε -> compute_forces(refinement.basis,
+                                                    refinement.ψ .+ ε.*refinement.δψ,
+                                                    refinement.occupation;
+                                                    ρ=refinement.ρ + ε.*refinement.δρ), 0)
+    forces - dF
+end
+function refine_forces(refinement::RefinementResult, scfres::NamedTuple)
```
mfherbst, line 142: It somehow feels weird that both `refinement` and `scfres` are passed only for `scfres` to essentially be passed through to the force computation.

src/postprocess/refine.jl
```diff
@@ -0,0 +1,144 @@
...
+    ψ, occ = select_occupied_orbitals(basis, scfres.ψ, scfres.occupation; threshold=occ_threshold)
+    ψr = transfer_blochwave(ψ, basis, basis_ref)
+    ρr = transfer_density(scfres.ρ, basis, basis_ref)
+    _, hamr = energy_hamiltonian(basis_ref, ψr, occ; ρ=ρr)
+
+    # Compute the residual R(P)
+    # This recomputes ρr internally, but it's only a minor inefficiency.
+    res = compute_projected_gradient(basis_ref, ψr, occ)
+
+    # Compute M^{-1} R(P), with M^{-1} defined in [CDKL2021]
```
mfherbst, line 53: Explain inline.

src/postprocess/refine.jl
Technici4n: I refactored this a bit. You may not like the triple nesting of functions but at least it's self contained with a comment on the top explaining what this "inversion of M" even means. 😄

src/occupation.jl
```diff
@@ -175,3 +175,13 @@ function compute_fermi_level(basis::PlaneWaveBasis{T}, eigenvalues, ::FermiZeroT
 
     εF
 end
+
+"""
+Check that all orbitals are fully occupied.
+"""
+function check_full_occupation(basis::PlaneWaveBasis, occupation)
+    filled_occ = filled_occupation(basis.model)
+    for occ_k = occupation
```
mfherbst, line 184: `occk` and `for occk in `

src/postprocess/refine.jl
```diff
@@ -0,0 +1,159 @@
...
+
+    M applied to the i-th band is defined by
+      M_i = P^⟂ T_i^{1/2} P^⟂ T_i^{1/2} P^⟂,
+    with the diagonal operator
+      T_i = kinetic energy + mean kinetic energy of i-th band.
+
+    To invert M_i, use P^⟂ T_i^{-1} P^⟂ as the preconditioner.
+
+    See [CDKL2021] for more details.
+    """
```
mfherbst, line 72: Put as comments. Also for longer comments like this we prefer to have them *outside* the actual function to ensure that the actual code is condensed (one gets a better overview what's going on in this way).

src/postprocess/refine.jl
```diff
@@ -0,0 +1,159 @@
...
+        end
+
+        map(enumerate(res)) do (ik, resk)
+            P = PreconditionerTPA(basis_ref, basis_ref.kpoints[ik])
+            ψk = ψ[ik]
+            precondprep!(P, ψk)
+            δψk = similar(resk)
+            for n = 1:size(resk, 2)
+                # Apply M_i^{-1} to each band.
+                δψk[:, n] = apply_inv_M(ψk, P, resk[:, n], n)
```
mfherbst, line 105: Needs an `@view`

src/postprocess/refine.jl
```diff
@@ -0,0 +1,159 @@
...
+            proj_tangent_kpt!(resk, ψk)
+            op(x) = apply_M(ψk, Pk, x, n)
+            function f_ldiv!(x, y)
+                x .= proj_tangent_kpt(y, ψk)
+                x ./= (Pk.mean_kin[n] .+ Pk.kin)
+                proj_tangent_kpt!(x, ψk)
+            end
+            J = LinearMap{eltype(ψk)}(op, size(resk, 1))
+            δψk = IterativeSolvers.cg(J, resk, Pl=FunctionPreconditioner(f_ldiv!),
+                      verbose=false, reltol=0, abstol=1e-15)
```
mfherbst, line 94: Formatting

src/postprocess/refine.jl
```diff
@@ -0,0 +1,159 @@
...
+            end
+            δψk
+        end
+    end
+
+    # Compute M^{-1}_22 R_2(P) as an approximation of (Ω+K)^{-1}_22 R_2(P).
+    e2 = apply_inv_M(ψr, resHF)
+
+    # Apply Ω+K to M^{-1}_22 R_2(P)
+    Λ = map(enumerate(ψr)) do (ik, ψk)
```
mfherbst, line 115: Should be able to `map(hamr, \psir) do (hamk, \psik)`

src/occupation.jl
Technici4n: I think `for x = ` is akward but it was changed in many places in https://github.com/JuliaMolSim/DFTK.jl/pull/920 so I kept using it.

src/postprocess/refine.jl
Technici4n: I think that this should be kept as close as possible to where `M` is actually defined because there are many things going on. An alternative is to move `M` out but I don't think it's relevant in other places.

src/postprocess/refine.jl
Technici4n: Since in my tests the CG would always converge in 1 iteration, I changed `abstol` to `100*eps(T)` with max 20 iterations and a warning if that is not enough to converge. We might have to increase the constants once we start testing the method on more complex systems, TBD.

src/postprocess/refine.jl
Technici4n: Moved to a comment at the top of the file.

src/postprocess/refine.jl
```diff
@@ -0,0 +1,183 @@
...
+# then perform an approximate first order correction in the larger space.
+#
+# To compute the first order correction ("refinement") δP in the larger space,
+# one needs to invert (Ω+K)δP = -R(P) where (Ω+K) is the Hessian matrix and
+# R(P) = [P, [P, H(P)]] is the residual.
+# See newton.jl for an explanation of these quantities.
+#
+# The basis elements of the smaller basis set are called low frequency (_1 suffix).
+# The additional basis elements are all high frequency (_2 suffix).
+# We approximate (Ω+K)_22 by M_22 which is cheap to compute and invert, and (Ω+K)_21 by 0.
```
mfherbst, line 15: Here I would either rationalise in one sentence or refer again to the paper.

src/postprocess/refine.jl
```diff
@@ -0,0 +1,183 @@
...
+#
+# Solving the system then amounts to computing
+#   δP_2 = M^{-1}_22 R_2(P)
+#   δP_1 = (Ω+K)^{-1}_11 (R_1(P) - (Ω+K)_12 δP_2)
+# The inversion of (Ω+K)_11 is roughly as expensive as the SCF on the low frequency space.
+#
+# [CDKL2021]:
+#     E. Cancès, G. Dusson, G. Kemlin, and A. Levitt
+#     *Practical error bounds for properties in plane-wave electronic structure
+#     calculations* Preprint, 2021. [arXiv](https://arxiv.org/abs/2111.01470)
```
mfherbst, line 32: No longer a preprint, can you look up the reference.

src/postprocess/refine.jl
```diff
@@ -0,0 +1,183 @@
...
+
+"""
+Invert the metric operator M.
+"""
+function invert_refinement_metric(basis::PlaneWaveBasis{T}, ψ, res) where {T}
+    # Apply the M_i operator with i=n.
```
mfherbst, line 38: So its M_n ?

src/postprocess/refine.jl
```diff
@@ -0,0 +1,183 @@
...
+# R(P) = [P, [P, H(P)]] is the residual.
+# See newton.jl for an explanation of these quantities.
+#
+# The basis elements of the smaller basis set are called low frequency (_1 suffix).
+# The additional basis elements are all high frequency (_2 suffix).
+# We approximate (Ω+K)_22 by M_22 which is cheap to compute and invert, and (Ω+K)_21 by 0.
+#
+# The metric operator M is as follows:
+# M applied to the i-th band is defined by
+#   M_i = P^⟂ T_i^{1/2} P^⟂ T_i^{1/2} P^⟂,
```
mfherbst, line 19: Use n and m for band indices.

src/postprocess/refine.jl
```diff
@@ -0,0 +1,183 @@
...
+#     E. Cancès, G. Dusson, G. Kemlin, and A. Levitt
+#     *Practical error bounds for properties in plane-wave electronic structure
+#     calculations* Preprint, 2021. [arXiv](https://arxiv.org/abs/2111.01470)
+
+"""
+Invert the metric operator M.
+"""
+function invert_refinement_metric(basis::PlaneWaveBasis{T}, ψ, res) where {T}
+    # Apply the M_i operator with i=n.
+    function apply_M(ψk, Pk, δψnk, n)
```
mfherbst, line 39: Add !

src/postprocess/refine.jl
```diff
@@ -0,0 +1,183 @@
...
+        proj_tangent_kpt!(δψnk, ψk)
+        δψnk = sqrt.(Pk.mean_kin[n] .+ Pk.kin) .* δψnk
+        proj_tangent_kpt!(δψnk, ψk)
+        δψnk = sqrt.(Pk.mean_kin[n] .+ Pk.kin) .* δψnk
+        proj_tangent_kpt!(δψnk, ψk)
+    end
+
+    # Apply the M_i^{-1} operator with i=n.
```
mfherbst, line 47: See above

src/postprocess/refine.jl
```diff
@@ -0,0 +1,183 @@
...
+        op(x) = apply_M(ψk, Pk, x, n)
+        function f_ldiv!(x, y)
+            x .= proj_tangent_kpt(y, ψk)
+            x ./= (Pk.mean_kin[n] .+ Pk.kin)
+            proj_tangent_kpt!(x, ψk)
+        end
+        J = LinearMap{eltype(ψk)}(op, size(resk, 1))
+        # This CG seems to converge very quickly in practice (often 1 iteration).
+        # Set a very low tolerance and warn if the iteration count is too high.
+        (δψk, history) = IterativeSolvers.cg(J, resk;
```
mfherbst, line 59: Maybe not a bad idea to also use our internal CG.

src/postprocess/refine.jl
```diff
@@ -0,0 +1,183 @@
...
+        proj_tangent_kpt!(resk, ψk)
+        op(x) = apply_M(ψk, Pk, x, n)
+        function f_ldiv!(x, y)
+            x .= proj_tangent_kpt(y, ψk)
+            x ./= (Pk.mean_kin[n] .+ Pk.kin)
+            proj_tangent_kpt!(x, ψk)
+        end
+        J = LinearMap{eltype(ψk)}(op, size(resk, 1))
+        # This CG seems to converge very quickly in practice (often 1 iteration).
+        # Set a very low tolerance and warn if the iteration count is too high.
```
mfherbst, line 58: Obvious from the code

src/postprocess/refine.jl
```diff
@@ -0,0 +1,183 @@
...
+end
+
+"""
+Result of calling the [`refine_scfres`](@ref) function.
+- `basis`: Refinement basis, larger than the basis used to
+           run a first [`self_consistent_field`](@ref) computation.
+- `ψ`, `ρ`, `occupation`: Quantities from the scfres, transferred to the refinement basis
+                          and with virtual orbitals removed.
+- `δψ`, `δρ`: First order corrections to the wavefunctions and density.
+              The sign is such that the refined quantities are ψ - δψ and ρ - δρ.
```
mfherbst, line 90: Change as discussed

src/postprocess/refine.jl
mfherbst: ok for now. with the explanatory text it should be fine.

src/postprocess/refine.jl
Technici4n: I changed it to return both `F` and `\deltaF` for forces. (And `refine_density` is gone since `\rho` and `\delta\rho` are both in the refinement info already).

DiffResults support will be for another time.

src/postprocess/refine.jl
Technici4n: Done

src/postprocess/refine.jl
Technici4n: Changed the reference in the code there. I just realized that I still need to change it in the example too!

src/postprocess/refine.jl
Technici4n: I changed the sign.

src/postprocess/refine.jl
Technici4n: Done

src/postprocess/refine.jl
Technici4n: Added a new `ΩpK_res` field to RefinementResult which contains whatever the CG returned.

mfherbst (commented):
Getting there

src/postprocess/refine.jl
```diff
@@ -0,0 +1,231 @@
...
+Default callback function for `refine_scfres`,
+which prints a convergence table.
+"""
+struct RefinementDefaultCallback
+    prev_time::Ref{UInt64}
+end
+function RefinementDefaultCallback()
+    RefinementDefaultCallback(Ref(zero(UInt64)))
+end
+function (cb::RefinementDefaultCallback)(info)
```
Technici4n, line 113: There should be a check for the MPI master process.

src/postprocess/refine.jl
```diff
@@ -198,6 +198,13 @@ function refine_scfres(scfres, basis_ref::PlaneWaveBasis{T};
     RefinementResult(basis_ref, ψr, ρr, occ, schur_residual, δρ, ΩpK_res)
 end
 
+"""
+Refine energies using a [`RefinementResult`](@ref).
+"""
+function refine_energies(refinement::RefinementResult)
+    energy(refinement.basis, refinement.ψ + refinement.δψ, refinement.occupation; ρ=refinement.ρ + refinement.δρ).energies
```
mfherbst, line 202: Line too long.

src/postprocess/refine.jl
```diff
@@ -0,0 +1,216 @@
...
+    δψ
+    δρ
+    ΩpK_res
+end
+
+"""
+Default callback function for `refine_scfres`,
+which prints a convergence table.
+"""
+struct RefinementDefaultCallback
```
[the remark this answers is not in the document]
Technici4n: Moved to `ResponseCallback` as discussed.

src/DFTK.jl
```diff
@@ -7,6 +7,7 @@ using Printf
 using TimerOutputs
 using Unitful
 using UnitfulAtomic
+using DiffResults
```
mfherbst, line 10: If you only use it in one file, I prefer to put the `using` in that file only to keep it local to the usage.

src/postprocess/refine.jl
```diff
@@ -0,0 +1,203 @@
+# Refinement of some quantities of interest (density, forces) following the
+# strategy described in [CDKL2022].
+# We use a different sign convention for corrections δX,
```
mfherbst, line 3: Note that here we use a different sign convention in contrast to the paper ...

mfherbst (commented):
Nits

src/postprocess/refine.jl
mfherbst: Still needed ?

PR #1019, opened 2024-11-18
Remove dependency on DftFunctionals.change_parameters fallback behavior for functionals without parameters

niklasschmitz:
This anticipates the change in https://github.com/JuliaMolSim/DftFunctionals.jl/pull/17 to keep working for generic XC functionals without exposed parameters. The current DFTK implementation of DispatchFunctional depends on change_parameters falling back to the identity.


mfherbst:
@niklasschmitz You should bump the minimal version on DftFunctionals.jl once https://github.com/JuliaRegistries/General/pull/119685 is in.

Issue #514, opened 2021-08-15
Fix AD for small occupations


antoine-levitt:
See https://github.com/JuliaMolSim/DFTK.jl/pull/513#event-5159480887. Currently only done for Fermi-Dirac, should be done for other smearing types. Probably simplest to not define rules but have the function return 0 at overflows.

niklasschmitz:
Also for Fermi-Dirac, the current fix is only sufficient for first order, as second order derivatives for large x again run into Inf/Inf division NaNs:
```julia
using DFTK, ForwardDiff
f0(x) = Smearing.occupation(Smearing.FermiDirac(), x)
f1(x) = ForwardDiff.derivative(f0, x)
f2(x) = ForwardDiff.derivative(f1, x)

x = 400.0
f0(x), f1(x), f2(x)  # (1.9151695967140057e-174, -0.0, NaN)
```

antoine-levitt:
We should just hardcode 0/1 when overflow/underflow, which would remove the need for the ugly workaround in forwarddiff_rules?

mfherbst:
Sounds fair and make physical sense.

niklasschmitz:
That would simplify things for sure. As mentioned to me by @azadoks, that's also what QE does here specifying the `maxarg` to the exponential https://github.com/QEF/q-e/blob/6e083f7950e599a1a6788f0638610f88c1613670/Modules/wgauss.f90#L49. I can set up a small PR

antoine-levitt:
We have `    max_exp_arg = -log(eps(T)) + 5  # add some wiggle room` in ewald, which sounds sensible.

antoine-levitt:
Cute trick, let's go with that! (maybe `y = exp(x); y/(1+y)`? is the compiler sufficiently clever to recognize `exp` is pure?)

niklasschmitz:
[...]

PR #1022, opened 2024-11-20
Fix Fermi-Dirac occupation overflow and remove custom ForwardDiff rule (#514)

niklasschmitz:
From the discussion at (#514).


antoine-levitt:
Very nice!

mfherbst:
Thanks !

PR #1025, opened 2024-11-25
Speed-up symmetrize_ρ

abussy:
In some cases, the `symmetrize_ρ` function becomes a major bottleneck of the calculation (high symmetry, large cutoffs and/or on the GPU). Profiling reveals that ~ half of the time is spent in `index_G_vectors`, and ~ half on `cis2pi`.

This PR drastically reduces the time spent in `index_G_vectors`, by pre-computing the FFT start, end and lengths, which currently takes place every time the function is called (number of symmetries x number of G vectors). As a result, calls to `symmetrize_ρ` are about 40-50% faster.

Note that the time spent in `cis2pi` could also be minimized by pre-computing and storing a 2D array containing the factors for each pair of G vector and symmetry. While this is done in SIRIUS (and probably other PW codes), this creates a lot of noise and complication in the code base.

### DIFF
--- a/src/common/cis2pi.jl
+++ b/src/common/cis2pi.jl
@@ -1,14 +1,13 @@
 """Function to compute exp(2π i x)"""
 cis2pi(x) = cispi(2x)
-function cis2pi(x::AbstractFloat)
+function cis2pi(x::T) where {T <: AbstractFloat}
     # Special case when 2x is an integer, as exp(n*π*i) = +- 1. Saves expensive
     # exponential evaluations when called repeatedly
     if isinteger(2x)
-        isodd(2x) ? result = -Complex{typeof(x)}(1, 0) : result = Complex{typeof(x)}(1, 0)
+        return isodd(2x) ? -one(complex(T)) : one(complex(T))
     else
-        result = cispi(2x)
+        return cispi(2x)
     end
-    result
 end
 
 """Function to compute sin(2π x)"""
--- a/src/common/cis2pi.jl
+++ b/src/common/cis2pi.jl
@@ -1,5 +1,15 @@
 """Function to compute exp(2π i x)"""
 cis2pi(x) = cispi(2x)
+function cis2pi(x::AbstractFloat)
+    # Special case when 2x is an integer, as exp(n*π*i) = +- 1. Saves expensive
+    # exponential evaluations when called repeatedly
+    if isinteger(2x)
+        isodd(2x) ? result = -Complex{typeof(x)}(1, 0) : result = Complex{typeof(x)}(1, 0)
+    else
+        result = cispi(2x)
+    end
+    result
+end
 
 """Function to compute sin(2π x)"""
 sin2pi(x) = sinpi(2x)
--- a/src/PlaneWaveBasis.jl
+++ b/src/PlaneWaveBasis.jl
@@ -400,6 +400,21 @@ Returns nothing if outside the range of valid wave vectors.
     end
 end
 
+@inline function index_G_vectors(fft_start::Tuple, fft_stop:: Tuple, fft_lengths::Tuple,
+                                 G::AbstractVector{<:Integer})
+
+    # FFTs store wavevectors as [0 1 2 3 -2 -1] (example for N=5)
+    function G_to_index(length, G)
+        G >= 0 && return 1 + G
+        return 1 + length + G
+    end
+    if all(fft_start .<= G .<= fft_stop)
+        CartesianIndex(Tuple(G_to_index.(fft_lengths, G)))
+    else
+        nothing  # Outside range of valid indices
+    end
+end
+
 function index_G_vectors(basis::PlaneWaveBasis, G::AbstractVector{<:Integer})
     index_G_vectors(basis.fft_size, G)
 end
--- a/src/PlaneWaveBasis.jl
+++ b/src/PlaneWaveBasis.jl
@@ -400,22 +400,8 @@ Returns nothing if outside the range of valid wave vectors.
     end
 end
 
-@inline function index_G_vectors(fft_start::Tuple, fft_stop:: Tuple, fft_lengths::Tuple,
-                                 G::AbstractVector{<:Integer})
-
-    # FFTs store wavevectors as [0 1 2 3 -2 -1] (example for N=5)
-    function G_to_index(length, G)
-        G >= 0 && return 1 + G
-        return 1 + length + G
-    end
-    if all(fft_start .<= G .<= fft_stop)
-        CartesianIndex(Tuple(G_to_index.(fft_lengths, G)))
-    else
-        nothing  # Outside range of valid indices
-    end
-end
-
-function index_G_vectors(basis::PlaneWaveBasis, G::AbstractVector{<:Integer})
+# @inline is necessary here for the inner function to be inlined as well
+@inline function index_G_vectors(basis::PlaneWaveBasis, G::AbstractVector{<:Integer})
     index_G_vectors(basis.fft_size, G)
 end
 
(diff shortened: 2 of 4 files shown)


antoine-levitt:
I see this:
```
@inline function index_G_vectors(fft_size::Tuple, G::AbstractVector{<:Integer})
    # the inline declaration encourages the compiler to hoist these (G-independent) precomputations
```
(which git blame says I added in 2022) Does that not work anymore? (or did that never work at all?) Maybe it needs an extra `@inline` for the `index_G_vectors(basis::PlaneWaveBasis, G::AbstractVector{<:Integer})` method?

antoine-levitt:
Regarding the cis2pi, most of these are probably very simple (0, 1/2, 1). It doesn't look like those are special-cased right now (see sincospi in julia/base/special/trig.jl). Maybe that would help?

antoine-levitt:
> Note that the time spent in cis2pi could also be minimized by pre-computing and storing a 2D array containing the factors for each pair of G vector and symmetry. While this is done in SIRIUS (and probably other PW codes), this creates a lot of noise and complication in the code base.

That looks like a lot of RAM, but presumably the symmetries are only relevant for small systems where this is acceptable. It's also non-differentiable (G and tau can't move) so we won't have issues here. So it might be worth it to precompute...

abussy:
> (which git blame says I added in 2022) Does that not work anymore? (or did that never work at all?) Maybe it needs an extra @inline for the index_G_vectors(basis::PlaneWaveBasis, G::AbstractVector{<:Integer}) method?

It turns out that adding an extra `@inline` in front of `index_G_vectors(basis::PlaneWaveBasis, G::AbstractVector{<:Integer})` yields the same performance. I will go that route instead, as it is more compact and cleaner.

>Regarding the cis2pi, most of these are probably very simple (integers or half-integers). It doesn't look like those are special-cased right now (see sincospi in julia/base/special/trig.jl). Maybe that would help?

I'll have a look, thanks for pointing that out!

antoine-levitt:
> It turns out that adding an extra @inline in front of index_G_vectors(basis::PlaneWaveBasis, G::AbstractVector{<:Integer}) yields the same performance. I will go that route instead, as it is more compact and cleaner.

Nice. Can you add a regression test (eg with `@allocations`) to ensure that this doesn't get lost?

abussy:
Indeed, most (if not all) calls to `cis2pi()` are made with integers or half integers. As a result, I implemented a very simple special case directly in the loop.

With the proper use of `@inline` for `index_G_vectors` and the `cis2pi` special cases, this PR accelerates the density symmetrization by a factor of 2-3x.

I have not added an additional regression test yet.

antoine-levitt:
Why not just put the special case in cis2pi? It looks generally useful.

mfherbst:
Yeah, this should be in base julia.

antoine-levitt:
In general I don't know, it's probably bad for vectorization, but I don't think we have much vectorization of cis2pi stuff anyway.

abussy:
> Why not just put the special case in cis2pi? It looks generally useful.

Reason was: one needs to be a lot more careful about the return value/type, as simply returning a +/-1.0 scalar does not cut it (e.g. when cis2pi takes an array as argument, or anything else). There might be ways to make it general, but I am not fluent enough in Julia yet.

antoine-levitt:
Just do one(typeof(x)) or something like this. Cispi does not take arrays, the dot vectorizes for you

abussy:
Added special cases in the `cis2pi()` directly.

> Nice. Can you add a regression test (eg with @allocations) to ensure that this doesn't get lost?

I'm not sure how to go about this. Since `@inline` is only a compiler hint, there is a possibility that the code behaves differently depending on the machine.

antoine-levitt:
It shouldn't: `@inline` speaks to the julia compiler, which is not machine dependent (at least not much), unlike LLVM. It can behave differently depending on the julia version, which is precisely what we want to catch: if julia becomes unable to optimize this, we complain on the julia github, or modify our code.

src/common/cis2pi.jl
```diff
@@ -1,5 +1,15 @@
 """Function to compute exp(2π i x)"""
 cis2pi(x) = cispi(2x)
+function cis2pi(x::AbstractFloat)
```
mfherbst, line 3: (x::T) where {T <: AbstractFloat}

src/common/cis2pi.jl
```diff
@@ -1,5 +1,15 @@
...
+    # Special case when 2x is an integer, as exp(n*π*i) = +- 1. Saves expensive
+    # exponential evaluations when called repeatedly
+    if isinteger(2x)
+        isodd(2x) ? result = -Complex{typeof(x)}(1, 0) : result = Complex{typeof(x)}(1, 0)
```
mfherbst, line 7: Can just use `T` here
also use `one(complex(T))`
I would write
```suggestion
        return isodd(2x) ? -one(complex(T)) : one(complex(T))
```

src/common/cis2pi.jl
```diff
@@ -1,5 +1,15 @@
...
+    else
+        result = cispi(2x)
```
mfherbst, line 9:
```suggestion
        return cispi(2x)
```

src/common/cis2pi.jl
```diff
@@ -1,5 +1,15 @@
...
+    end
+    result
```
mfherbst, line 11: delete this.

src/common/cis2pi.jl
```diff
@@ -1,5 +1,15 @@
 """Function to compute exp(2π i x)"""
 cis2pi(x) = cispi(2x)
```
mfherbst, line 2: Is this needed at all. I mean could this not just be folded into the definition below ?

antoine-levitt: Yes, the special casing is incorrect for ForwardDiff (it will silently ignore the dual part)

abussy:
Applied @mfherbst suggestions. As for the test, I am still unsure. I tried something like:
```
@testset "Si" begin
        silicon = TestCases.silicon
        Si = ElementPsp(silicon.atnum, load_psp("hgh/lda/si-q4"))
        atoms = [Si, Si]
        model = model_DFT(silicon.lattice, atoms, silicon.positions;
                          functionals=[:lda_x, :lda_c_vwn])
        Ecut = 32
        kgrid = [1, 1, 1]
        basis = PlaneWaveBasis(model; Ecut, kgrid, symmetries_respect_rgrid=true) 
        ρ0    = guess_density(basis)
        alloc = @allocations ρ_symm = DFTK.symmetrize_ρ(basis, ρ0)
        @show alloc
    end
 ```

and the inlined version of `index_G_vectors` actually involves more allocations (with `@allocated` and `@allocations`). The difference is also <1%. So there is no obvious way of distinguishing the 2 versions this way. I am also afraid that testing for a hard number of allocations might be very brittle. One could probably look into the llvm ir generated code, but that's a pretty big effort to automatize.

mfherbst:
Not sure how `@allocations ` works, but maybe you need a $ to interpolate things in.

Otherwise test timings instead ?

abussy:
Testing the timings is pretty brittle too. Depending on the machine it is ran on, or whether that machine is busy or not, they may vary significantly. I am short of idea right now.

mfherbst:
True ... maybe we merge without a test.

@antoine-levitt any thoughts ?

antoine-levitt:
Can't we make a simple case that reproduces the issue and test the allocations on that? Otherwise sure let's merge

abussy:
I added a test that emulates the `accumulate_over_symmetries()` function, called once with DFTK's `index_G_vectors()`, and once with a locally defined function without the `@inline` directive. There, the difference in allocation is clear and reproducible.

antoine-levitt:
Nice, thanks, let's merge!

PR #1027, opened 2024-11-25
Inexact Krylov for linear response

bonans:


src/response/chi0.jl
```diff
@@ -42,16 +42,16 @@ function compute_χ0(ham; temperature=ham.basis.model.temperature)
     basis = ham.basis
     filled_occ = filled_occupation(basis.model)
     smearing = basis.model.smearing
-    n_spin   = basis.model.n_spin_components
```
mfherbst, line 45: You have a lot of these whitespace commits in your pr. Can you make sure you keep the original formatting where appropriate ?

mfherbst (commented):
Thanks for the pr. Before I take a detailed look, can you remove the noise from the for formatting ? Thanks :smile:

src/response/chi0.jl
bonans: Fixed

mfherbst: Thanks :+1:

PR #1028, opened 2024-11-26
Change DFTK thread count to a runtime setting

### LINKED ISSUE #1012: Setting a >1 threading preference will trigger a recompilation and make it fail
Reproduction steps:
- Install DFTK, open julia and run `using DFTK`. It precompiles fine. Close julia.
- Now open `julia -t2`. Run: `using DFTK; DFTK.setup_threading(; n_DFTK=2)`. Close julia.
- Now open `julia -t2` again and run `using DFTK`. It has to precompile again, and will fail with the following error:
```
ERROR: The following 1 direct dependency failed to precompile:

DFTK

Failed to precompile DFTK [acf6eb54-70d9-11e9-0013-234b7a5f5337] to "/home/bruno/.julia/compiled/v1.11/DFTK/jl_onZ99l".
ERROR: LoadError: You set your preference for DFTK threads using set_DFTK_threads!, but only ran julia with 1 threads.
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] get_DFTK_threads()
    @ DFTK ~/.julia/packages/DFTK/xNYFf/src/common/threading.jl:50
  [3] parallel_loop_over_range(fun::DFTK.var"#287#289"{Matrix{ComplexF64}, DFTK.DftHamiltonianBlock, Matrix{ComplexF64}, Array{Float64, 3}, Bool}, range::UnitRange{Int64}, storages::Vector{@NamedTuple{ψ_reals::Array{ComplexF64, 3}}})
    @ DFTK ~/.julia/packages/DFTK/xNYFf/src/common/threading.jl:72
  [4] macro expansion
    @ ~/.julia/packages/DFTK/xNYFf/src/terms/Hamiltonian.jl:144 [inlined]
  [5] mul!(Hψ::Matrix{ComplexF64}, H::DFTK.DftHamiltonianBlock, ψ::Matrix{ComplexF64})
    @ DFTK ~/.julia/packages/TimerOutputs/NRdsv/src/TimerOutput.jl:237
  [6] macro expansion
    @ ~/.julia/packages/DFTK/xNYFf/src/eigen/lobpcg_hyper_impl.jl:350 [inlined]
  [7] LOBPCG(A::DFTK.DftHamiltonianBlock, X::Matrix{ComplexF64}, B::LinearAlgebra.UniformScaling{Bool}, precon::DFTK.PreconditionerTPA{Float64}, tol::Float64, maxiter::Int64; miniter::Int64, ortho_tol::Float64, n_conv_check::Int64, display_progress::Bool)
    @ DFTK ~/.julia/packages/TimerOutputs/NRdsv/src/TimerOutput.jl:237
[other stuff below]
```

Looks like the precompilation workload only runs with one thread.

Technici4n:
Fixes #1012 by changing the number of DFTK threads to a runtime setting.

Do we care about potential races when calling `set_DFTK_threads!()` in parallel? I suppose not. I don't want to make it an atomic.


src/common/threading.jl
```diff
@@ -6,51 +6,45 @@ Setup the number of threads used by DFTK's own threading (`n_DFTK`), by BLAS (`n
...
-Convenience function to disable all threading in DFTK and assert that Julia threading
-is off as well.
+Convenience function to disable all threading in DFTK.
 """
 function disable_threading()
     setup_threading(;n_fft=1, n_blas=1, n_DFTK=1)
 end
 
+# TODO: is a single write to an Int64 atomic?
+DFTK_threads = 0
```
mfherbst, line 29: Should be const for sure.

But also this is global state. Do we really want this?

mfherbst: Apparently according to Valentin the way to go is https://github.com/vchuravy/ScopedValues.jl, which will become part of base in 1.11. Can you look into this ?

Technici4n: Funny how that's actually inspired from a Java API. 😄

I'm not sure that
```jl
with(DFTK_THREADS => 1) do
    # All computations here
end
```
is very user friendly however, especially since it's inconsistent with BLAS and FFTW threads.

Technici4n: Making it `const` is a bad idea, the point is that it's actually mutable (and will be mutated by `set_DFTK_threads!`:
```
WARNING: redefinition of constant DFTK.DFTK_threads. This may fail, cause incorrect answers, or produce other errors.
```

mfherbst: const does not mean that it's immutable if you make it a `Ref`. It most importantly it ensures the type does not change.

also `with(DFTK_THREADS => 1) do` can be wrapped.

Technici4n: I talked to Valentin again, and the ideal thing to do would be to split work based on statically known block sizes (wherever we call `parallel_loop_over_range` at least), and let the Julia scheduler handle this.

mfherbst:
Ok, let's do it like this for now. We should really revisit our HPC setup and scheduling (like have some proper structured way of configuring this).

PR #1031, opened 2024-12-04
Enable force calculations on GPU runs

abussy:
Unfortunately, when attempting to calculate forces with DFTK's GPU architecture, the program simply crashes. This PR fixes this issue in the simplest way possible: the offending operations are executed on the CPU, and the data is moved back and forth. While this may not be optimal, it makes the application usable. Future profiling/benchmarking will reveal whether optimization is necessary.

This was thoroughly tested, including in combination with `GeometryOptimization`. Added a couple of regression tests for forces on GPUs as well.


mfherbst:
Not awesome, but ok, let's do it like this for now.

PR #1036, opened 2024-12-18
Minor optimization of nonlocal term

abussy:
When using the built-in DFTK.timer as a rough profiler, it appears that the `nonlocal` term involves a lot more memory allocations than any other term in the Hamiltonian. This is particularly obvious for the H*psi product in LOBPCG and the forces.

This PR minimizes these memory allocations. A minor speed-up of ~10% can be observed on the H*psi product, and a larger speedup of up to 50% on the forces (on a handful of test cases).

### DIFF
--- a/src/terms/nonlocal.jl
+++ b/src/terms/nonlocal.jl
@@ -75,17 +75,26 @@ end
             ψk = to_cpu(ψ[ik])
             occupationk = to_cpu(occupation[ik])
             form_factors = build_projector_form_factors(element.psp, G_plus_k_cart)
+
+            # Pre-allocation of large arrays
+            δHψk = similar(ψk)
+            P = similar(form_factors)
+            dPdR = similar(form_factors)
+            structure_factors = similar(form_factors, length(G_plus_k))
+            twoπp = similar(form_factors, length(G_plus_k))
+            
             for idx in group
                 r = model.positions[idx]
-                structure_factors = [cis2pi(-dot(p, r)) for p in G_plus_k]
-                P = structure_factors .* form_factors ./ sqrt(unit_cell_volume)
+                map!(p -> cis2pi(-dot(p, r)), structure_factors, G_plus_k)
+                P .= structure_factors .* form_factors ./ sqrt(unit_cell_volume)
 
                 forces[idx] += map(1:3) do α
-                    dPdR = [-2T(π)*im*p[α] for p in G_plus_k] .* P
-                    δHψk = P * (C * (dPdR' * ψk))
-                    -sum(occupationk[iband] * basis.kweights[ik] *
-                             2real(dot(ψk[:, iband], δHψk[:, iband]))
-                         for iband=1:size(ψk, 2))
+                    map!(p -> -2π*im*p[α], twoπp, G_plus_k)
+                    dPdR .= twoπp .* P
+                    mul!(δHψk, P, C * (dPdR' * ψk))
+                    @views -sum(occupationk[iband] * basis.kweights[ik] *
+                                2real(dot(ψk[:, iband], δHψk[:, iband]))
+                            for iband=1:size(ψk, 2))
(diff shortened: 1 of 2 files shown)


src/terms/operators.jl
```diff
@@ -108,7 +108,7 @@ struct NonlocalOperator{T <: Real, PT, DT} <: RealFourierOperator
     D::DT
 end
 function apply!(Hψ, op::NonlocalOperator, ψ)
-    Hψ.fourier .+= op.P * (op.D * (op.P' * ψ.fourier))
+    mul!(Hψ.fourier, op.P, (op.D * (op.P' * ψ.fourier)), 1, 1)
```
mfherbst, line 111: Is using `1` (integer) type-stable here ?

mfherbst:
Thanks, looks good. I just have one nit comment.

src/terms/operators.jl
mfherbst: My point is that some form of `one(T)` would be better ...

src/terms/nonlocal.jl
```diff
@@ -75,17 +75,26 @@ end
...
+            # Pre-allocation of large arrays
+            δHψk = similar(ψk)
+            P = similar(form_factors)
+            dPdR = similar(form_factors)
+            structure_factors = similar(form_factors, length(G_plus_k))
+            twoπp = similar(form_factors, length(G_plus_k))
+            
             for idx in group
                 r = model.positions[idx]
-                structure_factors = [cis2pi(-dot(p, r)) for p in G_plus_k]
```
mfherbst, line 80: I bet we have other places, where we are missing this opportunity, too. Could you have a quick look @abussy ?

src/terms/operators.jl
antoine-levitt: I don't see why not: int is always weaker than floating point types, e.g. 1 + Float16(1.) isa Float16

antoine-levitt:
The Hpsi is a no-brainer, thanks for spotting it. The pre-allocation thing I'm not such a big fan of... Have you tried these fancy things that people mentioned at the workshop that temporarily take over the allocator?

mfherbst:
> Have you tried these fancy things that people mentioned at the workshop that temporarily take over the allocator?

Worth trying ... but some of these packages seem quite brittle at this moment. I'm not sure to what extend it's a good idea to commit into them unless we also fancy getting potentially involved with fixing the bugs they may have.

src/terms/operators.jl
abussy: I didn't check type stability. I've seen the same pattern elsewhere in the code though (e.g. https://github.com/JuliaMolSim/DFTK.jl/blob/15b4d41d37204299646ad551d8c6dbd9a98ec85e/src/eigen/lobpcg_hyper_impl.jl#L262C9-L262C31). So if that is not safe, we probably need more care elsewhere too.

abussy:
> The Hpsi is a no-brainer, thanks for spotting it. The pre-allocation thing I'm not such a big fan of... Have you tried these fancy things that people mentioned at the workshop that temporarily take over the allocator?

I'm not sure I was at that talk, or at least I don't remember it being mentioned. Do you have some pointer?

> Also worth noting that there seems to be recent developments in julia for making pre-allocation less necessary, so maybe this will just go away at some point

That would be great. But for now, and especially in the case of the forces, there is a noticeable impact on performance. Also, I would argue that this paves the way towards a GPU implementation of the forces. Tim Besard warned me against excessive allocations on the GPU.

abussy:
This last commit enables calculation of nonlocal forces on the GPU. I add this here since I mentioned it earlier, and the code modification is minimal. In terms of performance, there is a factor ~10 between GPU and CPU. With this change, calculating forces on the GPU becomes reasonable (nonlocal forces was the major bottleneck before).

PR #1037, opened 2024-12-18
Minor optimization of the LOBPCG solver

abussy:
This PR was also motivated by monitoring memory allocations in the built-in DFTK.timer, and the memory allocated by `ortho! X vs Y` in particular.

This orthogonalization function has a loop, where potentially large matrix allocations take place at each step. Since these matrices keep the same dimensions, it makes more sense to have a single allocation followed by in-place multiplications.

Additionally, it was noted that quite some time is spent in `rayleigh_ritz`, in a matrix-matrix multiplication known to result in an Hermitian product. Since matrices are in a non-standard `LazyHcat` format, no automatic optimization can take place. This PR adds a new function `mul_hermi` for `LazyHcat` matrices, which only computes the upper block diagonal, and populates the rest of the function by adding its adjoint. This results in savings of the order of 30%.

### DIFF
--- a/src/eigen/lobpcg_hyper_impl.jl
+++ b/src/eigen/lobpcg_hyper_impl.jl
@@ -106,25 +106,21 @@ Base.:*(Aadj::Adjoint{T,<:LazyHcat}, B::AbstractMatrix) where {T} = Aadj * LazyH
     rows = size(A, 2)
     cols = size(B, 2)
     ret = similar(B.blocks[1], rows, cols)
-    fill!(ret, zero(T))
 
-    orow = 0  # row offset
-    for (ia, blA) in enumerate(A.blocks)
-        ocol = 0  # column offset
-        for (ib, blB) in enumerate(B.blocks)
-            ib > ia && continue
-            fac = one(T)
-            if ia == ib fac = T(0.5) end
-            ret[orow .+ (1:size(blA, 2)), ocol .+ (1:size(blB, 2))] .= fac .* (blA' * blB)
-            ocol += size(blB, 2)
+    ocol = 0  # column offset
+    for (ib, blB) in enumerate(B.blocks)
+        orow = 0  # row offset
+        for (ia, blA) in enumerate(A.blocks)
+            ib < ia && continue
+            ret[orow .+ (1:size(blA, 2)), ocol .+ (1:size(blB, 2))] .= blA' * blB
+            orow += size(blA, 2)
         end
-        orow += size(blA, 2)
+        ocol += size(blB, 2)
     end
-    # populate the lower diagonal with conjugate
-    ret + adjoint(ret)
+    Hermitian(ret)
 end
 
-mul_hermi(Aadj::AbstractArray{T}, B::AbstractArray{T}) where {T} = Aadj * B
+mul_hermi(Aadj::AbstractArray{T}, B::AbstractArray{T}) where {T} = Hermitian(Aadj * B)
 
 @views function *(Ablock::LazyHcat, B::AbstractMatrix)
     res = Ablock.blocks[1] * B[1:size(Ablock.blocks[1], 2), :]  # First multiplication
@@ -138,20 +134,13 @@ end
 
 @views function LinearAlgebra.mul!(res::AbstractMatrix, Ablock::LazyHcat,
                                    B::AbstractVecOrMat, α::Number, β::Number)
-    mul!(res, Ablock.blocks[1], B[1:size(Ablock.blocks[1], 2), :], α, β) # First multiplication
-    offset = size(Ablock.blocks[1], 2)
-    for block in Ablock.blocks[2:end]
-        mul!(res, block, B[offset .+ (1:size(block, 2)), :], α, 1)
-        offset += size(block, 2)
-    end
-    res
+    mul!(res, Ablock*B, I, α, β)
 end
 
 # Perform a Rayleigh-Ritz for the N first eigenvectors.
 @timing function rayleigh_ritz(X, AX, N)
     XAX = mul_hermi(X', AX)
-    @assert !any(isnan, XAX)
-    rayleigh_ritz(Hermitian(XAX), N)
+    rayleigh_ritz(XAX, N)
 end
 @views function rayleigh_ritz(XAX::Hermitian, N)
     # Fallback: Use whatever is the default dense eigensolver.
@@ -181,7 +170,7 @@ end
 # (which implies that X'BX is relatively well-conditioned, and
 # therefore that it is safe to cholesky it and reuse the B apply)
 function B_ortho!(X, BX)
-    O = Hermitian(mul_hermi(X'*BX))
+    O = mul_hermi(X', BX)
     U = cholesky(O).U
     @assert !any(isnan, U)
     rdiv!(X, U)
@@ -205,7 +194,7 @@ normest(M) = maximum(abs.(diag(M))) + norm(M - Diagonal(diag(M)))
     success = false
     nchol = 0
     while true
-        O = Hermitian(mul_hermi(X', X))
+        O = mul_hermi(X', X)
         try
             R = cholesky(O).U
             nchol += 1
--- a/src/eigen/lobpcg_hyper_impl.jl
+++ b/src/eigen/lobpcg_hyper_impl.jl
@@ -82,8 +82,8 @@ Base.adjoint(A::LazyHcat) = Adjoint(A)
 
 @views function Base.:*(Aadj::Adjoint{T,<:LazyHcat}, B::LazyHcat) where {T}
     A = Aadj.parent
-    rows = size(A)[2]
-    cols = size(B)[2]
+    rows = size(A, 2)
+    cols = size(B, 2)
     ret = similar(A.blocks[1], rows, cols)
 
     orow = 0  # row offset
@@ -101,24 +101,24 @@ end
 # Special case of Hermitian result: can only actively compute the block upper diagonal
 @views function mul_hermi(Aadj::Adjoint{T,<:LazyHcat}, B::LazyHcat) where {T}
     A = Aadj.parent
-    rows = size(A)[2]
-    cols = size(B)[2]
+    rows = size(A, 2)
+    cols = size(B, 2)
     ret = similar(B.blocks[1], rows, cols)
-    fill!(ret, 0)
+    fill!(ret, zero(T))
 
     orow = 0  # row offset
     for (ia, blA) in enumerate(A.blocks)
         ocol = 0  # column offset
         for (ib, blB) in enumerate(B.blocks)
-            if ib > ia continue end
-            fac = 1.0
-            if ia == ib fac = 0.5 end
-            mul!(ret[orow .+ (1:size(blA, 2)), ocol .+ (1:size(blB, 2))], adjoint(blA), blB, fac, 0)
+            ib > ia && continue
+            fac = one(T)
+            if ia == ib fac = T(0.5) end
+            mul!(ret[orow .+ (1:size(blA, 2)), ocol .+ (1:size(blB, 2))], blA', blB, fac, 0)
             ocol += size(blB, 2)
         end
         orow += size(blA, 2)
     end
-    #populate the lower diagonal with conjugate
+    # populate the lower diagonal with conjugate
     ret + adjoint(ret)
 end
 
--- a/src/eigen/lobpcg_hyper_impl.jl
+++ b/src/eigen/lobpcg_hyper_impl.jl
@@ -76,8 +76,7 @@ function Base.size(A::LazyHcat)
     (n, m)
 end
 
-Base.Array(A::LazyHcat) = stack(A.blocks)
-
+Base.Array(A::LazyHcat)   = stack(A.blocks)
 Base.adjoint(A::LazyHcat) = Adjoint(A)
 
 # Computes A*B matrix product for LazyHcat type. Special case if product is assumed to be Hermitian
@@ -106,17 +105,8 @@ Base.adjoint(A::LazyHcat) = Adjoint(A)
     end
 end
 
-function Base.:*(A::Adjoint{T,<:LazyHcat}, B::LazyHcat) where {T}
-    _mul(A, B)
-end
-
+Base.:*(A::Adjoint{T,<:LazyHcat}, B::LazyHcat) where {T}       = _mul(A, B)
 Base.:*(A::Adjoint{T,<:LazyHcat}, B::AbstractMatrix) where {T} = A * LazyHcat(B)
-
-mul_hermi(A, B) = Hermitian(A * B)
-
-function mul_hermi(A::Adjoint{T,<:LazyHcat}, B::LazyHcat) where {T}
-    _mul(A, B; hermitian=Val(true))
-end
 
 @views function *(Ablock::LazyHcat, B::AbstractMatrix)
     res = Ablock.blocks[1] * B[1:size(Ablock.blocks[1], 2), :]  # First multiplication
@@ -128,9 +118,14 @@ end
     res
 end
 
-@views function LinearAlgebra.mul!(res::AbstractMatrix, Ablock::LazyHcat,
-                                   B::AbstractVecOrMat, α::Number, β::Number)
+function LinearAlgebra.mul!(res::AbstractMatrix, Ablock::LazyHcat,
+                            B::AbstractVecOrMat, α::Number, β::Number)
     mul!(res, Ablock*B, I, α, β)
+end
+
+mul_hermi(A, B) = Hermitian(A * B)
+function mul_hermi(A::Adjoint{T,<:LazyHcat}, B::LazyHcat) where {T}
+    _mul(A, B; hermitian=Val(true))
 end
 
 # Perform a Rayleigh-Ritz for the N first eigenvectors.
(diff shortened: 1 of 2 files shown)


mfherbst:
> This results in savings of the order of 30%.

Of time to do the Reighleigh-Ritz step ?

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -90,16 +90,60 @@ Base.adjoint(A::LazyHcat) = Adjoint(A)
...
         end
         orow += size(blA, 2)
     end
     ret
 end
 
+# Special case of Hermitian result: can only actively compute the block upper diagonal
+@views function mul_hermi(Aadj::Adjoint{T,<:LazyHcat}, B::LazyHcat) where {T}
+    A = Aadj.parent
+    rows = size(A)[2]
```
mfherbst, line 104:
```suggestion
    rows = size(A, 2)
```

and similar for cols.

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -90,16 +90,60 @@ Base.adjoint(A::LazyHcat) = Adjoint(A)
...
+            mul!(ret[orow .+ (1:size(blA, 2)), ocol .+ (1:size(blB, 2))], adjoint(blA), blB)
             ocol += size(blB, 2)
         end
         orow += size(blA, 2)
     end
     ret
 end
 
+# Special case of Hermitian result: can only actively compute the block upper diagonal
+@views function mul_hermi(Aadj::Adjoint{T,<:LazyHcat}, B::LazyHcat) where {T}
```
mfherbst, line 102: I think you are assuming that `B::LazyHcat{T}`, i.e. both matrices are of the same element type.

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -90,16 +90,60 @@ Base.adjoint(A::LazyHcat) = Adjoint(A)
...
+    A = Aadj.parent
+    rows = size(A)[2]
+    cols = size(B)[2]
+    ret = similar(B.blocks[1], rows, cols)
+    fill!(ret, 0)
```
mfherbst, line 107:
```suggestion
    fill!(ret, zero(T))
```

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -90,16 +90,60 @@ Base.adjoint(A::LazyHcat) = Adjoint(A)
...
+
+    orow = 0  # row offset
+    for (ia, blA) in enumerate(A.blocks)
+        ocol = 0  # column offset
+        for (ib, blB) in enumerate(B.blocks)
+            if ib > ia continue end
```
mfherbst, line 113:
```suggestion
            ib > ia && continue
```

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -90,16 +90,60 @@ Base.adjoint(A::LazyHcat) = Adjoint(A)
...
+            fac = 1.0
```
mfherbst, line 114: one(T)

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -90,16 +90,60 @@ Base.adjoint(A::LazyHcat) = Adjoint(A)
...
+            if ia == ib fac = 0.5 end
```
mfherbst, line 115: same thing

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -90,16 +90,60 @@ Base.adjoint(A::LazyHcat) = Adjoint(A)
...
+            mul!(ret[orow .+ (1:size(blA, 2)), ocol .+ (1:size(blB, 2))], adjoint(blA), blB, fac, 0)
```
mfherbst, line 116: `0` is an integer -> `zero(T)`

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -90,16 +90,60 @@ Base.adjoint(A::LazyHcat) = Adjoint(A)
...
+            ocol += size(blB, 2)
+        end
+        orow += size(blA, 2)
+    end
+    #populate the lower diagonal with conjugate
```
mfherbst, line 121: Space after `#`

src/eigen/lobpcg_hyper_impl.jl
mfherbst: Also, I wonder if this could be handled more transparently if one had a version
`mul!(::Hermitian, ::Adjoint{T, <:LazyHcat}, ::LazyHcat)`, but maybe that's non-standard and overengineering ...

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -90,16 +90,60 @@ Base.adjoint(A::LazyHcat) = Adjoint(A)
     for blA in A.blocks
         ocol = 0  # column offset
         for blB in B.blocks
-            ret[orow .+ (1:size(blA, 2)), ocol .+ (1:size(blB, 2))] .= blA' * blB
+            mul!(ret[orow .+ (1:size(blA, 2)), ocol .+ (1:size(blB, 2))], adjoint(blA), blB)
```
antoine-levitt, line 93: you can do blA', it'll just lower to adjoint

abussy:
> In particular mul_hermi is actually quite a bit of code and I wonder if that's really worth it in terms of performance. After all these operations should be way cheaper than a Hamiltonian-vector product, no ?

Overall, for multiple test systems, I observed a ~2.5 % speedup on the overall timings. While this is not huge by itself, multiple such small optimizations over the code may add up to something significant.

mfherbst:
@abussy I think for `mul_hermi` an example showing a clear difference in timings would be good if you happen to have one handy. Just for @antoine-levitt  to get convinced whether the extra code is worth it :smile:.

abussy:
I cleaned-up this PR to make it the least disruptive possible. I only kept what I think is important, namely the `mul_hermi()` function for the product of `LazyHcat` matrices known to be Hermitian, and a redefinition of `mul!(A, B, C, alpha, beta)`. In the latter case, there was an explicit MM with the identity (clearly a waste).

For the `mul_hermi()` function, I observe up to 4% efficiency gains on the SCF. For a single additional function of 20 lines, I think it's a bargain. Making small steps like this can eventually lead to significant gains. That's my opinion at least.

As @mfherbst suggested, I am also providing an example. Generally, the speed-up will be the most visible for systems with a large number of electrons (and/or high Ecut). For the aluminium supercell below, I gained about 30s of runtime (on average, out of ~700, on my laptop).

```
using DFTK                                                                                           
using PseudoPotentialData                                                                            
setup_threading()                                                                                    
                                                                                                     
Ecut = 32.0                                                                                          
kgrid = [1, 1, 1]                                                                                    
maxiter = 10                                                                                         
tol = 1.0e-8                                                                                         
                                                                                                     
factor = 4                                                                                           
a = 3.8267                                                                                           
                      [1.0 0.0 1.0];                                                                 
                      [1.0 1.0 0.0]]                                                                 
end                                                                                                  
                                                                                                     
model = model_DFT(lattice, atoms, positions; temperature=1e-4,                                       
                                                                                                     
#actual calculations                                                                                 
DFTK.reset_timer!(DFTK.timer)                                                                        
basis  = PlaneWaveBasis(model; Ecut, kgrid)                                                          
scfres = self_consistent_field(basis; maxiter=maxiter, tol=tol);                                     
@show DFTK.timer
```

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -149,7 +181,7 @@ end
 # (which implies that X'BX is relatively well-conditioned, and
 # therefore that it is safe to cholesky it and reuse the B apply)
 function B_ortho!(X, BX)
-    O = Hermitian(X'*BX)
+    O = Hermitian(mul_hermi(X'*BX))
```
mfherbst, line 184: Oh also ... if we put the `Hermitian` wrapper inside the `mul_hermi` function, than we can even avoid computing the lower triangle at all as the wrapper will make sure it's never accessed.

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -110,14 +136,20 @@ Base.:*(Aadj::Adjoint{T,<:LazyHcat}, B::AbstractMatrix) where {T} = Aadj * LazyH
...
-    mul!(res, Ablock*B, I, α, β)
+@views function LinearAlgebra.mul!(res::AbstractMatrix, Ablock::LazyHcat,
+                                   B::AbstractVecOrMat, α::Number, β::Number)
+    mul!(res, Ablock.blocks[1], B[1:size(Ablock.blocks[1], 2), :], α, β) # First multiplication
+    offset = size(Ablock.blocks[1], 2)
+    for block in Ablock.blocks[2:end]
+        mul!(res, block, B[offset .+ (1:size(block, 2)), :], α, 1)
+        offset += size(block, 2)
+    end
+    res
```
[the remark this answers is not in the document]
abussy: You are correct, I went too fast there. I reverted it.

abussy:
Implemented the suggested changes. Requires the removal of the `@assert !any(isnan, XAX)` in `rayleigh_ritz()` before the diagonalization (because of disallowed scalar access on GPU, when XAX is already wrapped as Hermitian). I believe this is safe, as the `eigen()` function fails loudly with `LoadError: ArgumentError: matrix contains Infs or NaNs` when NaNs are present.

mfherbst:
This version is fine with me.

> Requires the removal of the `@assert !any(isnan, XAX)`

This is probably ok, but I recall we added it because of the GPU version (where apparently cuBlas did not check this back when). Maybe you could check whether this is still the case. If cuBlas now also fails loadly, I have no concerns.

@antoine-levitt Ok with this PR as it stands ?

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -100,6 +100,28 @@ end
 
 Base.:*(Aadj::Adjoint{T,<:LazyHcat}, B::AbstractMatrix) where {T} = Aadj * LazyHcat(B)
 
+# Special case of Hermitian result: can only actively compute the block upper diagonal
```
antoine-levitt, line 103: Describe what the function does instead of how it's implemented: something like "Computes A*B, assuming the result is hermitian"

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -100,6 +100,28 @@ end
...
+@views function mul_hermi(Aadj::Adjoint{T,<:LazyHcat}, B::LazyHcat) where {T}
```
antoine-levitt, line 104: Aadj -> A

antoine-levitt: (mul does A*B, not Aadj*B)

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -100,6 +100,28 @@ end
...
+            ib < ia && continue
+            ret[orow .+ (1:size(blA, 2)), ocol .+ (1:size(blB, 2))] .= blA' * blB
+            orow += size(blA, 2)
+        end
+        ocol += size(blB, 2)
+    end
+    Hermitian(ret)
+end
+
+mul_hermi(Aadj::AbstractArray{T}, B::AbstractArray{T}) where {T} = Hermitian(Aadj * B)
```
antoine-levitt, line 123: put the generic before the specialization

antoine-levitt: (also no need to type them as abstractarray, or get T just do mul_hermi(A,B) = Hermitian(A*B)

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -110,16 +132,15 @@ Base.:*(Aadj::Adjoint{T,<:LazyHcat}, B::AbstractMatrix) where {T} = Aadj * LazyH
...
-                            B::AbstractVecOrMat, α::Number, β::Number)
+@views function LinearAlgebra.mul!(res::AbstractMatrix, Ablock::LazyHcat,
+                                   B::AbstractVecOrMat, α::Number, β::Number)
     mul!(res, Ablock*B, I, α, β)
 end
 
 # Perform a Rayleigh-Ritz for the N first eigenvectors.
 @timing function rayleigh_ritz(X, AX, N)
-    XAX = X' * AX
-    @assert !any(isnan, XAX)
```
antoine-levitt, line 121: Why? This is useful to debug cases where A*x is badly implemented and runs  into NaNs.

antoine-levitt (commented):
Nice PR, some nits but good to go. Mul_hermi is a generally useful function we should ideally wrap more generally, but the problem is that this is a blind spot for BLAS (there's no BLAS routine for A*B, assuming the result is hermitian; there is for A^T A however, which is called automatically by julia) so it's fine to hardcode something in this case.

src/eigen/lobpcg_hyper_impl.jl
abussy: True. To be fair, I am copying the convention from the original general function https://github.com/JuliaMolSim/DFTK.jl/blob/dc62a84d0ba7f01c6b8307e4c0b26c38e78cdeae/src/eigen/lobpcg_hyper_impl.jl#L83 I guess I could change it there too

antoine-levitt: Yes that'd be good!

antoine-levitt:
Huh, I didn't see that this was a copy-paste. You can have a `_mul(A,B, hermitian=Val(false)` with `if hermitian ...` inside and then just dispatch the * and mul_hermi methods to that.

mfherbst:
Great idea @antoine-levitt. I would make it a kwarg, though i.e. `_mul(A,B; hermitian=Val(false) )` and then the check is `if hermitian isa Val{true}`

abussy:
I added a generic `mul_()` function that dispatches the general and Hermitian case. From my understanding, it is more efficient to branch out with `if hermitian isa Val{true}` a single time rather than in the loop. Correct me if I'm wrong, and I can reduce code duplication.

> Maybe you could check whether this is still the case. If cuBlas now also fails loadly, I have no concerns.

Unfortunately, cuBLAS let's this slide and goes ahead with NaNs. It turns out that using `@assert !any(isnan, UpperTriangular(parent(XAX)))` does not trigger GPU scalar access error, while fulfilling the safety requirements.

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -80,25 +80,51 @@ Base.Array(A::LazyHcat) = stack(A.blocks)
...
+
+    if hermitian isa Val{true}
+        # Only popuplate the upper block diagonal in Hermitian case
         ocol = 0  # column offset
-        for blB in B.blocks
-            ret[orow .+ (1:size(blA, 2)), ocol .+ (1:size(blB, 2))] .= blA' * blB
+        for (ib, blB) in enumerate(B.blocks)
+            orow = 0  # row offset
+            for (ia, blA) in enumerate(Ap.blocks)
+                ib < ia && continue
```
mfherbst, line 96: Isn't this (and the `Hermitian` wrapper) the only line that is not duplicate between the two versions of the `_mul` function ?

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -80,25 +80,51 @@ Base.Array(A::LazyHcat) = stack(A.blocks)
...
             ocol += size(blB, 2)
         end
-        orow += size(blA, 2)
+        ret
     end
-    ret
 end
 
-Base.:*(Aadj::Adjoint{T,<:LazyHcat}, B::AbstractMatrix) where {T} = Aadj * LazyHcat(B)
+@views function Base.:*(A::Adjoint{T,<:LazyHcat}, B::LazyHcat) where {T}
```
mfherbst, line 117: The `@views` here does nothing and can be removed

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -80,25 +80,51 @@ Base.Array(A::LazyHcat) = stack(A.blocks)
...
+    _mul(A, B)
+end
+
+Base.:*(A::Adjoint{T,<:LazyHcat}, B::AbstractMatrix) where {T} = A * LazyHcat(B)
+
+mul_hermi(A, B) = Hermitian(A * B)
+
+@views function mul_hermi(A::Adjoint{T,<:LazyHcat}, B::LazyHcat) where {T}
```
mfherbst, line 125: Same, `@views` can be removed

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -110,16 +136,16 @@ Base.:*(Aadj::Adjoint{T,<:LazyHcat}, B::AbstractMatrix) where {T} = Aadj * LazyH
...
     mul!(res, Ablock*B, I, α, β)
 end
 
 # Perform a Rayleigh-Ritz for the N first eigenvectors.
 @timing function rayleigh_ritz(X, AX, N)
-    XAX = X' * AX
-    @assert !any(isnan, XAX)
-    rayleigh_ritz(Hermitian(XAX), N)
+    XAX = mul_hermi(X', AX)
+    @assert !any(isnan, UpperTriangular(parent(XAX)))
```
mfherbst, line 147: Ok, ... but this sounds a bit like a missing method in base julia. I mean why should `isnan` work for `UpperTriangular` and not for `Hermitian` ...

src/eigen/lobpcg_hyper_impl.jl
abussy: Yes, that's the only difference

src/eigen/lobpcg_hyper_impl.jl
abussy: I agree, there is no good reason why this behaves this way

mfherbst (approved):
LGTM

PR #1038, opened 2024-12-18
Be more conservative when selecting initial number of bands

mfherbst:
@abussy This is a hopefully improved version being more conservative on the initial number of bands, but allowing the adaptive algorithm to still easily request more bands if needed.

Can you check this improves your performance a bit ?

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -59,6 +59,7 @@ PAGES = [
     "Tips and tricks" => [
         # Resolving convergence issues, what solver to use, improving performance or
         # reliability of calculations.
+        "tricks/achieving_convergence.md",
         "tricks/parallelization.md",
         "tricks/scf_checkpoints.jl",
         "tricks/compute_clusters.md",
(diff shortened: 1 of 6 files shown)


mfherbst:
Hmmm ... not sure. In most of my quick tests the SCF converges in less steps in the old algorithm, so not sure this is helpful overall ... as the old implementation needed less overall Hamiltonian applications.

abussy:
I see significant speed-ups (~20%) on my aluminium test case (Gaussian smearing, T=0.001 Ha). Admittedly, this is to be expected since it triggered this PR.

I wonder whether starting from a more conservative number of band and allowing it to decrease would be more robust.

mfherbst:
> I see significant speed-ups  (~20%) 

Ok, but this concerns "time to 10 SCF steps done" or "time to converging to 1e-8 in the density". Because my point was that with this PR the former is true but the latter not (i.e. in the old version the convergence rate was higher).

mfherbst:
@abussy This is actually another case where some form of automatic benchmark / test case crunching system would be really helpful. I see that this could help sometimes, but I'm not convinced it is in general the right thing to do and at least right now I lack the time to test old versus new systematically on a few representative systems to get a more educated idea.

PR #1039, opened 2024-12-21
Mention response and DFPT more prominently

mfherbst:


antoine-levitt:
Mention phonons?

mfherbst:
Good point. We should have really an example on that, actually. Do you happen to have one ?

antoine-levitt:
No, but basically it's just calling the phonons function after an scf I think

PR #1043, opened 2024-12-30
Move from IterativeSolvers to KrylovKit

mfherbst:


antoine-levitt:
Nice! CC @ClementineBarat as it might impact what she's doing. Maybe preconditioning is an issue, but for linear solve you can just solve PAx=Pb...

mfherbst:
Why should this make any difference at all ? It's changing from one GMRES to another ?

abussy:
A nice side effect of this PR is that default `LdosMixing` now works on GPUs!

PR #1044, opened 2025-01-01
Automatically use rcut from PseudoFamily when possible

Technici4n:
Happy new year! 😀


src/pseudo/load_psp.jl
```diff
@@ -1,6 +1,24 @@
+using PseudoPotentialData
```
mfherbst, line 1: Can you avoid this dependency by making it an extmodule ?

mfherbst (commented):
Thanks ! Looks good !

src/pseudo/load_psp.jl
Technici4n: Good idea, I didn't think of it.

mfherbst (commented):
One more small thing: Can you add a test where you set the rcut from the kwargs in load_psp just to ensure this works and always has preference (should be in this implementation, but just to be future-proof).

ext/DFTKPseudoPotentialDataExt.jl
```diff
@@ -0,0 +1,11 @@
+module DFTKPseudoPotentialDataExt
+
+using DFTK
+using PseudoPotentialData
+
+function DFTK.extra_psp_kwargs(family::PseudoFamily, element::Symbol)
```
mfherbst, line 6: Also a more suggestive name is probably extra_pseudometa_kwargs

PR #1049, opened 2025-01-15
Add more convergence history to scfres

ClementineBarat:
This PR adds history_εF to scfres and the possibility to save the convergence history of other data by passing a dictionary of functions that are evaluated and saved at each scf iteration.


antoine-levitt:
Hm, the API feels a bit ad hoc, and self_consistent_field already has lots of stuff in it. Can we somehow merge this with the callback API? Ie the callback would add stuff to `info`, and we'd (optionally) save the result of all callbacks or something?

Technici4n:
A bit ugly, but something like this might also work?
```jl
history_extra = []
callback = info -> push!(history_extra, info.<what needs to be saved>)`
self_consistent_field(...; callback)
```

ClementineBarat:
I added a specific callback that takes care of filling an history_extra dictionary.

src/scf/scf_callbacks.jl
```diff
@@ -104,6 +104,45 @@ function (cb::ScfDefaultCallback)(info)
     info
 end
 
+"""
+Callback that saves more convergence history
+- `use_default`: Whether or not to also use the default callback to print a convergence
```
mfherbst, line 109: This is not needed, you can chain callbacks (see https://docs.dftk.org/stable/examples/scf_callbacks/#Monitoring-self-consistent-field-calculations).

src/scf/scf_callbacks.jl
```diff
@@ -104,6 +104,45 @@ function (cb::ScfDefaultCallback)(info)
...
+        cb.default(info)
+    end
+    if info.stage == :iterate
+        for (key, fun) in cb.history_extra_functions
+            extra = fun(; info...)
+            cb.history_extra[key] = push!(cb.history_extra[key], extra)
+        end
+    end
+    if info.stage == :finalize
+        info = merge(info, (; cb.history_extra))
```
mfherbst, line 141: Why not merge the content rather than the dict ?

mfherbst (commented):
I'm ok to include the `εF` if you need it @ClementineBarat, but about the `ScfSaveHistory` struct I am not sure. What is the advantage of this struct over just writing a bunch of callbacks directly ?

Just to be clear: My problem is that you inevitably have to make some decisions when designing such a struct (e.g. you choose to call functions as `fun(; info...)`) which  work very well in the use case you have in mind, but which may not work well for others. Then others anyway have to write some bloat code to connect the SCF to their callback function.

ClementineBarat:
Yes that make sense, no need to add my custom callback to DFTK. I removed all changes except for adding the history_εF but i can do without if it is just for me.

src/scf/scf_callbacks.jl
ClementineBarat: Ok nice

mfherbst:
I think it can be useful also for occupation functions, but @antoine-levitt may have an opinion here.

@antoine-levitt If you are ok, I'd say we merge.

src/scf/self_consistent_field.jl
```diff
@@ -180,11 +180,15 @@ Overview of parameters:
...
+
+        # Update info with history
         history_Etot = vcat(info.history_Etot, energies.total)
         history_Δρ = vcat(info.history_Δρ, norm(Δρ) * sqrt(basis.dvol))
+        history_εF = vcat(info.history_εF, εF)
         n_matvec = info.n_matvec + nextstate.n_matvec
-        info_next = merge(info_next, (; energies, history_Etot, history_Δρ, n_matvec))
-
+        info_next = merge(info_next, (; energies, history_Etot, history_Δρ, history_εF,
+            n_matvec))
```
mfherbst, line 190: Here the indention is weird, I'd do
```suggestion
        info_next = merge(info_next, (; energies, history_Etot, history_Δρ, history_εF,
                                        n_matvec))
```

src/scf/self_consistent_field.jl
```diff
@@ -202,8 +206,7 @@ Overview of parameters:
 
     info_init = (; ρin=ρ, ψ=ψ, occupation=nothing, eigenvalues=nothing, εF=nothing, 
                    n_iter=0, n_matvec=0, timedout=false, converged=false,
-                   history_Etot=T[], history_Δρ=T[])
-
```
mfherbst, line 206: Here you're killing an empty line, which I think would help the structure of the code.

antoine-levitt:
scfres already has quite a lot of stuff, maybe we can have a convergence_history field and move all the convergence logging stuff to there?

mfherbst:
>  maybe we can have a convergence_history field and move all the convergence logging stuff to there?

Ok for me, sounds reasonable.

mfherbst:
Is this still needed / worked upon or can we close this ?

ClementineBarat:
Yes we can

PR #1053, opened 2025-02-04
Change default to ResponseOptions(verbose=true)

niklasschmitz:
I find myself always manually passing ResponseOptions(verbose=true), whenever I use ForwardDiff for response. Given that the response solver runs similarly long as the SCF itself (or even multiple times in the case of multiple partials), and the SCF is verbose by default, here I propose to make the response verbose by default too.


antoine-levitt:
Sure, makes sense!

PR #1055, opened 2025-02-05
Fix energy cutoff smearing example

niklasschmitz:
The energy cutoff smearing example got silently broken in one of the recent upgrades, resulting in the lattice not being scaled. This PR fixes this.

The old plot:
a703f954

I also increased the kgrid to (4,4,4) and recomputed the reference Ecut=100 results. Here is the new plot:
cutoff_smearing


antoine-levitt:
Woops that's bad, thanks for noticing!

PR #1056, opened 2025-02-06
Enable AtomicLocal term computation on the GPU

abussy:
Currently, calculations on the GPU are dominated by the `PlaneWaveBasis` instantiation and forces (especially for large systems with few k-points). One of the terms not ported to the GPU is the `AtomicLocal` term. This PR addresses that.

For efficiency reasons, the form factors are calculated and stored for unique values of |G+q| rather than for each G vector. The current format is a dictionary not fit for the GPU. In this PR, the `atomic_local_form_factor()` function was modified to return an array of the same dimension, and an auxiliary array mapping any G vector to a unique |G + q| value.

`compute_local_potential()` and `forces_local()` were modified to accommodate the GPU compatible form factors, and run efficiently on the GPU (no data movement, reduced allocations).

As a result, the timings for the `AtomicLocal` terms go from main hurdle to negligible on the GPU. CPU timings are marginally improved.

If these changes are accepted, a similar treatment can be applied to the `XC` term, which now becomes the main bottleneck.

### DIFF
--- a/src/terms/local.jl
+++ b/src/terms/local.jl
@@ -71,19 +71,32 @@ function (external::ExternalFromFourier)(basis::PlaneWaveBasis{T}) where {T}
 end
 
 # Returns the form factors at unique values of |G + q| (in Cartesian coordinates).
-# Uses a hash map for O(1) lookup.
+# Additionally, returns a mapping from any G index to a unique |G + q|
 function atomic_local_form_factors(basis::PlaneWaveBasis{T}, Gqs_cart::AbstractArray) where{T}
-    form_factors = IdDict{Tuple{Int,T},T}()  # IdDict for Dual compatibility
-    for G in Gqs_cart
+    iG2norm = zeros(Int, length(Gqs_cart))
+    norms = IdDict{T, Int}()
+    for (iG, G) in enumerate(Gqs_cart)
         p = norm(G)
+        if !haskey(norms, p)
+            norms[p] = length(norms) + 1
+        end
+        iG2norm[iG] = norms[p]
+    end
+
+    form_factors = zeros(T, length(norms), length(basis.model.atom_groups))
+    for (iG, G) in enumerate(Gqs_cart)
+        p = norm(G)
+        inorm = iG2norm[iG]
         for (igroup, group) in enumerate(basis.model.atom_groups)
-            if !haskey(form_factors, (igroup, p))
+            if form_factors[inorm, igroup] ≈ zero(T)
                 element = basis.model.atoms[first(group)]
-                form_factors[(igroup, p)] = local_potential_fourier(element, p)
+                form_factors[inorm, igroup] = local_potential_fourier(element, p)
             end
         end
     end
-    form_factors
+    form_factors = to_device(basis.architecture, form_factors)
+    iG2norm = to_device(basis.architecture, iG2norm)
+    (; form_factors, iG2norm)
 end
 
 ## Atomic local potential
@@ -104,26 +117,30 @@ function compute_local_potential(basis::PlaneWaveBasis{T}; positions=basis.model
     # and a structure factor e^{-i G·r}
     model = basis.model
     Gqs_cart = [model.recip_lattice * (G + q) for G in to_cpu(G_vectors(basis))]
-    # TODO Bring Gqs_cart on the CPU for compatibility with the pseudopotentials which
-    #      are not isbits ... might be able to solve this by restructuring the loop
 
-    form_factors = atomic_local_form_factors(basis, Gqs_cart)
+    form_factors, iG2norm = atomic_local_form_factors(basis, Gqs_cart)
+    Gqs = map(G -> G+q, G_vectors(basis))
 
-    Gqs = [G + q for G in to_cpu(G_vectors(basis))]  # TODO Again for GPU compatibility
-    pot_fourier = map(enumerate(Gqs)) do (iG, G)
-        p = norm(Gqs_cart[iG])
-        pot = sum(enumerate(model.atom_groups)) do (igroup, group)
-            structure_factor = sum(r -> cis2pi(-dot(G, r)), @view positions[group])
-            form_factors[(igroup, p)] * structure_factor
+    Tpot = promote_type(eltype(form_factors), eltype(eltype(positions)))
+    pot = to_device(basis.architecture, zeros(Complex{Tpot}, length(Gqs)))
+    pot_tmp = similar(pot)
+    indices = to_device(basis.architecture, collect(1:length(Gqs)))
+
+    for (igroup, group) in enumerate(model.atom_groups)
+        for r = positions[group]
+            ff_group = @view form_factors[:, igroup]
+            map!(iG -> cis2pi(-dot(Gqs[iG], r))*ff_group[iG2norm[iG]], pot_tmp, indices)
+            pot .+= pot_tmp./sqrt(model.unit_cell_volume)
         end
-        pot / sqrt(model.unit_cell_volume)
     end
 
+    pot_fourier = reshape(pot, basis.fft_size)
+
     if iszero(q)
-        enforce_real!(pot_fourier, basis)  # Symmetrize coeffs to have real iFFT
-        return irfft(basis, to_device(basis.architecture, pot_fourier))
+        enforce_real!(pot, basis)  # Symmetrize coeffs to have real iFFT
+        return irfft(basis, pot_fourier)
     else
-        return ifft(basis, to_device(basis.architecture, pot_fourier))
+        return ifft(basis, pot_fourier)
     end
 end
 (::AtomicLocal)(basis::PlaneWaveBasis{T}) where {T} =
@@ -136,28 +153,33 @@ function compute_forces(::TermAtomicLocal, basis::PlaneWaveBasis{T}, ψ, occupat
 end
 @timing "forces: local" function forces_local(S, basis::PlaneWaveBasis{T}, ρ, q) where {T}
     model = basis.model
-    recip_lattice = model.recip_lattice
-    ρ_fourier = to_cpu(fft(basis, total_density(ρ)))
     real_ifSreal = S <: Real ? real : identity
 
-    # TODO: currently, local forces entierly computed on the CPU.
-    #       This might not be optimal. See compute_local_potential() too.
     Gqs_cart = [model.recip_lattice * (G + q) for G in to_cpu(G_vectors(basis))]
-    form_factors = atomic_local_form_factors(basis, Gqs_cart)
+    form_factors, iG2norm = atomic_local_form_factors(basis, Gqs_cart)
+
+    Gqs = map(G -> G+q, G_vectors(basis))
+    ρ_fourier = reshape(fft(basis, total_density(ρ)), length(Gqs))
+
+    indices = to_device(basis.architecture, collect(1:length(Gqs)))
+    ρpot = similar(ρ_fourier)
+    twoπp = similar(ρ_fourier)
 
     # energy = sum of form_factor(G) * struct_factor(G) * rho(G)
     # where struct_factor(G) = e^{-i G·r}
     forces = [zero(Vec3{S}) for _ = 1:length(model.positions)]
     for (igroup, group) in enumerate(model.atom_groups)
-        element = model.atoms[first(group)]
         for idx in group
             r = model.positions[idx]
-            forces[idx] = -real_ifSreal(sum(conj(ρ_fourier[iG])
-                                              * form_factors[(igroup, norm(Gqs_cart[iG]))]
-                                              * cis2pi(-dot(G + q, r))
-                                              * (-2T(π)) * (G + q) * im
-                                              / sqrt(model.unit_cell_volume)
-                                        for (iG, G) in enumerate(to_cpu(G_vectors(basis)))))
+
+            ff_group = @view form_factors[:, igroup]
+            map!(iG -> cis2pi(-dot(Gqs[iG], r))*conj(ρ_fourier[iG])*
+                      ff_group[iG2norm[iG]], ρpot, indices)
+
+            forces[idx] += map(1:3) do α
+                map!(p -> -2π*im*p[α], twoπp, Gqs)
+                -real_ifSreal(sum(ρpot .* twoπp)/sqrt(model.unit_cell_volume))
+            end
         end
     end
     forces


src/terms/local.jl
```diff
@@ -71,19 +71,32 @@ function (external::ExternalFromFourier)(basis::PlaneWaveBasis{T}) where {T}
...
-    form_factors = IdDict{Tuple{Int,T},T}()  # IdDict for Dual compatibility
-    for G in Gqs_cart
+    iG2norm = zeros(Int, length(Gqs_cart))
+    norms = IdDict{T, Int}()
+    for (iG, G) in enumerate(Gqs_cart)
         p = norm(G)
+        if !haskey(norms, p)
+            norms[p] = length(norms) + 1
+        end
+        iG2norm[iG] = norms[p]
```
mfherbst, line 83: Look at `get!` which does exactly this in one line.

src/terms/local.jl
```diff
@@ -71,19 +71,32 @@ function (external::ExternalFromFourier)(basis::PlaneWaveBasis{T}) where {T}
...
+            if form_factors[inorm, igroup] ≈ zero(T)
                 element = basis.model.atoms[first(group)]
-                form_factors[(igroup, p)] = local_potential_fourier(element, p)
+                form_factors[inorm, igroup] = local_potential_fourier(element, p)
             end
         end
     end
-    form_factors
+    form_factors = to_device(basis.architecture, form_factors)
+    iG2norm = to_device(basis.architecture, iG2norm)
```
mfherbst, line 98: This can lead to type instability (the type of a variable changes). The compiler may be smart enough to figure it out here, though ... but I would check.

src/terms/local.jl
```diff
@@ -104,26 +117,30 @@ function compute_local_potential(basis::PlaneWaveBasis{T}; positions=basis.model
     # and a structure factor e^{-i G·r}
     model = basis.model
     Gqs_cart = [model.recip_lattice * (G + q) for G in to_cpu(G_vectors(basis))]
```
mfherbst, line 119: If this is no longer needed in the rest of the function this should be moved into `atomic_local_form_factors` in my opinion.

src/terms/local.jl
```diff
@@ -104,26 +117,30 @@ function compute_local_potential(basis::PlaneWaveBasis{T}; positions=basis.model
...
-    # TODO Bring Gqs_cart on the CPU for compatibility with the pseudopotentials which
-    #      are not isbits ... might be able to solve this by restructuring the loop
 
-    form_factors = atomic_local_form_factors(basis, Gqs_cart)
+    form_factors, iG2norm = atomic_local_form_factors(basis, Gqs_cart)
+    Gqs = map(G -> G+q, G_vectors(basis))
+
+    Tpot = promote_type(eltype(form_factors), eltype(eltype(positions)))
+    pot = to_device(basis.architecture, zeros(Complex{Tpot}, length(Gqs)))
+    pot_tmp = similar(pot)
```
mfherbst, line 126: Add a comment this is to avoid allocations.

src/terms/local.jl
```diff
@@ -104,26 +117,30 @@ function compute_local_potential(basis::PlaneWaveBasis{T}; positions=basis.model
     # and a structure factor e^{-i G·r}
     model = basis.model
     Gqs_cart = [model.recip_lattice * (G + q) for G in to_cpu(G_vectors(basis))]
-    # TODO Bring Gqs_cart on the CPU for compatibility with the pseudopotentials which
-    #      are not isbits ... might be able to solve this by restructuring the loop
 
-    form_factors = atomic_local_form_factors(basis, Gqs_cart)
+    form_factors, iG2norm = atomic_local_form_factors(basis, Gqs_cart)
```
mfherbst, line 121: Not sure if iG2norm is the best name. It essentially maps from iG to the location where the appropriate form factor element is stored.

src/terms/local.jl
```diff
@@ -71,19 +71,32 @@ function (external::ExternalFromFourier)(basis::PlaneWaveBasis{T}) where {T}
...
+        iG2norm[iG] = norms[p]
+    end
+
+    form_factors = zeros(T, length(norms), length(basis.model.atom_groups))
+    for (iG, G) in enumerate(Gqs_cart)
+        p = norm(G)
+        inorm = iG2norm[iG]
         for (igroup, group) in enumerate(basis.model.atom_groups)
-            if !haskey(form_factors, (igroup, p))
+            if form_factors[inorm, igroup] ≈ zero(T)
```
mfherbst, line 91: it should be exactly zero, no ?

It is unexepected to me to equal "zero" and "unitialised", I'm not yet sure what to think about it right now. But let's just think again if there is another way.

src/terms/local.jl
```diff
@@ -104,26 +117,30 @@ function compute_local_potential(basis::PlaneWaveBasis{T}; positions=basis.model
...
-    pot_fourier = map(enumerate(Gqs)) do (iG, G)
-        p = norm(Gqs_cart[iG])
-        pot = sum(enumerate(model.atom_groups)) do (igroup, group)
-            structure_factor = sum(r -> cis2pi(-dot(G, r)), @view positions[group])
-            form_factors[(igroup, p)] * structure_factor
+    for (igroup, group) in enumerate(model.atom_groups)
+        for r = positions[group]
+            ff_group = @view form_factors[:, igroup]
+            map!(iG -> cis2pi(-dot(Gqs[iG], r))*ff_group[iG2norm[iG]], pot_tmp, indices)
+            pot .+= pot_tmp./sqrt(model.unit_cell_volume)
```
mfherbst, line 133: spacings

src/terms/local.jl
```diff
@@ -136,28 +153,33 @@ function compute_forces(::TermAtomicLocal, basis::PlaneWaveBasis{T}, ψ, occupat
...
     Gqs_cart = [model.recip_lattice * (G + q) for G in to_cpu(G_vectors(basis))]
-    form_factors = atomic_local_form_factors(basis, Gqs_cart)
+    form_factors, iG2norm = atomic_local_form_factors(basis, Gqs_cart)
+
+    Gqs = map(G -> G+q, G_vectors(basis))
+    ρ_fourier = reshape(fft(basis, total_density(ρ)), length(Gqs))
+
+    indices = to_device(basis.architecture, collect(1:length(Gqs)))
+    ρpot = similar(ρ_fourier)
+    twoπp = similar(ρ_fourier)
```
mfherbst, line 166: ρ and p are hard to distinguish. add an _ to make this visually clearer.

src/terms/local.jl
```diff
@@ -136,28 +153,33 @@ function compute_forces(::TermAtomicLocal, basis::PlaneWaveBasis{T}, ψ, occupat
...
-                                              / sqrt(model.unit_cell_volume)
-                                        for (iG, G) in enumerate(to_cpu(G_vectors(basis)))))
+
+            ff_group = @view form_factors[:, igroup]
+            map!(iG -> cis2pi(-dot(Gqs[iG], r))*conj(ρ_fourier[iG])*
+                      ff_group[iG2norm[iG]], ρpot, indices)
+
+            forces[idx] += map(1:3) do α
+                map!(p -> -2π*im*p[α], twoπp, Gqs)
+                -real_ifSreal(sum(ρpot .* twoπp)/sqrt(model.unit_cell_volume))
```
mfherbst, line 181: Should we not go all the way and also avoid the allocation of `twoπρ and the temporary in `ρpot .* twoπρ` by dirrectly doing the sum ?

src/terms/local.jl
```diff
@@ -136,28 +153,33 @@ function compute_forces(::TermAtomicLocal, basis::PlaneWaveBasis{T}, ψ, occupat
...
-        element = model.atoms[first(group)]
         for idx in group
             r = model.positions[idx]
-            forces[idx] = -real_ifSreal(sum(conj(ρ_fourier[iG])
-                                              * form_factors[(igroup, norm(Gqs_cart[iG]))]
-                                              * cis2pi(-dot(G + q, r))
-                                              * (-2T(π)) * (G + q) * im
-                                              / sqrt(model.unit_cell_volume)
-                                        for (iG, G) in enumerate(to_cpu(G_vectors(basis)))))
+
```
mfherbst, line 174: This new code is generally harder to read than what we had. To what extend is this needed to get it to work / performance ?

mfherbst (commented):
Very nice idea. I'm a little suspicious about the iG2norm and the way you check for uninitialised form factors.

Another idea here is to revert the direction: For `Gqs[iG]` we already need to do random access. Can we not just iterate over all form factor elements (unique by |G+q|) in their natural order and have another array that specifies the matching indices `iG` or is this just making a mess ?

abussy:
> Another idea here is to revert the direction: For Gqs[iG] we already need to do random access. Can we not just iterate over all form factor elements (unique by |G+q|) in their natural order and have another array that specifies the matching indices iG or is this just making a mess ?

I'm afraid this would complicate things. We would need some kind of array of array to store the iG indices for each |G+q|, which would require a double loop down the line. In the light of GPU computing, this would also lead to more smaller kernel launches (less efficient in principle).

src/terms/local.jl
abussy: I agree the code loses in clarity. The main challenge for the GPU is to use compatible constructs, and an explicit for loop is not one of them. My hope is that the comment describing the formula before the loop is enough to guide the reader. If not, an additional comment could be added before each `map!` describing what it does

src/terms/local.jl
abussy: I'm always reluctant to use exact comparison with floats, but maybe this is fine in Julia. As for the "zero" = "uninitialized", I'll give it some thoughts.

abussy: I came up with a work around. It's a bit clunky, but more explicit.

src/terms/local.jl
```diff
@@ -71,19 +71,35 @@ function (external::ExternalFromFourier)(basis::PlaneWaveBasis{T}) where {T}
...
+# Additionally, returns a mapping from any G index to a unique |G + q| index
+function atomic_local_form_factors(basis::PlaneWaveBasis{T}; q=zero(Vec3{T})) where{T}
+    Gqs_cart = [basis.model.recip_lattice * (G + q) for G in to_cpu(G_vectors(basis))]
+
+    iG2inorm_cpu = zeros(Int, length(Gqs_cart))
+    norm_indices = IdDict{T, Int}()
+    for (iG, G) in enumerate(Gqs_cart)
+        p = norm(G)
+        get!(norm_indices, p, length(norm_indices) + 1)
+        iG2inorm_cpu[iG] = norm_indices[p]
```
mfherbst, line 83:
```suggestion
        iG2inorm_cpu[iG] = get!(norm_indices, norm(G), length(norm_indices) + 1)
```
or is that too loaded of a line ?

src/terms/local.jl
```diff
@@ -71,19 +71,35 @@ function (external::ExternalFromFourier)(basis::PlaneWaveBasis{T}) where {T}
...
 # Returns the form factors at unique values of |G + q| (in Cartesian coordinates).
-# Uses a hash map for O(1) lookup.
-function atomic_local_form_factors(basis::PlaneWaveBasis{T}, Gqs_cart::AbstractArray) where{T}
-    form_factors = IdDict{Tuple{Int,T},T}()  # IdDict for Dual compatibility
-    for G in Gqs_cart
+# Additionally, returns a mapping from any G index to a unique |G + q| index
+function atomic_local_form_factors(basis::PlaneWaveBasis{T}; q=zero(Vec3{T})) where{T}
+    Gqs_cart = [basis.model.recip_lattice * (G + q) for G in to_cpu(G_vectors(basis))]
+
+    iG2inorm_cpu = zeros(Int, length(Gqs_cart))
```
mfherbst, line 78: Why not `iG2iform_cpu`, essentially that's the whole point, no ? To get a mapping from G to the correct form factor value ?
Or do you need anywhere the property that it's a mapping from G to unique G+q ?

src/terms/local.jl
```diff
@@ -71,19 +71,35 @@ function (external::ExternalFromFourier)(basis::PlaneWaveBasis{T}) where {T}
...
+    norm_indices = IdDict{T, Int}()
+    for (iG, G) in enumerate(Gqs_cart)
+        p = norm(G)
+        get!(norm_indices, p, length(norm_indices) + 1)
+        iG2inorm_cpu[iG] = norm_indices[p]
+    end
+
+    form_factors_cpu = zeros(T, length(norm_indices), length(basis.model.atom_groups))
+    visited_norms = IdDict{T, Int}()
+    for (iG, G) in enumerate(Gqs_cart)
```
mfherbst, line 88: Can't you just loop over `norm_indices` ? This directly gives you `inorm` and `p` and moreover it's unique wrt. `p`.

src/terms/local.jl
```diff
@@ -136,28 +154,33 @@ function compute_forces(::TermAtomicLocal, basis::PlaneWaveBasis{T}, ψ, occupat
...
-            forces[idx] = -real_ifSreal(sum(conj(ρ_fourier[iG])
-                                              * form_factors[(igroup, norm(Gqs_cart[iG]))]
-                                              * cis2pi(-dot(G + q, r))
-                                              * (-2T(π)) * (G + q) * im
-                                              / sqrt(model.unit_cell_volume)
-                                        for (iG, G) in enumerate(to_cpu(G_vectors(basis)))))
+
+            ff_group = @view form_factors[:, igroup]
+            map!(iG -> cis2pi(-dot(Gqs[iG], r)) * conj(ρ_fourier[iG]) *
+                      ff_group[iG2inorm[iG]], ρ_pot, indices)
```
mfherbst, line 178: I think the 
```julia
map!(ρ_pot, indices) do iG
    stuff
end
```
syntax may be easier to read here.

src/terms/local.jl
```diff
@@ -136,28 +154,33 @@ function compute_forces(::TermAtomicLocal, basis::PlaneWaveBasis{T}, ψ, occupat
...
+
+            forces[idx] += map(1:3) do α
+                map!(iG -> -2π*im*Gqs[iG][α] * ρ_pot[iG], twoπp_ρ_pot, indices)
+                -real_ifSreal(sum(twoπp_ρ_pot) / sqrt(model.unit_cell_volume))
```
mfherbst, line 182: I don't see why you need the temporary here. Can't you just
```julia
tmp = sum(indices) do iG
    -2π*im*Gqs[iG][α] * ρ_pot[iG]
end
- real_ifSreal(tmp) / sqrt(model.unit_cell_volume)
```

src/terms/local.jl
mfherbst: Makes sense. Yes, I think the comment helps.

src/terms/local.jl
mfherbst: I think this goes in the right direction, I made another suggestion.

src/terms/local.jl
abussy: No, it's always tied to the form factor. `iG2iform_cpu` sounds good

src/terms/local.jl
abussy: Indeed, that will make things cleaner

PR #1057, opened 2025-02-06
Use DifferentationInterface instead of DiffResults

Technici4n:
It's a bit nicer since the result struct doesn't need to be pre-allocated with the right dimensions.


mfherbst (commented):
All up for it, `DifferentiationInterface` is certainly the more modern way to do this.

We need to fix the tests though.

BTW in https://github.com/JuliaMolSim/DftFunctionals.jl we also use `DiffResults` extensively. May be good to replace it there, too (but there the allocations matter, so one has to be careful).

src/postprocess/refine.jl
```diff
@@ -172,14 +172,13 @@ The refined energies can be obtained by E + dE.
...
-                                    zeros(T, length(term_names)))
+
     f(ε) = energy(refinement.basis,
                   refinement.ψ + ε.*refinement.δψ,
                   refinement.occupation;
                   ρ=refinement.ρ + ε.*refinement.δρ).energies.values
-    result = ForwardDiff.derivative!(result, f, zero(T))
-    (; E=Energies(term_names, result.value), dE=Energies(term_names, result.derivs[1]))
+    result = value_and_derivative(f, AutoForwardDiff(), zero(T))
+    (; E=Energies(term_names, result[1]), dE=Energies(term_names, result[2]))
```
mfherbst, line 181: I don't like the `[1]` and `[2]` thing here. Can you not unpack the tuple (maybe it's even a named tuple) ?

Technici4n: Better now?

PR #1058, opened 2025-02-06
Enable GPU pre-compilation

abussy:
With the latest Julia and JuliaGPU versions, it is now possible to pre-compile GPU workloads. This helps reduce the startup time of GPU calculations, which has been a major issue until now.

On the GH200 architecture, this amounts to a reduction of ~40 seconds (!) per SCF calculation, regardless of the input details. I expect even larger gains on older computers.

### DIFF
--- a/ext/DFTKCUDAExt.jl
+++ b/ext/DFTKCUDAExt.jl
@@ -1,4 +1,5 @@
 module DFTKCUDAExt
+using PrecompileTools
 using CUDA
 import DFTK: GPU, DispatchFunctional
 using DftFunctionals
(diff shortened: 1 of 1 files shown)


ext/DFTKCUDAExt.jl
```diff
@@ -15,4 +16,33 @@ for fun in (:potential_terms, :kernel_terms)
...
+        magnetic_moments = [2, -2]
+
+        @compile_workload begin
+            model = model_DFT(lattice, atoms, positions;
+                              functionals=LDA(), magnetic_moments,
+                              temperature=0.1, spin_polarization=:collinear)
+            basis = PlaneWaveBasis(model; Ecut=5, kgrid=[2, 2, 2], architecture=GPU(CuArray))
+            ρ0 = guess_density(basis, magnetic_moments)
+            scfres = self_consistent_field(basis; ρ=ρ0, tol=1e-2, maxiter=3, callback=identity)
+            compute_forces_cart(scfres)
```
mfherbst, line 42: Would it make sense to put this code somehow into a single location (thinking of a non-exported function in DFTK).

My rationale is that this we have to essentially do for every architecture (CPU, CUDA, maybe later AMDGPU) so just copying it could be come an annoying maintainance burden.

But maybe I am overthinking this.

mfherbst (commented):
Awesome, thanks very much !

mfherbst:
I made a slightly polished version in #1060.

PR #1060, opened 2025-02-14
 Enable GPU pre-compilation

mfherbst:
Superseeds #1058 based on @abussy's work.

### DIFF
--- a/.github/workflows/CompatHelper.yaml
+++ b/.github/workflows/CompatHelper.yaml
@@ -3,12 +3,17 @@ name: CompatHelper
 on:
   schedule:
     - cron: '00 00 * * *'
+  workflow_dispatch:
+
+permissions:
+  contents: write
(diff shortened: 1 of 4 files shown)


mfherbst:
@abussy

I am really confused by this. Without the check in https://github.com/JuliaMolSim/DFTK.jl/pull/1060/commits/d5c3d1389453e9e83527ed0e607d8e1ba2f9db4b the precompilation fails because some methods defined in `LibxcCudaExt` are not available.

However, this suggests that with this additional check, GPU precompilation will be a avoid operation --- which is also what I find in my experiments (i.e. that on our GPU machine I see no difference with and without this PR).

mfherbst:
Ok, thanks, I'll try that.

mfherbst:
Cool, thanks it works. Saves about a factor of 5 in the startup cost, very impressive.

PR #1061, opened 2025-02-14
Port XC term instantiation and forces to GPU

abussy:
Applied similar strategy as PR #1056 in order to port the XC term to the GPU.

The `_force_xc` function makes this a little bit awkward for the forces.

### DIFF
--- a/src/density_methods.jl
+++ b/src/density_methods.jl
@@ -101,8 +101,7 @@ end
 # densities.
 function atomic_total_density(basis::PlaneWaveBasis{T}, method::AtomicDensity;
                               coefficients=ones(T, length(basis.model.atoms))) where {T}
-    form_factors = atomic_density_form_factors(basis, method)
-    atomic_density_superposition(basis, form_factors; coefficients)
+    atomic_density_superposition(basis, method; coefficients)
 end
 
 # Build a spin density from a superposition of atomic densities and provided magnetic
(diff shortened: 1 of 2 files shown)


src/density_methods.jl
```diff
@@ -134,50 +133,63 @@ function atomic_spin_density(basis::PlaneWaveBasis{T}, method::AtomicDensity,
...
+
+    iG2ifnorm_cpu = zeros(Int, length(G_cart))
+    norm_indices = IdDict{T, Int}()
+    for (iG, G) in enumerate(G_cart)
+        p = norm(G)
+        iG2ifnorm_cpu[iG] = get!(norm_indices, p, length(norm_indices) + 1)
+    end
+
+    form_factors_cpu = zeros(T, length(norm_indices), length(basis.model.atom_groups))
+    for(p, ifnorm) in norm_indices
```
mfherbst, line 183:
```suggestion
    for (p, ifnorm) in norm_indices
```

src/density_methods.jl
```diff
@@ -134,50 +133,63 @@ function atomic_spin_density(basis::PlaneWaveBasis{T}, method::AtomicDensity,
...
     enforce_real!(ρ, basis)  # Symmetrize Fourier coeffs to have real iFFT
-    irfft(basis, ρ)
+    irfft(basis, reshape(ρ, basis.fft_size))
 end
 
+"""
+Returns the form factors at unique values of |G + q| (in Cartesian coordinates).
+Additionally, returns a mapping from any G index to the corresponding entry in the form_factors array.
+"""
 function atomic_density_form_factors(basis::PlaneWaveBasis{T},
```
mfherbst, line 171: Is there some point in generalising this with the function you adapted in  #1056. I mean it is pretty much the same code or am I missing something ?

src/terms/xc.jl
```diff
@@ -163,42 +163,51 @@ end
...
             r = model.positions[iatom]
-            forces[iatom] = _force_xc(basis, Vxc_fourier, form_factors, igroup, r)
+            forces[iatom] = _force_xc(basis, Vxc_fourier, form_factors, iG2ifnorm,
+                                      igroup, r, Gs, work, indices)
         end
     end
     forces
 end
 
 # Function barrier to work around various type instabilities.
```
mfherbst, line 195: Perhaps what you did solves the type instabilities. Can you check that ?

In case it does we can inline the function, making this less clunky.

src/density_methods.jl
abussy: Indeed, it is very similar. Only difference is the actual method for the calculation of the form factors (and the optional + q for atomic_local). I guess a single function could be used, with the actual form factor function name passed as (keyword) argument.

Given the order of file inclusion on `DFTK.jl`, it would probably need to be defined in `local.jl` or some other file up the tree.

mfherbst: Or a separate file atomic_form_factors.jl or similar ... there is some point in using this for computing projectors for DOS calculations as well if I don't misremember.

src/terms/xc.jl
abussy: Unfortunately, the type instability is still present. I changed the `_force_xc()` function to include the loop over atoms and make its call less clunky

src/density_methods.jl
abussy: There might also be some application for the `AtomicNonlocal` term, which currently computes form factors for all G vectors, and not only for unique norms. I'd suggest to tackle this in a separate PR

src/terms/xc.jl
```diff
@@ -163,44 +163,47 @@ end
...
-                   if has_core_density(model.atoms[first(group)])]
+                   if has_core_density(basis.model.atoms[first(group)])]
     @assert !isnothing(nlcc_groups)
 
-    TT = promote_type(T, eltype(Vxc_real))
-    forces = [zero(Vec3{TT}) for _ = 1:length(model.positions)]
-    for (igroup, group) in nlcc_groups
-        for iatom in group
-            r = model.positions[iatom]
-            forces[iatom] = _force_xc(basis, Vxc_fourier, form_factors, igroup, r)
```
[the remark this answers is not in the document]
abussy: I have observed drastic performance drops when unwrapping the `_force_xc` function (factor ~3x on the xc forces). This new version of the function barrier performs the same as the original. I attributed that to type instability. Granted, I did not test measure on phonon calculations. I will have a look.

src/terms/xc.jl
```diff
@@ -163,44 +163,47 @@ end
...
     TT = promote_type(T, real(U))
-    f  = zero(Vec3{TT})
-    for (iG, (G, G_cart)) in enumerate(zip(to_cpu(G_vectors(basis)), to_cpu(G_vectors_cart(basis))))
-        f -= real(conj(Vxc_fourier[iG])
-                  .* form_factors[(igroup, norm(G_cart))]
-                  .* cis2pi(-dot(G, r))
-                  .* (-2T(π)) .* G .* im
-                  ./ sqrt(basis.model.unit_cell_volume))
+    Gs = G_vectors(basis)
+    work = to_device(basis.architecture, zeros(Complex{TT}, length(Gs)))
```
mfherbst, line 186: Can't you allocate the zeros directly on the device (see `zeros_like` function in DFTK)

src/terms/xc.jl
```diff
@@ -163,44 +163,47 @@ end
...
+        for iatom in group
+            r = basis.model.positions[iatom]
+            ff_group = @view form_factors[:, igroup]
+            map!(work, indices) do iG
+                cis2pi(-dot(Gs[iG], r)) * conj(Vxc_fourier[iG]) * ff_group[iG2ifnorm[iG]]
+            end
+
+            forces[iatom] += map(1:3) do α
+                tmp = sum(indices) do iG
+                    -2π*im*Gs[iG][α] * work[iG]
```
mfherbst, line 200: Just realised: Isn't this scalar indexing ? Why does this even work ? Is this actually compiled to a GPU kernel ... that would be pretty amazing if yes.

src/terms/xc.jl
```diff
@@ -163,44 +163,47 @@ end
...
-        Vxc_fourier = to_cpu(fft(basis, mean(Vxc_real, dims=4)))
+        Vxc_fourier = fft(basis, mean(Vxc_real, dims=4))
     end
 
-    model = basis.model
-    form_factors = atomic_density_form_factors(basis, CoreDensity())
+    form_factors, iG2ifnorm = atomic_density_form_factors(basis, CoreDensity())
     nlcc_groups = [(igroup, group) for (igroup, group) in enumerate(basis.model.atom_groups)
-                   if has_core_density(model.atoms[first(group)])]
+                   if has_core_density(basis.model.atoms[first(group)])]
```
mfherbst, line 174: Maybe use `filter` here ... and then only add the `igroup` in the `for` loop in `_forces_xc` below. I find it a little strange that the index is added, but this is not really used for anything in this function, but only in the called function. That's a little surprising.

src/terms/xc.jl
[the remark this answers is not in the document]
mfherbst: Yeah, just use Cthulhu with `@descend_code_warntype`. I think the main issue is whether it figures out the type of `form_factors` and `iG2ifnorm` inside the called function `_forces_xc` specifically. If yes we should be good, if no, then the hot maps in the inner loop will do generic arithmetic.

The phonon case essentially is the case where `T != U`. In standard DFT we should have `T == U`.

src/terms/xc.jl
abussy: This `sum()` construct does indeed run on the GPU, with a `partial_mapreduce_grid` CUDA kernel. As far as I understand, this is an alias for something like `sum(-2π*im .* Gs .* work)`, which is an array operation allowed on the GPU.
Additionally, trying to access elements of `Gs` or `work` outside of the sum leads to scalar indexing errors.

src/terms/xc.jl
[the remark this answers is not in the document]
abussy: I am a bit confused here. In phonon calculations, the `compute_dynmat` function is called for each term, and within it forces are calculated. However, there is no definition of `compute_dynmat` in `xc.jl`, and it therefore default to the `terms.jl` default (returns nothing). So the XC `compute_forces` function is never called in phonon calculations.

Moreover, there is not a single test in DFTK where `T != U`. Could this be an ancient artifact?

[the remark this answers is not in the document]
mfherbst: It comes from here: https://github.com/JuliaMolSim/DFTK.jl/pull/766 so it is actually way older than the phonon stuff, sorry for setting you down the wrong road.

I am pretty sure the case `T != U` up in sensitivity and AD scenarios, but let's not be held back too much by this. If in a fairly standard case `@descend_code_warntype` flags up nothing serious, I'd be happy to go along with this.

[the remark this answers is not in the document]
mfherbst: I think there is actually a type instability still. The `forces` array is type-identified as `Vector{Any}` for some reason. I just ran an LDA silicon example and did
```julia
@descend_code_warntype compute_forces(basis.terms[end], basis, scfres.ψ, scfres.occupation; scfres.ρ)
```
where `basis.terms[end]` is a `TermXc`. I did not further identify where this comes from, but in any case the main computation seems to be done in a type-stable manner.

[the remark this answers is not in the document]
mfherbst: I fixed it with a type annotation. All happy now :).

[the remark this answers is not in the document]
abussy: It turns out that similar type instabilities show up for the `AtomicLocal` and the `AtomicNonlocal` terms, when using `@descend_code_warntype`. There too, adding a type annotation solves it. I'll add it as part of this PR.

Additionally, I'll investigate further whether the `_forces_xc` function barrier is necessary or not.

PR #1062, opened 2025-02-14
Minor optimization of FFTGrid initialization

abussy:
With PR #1056 and #1061, the bottleneck of `PlaneWaveBasis` instantiation becomes the creation of `r_vectors` at `FFTGrid` initialization. For systems with high cutoffs, this is non-negligible.

This PR enables the calculation of the `r_vectors` on the GPU. Additionally, it greatly speeds-up the CPU version by precomputing and recasting the inverse of `fft_size`.

### DIFF
--- a/src/fft.jl
+++ b/src/fft.jl
@@ -89,7 +89,8 @@ function FFTGrid(fft_size::Tuple{Int, Int, Int}, unit_cell_volume::T,
 
     VT = value_type(T)
     fac = 1 ./ VT.(fft_size)
-    r_vectors = map(idx -> Vec3{VT}(fac .* (idx.I .- (1, 1, 1))), CartesianIndices(fft_size))
+    r_vectors_cpu = map(idx -> Vec3{VT}(fac .* (idx.I .- (1, 1, 1))), CartesianIndices(fft_size))
+    r_vectors = to_device(arch, r_vectors_cpu)
 
     FFTGrid{T, VT, typeof(Gs), typeof(r_vectors)}(fft_size, opFFT, ipFFT, opBFFT, ipBFFT,
                                                   fft_normalization, ifft_normalization,
--- a/src/fft.jl
+++ b/src/fft.jl
@@ -88,9 +88,8 @@ function FFTGrid(fft_size::Tuple{Int, Int, Int}, unit_cell_volume::T,
     fft_normalization  = sqrt(unit_cell_volume) / length(ipFFT)
 
     VT = value_type(T)
-    r_vectors = [(Vec3{VT}(idx.I) .- (1, 1, 1)) ./ VT.(fft_size)
-                 for idx in CartesianIndices(fft_size)]
-    r_vectors = to_device(arch, r_vectors)
+    fac = 1 ./ VT.(fft_size)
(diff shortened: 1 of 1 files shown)


src/fft.jl
```diff
@@ -88,9 +88,8 @@ function FFTGrid(fft_size::Tuple{Int, Int, Int}, unit_cell_volume::T,
     fft_normalization  = sqrt(unit_cell_volume) / length(ipFFT)
 
     VT = value_type(T)
-    r_vectors = [(Vec3{VT}(idx.I) .- (1, 1, 1)) ./ VT.(fft_size)
-                 for idx in CartesianIndices(fft_size)]
-    r_vectors = to_device(arch, r_vectors)
+    fac = 1 ./ VT.(fft_size)
+    r_vectors = map(idx -> Vec3{VT}(fac .* (idx.I .- (1, 1, 1))), CartesianIndices(fft_size))
```
mfherbst, line 92: I may be blind, but why does this now do computation on the GPU ?

abussy: You are right, it does not. I have been blindsided because it is so much faster, and because removing the `to_device()` function does not break the GPU tests, nor any example I ran.

Browsing the code, it is not clear to me whether `r_vectors` should be a GPU array at all or not, actually

mfherbst: Maybe it does not need to be. For consistency it is, however, very strange if `r_vectors(basis)` returns a CPU array but `G_vectors(basis)` does return a GPU array. Could the solution be to do the computation on the CPU and then only transfer it to the GPU thereafter ?

abussy: Given the speed of the operation, this is very reasonable. And we get peace of mind and consistency.

src/fft.jl
```diff
@@ -89,7 +89,8 @@ function FFTGrid(fft_size::Tuple{Int, Int, Int}, unit_cell_volume::T,
 
     VT = value_type(T)
     fac = 1 ./ VT.(fft_size)
-    r_vectors = map(idx -> Vec3{VT}(fac .* (idx.I .- (1, 1, 1))), CartesianIndices(fft_size))
+    r_vectors_cpu = map(idx -> Vec3{VT}(fac .* (idx.I .- (1, 1, 1))), CartesianIndices(fft_size))
```
mfherbst, line 92: Add a comment that the reason here is speed. Usually we do these things, since they cannot be done easily on the GPU or since we cannot be bothered to do it on the GPU.

abussy:
Not exactly. My tests have shown that simply taking out `fac` from the original construct, i.e.
```
VT = value_type(T)
fac = 1 ./ VT.(fft_size)
r_vectors = [(Vec3{VT}(idx.I) .- (1, 1, 1)) .* fac for idx in CartesianIndices(fft_size)]
```
is considerably slower than the proposed change (factor ~5x). I am not sure why that is, but I suspect it has to do with with creating a `Vec3{VT}` as the last step. In fact, the following is as efficient as the `map` construct I initially proposed:

```
fac = 1 ./ VT.(fft_size)
r_vectors = [Vec3{VT}(fac .* (idx.I .- (1, 1, 1))) for idx in CartesianIndices(fft_size)]
```
I am happy to switch to the above, if that keeps the code clearer

mfherbst:
In my opinion it does, yes.

PR #1068, opened 2025-02-24
GPU optimization of LOBPCG

abussy:
This PR is the result of a detailed profiling of the LOBPCG solver with NVIDIA's Nsight Systems. It allowed for the identification of various hot spots, where code is very slow during GPU runs.

In particular, there are many instances of explicit loops over matrix columns. This access pattern is not ideal, as the massive parallelism of the GPU is not fully exploited. Array operations on the whole matrix are far more efficient.

I measured speed-ups of the order of 30% on the whole LOBPCG iterative solver. Excluding the cost of the H x Psi product (not modified in this PR), the speed-ups reach 50%.

Unfortunately, this comes at the cost of some code readability. I left comments describing what is calculated when necessary.

Finally, I scrapped 2 loops using DFTK's custom threading (in `ortho! X vs Y` and `ldiv!` for the preconditioner). I made sure the effect is negligible on CPU runs ( tested with the default `n_DFTK` = `n_blas` thread option). It seems that simple BLAS threading on large array operations is quite efficient by itself.


src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -310,10 +305,17 @@ end
...
+    num = sum(conj(X) .* AX, dims=1)
+    den = sum(conj(X) .* BX, dims=1)
+    vec(real.(num ./ den))
+end
 
 function final_retval(X, AX, BX, resid_history, niter, n_matvec)
-    λ = @views [real((X[:, n]'*AX[:, n]) / (X[:, n]'BX[:, n])) for n=1:size(X, 2)]
-    λ_device = oftype(X[:, 1], λ)  # Offload to GPU if needed
+    λ_device = compute_λ(X, AX, BX)
+    λ = oftype(ones(eltype(λ_device), 1), λ_device)
```
antoine-levitt, line 318: This will need to be merged with master

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -324,8 +326,10 @@ function final_retval(X, AX, BX, resid_history, niter, n_matvec)
         BX = BX[:, p]
         resid_history = resid_history[p, :]
     end
+    # compute norms of column vectors with array operations for GPU performance
+    norms = vec(sqrt.(sum(abs2, residuals; dims=1)))
     (; λ=λ_device, X, AX, BX,
-     residual_norms=norm.(eachcol(residuals)),
+     residual_norms=oftype(ones(eltype(norms), 1), norms),
```
antoine-levitt, line 332: Add a `to_cpu` method?

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -249,24 +249,19 @@ end
...
-    dropped = Int[]
-    for i=1:size(X,2)
-        n = norm(@views X[:,i])
-        if n <= tol
-            X[:,i] = randn(T, size(X,1))
-            push!(dropped, i)
-        end
-    end
+    # Using array operations for GPU performance
+    norms = vec(sqrt.(sum(abs2, X; dims=1)))
```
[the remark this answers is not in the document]
antoine-levitt: Wait, does that not work? I can't test I don't have a GPU here.

antoine-levitt:
Hm, I'm kind of split on this.
- Looks like GPU style clashes with threading in a number of places. Threading has been mostly an annoyance, and for very limited results. I think I'm fine with just scrapping threading from DFTK altogether, and pointing people to GPU if they want to do large-scale.
- It's pretty unfortunate if GPU forces us towards a matlab/numpy style. Those dot products are not very nice... I'd rather use things like eachcol, map, etc.

Also don't add comments at each place we use vector-style operations: we should converge on one uniform style and use it without comments.

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -310,10 +305,17 @@ end
     X
 end
 
+# Computes λ = real((X' * AX) / (X' *BX)), for each column of X
+function compute_λ(X, AX, BX)
+    # using array operations for GPU performance
+    num = sum(conj(X) .* AX, dims=1)
+    den = sum(conj(X) .* BX, dims=1)
+    vec(real.(num ./ den))
+end
```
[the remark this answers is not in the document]
abussy: It did not show up as a an issue when I was testing the CPU performance of this PR. This might be the case sometimes, maybe. As for the norms above, the performance gain on the GPU is massive.

For code that could be problematic on the CPU, one option would be to have a GPU definition of it in `DFTKCUDAExt.jl` and dispatch to it when needed. The main issue with it is code inflation though.

src/eigen/lobpcg_hyper_impl.jl
abussy: Valid point, it would make a lot of sense here

src/eigen/lobpcg_hyper_impl.jl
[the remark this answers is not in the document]
antoine-levitt: Is this an issue with GPUArrays (that it doesn't implement the right overloads) or something more intrinsic?

mfherbst:
Yeah same here. It's a shame this has such an impact on readability. From my point of view the key points are:

- Avoid code duplication: I think it's better to have a bit of ugly code at selected places (with nice comments and a common convention of course) rather than a GPU and a CPU version with different code. Maybe for some primitives we use often (e.g. column-wise norms or outer products) we can invent functions that are easier to understand and hide the array syntax a little.

- Avoid performance regressions on CPU: For small arrays this may not matter but we should test on non-trivial calculations if the extra temporaries don't end up killing us

mfherbst (commented):
Not great, but within what we discussed looks like a good solution to me.

@antoine-levitt may also have some thoughts.

mfherbst:
If there really turns out to be a bad CPU perf issue can we specialise for CPU maybe, but certainly I'd try to avoid that.

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -33,6 +33,14 @@
 # other eigenvectors (which is not the case in many - all ? - other
 # implementations)
 
+# - The massive parallelism of the GPU can only be fully expoloited when
```
antoine-levitt, line 36: typo

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -33,6 +33,14 @@
...
+# operating on whole arrays. For performance reasons, one should avoid
+# explicitly looping over columns or element. For example, computing
+# the norms of the columns of an array A should be dones as:
```
antoine-levitt, line 39: typo

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -247,26 +255,22 @@ normest(M) = maximum(abs.(diag(M))) + norm(M - Diagonal(diag(M)))
     (; X, nchol, growth_factor)
 end
 
+# Calculate the norms of the columns of an array X
+function colnorms(X::AbstractArray{T}) where{T}
```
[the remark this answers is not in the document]
antoine-levitt: you don't need to type X

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -501,10 +504,9 @@ end
         end
 
         # Quick sanity check
-        for i = 1:size(X, 2)
-            @views if abs(BX[:, i]'X[:, i] - 1) >= sqrt(eps(real(eltype(X))))
-                error("LOBPCG is badly failing to keep the vectors normalized; this should never happen")
-            end
+        diffs = abs.(sum(conj(BX) .* X, dims=1) .-1)
```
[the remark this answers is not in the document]
antoine-levitt: Should we have a `diag_prod` function?

src/eigen/preconditioners.jl
```diff
@@ -28,7 +28,7 @@ mutable struct PreconditionerTPA{T <: Real}
     basis::PlaneWaveBasis
     kpt::Kpoint
     kin::AbstractVector{T}  # kinetic energy of every G
-    mean_kin::Union{Nothing, Vector{T}}  # mean kinetic energy of every band
+    mean_kin::Union{Nothing, AbstractVector{T}}  # mean kinetic energy of every band
```
antoine-levitt, line 31: This should probably be specialized, ie parametrize the type with TA and have kin::TA, mean_kin::Union{Nothing, TA}. But it doesn't matter much.

src/eigen/preconditioners.jl
```diff
@@ -50,9 +50,8 @@ end
     if P.mean_kin === nothing
         ldiv!(Y, Diagonal(P.kin .+ P.default_shift), R)
     else
-    parallel_loop_over_range(1:size(Y, 2)) do n
-            Y[:, n] .= P.mean_kin[n] ./ (P.mean_kin[n] .+ P.kin) .* R[:, n]
-        end
+        # same as loop Y[:, n] .= P.mean_kin[n] ./ (P.mean_kin[n] .+ P.kin) .* R[:, n]
+        Y .= (P.mean_kin' ./ (P.mean_kin' .+ P.kin)) .* R
```
antoine-levitt, line 54: That's the kind of matlabism/numpyisms that I was glad to leave behind when going to julia... sigh... I don't have a good solution here.

src/eigen/preconditioners.jl
```diff
@@ -64,16 +63,16 @@ ldiv!(P::PreconditionerTPA, R) = ldiv!(R, P, R)
...
+        Y .= ((P.mean_kin' .+ P.kin) ./ P.mean_kin') .* R
     end
     Y
 end
 (Base.:*)(P::PreconditionerTPA, R) = mul!(copy(R), P, R)
 
 function precondprep!(P::PreconditionerTPA, X::AbstractArray)
-    P.mean_kin = [real(dot(x, Diagonal(P.kin), x)) for x in eachcol(X)]
+    # real(dot(x, Diagonal(P.kin), x)), for all columns x of X (efficient on the GPU)
+    P.mean_kin = vec(real(sum(conj(X) .* (P.kin .* X); dims=1)))
```
antoine-levitt, line 75: Maybe we should have a `diagprod(X, P.kin, X)` method

antoine-levitt (commented):
Thanks for the PR. Not very satisfactory, but pluses outweigh minuses, unfortunately...

antoine-levitt:
Also careful in refactoring that lobpcg should be self-contained (not use too much stuff from other parts of DFTK) so that users can steal it for their own projects. We really should pull it off to its own project and depend on it. We had a plan to integrate it with either iterativesolvers or krylovkit but I think we should just give up on that and split it off.

mfherbst:
Sounds very reasonable, thanks @abussy .

abussy:
With this latest commit, I did the following:
- moved performance critical code into low-level functions (e.g. `compute_λ`, `diag_prod`), written in a CPU and Human friendly manner
- Changed the `DFTKCUDAExt` to a directory, in which the above function are re-implemented in a GPU optimized way. They will be called appropriately with multiple dispatch.

In the spirit of splitting LOBPCG from DFTK in the future, I also removed usage of DFTK's explicit threading. Note that I kept using DFTK's `to_cpu` function in order to avoid redefinition warnings. Since this is a one line function, it will be easy to move it to a LOBPCG module.

Finally, I kept the calculation of norms in the function `columnwise_norms` as an array operation, as it performs better than `norm.(eachcol(X))`, even on the CPU (marginally so, but still).

src/eigen/preconditioners.jl
```diff
@@ -11,6 +11,19 @@
 import LinearAlgebra: ldiv!
 import LinearAlgebra: mul!
 
+# Returns dot(A[:, i], Diagonal(P.kin), B[:, i])), for all columns of A, B
```
mfherbst, line 14: Mismatch between comment and implementation.

Also I would actually do this like the `dot` function. I.e. have a two-argument version (where you just call the two-argument `dot`) and a three-argument version, where you call the three-argument `dot`. I think the three-argument `dot` with an `I` (uniform scaling) compiles away, so you can just have that as your implementation and then dispatch `diag_prod(A, B) = diag_prod(A, I, B)`.

src/eigen/preconditioners.jl
```diff
@@ -11,6 +11,19 @@
...
+@views function diag_prod(A::AbstractArray{T}, B::AbstractArray{T}; diag=nothing) where {T}
```
mfherbst, line 15: Maybe we should move all of these helper functions to a single file in `src/common`. Because some being in `preconditioners.jl` and some in LOBPCG does not really help finding them in the future.

src/eigen/preconditioners.jl
mfherbst: Agree. I'd have this:

```suggestion
mutable struct PreconditionerTPA{T <: Real, Tkin <: AbstractVector{T}}
    basis::PlaneWaveBasis
    kpt::Kpoint
    kin::Tkin  # kinetic energy of every G
    mean_kin::Tkin  # mean kinetic energy of every band or arbitrary shift
end
```
and I would pre-set the `mean_kin` with a single elements set to the default shift of 1 or something like that.

src/eigen/preconditioners.jl
```diff
@@ -50,7 +63,7 @@ end
     if P.mean_kin === nothing
         ldiv!(Y, Diagonal(P.kin .+ P.default_shift), R)
     else
-    parallel_loop_over_range(1:size(Y, 2)) do n
```
mfherbst, line 53: Removing parallelisation ok ?

mfherbst (commented):
Some small comments, but mostly fine.

src/eigen/preconditioners.jl
abussy: Maybe in a new file `src/eigen/common.jl` ? This would keep the locality of functions needed by LOBPCG for the moment when it branches out of DFTK.

src/eigen/preconditioners.jl
abussy: I need to double check performance, but this also goes in the direction of eventually branching out LOBPCG  from DFTK.

mfherbst: I see, but I suppose the preconditioner would remain in DFTK in any case.

abussy:
Implemented most of the points brought up by @mfherbst:
- Moved help linear algebra functions into their own file in `src/eigen/linalg.jl`
- Reintroduced DFTK thread parallelization in the preconditioner: my measurements show that this is faster than BLAS threading alone, and if it stays in DFTK, it is not a problem. Overall, this has a close to negligible impact on timings.
- Wrote separate functions for 2- and 3-arguments `diag_prod`

I however did not change the `PreconditionerTPA` struct yet. Forcing the `mean_kin` field to be an `AbstracVector{T}` complicates the handling of the preconditioner application in the special case when there is only a default shift. Currently, we have
```
if P.mean_kin === nothing
    ldiv!(Y, Diagonal(P.kin .+ P.default_shift), R)
else
```
but to get the same with the above constraint, I do not find better than
```
if length(P.mean_kin) == 1 && P.mean_kin[1] == P.default_shift
    ldiv!(Y, Diagonal(P.kin .+ P.default_shift), R)
else
```
which looks very clunky to me.

src/eigen/linalg.jl
```diff
@@ -0,0 +1,19 @@
...
+
+# Returns a vector of dot(A[:, i], B[:, i]), for all columns of A, B
+@views function diag_prod(A::AbstractArray{T}, B::AbstractArray{T}) where {T}
+    @assert size(A) == size(B)
+    [real(dot(A[:, i], B[:, i])) for i = 1:size(A, 2)]
+end
+
+# Returns a vector of dot(A[:, i], Diagonal(diag), B[:, i]), for all columns of
+# A, B, and a vector diag
+@views function diag_prod(A::AbstractArray{T}, diag::AbstractVector,
```
mfherbst, line 14: Why not explicitly have `diag::Diagonal` and than pass it through to `dot` ? Or rather here it could even be D::AbstractMatrix without much changes, right ? (May not be so easy for the GPU implementation, though)

mfherbst:
> I however did not change the PreconditionerTPA struct yet. Forcing the mean_kin field to be an AbstracVector{T} complicates the handling of the preconditioner application in the special case when there is only a default shift. Currently, we have

Yeah ok, but I'd still use type templates instead of the `AbstractArray` objects as antoine suggested.

src/eigen/linalg.jl
abussy: I could probably pass an `AbstractMatrix` to the function, and have a special case with a `isdiag()` test on the GPU (and otherwise do the full MM). Not sure if it is worth the extra code.

mfherbst: Just specialise on the type ::Diagonal for GPU ? From the diagonal object you can extract the actual vector as a GPU array

abussy:
- Made `diag_prod` signature more like `dot` by taking a matrix rather than a vector
- Parametrized type of `kin` and `mean_kin` in `PreconditionerTPA`

src/eigen/linalg.jl
```diff
@@ -0,0 +1,19 @@
...
+    @assert size(A) == size(B)
+    [real(dot(A[:, i], B[:, i])) for i = 1:size(A, 2)]
+end
+
+# Returns a vector of dot(A[:, i], M, B[:, i]), for all columns of
+# A, B, and matrix M
+@views function diag_prod(A::AbstractArray{T}, M::AbstractMatrix,
+                          B::AbstractArray{T}) where {T}
+    @assert size(A) == size(B)
+    @assert size(M, 2) == size(B, 1)
```
antoine-levitt, line 17: I wouldn't do this (and I would not type M) because M can be a bit more abstract (eg I or something)

ext/DFTKCUDAExt/lobpcg.jl
```diff
@@ -0,0 +1,47 @@
+### GPU-specific implementations of functions called during LOBPCG
```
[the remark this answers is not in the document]
antoine-levitt: Sorry I see you've done this already.

antoine-levitt (commented):
LGTM, thanks!

mfherbst:
A general point that came to me: These GPU specific optimisations should be helpful for AMD as well, right? So perhaps we should not make the types CuArrays but GpuArrays (and thus not put the code into the extmodule)?

abussy:
Indeed, these optimizations should be vendor agnostic. I have access to some Mi200 GPUs, I'll try to run there and confirm it works/performs. If so, I suggest creating a new subdirectory in `src/common` to put the generic GPU optimizations (maybe `src/common/gpu_opts`). Then the DFTKCUDAExt could also be reverted to a single file.

mfherbst:
Yeah or src/common/linalg.jl or src/eigen/linalg.jl with both versions: CPU and GPU.

src/eigen/linalg.jl
```diff
@@ -0,0 +1,19 @@
+# Calculate the norms of the columns of an array
+function columnwise_norms(X::AbstractArray{T}) where{T}
+    vec(sqrt.(sum(abs2, X; dims=1)))
+end
+
+# Returns a vector of dot(A[:, i], B[:, i]), for all columns of A, B
```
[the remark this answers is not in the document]
abussy: mmh, I personally feel like `columnwise_norms` is more descriptive

mfherbst:
> Regarding the location, I put it in src/common/gpu_opts/lobpcg.jl, rather than in src/eigen/linalg.jl with the rest of the helper functions. My reasoning is that there will be more of these coming (e.g. symmetry in the next PR), and I think it is easier to navigate if all GPU specific implementations are in the same place.

Ok, I see. Then I'd not have this in common and rather use a separate subdirectory `src/gpu` with all these files inside. I would also move `src/workarounds/gpu_arrays.jl` in there. Maybe good to keep the naming, i.e. if there is ` src/eigen/linalg.jl` there should be `src/gpu/linalg.jl` ?

abussy:
> Maybe good to keep the naming between cpu and gpu consistent, i.e. if there is src/eigen/linalg.jl there should be src/gpu/linalg.jl with the respective GPU things ?

That would probably be the clearest solution. Only caveat is possible file proliferation:  `linalg.jl`, `lobpcg.jl` and `preconditioner.jl` for this PR alone. I think it is ok, since writing GPU specific code is only done when truly necessary.

mfherbst:
> Only caveat is possible file proliferation

True, but to me the preconditioner and lobpcg GPU things make sense to just be in a file `src/gpu/linalg.jl` ... but whatever, probably not too important after all. I think what matters is that the GPU functions should not start to scatter around multiple locations in the code base, but are in one clearly identified file or folder.

src/eigen/linalg.jl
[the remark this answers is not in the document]
antoine-levitt: This is not about that function it's another one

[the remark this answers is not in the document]
abussy: Oh, my bad. Right, that makes sense, especially since `diag_prod` was generalized to non-diagonal matrices.

mfherbst:
I just removed the assertions for the size checks. Such functions are pretty low-level and either the called functions anyway perform such size checks or the size agreement is ensured by the surrounding algorithm. Such `@assert` insertions will in general only discourage the compiler from inlineing.

@abussy Thanks very much for the good work, laying the foundation for more GPU improvements in the future.

src/eigen/linalg.jl
```diff
@@ -0,0 +1,15 @@
...
+@views function columnwise_dots(A::AbstractArray{T}, B::AbstractArray{T}) where {T}
+    [real(dot(A[:, i], B[:, i])) for i = 1:size(A, 2)]
+end
+
+# Returns a vector of real(dot(A[:, i], M, B[:, i])), for all columns of
+# A, B, and matrix M
+@views function columnwise_dots(A::AbstractArray{T}, M, B::AbstractArray{T}) where {T}
+    [real(dot(A[:, i], M, B[:, i])) for i = 1:size(A, 2)]
+end
```
Technici4n, line 15: What is the reasoning for the `real` calls here?

Technici4n: (asking because I don't see them in the GPU version - did I miss something?)

abussy: I think that's most likely a mistake. I assume it comes from the original preconditioner code: https://github.com/JuliaMolSim/DFTK.jl/blob/3c34d3855f6fec995edc3681caa075a4bd35ed1e/src/eigen/preconditioners.jl#L75-L77
But `real` is still taken there now, so I think it can go.

Technici4n: Do you want me to open an issue to remember?

abussy: I'll take care of it immediately

PR #980, opened 2024-06-05
Tweak bookkeeping in LOBPCG

antoine-levitt:
@mfherbst does that fix https://github.com/JuliaMolSim/DFTK.jl/issues/979?

PR #1071, opened 2025-02-27
Fix GPU runs

abussy:
PR #980 breaks GPU runs by doing element-wise access on GPU arrays. This fixes it by copying the offending array to the host first.

### DIFF
--- a/src/eigen/lobpcg_hyper_impl.jl
+++ b/src/eigen/lobpcg_hyper_impl.jl
@@ -311,7 +311,7 @@ end
 end
 
 function final_retval(X, AX, BX, λ, resid_history, niter, n_matvec)
-    λ_host = oftype(ones(eltype(λ), 1), λ) # Copy to CPU for element-wise access
+    λ_host = oftype(ones(eltype(λ), 1), λ)  # Copy to CPU for element-wise access
     if !issorted(λ_host)
         p = sortperm(λ_host)
         λ_host = λ_host[p]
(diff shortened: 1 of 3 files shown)


src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -310,17 +310,17 @@ end
     X
 end
 
-
 function final_retval(X, AX, BX, λ, resid_history, niter, n_matvec)
-    if !issorted(λ)
-        p = sortperm(λ)
-        λ = λ[p]
+    λ_host = oftype(ones(eltype(λ), 1), λ) # Copy to CPU for element-wise access
```
mfherbst, line 314: Since you are not copying it back to the device: Is this what we want ? Is this maybe what we do anyway in diag.jl (that is putting the eigenvalues on the cpu) ?

mfherbst: Alternative is maybe to return \lambda[sortperm] ?

abussy: There was actually a mistake, in which the unsorted eigenvalues where returned. This is fixed now (but was not caught by the tests).

As for not copying back to the GPU, this is indeed consistent with `diag.jl`, where the eigenvalues are transferred to the host anyways. So doing it at this stage isn't a performance overhead, and it is what we want.

mfherbst: Can you just add an appropriate note in the LOBPCG docstring? To me this would otherwise come as a surprise.

Issue #1064, opened 2025-02-17
Rename all mentions of HGH in the code to GTH.


raphaelzstone:
Here are all the files that I would need to change to phase out hgh pseudopotential files. Some of the actions I need to take I'm not totally sure about, so some direction would be appreciated based on what I have down for each of the files. At this point I've mostly just noted which files have hgh mentions that need to be reconciled, and took a guess about what or how to reconcile that issue, but I may not totally understand how hgh and gth files work with DFTK. Please advise or confirm for my approach.

test/PspHgh.jl – Probably just needs the file name changed to PspGth, and all references would need changing too.
test/list_psp.jl – I think just change the name to GTH in line 1.
src/pseudo/PspHgh.jl – Maybe a bit complicated. Not totally sure. This probably also encodes for gth files. Not totally sure.
src/pseudo/load_psp.jl – Remove .hgh file abilities.
src/workarounds/forwarddiff_rules.jl – Calling Psphgh, fixed with test/psphgh.jl renaming/structuring.
data/psp/hgh – Just remove all the pseudopotentials? Or maybe just change the names to gth? Not really clear on this.
examples/pseudopotentials.jl – Alter the example so it doesn't use .hgh files (only example with .hgh reference).
docs/features.md – Remove HGH mention (only mention of hgh in docs).

mfherbst:
I am not sure I fully understand your issue.

The `data/psp/hgh` files are essentially only provided for backward compatibility. The recommended way to use Goedecker-type pseudos is already via [PseudoPotentialData.jl](https://juliamolsim.github.io/PseudoPotentialData.jl/stable/) where they are already called "gth" files. The extension is essentially arbitrary. In our ecosystem "something.hgh" is the same format as "something.gth", namely the format used by CP2K. But there are in general no file format conventions for this and people do random things.

Yes, in DFTK we still use the acronym "HGH" at plenty of places and it would probably better to rename this consistently to "GTH", I agree with that. Would that be what you are after ?

raphaelzstone:
Ok, that makes sense. So should data/psp/hgh have the name changed for consistency? and should the examples/pseudopotentials.jl file keep the hgh example it currently uses, or should both examples be from PseudoPotentialData.jl? So then an edited list to change the HGH acronym to be GTH for consistency. Let me know if this looks good.

test/PspHgh.jl – Change file name and references to this file
test/list_psp.jl – I think just change the name to GTH in line 1.
[...]

PR #1075, opened 2025-03-04
Hgh Cleanup

raphaelzstone:
First of probably a few iterations for #1064 
I mostly changed basic things, some naming and such, and some references to hgh vs gth pseudopotentials. The next step in my mind would be to make a duplicate file of `src/pseudo/PspHgh.jl` called PspGth with all the same functionality, and update load_psp.jl accordingly, rerouting most loading through that. Then just putting warnings throughout PspHgh so if anyone is loading .hgh files (like from the data folder) there is a warning.
In the whole I'm still unfamiliar with both github and the process of deprecation, so let me know if theres stuff I should be doing, best practices I'm missing, and also advice on if these next steps are pertinent right now, or if they're even reasonable thought processes.

### DIFF
--- a/examples/pseudopotentials.jl
+++ b/examples/pseudopotentials.jl
@@ -6,7 +6,7 @@
 #
 # Currently, DFTK supports norm-conserving (NC) PSPs in
 # separable (Kleinman-Bylander) form. Two file formats can currently
-# be read and used: analytical Hartwigsen-Goedecker-Hutter (HGH) PSPs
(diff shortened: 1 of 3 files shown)


mfherbst:
So far all looks good to me. Happy to merge this if you want to do this in multiple PRs, but you can also do it in one go ... as you wish.

raphaelzstone:
I think I'll do a separate PR for the next steps since it'll be a bit more complicated for me.

PR #1077, opened 2025-03-05
Fix force symmetrisation bug

mfherbst:
Partial fix for #1066


mfherbst:
Hmmm. The immediate bug is fixed, but there is still something funny as the geometry optimisation example in #1066 does not converge at all.

PR #1078, opened 2025-03-11
Added kweights to gradient and precond. in direct_minimization.jl

elorannug:
`direct_minimization.jl` modified as per the meeting last week. This is so the new version can be tested on harder problems by Michael. The script below runs the new direct minimization. 
```julia
using DFTK, AtomsBuilder, PseudoPotentialData, Random
Random.seed!(42)
pseudopotentials = PseudoFamily("dojo.nc.sr.pbesol.v0_4_1.standard.upf")
model = model_DFT(bulk(:Si); functionals=PBEsol(), pseudopotentials)
kg = 5
basis = PlaneWaveBasis(model; Ecut=20, kgrid=[kg, kg, kg])

dir_min_run = direct_minimization(basis; tol = 1e-6)
```

### DIFF
--- a/src/scf/direct_minimization.jl
+++ b/src/scf/direct_minimization.jl
@@ -34,19 +34,20 @@ struct DMPreconditioner
     Nk::Int
     Pks::Vector # Pks[ik] is the preconditioner for k-point ik
     unpack::Function
+    kweights::Vector
 end
 function LinearAlgebra.ldiv!(p, P::DMPreconditioner, d)
     p_unpack = P.unpack(p)
     d_unpack = P.unpack(d)
     for ik = 1:P.Nk
-        ldiv!(p_unpack[ik], P.Pks[ik], d_unpack[ik])
+        ldiv!(p_unpack[ik], P.Pks[ik], ((P.kweights[ik])^-1)*d_unpack[ik])
     end
     p
 end
 function LinearAlgebra.dot(x, P::DMPreconditioner, y)
     x_unpack = P.unpack(x)
     y_unpack = P.unpack(y)
-    sum(dot(x_unpack[ik], P.Pks[ik], y_unpack[ik])
+    sum(dot(P.kweights[ik]*x_unpack[ik], P.Pks[ik], y_unpack[ik])
         for ik = 1:P.Nk)
 end
 function precondprep!(P::DMPreconditioner, x)
@@ -150,7 +151,7 @@ function direct_minimization(basis::PlaneWaveBasis{T};
             G = unsafe_unpack(G)
             for ik = 1:Nk
                 mul!(G[ik], ham.blocks[ik], ψ[ik])
-                G[ik] .*= 2*filled_occ
+                G[ik] .*= 2 * filled_occ * basis.kweights[ik]
             end
         end
         energies.total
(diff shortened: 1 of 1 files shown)


src/scf/direct_minimization.jl
```diff
@@ -34,19 +34,20 @@ struct DMPreconditioner
     Nk::Int
     Pks::Vector # Pks[ik] is the preconditioner for k-point ik
     unpack::Function
+    kweights::Vector
```
mfherbst, line 37: Isn't it easier to just include this into the `Pks` directly ? Maybe with a constructor
```julia
function DMPreconditioner(Pks, kweights, unpack)
   DMPreconditioner(length(Pks), Pks .* kweights, unpack)
end
```
or something similar ?

elorannug: Tldr: More elegant implementations are possible, but this works good enough for now.😅

Thanks for the quick response! I made the same modifications as I did in the code presented in meeting last week, before I had taken a more thorough look into the preconditioner. That way the results presented in the meeting should still hold in this PR. I also think the changes I made are pretty readable, which I think helps in debugging. 

Something like you propose here is definitely more elegant, but if you want to do it that way instead instead of my code, you'll have to be a little careful. The `Pks` you refer to from the `DMPreconditioner` struct is a `PreconditionerTPA` struct for every k-point by default:
```julia
mutable struct PreconditionerTPA{T <: Real}
    basis::PlaneWaveBasis
    kpt::Kpoint
    kin::AbstractVector{T}  # kinetic energy of every G
    mean_kin::Union{Nothing, Vector{T}}  # mean kinetic energy of every band
    default_shift::T  # if mean_kin is not set by `precondprep!`, this will be used for the shift
end
```
Since this struct contains the `basis`, in our overloaded functions for `ldiv!` and `dot`, we could do `P.Pks[ik].basis.kweights[ik]`. That way we would not have to store `kweights` in the `DMPreconditioner`. Doing it this way on my machine, the code converges to the same point, with the same number of iterations, and in the same amount of time.

The issue of having `P.Pks[ik].basis.kweights[ik]` in the overloads of `ldiv!` and `dot` vs adding it to the `precondprep!` is small for the purposes of debugging, but definitely doable. 

Note: If you were to do this, the code would not work for `prec_type=PreconditionerNone` anymore, so we would have to account for that as well.

Do you want me to do a new PR with any or all of these changes, or is it fine as it is for now?

mfherbst: Ah true, I forgot that here the Pks are not plain matrices. In that case what you propose is generally good, I think.

src/scf/direct_minimization.jl
```diff
@@ -34,19 +34,20 @@ struct DMPreconditioner
...
 end
 function LinearAlgebra.ldiv!(p, P::DMPreconditioner, d)
     p_unpack = P.unpack(p)
     d_unpack = P.unpack(d)
     for ik = 1:P.Nk
-        ldiv!(p_unpack[ik], P.Pks[ik], d_unpack[ik])
+        ldiv!(p_unpack[ik], P.Pks[ik], ((P.kweights[ik])^-1)*d_unpack[ik])
```
mfherbst, line 43: To avoid allocating memory here can you do:
```suggestion
        ldiv!(p_unpack[ik], P.Pks[ik], d_unpack[ik])
        ldiv!(P.kweights[ik], p_unpack[ik])
```

elorannug: Smart! Shall i add that to the PR or something? (I'm not well versed in contributing to professional public repos😅) I look forward to see how all the changes may impact the performance of `direct_minimization()` on your experiments.

mfherbst: Yes please !

PR #1080, opened 2025-03-11
FD test forces

mfherbst:
@Technici4n @antoine-levitt 

This is the failing test case I have been talking about.

It's still a bit messy, I'll try to clean it up later.

If I remove AtomicNonLocal from the terms all is good, so the error is almost certainly in that term.


antoine-levitt:
Doesn't get better for higher ecuts either

antoine-levitt:
Even fails when I hardcode lmax to 0

antoine-levitt:
Even fails when I set the lattice to be completely cubic!

antoine-levitt:
I don't understand, why should the finite difference energy be zero here? I don't see any obvious symmetries

PR #1081, opened 2025-03-13
Fix nonlocal forces and add test case

Technici4n:
Fix the forces issue from https://github.com/JuliaMolSim/DFTK.jl/issues/1066 and add a test case checking that `compute_forces` is indeed the partial derivative of `energy_ops`.

The issue was in `symmetrize_forces`. 3 years old, introduced in https://github.com/JuliaMolSim/DFTK.jl/commit/157dac8cb74ef72f38a48dd6aadfee1e818d3296 😄


test/forces.jl
```diff
@@ -159,3 +159,48 @@ end
...
+
+    function get_term_forces(lattice, atoms, positions; kgrid=[1,1,1], Ecut=4, symmetries=false, ε=1e-8)
+        model = model_DFT(lattice, atoms, positions; functionals=LDA(), symmetries, temperature=1e-3)
+        basis = PlaneWaveBasis(model; kgrid, Ecut)
+
+        scfres = self_consistent_field(basis; tol=1e-7)
+
+        map(1:length(basis.terms)) do iterm
+            test_atom = rand(1:length(model.atoms))
+            test_dir = randn(3)
```
mfherbst, line 174: rand is better
Also for MPI tests to succeed you need to distribute randomness across processors (see other tests)

mfherbst (commented):
Thanks !

test/forces.jl
Technici4n: Indeed, thanks!

test/testcases.jl
```diff
@@ -126,6 +126,21 @@ o2molecule = merge(o2molecule,
                    (; atoms=fill(ElementPsp(o2molecule.atnum,
                                             load_psp(o2molecule.psp_gth)), 2)))
 
+tio2 = (;
```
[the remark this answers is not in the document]
Technici4n: Done :) The extxyz format is not great, but it at least seems to support a `comment="whatever"` entry on the second line, to explain where the structure came from. I put the file in the `test/structures` subfolder following #1080.

PR #1082, opened 2025-03-14
Fix symmetries & symmetry-breaking with ForwardDiff

niklasschmitz:
It is finally time to fix symmetries. This PR will use 
 simple approach of just filtering out the symmetries that break the dual part also from the primal SCF. Later we can do better, by keeping primal symmetries for the primal and doing a partial unfold of the primal result.

- [x] Add failing test case
- [ ] Fix ForwardDiff rule for symmetries

### DIFF
--- a/test/forwarddiff.jl
+++ b/test/forwarddiff.jl
@@ -257,3 +257,28 @@ end
     derivative_fd = ForwardDiff.derivative(compute_force, 0.0)
     @test norm(derivative_ε - derivative_fd) < 1e-4
 end
+
+
+@testitem "Symmetry-breaking perturbation using ForwardDiff" #=
+    =#    tags=[:dont_test_mpi] setup=[TestCases] begin
+
+    function run_scf(ε)
+        v = ε * [0., 0., 0., 0., 0., 1.]
+        lattice = voigt_strain_to_full(v) * aluminium.lattice
+        model = model_DFT(lattice, aluminium.atoms, aluminium.positions;
+                    functionals=LDA(), temperature=1e-2, smearing=Smearing.Gaussian(),
+                    kinetic_blowup=BlowupCHV())
+        basis = PlaneWaveBasis(model; Ecut=5, kgrid=[2, 2, 2])
+        self_consistent_field(basis; tol=1e-10)
+    end
+
+    h = 1e-4
+    scfres0 = run_scf(0.)
+    scfres1 = run_scf(-h)
+    scfres2 = run_scf(+h)
+    δρ_finitediff = (scfres2.ρ - scfres1.ρ) / 2h
+
+    δρ = ForwardDiff.derivative(ε -> run_scf(ε).ρ, 0.)
+
+    @test norm(δρ - δρ_finitediff) * sqrt(scfres0.basis.dvol) < 1e-7
+end


mfherbst:
Related to #819

Technici4n:
I took a bit of time to polish and push what I did wrote a couple of months ago, should fix the issue!

The handling of atom displacements is suboptimal (you need to make the lattice vectors duals too) but that can be fixed later, and was already an issue before.

src/SymOp.jl
```diff
@@ -74,8 +74,10 @@ function check_group(symops::Vector; kwargs...)
             error("check_group: symop $s with inverse $(inv(s)) is not in the group")
         end
         for s2 in symops
-            if !is_approx_in_symops(s*s2) || !is_approx_in_symops(s2*s)
-                error("check_group: product is not stable")
+            if !is_approx_in_symops(s*s2)
+                error("check_group: product is not stable: $(s*s2) is not in the group")
+            elseif !is_approx_in_symops(s2*s)
+                error("check_group: product is not stable: $(s2*s) is not in the group")
```
niklasschmitz: Isn't this second case already covered by the double loop?

Technici4n: Indeed it is

src/workarounds/forwarddiff_rules.jl
```diff
@@ -71,12 +71,81 @@ end
...
+                                     tol_symmetry, kwargs...)
+    remove_broken_symmetries(lattice, atoms, positions, symmetries; tol_symmetry)
+end
+
+function remove_broken_symmetries(lattice, atoms, positions,
+                                  symmetries; tol_symmetry=SYMMETRY_TOLERANCE)
+    filter(symmetries) do symmetry
+        !is_symmetry_broken(lattice, atoms, positions, symmetry; tol_symmetry)
+    end
+end
```
mfherbst: Why don't you inline this function above ? It's basically exactly what the `filter` function is for ?

src/workarounds/forwarddiff_rules.jl
```diff
@@ -71,12 +71,81 @@ end
...
+
+"""
+Return `true` if a symmetry that holds for the primal part is broken by
+a perturbation in the lattice or in the positions, `false` otherwise.
+"""
+function is_symmetry_broken(lattice, atoms, positions, symmetry::SymOp; tol_symmetry)
```
mfherbst: Since this is specific to Duals and would not even work with non-duals (I think), I'd annotate the type of at least one argument to make that clear.

The name is perhaps also a little too generic. How about `symmetry_broken_by_dual` or sth like that.

mfherbst: Also this function feels like possible code duplication with https://github.com/JuliaMolSim/DFTK.jl/blob/forwarddiff-symmetry-breaking/src/symmetry.jl#L118 that one may be able to avoid.

Symmetries are already annoying enough, better avoid spreading out the logic too much in case we want to fix things / change things later.

src/workarounds/forwarddiff_rules.jl
```diff
@@ -71,12 +71,81 @@ end
...
+    # For any lattice atom at position x, W*x + w should be in the lattice.
+    # In cartesian coordinates, with a perturbed lattice A = A₀ + εA₁,
+    # this means that for any atom position xcart in the unit cell and any 3 integers m,
+    # there should be an atom at position ycart and 3 integers n such that:
```
mfherbst: `n` and `m` three integers is a bit unusual, one would rather expect a scalar integer for these (index). Better use another symbol.

src/workarounds/forwarddiff_rules.jl
```diff
@@ -71,12 +71,81 @@ end
...
+    # Wcart * (xcart + A*m) + wcart = ycart + A*n
+    # where
+    # - Wcart = A₀ * W * A₀⁻¹; note that W is an integer matrix
+    # - wcart = A₀ * w.
```
mfherbst: Dashes are always misleading in such cases (minus sign), better use `*` and indent.

src/workarounds/forwarddiff_rules.jl
```diff
@@ -71,12 +71,81 @@ end
...
+    # The strategy is then to check that:
+    # 1. A⁻¹ * Wcart * A is still an integer matrix (i.e. no dual part),
+    #    such that any change in m is easily compensated for in n.
+    # 2. The primal component of (*), i.e. with ε=0, is already known to hold.
+    #    Since n does not have a dual component, it is enough to check that
+    #    the dual part of the following is 0:
+    #      A⁻¹ * Wcart * A*x + A⁻¹ * wcart - y 
+
+    lattice_primal = ForwardDiff.value.(lattice)
+    W = inv(lattice) * lattice_primal * symmetry.W * inv(lattice_primal) * lattice
```
mfherbst: Just to avoid nasty issues later: This will not work for 1D or 2D settings. Best make an assertion somewhere or use the custom lattice inversion functions we have.

mfherbst: Don't like the variable names. You are doing something to a W and then you call it W ... potentially confusing.

src/workarounds/forwarddiff_rules.jl
```diff
@@ -71,12 +71,81 @@ end
...
+    for group in atom_groups
+        positions_group = positions[group]
+        for position in positions_group
+            # Find symmetric image; logic derived from symmetrize_forces
+            other_at = symmetry.W \ (ForwardDiff.value.(position) - symmetry.w)
+            smallest_deviation, i_other_at = findmin(positions_group) do at
+                δat = ForwardDiff.value.(at) - other_at
+                maximum(abs, δat - round.(δat))
+            end
+            @assert smallest_deviation < 10tol_symmetry
```
mfherbst: I recall this routine I had to tweak a few times already and it's a little heuristic, so further improvements may be necessariy. Again we should make sure to only have this code once in DFTK to avoid fixing it at one point, but not at another in the future ... these bugs are already spurious enough when they appear.

test/forwarddiff.jl
```diff
@@ -260,3 +260,150 @@ end
...
+    mirroryz = rotyz * rotyz
+    # Mirror y
+    mirrory = SymOp([1 0 0; 0 -1 0; 0 0 1], [0., 0., 0.])
+    # Translation by 0.5 in the x direction
+    transx = SymOp(diagm([1, 1, 1]), [0.5, 0., 0.])
+
+    is_approx_in(collection, s1) = any(s -> isapprox(s, s1), collection)
+
+    # Generate the full group
+    function complete_group(symops)
```
mfherbst: Do we not have such a function inside DFTK ... perhaps we should ?

test/forwarddiff.jl
```diff
@@ -260,3 +260,150 @@ end
...
+                    check(s*t)
+                    check(t*s)
+                end
+            end
+            if isempty(to_add)
+                return completed_group
+            end
+            append!(completed_group, to_add)
+        end
+        DFTK.check_group(completed_group)
```
mfherbst: Correct return value ?

test/forwarddiff.jl
mfherbst: could be useful for cases where one wants to build one's own symmetry group manually.

test/forwarddiff.jl
```diff
@@ -260,3 +260,150 @@ end
...
+        for s in expected_symmetries
+            @test is_approx_in(filtered_symmetries, s)
+        end
+        for s in filtered_symmetries
+            @test is_approx_in(expected_symmetries, s)
+        end
+    end
+
+    # Instantiate Dual to test with perturbations
+    ε = ForwardDiff.Dual{ForwardDiff.Tag{Nothing, Float64}}(0.0, 1.0)
```
mfherbst, line 303: You could be a little more nasty and use perturbations that are closer to the symmetry tolerance to test the limits.

mfherbst (commented):
Nice.

mfherbst:
> The handling of atom displacements is suboptimal (you need to make the lattice vectors duals too) but that can be fixed later, and was already an issue before.

Add broken test ?

src/workarounds/forwarddiff_rules.jl
Technici4n: I did it like that because the test calls `remove_broken_symmetries` directly. What we can do is inline `is_symmetry_broken` into `remove_broken_symmetries`, but I would rather not inline a complicated function inside a simple function.

src/workarounds/forwarddiff_rules.jl
Technici4n: Either the lattice or the atoms can have in principle have duals (potentially not even reflected in the type, e.g. `[0 0.5+epsilon 0]` will have `Vector{Any}` as the type). Other parts of the code assume that the lattice needs to have a dual (even if only the atoms are perturbed), but for this low-level building block it felt unnecessarily restrictive.

Agreed on trying to deduplicate the code with `symmetrize_forces`.

test/forwarddiff.jl
Technici4n: Potentially, but this is not really the goal of this test. Here I just want to ensure that we correctly filter out broken symmetries, not try to play with the heuristic thresholds.

src/workarounds/forwarddiff_rules.jl
Technici4n: Changed to `u` and `v`.

src/workarounds/forwarddiff_rules.jl
Technici4n: I think that `symmetry.w` is the primal part of `w`, so it's a bit tricky to name these variables in a different way. The logic is tricky, idk how much the variable names can really help there.

src/SymOp.jl
```diff
@@ -74,10 +74,41 @@ function check_group(symops::Vector; kwargs...)
...
-                error("check_group: product is not stable")
+            if !is_approx_in_symops(s*s2)
+                error("check_group: product is not stable: $(s*s2) is not in the group")
             end
         end
     end
     symops
 end
+
+is_approx_in(group, symop; kwargs...) = any(s -> isapprox(s, symop; kwargs...), group)
```
mfherbst: I would do it the other way round (see `occursin` function in Julia ... first argument is needle, the second is the haystack)

src/SymOp.jl
```diff
@@ -74,10 +74,41 @@ function check_group(symops::Vector; kwargs...)
...
+            error("Could not complete group in $maxiter iterations")
+        end
+        to_add = []
+        function check(s1)
+            if !is_approx_in(completed_group, s1; kwargs...) && !is_approx_in(to_add, s1; kwargs...)
+                push!(to_add, s1)
+            end
+        end
+        # Identity always needs to be there!
+        check(SymOp(diagm([1, 1, 1]), [0., 0., 0.]))
```
mfherbst: `one(SymOp)`

src/SymOp.jl
```diff
@@ -74,10 +74,41 @@ function check_group(symops::Vector; kwargs...)
...
+
+function complete_symop_group(symops; maxiter=10, kwargs...)
+    completed_group = Vector(symops)
+
+    for it = 1:maxiter
+        if it == maxiter
+            error("Could not complete group in $maxiter iterations")
+        end
+        to_add = []
+        function check(s1)
```
mfherbst: Don't like the closure on `to_add` and implicit state change here, I'd move the function definition out of the for loop.
Regarding the name: I'd say the main effect is adding the missing elements and not checking something.

src/workarounds/forwarddiff_rules.jl
```diff
@@ -71,12 +71,75 @@ end
...
+    # The strategy is then to check that:
+    # 1. A⁻¹ * Wcart * A is still an integer matrix (i.e. no dual part),
+    #    such that any change in u is easily compensated for in v.
+    # 2. The primal component of (*), i.e. with ε=0, is already known to hold.
+    #    Since v does not have a dual component, it is enough to check that
+    #    the dual part of the following is 0:
+    #      A⁻¹ * Wcart * A*x + A⁻¹ * wcart - y 
+
+    lattice_primal = ForwardDiff.value.(lattice)
+    W = compute_inverse_lattice(lattice) * lattice_primal * symmetry.W * compute_inverse_lattice(lattice_primal) * lattice
```
mfherbst: Line length

PR #1085, opened 2025-04-14
Update tutorial.jl

m-schwendt:
The way it was written corresponded to a slice along the y-axis instead of the x-axis.


mfherbst:
Thanks !

mfherbst:
Thanks for the feedback, much appreciated !

PR #1088, opened 2025-04-30
Reduce RAM usage of nonlocal term

### LINKED ISSUE #1032: Stress computation can use a lot of RAM
I was running a computation whose basis alone takes 80 Gb of RAM, most of it being used by the `P` matrices in the nonlocal term.

Since `compute_stresses_cart` computes the gradient with ForwardDiff, it needs 7x as much additional memory (primal part + 6 dual parts). Unfortunately, 80 x 7 Gb was too much memory for me so my process got killed. :(

Could there be a way to reduce the memory usage of stress computations without sacrificing (too much) performance?

Technici4n:
Still super super WIP.

The problem lies with how the `P` matrices are stored in the nonlocal term. Each column of the `P` matrix corresponds to a projector (and each psp usually has multiple projectors). Multiple atoms with the same psp have the same projectors except for a structure factor $e^{-(G+k)\cdot r_{\text{atom}}}$.

The basic idea in this PR is to store the projectors without the phase factor, and apply the structure factor on-the-fly. The cleanest way to do it that I could find is to use a custom matrix-like type for `P` such that from the outside `P` still looks like a dense matrix.

Not sure yet what the best way to apply the structure factor on-the-fly would be. Either we use a temporary vector to hold `ψ[:,iband] .* structure_factor` or we write a loop but then the question is whether it will work on the GPU.

Fixes #1032.

TODO:
- [ ] Actually fix the issue and make the code work.
- [ ] Check that the CPU performance and memory allocations are still fine.
- [ ] Check that the memory usage went down.
- [ ] Check that the GPU performance is still fine.


Technici4n:
Here is an example showing the problem:

If we have oxygen from PseudoDojo there are 13 projectors. Now assume we perform a computation with roughly $10^4$ plane waves per $k$-point, and with $10^3$ $k$-points.

Each `P` matrix would have $13 \cdot 10^4$ entries. With $10^3$ such matrices (one per $k$-point) and considering that each entry is a complex number which requires 16 bytes, this gives a total memory usage of roughly 2 gigabytes.

If we have the same setting but 8 oxygen atoms the memory usage goes to 16 gigabytes. With this PR, the goal is that it should remain around 2 gigabytes.

antoine-levitt:
So this is a performance memory tradeoff. I did it this way because then you use full blas3 operations. I remember something like this gave a huge boost on abinit, but possibly it was blas3ifying the operation on all the bands at once that gave the performance boost (as opposed to band by band as it was done before). So I'm not opposed to switching but keep in mind the performance implications and benchmark it.

antoine-levitt:
See https://docs.abinit.org/variables/dev/#use_gemm_nonlop

Technici4n:
Nice reference, maybe that lets me find how abinit does it. We also need to be careful not to destroy GPU performance. 😓

antoine-levitt:
If you want to look, it's in the opernla routine (I worked on that code a long time ago, it still haunts my dreams)

mfherbst (commented):
Nice. Yeah the BLAS 2  versus BLAS 3 tradeoff is something to be understood a little better by benchmarks before we commit on one way to do this.

src/terms/nonlocal.jl
```diff
@@ -85,10 +87,10 @@ function LinearAlgebra.mul!(C::AbstractMatrix, A::NonlocalProjectors, B::Abstrac
     ψ_scratch = A.ψ_scratch
     for p in A.projectors
         for proj in eachcol(p.projectors)
-            # TODO: what allocates here? and does it use BLAS?
+            # TODO: does this use BLAS?
             ψ_scratch .= p.structure_factors .* proj
-            for iband in size(B, 2)
-                C[:, iband] .+= ψ_scratch .* (α * B[iproj, iband])
+            for iband in axes(B, 2)
+                @views C[:, iband] .+= ψ_scratch .* (α * B[iproj, iband])
```
mfherbst: 5-argument `mul!` ?

test/PspUpf.jl
```diff
@@ -250,3 +250,42 @@ end
         end
     end
 end
+
+@testitem "Test nonlocal term operations" tags=[:psp] setup=[mPspUpf] begin
```
Technici4n, line 254: Not sure about this test. The goal was just for me to be able to try my changes with <5s waiting time.

src/terms/nonlocal.jl
```diff
@@ -282,6 +381,7 @@ function PDPψk(basis, positions, psp_groups, kpt, kpt_minus_q, ψk)
     D = build_projection_coefficients(basis, psp_groups)
     P = build_projection_vectors(basis, kpt, psp_groups, positions)
     P_minus_q = build_projection_vectors(basis, kpt_minus_q, psp_groups, positions)
+    # TODO: probably needs an extra parenthesis to first compute P'ψ
```
Technici4n, line 384: So... I noticed that Julia has custom `*` overloads for matrix multiplication with more than 2 operands to select the chain of operations that will minimize the total cost. Presumably it will almost always compute `P_minus_q' * ψk` first. But if it doesn't we are in trouble, so this should probably be changed to `(P_minus_q' * ψk)`.

src/terms/nonlocal.jl
```diff
@@ -144,6 +144,110 @@ function build_projection_coefficients(T::Type, psp::NormConservingPsp)
...
+allocating the full matrix.
+This type extends AbstractMatrix, but it does not implement all
+the required methods, only those that were shown to be needed.
+In particular, random access to the matrix elements is not supported.
+"""
+struct NonlocalProjectors{T <: Real,
+                          ST <: AbstractVector{Complex{T}},
+                          PT <: AtomProjectors,
+                         } <: AbstractMatrix{Complex{T}}
+    # TODO: this is a real problem wrt. thread-safety, no?
```
Technici4n, line 170: How bad is this? `DftHamiltonianBlock` should handle it fine, but `GenericHamiltonianBlock` seems to be parallelizing over bands which will cause problems!

src/terms/nonlocal.jl
```diff
@@ -144,6 +144,126 @@ function build_projection_coefficients(T::Type, psp::NormConservingPsp)
...
+# Add a level of indirection here to avoid ambiguity with the mul! method provided by Julia.
+LinearAlgebra.mul!(C::AbstractVector, A::Adjoint{<:Any, <:NonlocalProjectors},
+                   ψk::AbstractVector) = _mul!(C, A, ψk)
+LinearAlgebra.mul!(C::AbstractMatrix, A::Adjoint{<:Any, <:NonlocalProjectors},
+                   ψk::AbstractMatrix) = _mul!(C, A, ψk)
+
+LinearAlgebra.mul!(C::AbstractVector, A::NonlocalProjectors, B::AbstractVector,
+                   α::Number, β::Number) = _mul!(C, A, B, α, β)
+LinearAlgebra.mul!(C::AbstractMatrix, A::NonlocalProjectors, B::AbstractMatrix,
+                   α::Number, β::Number) = _mul!(C, A, B, α, β)
```
mfherbst, line 216: Why is this needed ? Surprises me a bit.

src/terms/nonlocal.jl
```diff
@@ -144,6 +144,126 @@ function build_projection_coefficients(T::Type, psp::NormConservingPsp)
...
+
+function _mul!(C::AbstractVecOrMat, A::Adjoint{<:Any, <:NonlocalProjectors},
+               ψk::AbstractVecOrMat)
+    if size(C, 1) != size(A, 1) || size(A, 2) != size(ψk, 1) || size(ψk, 2) != size(C, 2)
+        throw(DimensionMismatch(lazy"A has size $(size(A)), B has size $(size(ψk)), C has size $(size(C))"))
+    end
+
+    iproj = 1
+    for at in A.parent.atoms
+        for proj in eachcol(at.projectors)
```
mfherbst, line 226: join these loops to avoid too deep nesting.

src/terms/nonlocal.jl
```diff
@@ -144,6 +144,126 @@ function build_projection_coefficients(T::Type, psp::NormConservingPsp)
...
+        nproj = size(at.projectors, 2)
+
+        Pwork = @view Pbuffer[:, 1:nproj]
+        Pwork .= at.structure_factors .* at.projectors
+        Bwork = if BT <: AbstractVector
+            @view(B[iproj:iproj+nproj-1])
+        else
+            @view(B[iproj:iproj+nproj-1, :])
+        end
+        mul!(C, Pwork, Bwork, α, 1)
```
mfherbst, line 261: why not put `β` here ?

src/terms/nonlocal.jl
```diff
@@ -144,6 +144,126 @@ function build_projection_coefficients(T::Type, psp::NormConservingPsp)
...
+    for at in A.atoms
+        nproj = size(at.projectors, 2)
+
+        Pwork = @view Pbuffer[:, 1:nproj]
+        Pwork .= at.structure_factors .* at.projectors
+        Bwork = if BT <: AbstractVector
+            @view(B[iproj:iproj+nproj-1])
+        else
+            @view(B[iproj:iproj+nproj-1, :])
+        end
```
mfherbst, line 260: this looks weird.

Technici4n:
Yes, I am not sure of the current implementation, it's quite complex.

We can additionally save quite a bit of RAM if we allow ourselves to recompute the spherical harmonics on-the-fly. In that case we will only need to store one form factor per angular moment component l instead of 2l+1. For PD Silicon this would bring the number of projector form factors that need to be stored from 18 down to 6.

PR #1090, opened 2025-05-07
Fix occupation insulator

### LINKED ISSUE #1089: Float instability of Fermi-level derivative of Chi0 for temperature & large band gap
This came up now with me computing DFPT for various structures (metals, semiconductors, insulators) with temperature=1e-3 irrespective of the structure. It worked fine on Al and Si, but NaCl (salt) failed once it got to the response solver. Here's a cooked down self-contained reproducer:
```julia
using DFTK
using PseudoPotentialData
using ForwardDiff
using LinearAlgebra
DFTK.setup_threading()

pseudopotentials = PseudoFamily("dojo.nc.sr.pbe.v0_5.standard.upf")
atoms = [
    ElementPsp(:Na, pseudopotentials),
    ElementPsp(:Cl, pseudopotentials),
]
positions = [[0.0, 0.0, 0.0], [0.5, 0.5, 0.5]]

a = 10.92
lattice = (a/2) * [0 1 1; 1 0 1; 1 1 0]
Ecut = 44

function compute_stress(strain)
    new_lattice = (1 + strain) * lattice
    model = model_DFT(new_lattice, atoms, positions;
                      functionals=PBE(), kinetic_blowup=BlowupCHV(),
                      temperature=1e-3, smearing=Smearing.Gaussian())
    basis = PlaneWaveBasis(model; Ecut, kgrid=(4, 4, 4))
    scfres = self_consistent_field(basis; tol=1e-9)
    stress = compute_stresses_cart(scfres)
    tr(stress)
end

@show compute_stress(0.)  # This works fine
@show ForwardDiff.derivative(compute_stress, 0.)  # This gives NaN in response solver
```
Tracking this NaN down, I think I found the issue in chi0.jl:
https://github.com/JuliaMolSim/DFTK.jl/blob/fa29008faa5402a6a4a3d865a872d99f140a72e8/src/response/chi0.jl#L277

For an insulator with a large band gap (such as NaCl with ~5.0 eV at PBE), we get **numerical underflow** in all the occupation derivatives, which in this case indeed leads to exact zeros `δocc_tot = 0.0` and `D = 0.0` thus `δocc_tot / D = NaN` .

It seems the quickest fix is just to check `iszero(D)` before that division. What I am not sure about is whether returning zero as the Fermi level derivative in this case is even valid. 

With fixed temperature, but increasing band gap the Fermi level should be converging to the center of the gap, right? That would mean that it should asymptotically observe half the derivative of the gap. Does this make sense?

mfherbst:
Fix #1089 by making sure large band gap + temperature returns a sane Fermi level and the insulator branches are taken for chi0 computations.

TODO:
- [x] Convert issue example into test within the response calculation tests.

### DIFF
--- a/src/Model.jl
+++ b/src/Model.jl
@@ -223,15 +223,17 @@ the contained parameters. This constructor is useful for changing the data
 type in the model or for changing `lattice` or `positions` in
 geometry/lattice optimisations.
 
-!!! warning
+!!! warning "Symmetries need to be updated manually."
     Unless a `symmetries` keyword argument is given
     this constructor copies the symmetries from `model`.
     This means if you update the `positions`, `lattice`, `atoms`
     or `terms` in a way that they *break* some of the original symmetries,
     you should either also supply matching `symmetries` or specify
     `symmetries=true`, which will re-determine the symmetries corresponding
-    to the new structure. In the case of spin do also re-supply the appropriate
-    `magnetic_moments` of the atoms.
+    to the new structure. In the case of spin do also re-supply the
+    appropriate `magnetic_moments` of the atoms.
+
+    This behaviour is slated to change in an upcoming breaking release.
 """
 function Model{T}(model::Model;
                   lattice::AbstractMatrix=model.lattice,
@@ -240,6 +242,11 @@ function Model{T}(model::Model;
                   symmetries=model.symmetries,
                   magnetic_moments=T[],
                   kwargs...) where {T <: Real}
+    # TODO In an upcoming breaking release come up with a way to store the symmetry-equivalent
+    #      atoms in the model (including the effect magnetic moments), such that if the atoms
+    #      do not change (the usual case) the default symmetry determination routine can still
+    #      work with updated positions without the need to re-provide the magnetic_moments.
+
     Model(T.(lattice), atoms, positions;
           model.model_name,
           model.n_electrons,
--- a/src/Model.jl
+++ b/src/Model.jl
@@ -218,25 +218,38 @@ end
     Model(model; [lattice, positions, atoms, kwargs...])
     Model{T}(model; [lattice, positions, atoms, kwargs...])
 
-Construct an identical model to `model` with the option to change some of the contained
-parameters. This constructor is useful for changing the data type in the model
-or for changing `lattice` or `positions` in geometry/lattice optimisations.
+Construct an identical model to `model` with the option to change some of
+the contained parameters. This constructor is useful for changing the data
+type in the model or for changing `lattice` or `positions` in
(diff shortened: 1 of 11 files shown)


src/Model.jl
```diff
@@ -218,25 +218,38 @@ end
...
 
-Construct an identical model to `model` with the option to change some of the contained
-parameters. This constructor is useful for changing the data type in the model
-or for changing `lattice` or `positions` in geometry/lattice optimisations.
+Construct an identical model to `model` with the option to change some of
+the contained parameters. This constructor is useful for changing the data
+type in the model or for changing `lattice` or `positions` in
+geometry/lattice optimisations.
+
+!!! warning
```
antoine-levitt, line 226: We should definitely do that by default and be able to turn it off via a flag

src/occupation.jl
```diff
@@ -122,60 +136,68 @@ function compute_fermi_level(basis::PlaneWaveBasis{T}, eigenvalues, ::FermiTwoSt
...
     end
 
     εF
 end
 
+
+"""
+Estimate a Fermi level by assuming (a) integer occupation and (b) an equal number of bands
+is filled for each k-point. This is largely the setting for occupation without temperature.
+Note, that while this function can be used in cases with spin or with temperature, there
```
antoine-levitt, line 174: Remove first comma

antoine-levitt: (or rather move it to "note that, while")

src/occupation.jl
```diff
@@ -122,60 +136,68 @@ function compute_fermi_level(basis::PlaneWaveBasis{T}, eigenvalues, ::FermiTwoSt
...
+is no guarantee that the Fermi-level estimated by this function *is* the actual Fermi level.
+Therefore this function should generally only be used as a starting point for other routines.
+"""
+function guess_fermi_level_intocc_(basis::PlaneWaveBasis, eigenvalues)
```
antoine-levitt, line 178: guess_fermi_level_insulator?

src/response/chi0.jl
```diff
@@ -246,6 +246,26 @@ function compute_αmn(fm, fn, ratio)
     ratio * fn / (fn^2 + fm^2)
 end
 
+function is_insulator(basis::PlaneWaveBasis, eigenvalues, εF::T;
```
antoine-levitt, line 249: "effective insulator"?

antoine-levitt (commented):
Looks good, thanks!

src/Model.jl
mfherbst: Agree. But it's breaking and more work than what it looks (due to the magnetic moments), so let's do it later.

PR #1094, opened 2025-05-14
Proposal for badly failing cases

mfherbst:
I know these failures should not happen (as the error message says :smile:), but they do ... and unfortunately often in longer-running workflows where it's not super easy to figure out why and further it's super annoying that the entire computation crashes because of this.

Here is one proposal to deal with it ... namely just fall back to an expensive, but hopefully accurate orthogonalisation method. One could also implement throwing some special exception, which can then be handled higher up (e.g. by perhaps aborting the LOBPCG at the current state and flagging non-convergence). Not sure, what is better.

CC @antoine-levitt


antoine-levitt:
Sure, that makes a lot of sense. My guess is if that routine is failing some other part will fail further down the line, so maybe always print the warning, but I agree it's better to keep going

PR #1097, opened 2025-05-16
GPU optimized symmetry operations

abussy:
When simulating large systems with symmetries, the symmetrization of the density is extremely slow. This is because the array is first transferred to the CPU, before a double loop over symmetries and G vectors takes place. By comparison to loops running on the GPU, this is slow, and the operation becomes a major bottleneck.

This PR introduces GPU specific implementations for `accumulate_over_symmetries!` and `lowpass_for_symemtry!`, using the `map!` construct to run on the device, and thus saving data transfers. The loop itself is also greatly accelerated. Note that these new implementations do run on the CPU, but slower than the current solution.

For illustration, using the input below, I observe these timings (seconds):

| GPU | master | this PR |
| --- | --- |---|
| A100|  195 | 14 |
| H100 | 135 | 9 |
| Mi200 | 212 | 26|

```
using DFTK
using PseudoPotentialData
using CUDA
# using AMDGPU
setup_threading(;)

arch = DFTK.GPU(CuArray)
# arch = DFTK.CPU(ROCArray)

Ecut = 32.0
kgrid = [2, 1, 1]
maxiter = 5
tol = 1.0e-8
temperature = 1e-4

factor = 4
a = 3.8267
                      [1.0 0.0 1.0];
                      [1.0 1.0 0.0]]
end

model = model_DFT(lattice, atoms, positions; temperature=temperature,

# warmup
scfres = self_consistent_field(basis; maxiter=3, tol=tol);

#actual calculations
DFTK.reset_timer!(DFTK.timer)
scfres = self_consistent_field(basis; maxiter=maxiter, tol=tol);
@show DFTK.timer
```


src/gpu/symmetry.jl
```diff
@@ -0,0 +1,47 @@
...
+    ρtmp = similar(ρ)
+
+    for symop in symmetries
+        isone(symop) && continue
+        map!(ρtmp, ρ, Gs) do ρ_i, G
+            idx = index_G_vectors(fft_size, symop.S * G)
+            isnothing(idx) ? zero(complex(T)) : ρ_i
+        end
+        ρ .= ρtmp
+    end
```
mfherbst, line 45: This feels quite wasteful. Is there no way to somehow do the loop over symmetries inside, such that not for each  symmetry the entire ρ array is copied ?

src/gpu/symmetry.jl
```diff
@@ -0,0 +1,47 @@
...
+        if isone(symop)
+            ρaccu .+= ρin
+            continue
+        end
+
+        invS = Mat3{Int}(inv(symop.S))
+        map!(ρtmp, Gs) do G
+            idx = index_G_vectors(fft_size, invS * G)
+            isnothing(idx) ? zero(complex(T)) : cis2pi(-T(dot(G, symop.τ))) * ρin[idx]
+        end
```
mfherbst, line 24: Apart from these 5 lines the function is basically identical to the CPU version. I'm wondering if it's not better to just duplicate these 5 lines into a GPU version as opposed to the entire `accumulate_over_symmetries!` function.

src/gpu/symmetry.jl
mfherbst: Essentially your goal is to filter out those G that cannot be generated by the symmetry operations.
Perhaps one could also think about computing (and storing) such a mask in a basis ? Than this operation would be essentially free on the GPU.

Of course depends how relevant this is (i.e. how many times is it called, how much of a bottle neck is it etc. )

mfherbst (commented):
Thanks, some thoughts.

src/gpu/symmetry.jl
abussy: > This feels quite wasteful. Is there no way to somehow do the loop over symmetries inside, such that not for each symmetry the entire ρ array is copied ?

This is a really good point, as this allows for a single kernel launch and less copies. I played around with this idea, and I managed to get some significant speedups:

| GPU | master | original PR | new loop order |
| --- | --- |---| --- |
| A100|  195 | 14 | 8.5 |
| H100 | 135 | 9 | 6.5 |

It seems however that `AMDGPU.jl` is quite more restrictive than `CUDA.jl` as for what it accepts within a `map!`. I am yet to find a generic implementation.

As of now, I am not sure precomputing and storing masks in the `PlaneWaveBasis` is worth the code complications. Especially after switching the order of the loops, the symmetrization does not appear to be a bottleneck.

abussy:
Changing the order of the loops reduces the number of kernel calls and large copies. As a result, performance is greatly improved:

| GPU | master | original PR | new loop order |
| --- | --- |---| --- |
| A100|  195 | 14 | 8.5 |
| H100 | 135 | 9 | 6.5 |
| Mi200 | 212 | 26 | 24 |

Note that this introduces some level of clunkiness.  For example, all data accessed within the `map!` mus be `isbit`. Therefore, various fields of the symmetry vector must first be copied in their own arrays and transferred to the GPU. The cost of these actual copies is negligible due to the small size of the data and the low frequency of the transfer (I compared timings to a case were I did these transfer once and for all, for peace of mind).

src/gpu/symmetry.jl
abussy: It might not be so relevant anymore with the latest commit. The functions look more different now.

src/gpu/symmetry.jl
```diff
@@ -0,0 +1,46 @@
...
+    ρtmp = similar(ρ)
+
+    symm_S = to_device(basis.architecture, [symop.S for symop in symmetries])
+
+    map!(ρtmp, ρ, Gs) do ρ_i, G
+        acc = ρ_i
+	    for S in symm_S
+            idx = index_G_vectors(fft_size, S * G)
+            acc *= isnothing(idx) ? zero(complex(T)) : one(complex(T))
+        end
```
mfherbst, line 41: Formatting off.

src/gpu/symmetry.jl
```diff
@@ -0,0 +1,46 @@
+using GPUArraysCore
+
+function accumulate_over_symmetries!(ρaccu::AbstractArray, ρin::AbstractArray,
+    basis::PlaneWaveBasis{T}, symmetries) where {T}
```
mfherbst, line 4: Formatting

mfherbst: Also, how much worse is this implementation compared to the CPU version ? Can we not just make this the CPU version, too ? It looks like it should not be very much worse than what we have.

src/gpu/symmetry.jl
```diff
@@ -0,0 +1,46 @@
...
+            τ = symm_τ[i_symm]
+            idx = index_G_vectors(fft_size, invS * G)
+            acc += isnothing(idx) ? zero(complex(T)) : cis2pi(-T(dot(G, τ))) * ρin[idx]
+        end
+        acc
+    end
+    ρaccu
+end
+
+function lowpass_for_symmetry!(ρ::AbstractGPUArray, basis::PlaneWaveBasis{T};
```
mfherbst, line 26: The only place where we need this is in `symmetrize_ρ`, where it comes right after `accumulate_over_symmetries!`. I think for the GPU version it would make a lot of sense to fuse these two functions into one and thus have a single `map!` going over all `Gs`. This you should be able to do with a boolean flag `do_lowpass`, which hopefully Julia is smart enough to constant-prop into the GPU kernel and fully compile away if set to `false`.

Again feel free to make this fused function also the CPU function if this does not hurt performance too much.

mfherbst (commented):
Nice. Your current implementation, spiked another idea: Fuse both functions. If it's too complicated don't bother, but I think it should work.

src/gpu/symmetry.jl
mfherbst: I agree. Avoiding code duplication is more important here. Also it's not all too bad. Just keep the appropriate comments to make sure it's clear that the GPU is the reason why things are done in a certain way.

src/symmetry.jl
```diff
@@ -269,49 +269,66 @@ function apply_symop(symop::SymOp, basis, ρin; kwargs...)
...
         end
-    end  # symop
+        acc
+    end
     ρaccu
 end
 
 # Low-pass filters ρ (in Fourier) so that symmetry operations acting on it stay in the grid
+# This function is optimized for CPU and GPU.
 function lowpass_for_symmetry!(ρ::AbstractArray, basis; symmetries=basis.symmetries)
```
mfherbst, line 308: Since this allocates in all cases: Can we call this `lowpass_for_symmetry` and make it return the `ρ` instead of in-place modifying it ?

src/symmetry.jl
```diff
@@ -269,49 +269,66 @@ function apply_symop(symop::SymOp, basis, ρin; kwargs...)
...
+    ρtmp = similar(ρ)
+
+    symm_S = to_device(basis.architecture, [symop.S for symop in symmetries])
+
+    # Perfrorms the following loop in an optimized way for CPU and GPU:
+    # for (ig, G) in enumerate(G_vectors(basis))
+    #     if index_G_vectors(basis, symop.S * G) === nothing
+    #         ρ[ig] = 0
+    #     end
+    # end
```
mfherbst: I think that's clear without the additional comment. I would only say it's done like this to be more performant on GPUs and CPUs.

src/symmetry.jl
abussy: Sure. It would also cascade down to `enforce_real!` that would become `enforce_real` for consistency. I have nothing against it.

abussy: It is actually possible to make `lowpass_for_symmetry!` non-allocating by passing the density to `map!` both as the source and the destination. This is only possible because each element of the destination only depends on the corresponding element of the source (i.e. it would not be possible in `accumulate_over_symmetries`, which is why I didn't think of it initially, I guess). Last commit addresses this.

mfherbst: This is great, thank you !

mfherbst:
The `G_vectors_generator` has always been a hack to get the symmetrisation fast. Since it's now done differently and this function no longer needed, we can completely remove it.

Yeah I agree the test does not really make sense here any more. Is it still worth it to test that `index_G_vectors` is properly inlined in the new order of the loops ?

PR #1098, opened 2025-05-16
Inexact Krylov methods and adaptive response

mfherbst:
Will superseed #1027 and #737:

- [x] Preconditioned GMRES implementation ok ?
- [x] Factor of 2 next to fn in Bonan's implementation ?
- [x] Balanced strategy: For the orbital term lower bound we should not use the *total* Nocc
- [x] Think about what happens if we use q ≠ 0. Are things ok ?
- [x] Testing: A more challenging test case (e.g. aluminium ?)
- [x] Use default SCF mixing as preconditioner
- [x] Cleanup code a bit more
- [x] Nicer print formatting
- [x] Example
- [x] Faster than before ?


src/response/gmres.jl
```diff
@@ -0,0 +1,146 @@
+using LinearMaps
+using KrylovKit: OrthonormalBasis, Orthogonalizer
+using LinearAlgebra: Givens
+
+# Perform an inexact matrix-vector product, ensuring that
+# maximum(norm, eachcol(Y - A*X)) < tol
```
antoine-levitt: Why infinity norm here?

src/response/gmres.jl
```diff
@@ -0,0 +1,146 @@
...
+inexact_mul(A, x; tol=0.0) = (; Ax=A * x, info=(; tol))
```
antoine-levitt: Name is a bit weird... mul_with_tolerance? Maybe say this is meant to be overloaded for specific types, because this looks like a helper function and it's not clear how it's meant to be used.

src/response/gmres.jl
```diff
@@ -0,0 +1,146 @@
...
+            comment = "Restart"
+        end
+        @printf("%4i  %6i  %6i  %15.8g  %15s\n", n_iter, length(info.restart_history),
+                info.k, info.resid_history[n_iter], comment)
+    end
+    info
+end
+
+@doc raw"""
+Preconditioned Inexact GMRES algorithm as discussed in [arxiv 2505.02319](https://arxiv.org/pdf/2505.02319).
```
antoine-levitt: Add high-level description?

antoine-levitt: move file to inexact_gmres.jl (to make clear it's not just a reimplementation of gmres)

src/response/gmres.jl
antoine-levitt: OK gotcha, sorry. Maybe "performs an inexact matrix-matrix product, ensuring that each matvec is accurate up to tol"?

src/response/gmres.jl
```diff
@@ -0,0 +1,146 @@
...
+At convergence ``\| \text{precon}^{-1} (A*x - b) \| < \text{tol}`` is ensured.
+
+Overview of parameters:
+- **krylovdim**: Maximal Krylov subspace dimension before restart
+- **tol**:  Convergence threshold
+- **s**: Initial guess for the smallest singular value of the upper Hessenberg matrix.
```
antoine-levitt: separate which criterion is "standard" and which is because of the inexact?

src/response/gmres.jl
```diff
@@ -0,0 +1,146 @@
...
+
+        while (n_iter < maxiter && k < m)  # Arnoldi loop
+            n_iter += 1
+
+            # Push new Krylov vector
+            push!(V, r / β)
+            k = length(V)
+
+            # Compute new Krylov vector and orthogonalise against subspace
+            tolA = tol * s / (3m * abs(y[k]))  # |y[k]| is the estimated residual norm
```
antoine-levitt: criteria based on the number of vectors are usually quite pessimistic (but I didn't do any tests here)

antoine-levitt: So basically this is a relative tolerance of `tol*s/(3m)`. Is this really worth the hassle having it depend on the iterations, compared to simply having the matvec perform using a relative tolerance of `tol/10` or something like this? Also maybe it would be clearer to have the `tol` in `inexact_matvec` be a relative tolerance? (relative to `b`)

src/response/gmres.jl
antoine-levitt: Is this a fork of an existing solver (krylovkit's?), with the inexact_matvec spliced in? If so maybe say it.

src/response/hessian.jl
```diff
@@ -159,53 +159,158 @@ that is return δψ where (Ω+K) δψ = rhs.
...
+Input parameters:
+- `tol`: Desired tolerance in the density variation and orbital variation
+- `mixing`: Mixing to use to precondition GMRES
+- `maxiter_sternheimer`: Maximal number of iterations for each of the Sternheimer solvers
+- `maxiter`: Maximal number of iterations for the Dyson equation solver
+   convergence, it will automatically stop.
+- `krylovdim`: Maximal Krylov subspace dimension in the Dyson equation solver
+- `s`: Initial guess for the smallest singular value of the upper Hessenberg matrix
+   in the Dyson equation solver.
+- `bandtolalg`: Algorithm for adaptive selection of Sternheimer tolerances
```
antoine-levitt, line 241: bandtol_alg?

src/response/hessian.jl
```diff
@@ -217,15 +322,66 @@ using a split algorithm, where `rhs` is typically
...
                        scfres.εF, scfres.eigenvalues, rhs;
-                       scfres.occupation_threshold, kwargs...)
+                       scfres.occupation_threshold, mixing,
+                       bandtolalg=BandtolBalanced(scfres), kwargs...)
+end
+
+# TODO This feels a bit hackish, because for all other entities
+#      (like Ω+K or χ0 or K) we don't have a struct, but just an apply function
+#      so why have a struct here ?
+struct DielectricAdjoint{Tρ, Tψ, Toccupation, TεF, Teigenvalues, Tq}
```
antoine-levitt, line 351: This is because inexact_mul is with a struct. You could also have inexact_mul be a function (ie gmres(inexact_mul, ...)). I don't have a strong opinion one side or the other.

src/response/gmres.jl
mfherbst: >  criteria based on the number of vectors are usually quite pessimistic (but I didn't do any tests here)

I have to admit we did not really question this, it comes out of the original Simoncini paper (m = krylovdim, so it's pretty much 10 or 20)

> compared to simply having the matvec perform using a relative tolerance of tol/10

Well, but this is exactly the point: Due to the factor `1 / y[m]` you can be cruder in later iterations and save between 30% and 40% of the cost.

> Also maybe it would be clearer to have the tol in inexact_matvec be a relative tolerance? (relative to b)

Why relative to `b` ? I guess one could argue the `tol` in `inexact_gmres` could be relative to `b`, but `mul_approximate` does not really know anything of `b`

src/response/hessian.jl
mfherbst: We don't really do this anywhere else (e.g. `fermialg`, `diagtolalg`). We can discuss this, but then it should be changed everywhere.

antoine-levitt: I'd argue fermialg is fine but diagtolalg should be diagtol_alg. bandtolalg is pretty hard to read (my brain keeps parsing it as band to lalg). But it's fine anyway don't bother.

src/response/gmres.jl
antoine-levitt: > I have to admit we did not really question this, it comes out of the original Simoncini paper (m = krylovdim, so it's pretty much 10 or 20)

Yeah, I'd just use a fudge factor here instead of depending on m, but it doesn't matter too much.

> Well, but this is exactly the point: Due to the factor 1 / y[m] you can be cruder in later iterations and save between 30% and 40% of the cost.

atol = tol/|y| but the rhs being matveced is of size |y|, so rtol = tol, no?

> Why relative to b ? I guess one could argue the tol in inexact_gmres could be relative to b, but mul_approximate does not really know anything of b

Sorry I meant wrt the vector currently being matvec'ed

mfherbst: > atol = tol/|y| but the rhs being matveced is of size |y|, so rtol = tol, no?

Not quite, actually `V[k]` always has norm `1` (line 106), but maybe I don't understand ?

> Sorry I meant wrt the vector currently being matvec'ed

Ah ok. I made the change, but I'm not sure this simplifies things. Anyway the relative and absolute tolerance business in the code is quite messy in the chi0 context. We should think about that carefully at some point and clean it up. In particular when dealing with multiple RHS or so this will matter, I guess.

mfherbst:
@antoine-levitt Unless I hear from you otherwise, I'll merge this as is.

antoine-levitt:
Sure! I'm not fully convinced it's the right approach yet, but I'm convinced it's better than what we had before so let's go

PR #1099, opened 2025-05-19
KernelAbstractions for FFTs

abussy:
Careful GPU profiling shows that a lot of time is wasted reshuffling arrays during K-point specific FFT calls. Since FFTs are a large part of `DftHamiltonian multiplication` and `compute_density`, accelerating them can have a large impact on performance, especially for systems with dense k-point meshes.

For illustration, a `ifft!` call currently looks like this in the profiler:

fft_master

The 3 blue squares on the right are the actual FFT, and the blue square on the left is the kernel filling the array with zeros. All the rest, i.e. the majority of the runtime, is spend in this operation: `f_real[Gvec_mapping] = f_fourier`. Something similar happens for `fft!`.

This PR introduces explicit GPU kernels for the reshuffling of data during FFTs. The `KernelAbstractions.jl` package is used in order to remain vendor agnostic (already an indirect dependency of DFTK through GPU packages). The kernel themselves are very simple, but their performance is much better. The same `ifft!` call now looks like the following in the profiler:

fft_kernel

For this to be possible, some light code uglification is necessary. The `Gvec_mapping` field of a given K-point must be available on the GPU at all time, and transferring it to the device at each call would defeat the purpose of this PR. As a solution for this, I added a `Gvec_mapping_gpu` field to the K-point struct (could also be called `Gvec_mapping_fft`). I decided to duplicate the data rather than moving the original `Gvec_mapping` field to the device, because the latter is assumed to be on the CPU in most places in the code, and it's inverse `mapping_inv` is not in a format suitable for the GPU (and it makes sense that back and forth ops are stored in the same manner).

Finally, here are some timings (s) for the input show below:

| GPU | master | this PR |
| --- | --- |--- |
| A100 | 44 | 33 |
| H100 | 51 | 40 |
|Mi200 |71 | 54 |

```
using DFTK
using PseudoPotentialData
using CUDA
setup_threading(;)

arch = DFTK.GPU(CuArray)

Ecut = 72.0
kgrid = [10, 10, 10] 
temp = 0.01
tol = 1.0e-8
maxiter = 10

a = 7.260
lattice = a * [[1. 0. 0.];
               [0. 1. 0.];
               [0. 0. 1.]]

Sr = ElementPsp(:Sr, PseudoFamily("dojo.nc.sr.pbe.v0_4_1.stringent.upf"))
V = ElementPsp(:V, PseudoFamily("dojo.nc.sr.pbe.v0_4_1.stringent.upf"))
O = ElementPsp(:O, PseudoFamily("dojo.nc.sr.pbe.v0_4_1.stringent.upf"))

atoms     = [Sr, V, O, O, O]
positions = [[0.5, 0.5, 0.5],
             [0.0, 0.0, 0.0],
             [0.5, 0.0, 0.0],
             [0.0, 0.5, 0.0],
             [0.0, 0.0, 0.5]]

model = model_DFT(lattice, atoms, positions; temperature=temp, functionals=PBE(),
                  smearing=DFTK.Smearing.Gaussian())

#warmup
scfres = self_consistent_field(basis; tol=tol, maxiter=3);

#actual calculations
DFTK.reset_timer!(DFTK.timer)
scfres = self_consistent_field(basis; tol=tol, maxiter=maxiter);
@show DFTK.timer
```


antoine-levitt:
Nice, but for large scale computations we might be better off just doing non variational computations on a cubic grid, which avoids this computation entirely.

mfherbst:
I think there is a sweet spot in between where you need variational data. Especially in data generation for ML consistency between energy and forces within a structure and across a dataset is quits important for example.

But we should really test "how bad" such inconsistencies would be for nonvariational.

abussy:
I've learnt to be wary of non variational calculations in my (more scientific) past, especially if gradients are involved.

Generally, I think that variational calculations should always be an option,  and probably the default one. Then, we might as well do it efficiently.

antoine-levitt:
Careful about the word variational here. Non variational in the dftk context just means that the discretization is non variational, ie the energy is not necessarily above the true one. Also note that even the "variational" energy is not strictly variational, because the xc energy is approximated on a grid. The fact that the discretization is not variational doesn't mean that the discrete orbitals are not: they still minimize the discretized energy, and the hellmann feynman theorem is still valid. So I wouldn't expect problems like that.

mfherbst:
I'd need a more careful review, but I think given the performance improvement that's a reasonable amount of code to add, don't you agree @antoine-levitt ?

src/fft.jl
```diff
@@ -108,13 +112,12 @@ function ifft!(f_real::AbstractArray3, fft_grid::FFTGrid, f_fourier::AbstractArr
...
-               Gvec_mapping::Vector{Int}, f_fourier::AbstractVector; normalize=true)
-    @assert length(f_fourier) == length(Gvec_mapping)
-    @assert size(f_real) == fft_grid.fft_size
+               Gvec_mapping::AbstractVector{Int}, f_fourier::AbstractVector;
+               normalize=true)
 
     # Pad the input data
     fill!(f_real, 0)
-    f_real[Gvec_mapping] = f_fourier
+    ifft_remap!(f_real, Gvec_mapping, f_fourier)
```
antoine-levitt, line 120: More meaningful names would be good. Something like fill_grid_with_kpt_data!()? Maybe also comment that this does `f_real[Gvec_mapping]=f_fourier`?

antoine-levitt (commented):
Sounds like the kernel overload should just be in GPUArrays, no? (ie x[I]=y should do what you implemented) Can we fix it there?

mfherbst:
I think kernelabstractions also has a cpu backend, so perhaps we can actually avoid special-casing the gpu code. I also have a few issues with the proposed code structure ... will review more later.

src/Kpoint.jl
```diff
@@ -4,16 +4,17 @@ More generally, a ``k``-point is a block of the Hamiltonian;
...
+    #                                # returned by the `spin_components` function
+    coordinate::Vec3{T}              # Fractional coordinate of k-point
+    G_vectors::GT                    # Wave vectors in integer coordinates (vector of Vec3{Int})
+    #                                # ({G, 1/2 |k+G|^2 ≤ Ecut})
+                                     # This is not assumed to be in any particular order
+    mapping::Vector{Int}             # Index of G_vectors[i] on the FFT grid:
+    #                                # G_vectors(basis)[kpt.mapping[i]] == G_vectors(basis, kpt)[i]
+    mapping_inv::Dict{Int, Int}      # Inverse of `mapping`:
+    #                                # G_vectors(basis)[i] == G_vectors(basis, kpt)[mapping_inv[i]]
+    mapping_gpu::AbstractVector{Int} # GPU copy of `mapping` for fast FFTs
```
mfherbst: I would make this a full type variable (similar to `GT`)  and included in the definition of Kpoint.
Also do we need the copy, i.e. can we not just put it always on the gpu ? I think we have a few other places but the FFTs where we need the `mapping`, but these are all not performance critical (e.g. IO), so the extra communication should be fine.

src/Kpoint.jl
```diff
@@ -35,20 +36,22 @@ function Kpoint(spin::Integer, coordinate::AbstractVector{<:Real},
...
 
     mapping_inv = Dict(ifull => iball for (iball, ifull) in enumerate(mapping))
-    Kpoint(spin, k, Gvecs_k, mapping, mapping_inv)
+    mapping_gpu = to_device(architecture, mapping)
+    Kpoint(spin, k, Gvecs_k, mapping, mapping_inv, mapping_gpu)
 end
 
 # Construct the kpoint with coordinate equivalent_kpt.coordinate + ΔG.
 # Equivalent to (but faster than) Kpoint(equivalent_kpt.coordinate + ΔG).
-function construct_from_equivalent_kpt(fft_size, equivalent_kpt, coordinate, ΔG)
```
mfherbst, line 43: That's quite a low-level function. Can't we instead pass architecture ?

src/gpu/fft.jl
```diff
@@ -0,0 +1,29 @@
...
+# f_real[Gvec_mapping] = f_fourier or f_fourier .= view(f_real, Gvec_mapping)
+# are not optimized on the GPU, and performance suffers when large numbers of
+# FFT calls take place (tens of thousands or more).
+
+import KernelAbstractions: get_backend, @kernel, @index, @Const
+
+# Mapping kernel for the K-point specific iFFT
+@kernel function ifft_remap_kernel!(f_real, @Const(Gvec_mapping), @Const(f_fourier))
+    i = @index(Global)
+    @inbounds f_real[Gvec_mapping[i]] = f_fourier[i]
```
mfherbst: I'm wondering if the main performance problem was actually that we did not put an `@inbounds` in the assignment `f_real[Gvec_mapping] = f_fourier`.

@abussy Can you check that the original DFTK code with a simple `@inbounds f_real[Gvec_mapping] = f_fourier` not already does the trick?

In any case if we make the inbounds assumption for the GPU version, we should do it for the CPU version, too.

src/fft.jl
```diff
@@ -100,6 +100,10 @@ end
 G_vectors(fft_grid::FFTGrid) = fft_grid.G_vectors
 r_vectors(fft_grid::FFTGrid) = fft_grid.r_vectors
 
+# Remapping function for the K-point specific iFFT
+function ifft_remap!(f_real::AbstractArray3, Gvec_mapping::Vector{Int}, f_fourier::AbstractVector)
+    f_real[Gvec_mapping] = f_fourier
```
mfherbst: `@inbounds`

src/gpu/fft.jl
abussy: Very good point. I had a look into the impact of using `@inbounds`, and it is significant. As a matter of fact, it matches the performance of the custom kernel, so there is no need to add the `KernelAbstractions.jl` dependency, nor the `src/gpu/fft.jl` file.

To fully match the performance of the PR in its current state, it is not enough simply add `@inbounds` to the master branch code. Profiling reveals that there is still a lot of time spent transferring the `Gvec_mapping` array from the CPU to the device at each FFT call.

I would propose the keep a version of `Gvec_mapping` on the GPU, but get rid of all the kernel specific code.

mfherbst: Sounds great to me !

src/Kpoint.jl
abussy: I agree that keeping a single instance of `mapping` on the GPU and transferring it to the CPU when needed would not impact performance.

The main reason I went that route is that `mapping_inv` is not GPU compatible, and that keeping the forward and backward operation in the same place makes sense to me.

PR #1101, opened 2025-05-22
Reuse stored projectors in AtomicNonlocal forces

abussy:
During the calculation of the AtomicNonlocal forces, the projectors are recomputed for each element and k-point. There is no good reason to do that since they are already stored in the TermAtomicNonlocal.

This mostly impacts GPU performance, as relatively slow CPU loops over G + k vectors are avoided. This PR might collide with #1088.

### DIFF
--- a/src/terms/nonlocal.jl
+++ b/src/terms/nonlocal.jl
@@ -46,7 +46,7 @@ end
     (; E, term.ops)
 end
 
-@timing "forces: nonlocal" function compute_forces(::TermAtomicNonlocal,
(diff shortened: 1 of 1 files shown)


mfherbst:
Yes, sure, makes sense to re-use what we already computed.

@Technici4n Any problems with merging this before #1088 ?

Technici4n:
It will be a bit of work to update #1088 after this, but conceptually this change makes sense, and I don't know when I will be done with #1088. So go for it!

PR #1103, opened 2025-05-24
Fix bug with TimerOutputs

mfherbst:


antoine-levitt:
Nice, timer outputs always messed with everything, now hopefully less...

Issue #1102, opened 2025-05-23
Proposal to replace `direct_minimization` and switch to Manopt solvers


kellertuer:
This semester I have  student (@elorannug) working on the direct minimization for scf and port it to Manopt.

In this issue I would like to check how to best continue. Before describing the current for of `direct_minimization` and a proposed new form, let's first state the main questions that I would like to discuss

Should the old method stay but be moved into an `Optim.jl` extension? That makes a new release at least only “mildly breaking”, a user has to load Optim and potentially Linesearches themselves.
The disadvantage is, if someone loads Optim and Manopt, both extensions have to work nicely together still.

What do you think @antoine-levitt ?

## Current form

In the current `scf/direct_minimization.jl` the user has an easy interface based on a `PlaneWaveInterface` that allows to set the usual `maxiter`/`tol` stopping criteria, choose a solver from Optim and (optionally= provide a start point.

The only keyword I personally do not fully understand is `alpha_guess` – is that a function to have a good initial step size guess for the line search?

## New form

The `basis` will stay the only mandatory parameter.

All existing kwargs would stay, we we just use them differently internally

### Keyword arguments

* `stopping_criterion=` will use the existing `maxiter`/`tol` to introduce a default stopping criterion based on that, but a user could also provide their own one
  one challenge might be that we currently have one that we do define ourselves – if that is only introduced in the extension, it would be quite private, and hence not so easy to use for an end user, but I think I know a trick
 * `solver=` provide an `AbstractManoptSolverState` to specify which solver to use
 * `peconditioner=` a manopt preconditioner, especially when using LBFGS (there is also a second one its, type)
 * `manifold=` specify whether to use `Stiefel` or `Grassmann`; one could also allow for their Generalized (i.e. weighted) versions
 * `kwargs...` can be used to decorate both problem and state, such that `debug=` and such would also work in this function call.
 
 Since this requires both Manopt anf Manifolds, I would propose to do this in a `DFTKManoptManifoldsExt.jl` with the above keywords. It would still only be “mildly bearking” for end users.
 
 Let me know what you think.

antoine-levitt:
[...]

PR #1105, opened 2025-06-03
[WIP, Discussion] `direct_minimization` with Manopt.jl

kellertuer:
Following the discussion in #1102, here is a work in progress on moving `direct_minimization` to an extension and to use `Manopt.jl`.

Due to the discussions in the linked issue I feel a bit unsure how and whether to continue. Currently left to do

* [ ] 1 document `direct_minimization` in `src/scf/` that it requires `Manopt` and `Manifolds` to work
* [ ] 2 (maybe) rename `ManoptPreconditioner` to maybe something more reasonable. It basically wraps a DFTK preconditioned 
* [ ] 3 `CostGradFunctor!` should get a nicer name and could maybe also become the fixed energy/gradient computation but also maybe stay a keyword argument
* [ ] 4 Document `CostGradFunctor!`
* [ ] 5 build a factory (https://manoptjl.org/stable/plans/#Manopt.ManifoldDefaultsFactory) for the `StopWhenDensityChangeLess` stopping criterion to postpone creation until we know the size of the density matrix – this makes the `_` keywords obsolete
* [ ] 6 clarify what the use of `alphaguess` is and what else than `nothing it can/should be.
* [ ] 7 the preconditioner has to be passed differently to gradient_descent/cg than Quasi newton, this needs to be handled somewhat.
* [ ] 8 discuss what to do in the end with the `info` or in other words what to return from that function.
* [ ] add tests and see where we can cover this in the docs

Points 1-5 and 9are merely to work though as soon as I find time
On 6 and 8 I need some help and feedback
7 is maybe a bit tricky to get right in general.

For now this function (and its helpers) are a bit work in progress – I open this a bit early, since I am not sure the philosophies of the packages agree well enough that this approach is useful. in the issue above it sounds currently like the idea here is that the philosophy is too different and that it might not fit.


kellertuer:
While I did not yet address the points above, I spent this evening debugging the code, now it is relatively easy to

* change the solver
* change the manifold
* change the linesearch
* change the retraction
* change the vector transport
* activate debug & record
* change the stopping criterion (though that “deactivates” the values passed to `tol` and `maxiter`

Every now and then it seems the gradient is not so correct because the step size collapses after a few iterations, I have to check that (or the example is too simple so it finishes after < 4 steps anyways).

6 is nearly done, just requires a bit of tolerance on the gradient descent side in manopt; 
Feedback on 5 and 7 would be nice; besides a bit of debugging the rest, only documentation and testing (and maybe a bit of renaming) is missing.

kellertuer:
I worked a bit mainly on a bit of testing (as far as my knowledge reaches), naming and documentation.

This should work for now, with the small caveat that for `GradientDescentState` (i.e. using gradient descent) as well as to use callbacks, we have to wait for the next version of Manopt.jl to be released.
Besides waiting for these, the PR is finished up to the following points to discuss

## Questions on the code
(also written within the code in the corresponding places)

* In the stopping criterion – if I just know the manifold, can I then already compute (at least the size/type) of the density  `ρ`? That would make a much nicer constructor only requiring the tolerance.
* Does it makes sense to be able to record `ρ` ? A user could then easily use `record = [:ρ] to record if we do an easy recorder (but maybe also no one likes that here)
* a debug is maybe not so useful, since it seems to be a large array, but its norm maybe?
* Should a user be able to provide their own cost/grad? (see ...and beyond below as well)
  Then we would have to change a few small things in the setup.
* What should the `debug_info` contain?

## Questions in general

* does DFTK have a certain code style I should apply?
* does DFTK want to use [`DocumenterCitations.jl`](https://github.com/JuliaDocs/DocumenterCitations.jl) for literature? I could set it up, either here or in a separate PR
* does DFTK want to use [`DocumenterInterlinks.jl`](https://github.com/JuliaDocs/DocumenterInterLinks.jl) to link to other-peoples-package-docs?
* where should the doc strings of this extension be put in the doc?
* Should we add for example the code above as a tutorial somewhere?
* It would be nice to mention Ole Gunnar somewhere, since the code here is based on his master thesis

## ...and beyond

I had a discussion with @jonas-pueschel to also include his CG approache; that however would first mean, that in severa places, especially in Armijo and Wolfe, the inner product with the gradient has to ne replaced an evaluation of the differentia, see https://github.com/JuliaManifolds/Manopt.jl/issues/482. Then we can probably either update the cost/grad here or provide a second alternative.

src/terms/Hamiltonian.jl
```diff
@@ -185,9 +185,24 @@ end
...
+
+Compute the energies and Hamiltonian.
+
+# Arguments
+- `basis::PlaneWaveBasis`: The plane wave basis to use.
+- `ψ`: The wavefunctions, as a vector of arrays, one for each k-point.
+- `occupation`: The occupation numbers for each k-point and band.
+
+The keyword arguments `kwargs...` are passed to [`ene_ops`](@ref).
+These can be used to pass additional information that might be useful for the energy terms to precompute
```
mfherbst: not just useful, sometimes even required. I would just drop the second sentence and say something like ... can be used to pass additional information to the terms.

src/terms/Hamiltonian.jl
```diff
@@ -215,7 +230,7 @@ kwargs is additional info that might be useful for the energy terms to precomput
     ham = Hamiltonian(basis, [HamiltonianBlock(basis, kpt, hks; scratch)
                               for (hks, kpt) in zip(hks_per_k, basis.kpoints)])
     energies = Energies(term_names, energy_values)
-    (; energies, ham)
+    return (; energies, ham)
```
mfherbst: Please don't, we generally don't do that.

src/terms/Hamiltonian.jl
```diff
@@ -185,9 +185,24 @@ end
...
+    (; energies, hamiltonien) = energy_hamiltonian(basis::PlaneWaveBasis, ψ, occupation; kwargs...)
+
+Compute the energies and Hamiltonian.
+
+# Arguments
+- `basis::PlaneWaveBasis`: The plane wave basis to use.
+- `ψ`: The wavefunctions, as a vector of arrays, one for each k-point.
+- `occupation`: The occupation numbers for each k-point and band.
+
+The keyword arguments `kwargs...` are passed to [`ene_ops`](@ref).
```
mfherbst: If you want, rename `ene_ops` to `energy_hamiltonian`. I wanted to do that anyway at some point as it makes more sense to not have two functions for that.

ext/DFTKManifoldsManoptRATExt.jl
```diff
@@ -0,0 +1,352 @@
+module DFTKManifoldsManoptRATExt
+    using DFTK
+    using Manopt
+    using Manifolds
+    using RecursiveArrayTools
+
+function __init__()
+    # A small trick to make the stopping criterion available globally
+    setglobal!(DFTK, :StopWhenDensityChangeLess, StopWhenDensityChangeLess)
```
mfherbst: Wow ... feels like a lot of magic. Is there no other way ?

ext/DFTKManifoldsManoptRATExt.jl
```diff
@@ -0,0 +1,352 @@
...
+"""
+struct ManoptPreconditionersWrapper!!{T,S}
+    Nk::Int
+    Pks::Vector{T}
+    kweights::Vector{S}
+end
+function (mp::ManoptPreconditionersWrapper!!)(M::ProductManifold,
+    Y,
+    p,
+    X)
```
mfherbst: We can bikeshed styling later, but that just makes code lengthy in my opinion.

ext/DFTKManifoldsManoptRATExt.jl
```diff
@@ -0,0 +1,352 @@
...
+    end
+    return Y
+end
+function (mp::ManoptPreconditionersWrapper!!)(M::ProductManifold, p, X)
+    Y = copy(M, p, X)
+    mp(M, Y, p, X)
+    return Y
+end
+"""
+    HartreeFockEnergyCostGrad{T,S,P,X,R,E,H}
```
mfherbst: HartreeFock is just one (out of many) methods of a similar mathematical structure ? I'd more generally call it just `EnergyCostGrad` ?

ext/DFTKManifoldsManoptRATExt.jl
```diff
@@ -0,0 +1,352 @@
...
+* `X::X`: The last gradient, stored as a `Manopt.jl` tangent vector, i.e. using an `ArrayPartition`
+* `ρ::R`: The last density, computed from the wave functions.
+* `energies::E`: The last vector of energies, computed from the wave functions.
+* `ham::H`: The last Hamiltonian, computed from the wave functions.
+"""
+mutable struct HartreeFockEnergyCostGrad{T,S,P,X,R,E,H}
+    basis::PlaneWaveBasis{T}
+    occupation::Vector{S}
+    Nk::Int
+    filled_occ::Int
```
mfherbst, line 79: I would not store this, `occupation` should be all you need

ext/DFTKManifoldsManoptRATExt.jl
```diff
@@ -0,0 +1,352 @@
...
+    ρ::R        # the last density
+    energies::E # the last vector of energies
+    ham::H      # the last Hamiltonian
+end
+# Function shared by both cost and gradient of cost:
+function _compute_density_energy_hamiltonian!(cgf::HartreeFockEnergyCostGrad, M::ProductManifold, p)
+    # Can we improve this by copying elementwise?
+    # copyto!(cgf.ψ, (copy(x) for x in p.x)) # deepcopyto!
+    # Maybe like
+    for i in eachindex(cgf.ψ)
```
mfherbst: We usually call `k`-point indices `ik`.

ext/DFTKManifoldsManoptRATExt.jl
```diff
@@ -0,0 +1,352 @@
...
+    Nk::Int
+    filled_occ::Int
+    ψ::P        # last iterate, While we usually have an  iterate `ArrayPartition` `p` iterate in Manopt, we store it as a vector, i.e. in the ψ format for DFTK
+    X::X        # last gradient
+    ρ::R        # the last density
+    energies::E # the last vector of energies
+    ham::H      # the last Hamiltonian
+end
+# Function shared by both cost and gradient of cost:
+function _compute_density_energy_hamiltonian!(cgf::HartreeFockEnergyCostGrad, M::ProductManifold, p)
```
mfherbst: One thing we should think about is MPI parallelism. In that case the `k`-points are distributed across processors. How would ManOpt react to that ? For us `compute_density` and `energy_hamiltonian` do the necessary synchronisation behind the scenes.

ext/DFTKManifoldsManoptRATExt.jl
```diff
@@ -0,0 +1,352 @@
...
+
+#
+#
+# Stopping Criteria
+get_parameter(objective::Manopt.AbstractManifoldCostObjective, s) = get_parameter(Manopt.get_cost_function(objective), s)
+get_parameter(energy_costgrad::HartreeFockEnergyCostGrad, s::Symbol) = get_parameter(energy_costgrad, Val(s))
+get_parameter(energy_costgrad::HartreeFockEnergyCostGrad, ::Val{:ρ}) = energy_costgrad.ρ
+
+# TODO/DISCUSS:
+# *  Can we also accept a manifold here and conclude the `ρ` from there?
```
mfherbst: Not sure, I understand the question.

ext/DFTKManifoldsManoptRATExt.jl
```diff
@@ -0,0 +1,352 @@
...
+    s = has_stopped ? "reached" : "not reached"
+    return "|Δρ| = $(c.last_change) < $(c.tolerance):\t$s"
+end
+function Base.show(io::IO, c::StopWhenDensityChangeLess)
+    return print(io, "StopWhenDensityChangeLess with threshold $(c.tolerance).\n    $(status_summary(c))")
+end
+
+# TODO/DISCUSS: Should we have Records / Debugs for
+# * ρ ? A user could then easily use `record = [:ρ] to record it
+# * a debug is maybe not so useful, since it seems to be a large array, but its norm maybe?
```
mfherbst: the norm of the change of ρ would be useful.

ext/DFTKManifoldsManoptRATExt.jl
```diff
@@ -0,0 +1,352 @@
...
+
+# TODO/Discuss:
+# * Should a user be able to provide their own cost/grad?
```
mfherbst: this is done via the basis, essentially.

We should discuss a bit how the generalisation to metals happens and what changes would be needed ... to make sure the interface does not end up being too specific for insulators.

ext/DFTKManifoldsManoptRATExt.jl
```diff
@@ -0,0 +1,352 @@
...
+
+Compute a minimizer of the Hartree-Fock energy functional using the direct minimization method.
+
+# Argument
+
+* `basis::DFTK.PlaneWaveBasis{T}`: The plane wave basis to use for the DFT calculation.
+
+# Keyword Arguments
+
+* `manifold_constructor=(n,k) -> Stiefel(n,k,ℂ)`: A function that constructs a single component of the product manifold, which is the domain of the energy functional (cost)
```
mfherbst: I don't think this needs to be too flexible on the user-facing side. It's more that one will have different "algorithms" that essentially view the problem following a different mathematical structure, but this will not only change the manifold, but also various other aspects of the solution algorithms.

ext/DFTKManifoldsManoptRATExt.jl
```diff
@@ -0,0 +1,352 @@
...
+  It maps the dimensions `(n,k)` to a manifold to be used per component. The default is the complex Stiefel manifold
+* `ψ=nothing`: The initial guess for the wave functions. If not provided, random orbitals will be generated.
+* `tol=1e-6`: stop when the change in the density is less than this tolerance. If you set the `stopping_criterion=` directly, this keyword has no effect.
+* `maxiter=1000`: The maximum number of iterations for the optimization. If you set the `stopping_criterion=` directly, this keyword has no effect.
+* `preconditioner=DFTK.PreconditionerTPA`: The preconditioner to use for the optimization.
+* `solver=QuasiNewtonState`: The solver to use for the optimization. Defaults to a quasi-Newton method.
+* `stopping_criterion=Manopt.StopAfterIteration(maxiter) | StopWhenDensityChangeLess(tol)`: The stopping criterion for the optimization.
+* `evaluation=InplaceEvaluation()`: The evaluation strategy for the cost and gradient.
```
mfherbst: Is there another one possible or does it even make sense ?

ext/DFTKManifoldsManoptRATExt.jl
```diff
@@ -0,0 +1,352 @@
...
+    ψ = isnothing(ψ) ? [DFTK.random_orbitals(basis, kpt, n_bands) for kpt in basis.kpoints] : ψ
+    occupation = [filled_occ * ones(T, n_bands) for _ = 1:Nk]
+    energies, ham = energy_hamiltonian(basis, ψ, occupation; ρ=ρ)
+
+    # Part II: Setup solver
+    #
+    #
+    # Initialize the product manifold
+    dimensions = size.(ψ) # Vector of touples of ψ dimensions
+    manifold_array = map(dim -> manifold_constructor(dim[1], dim[2]), dimensions)
```
mfherbst, line 435: Again here we should think what happens with MPI parallelism.

mfherbst (commented):
I left a few high-level comments. Maybe we should have a zoom to discuss a few details, that I don't get yet.

kellertuer:
Yes a zoom is probably the best, in Text it might take a few iterations too much.

src/terms/Hamiltonian.jl
kellertuer: Sure, no problem – for now this was just me trying to understand what it does and document it for the sake of documentation.

src/terms/Hamiltonian.jl
kellertuer: Write `return` ?
Interesting, that is from auto formatter when I paste lines. To me it makes it much clearer that this is not only the last line but also really the intended return value – and not an accidentally returned one just because it is the last line.

ext/DFTKManifoldsManoptRATExt.jl
kellertuer: I fear there is not. This is adapted from how Benoit Legat introduced JuMP/MOI to Manopt. The problem is

* extensions were meant to only _extend_ functions by new signatures
* all defined structs within an extension are therefore local and can basically not be accessed from outside

For this one – we can not define it outside of the extension since it needs both Manopt and DFTK, but if we would not do this trick, it could only be used as the default and never with other tolerances or in other places by a user (since it is only super locally defined within the extension).

For now I am not aware of a nicer solution here, in like not at all.

ext/DFTKManifoldsManoptRATExt.jl
kellertuer: This is again my auto formatter, I do not mingle with that myself usually, but feel free to format the code whatever you like.

ext/DFTKManifoldsManoptRATExt.jl
kellertuer: Sure I can work on those in the meantime.

ext/DFTKManifoldsManoptRATExt.jl
kellertuer: One has the type to easily switch to Grassmann for example, but if you ever have something real-valued or a more advanced change, you have to modify this.

ext/DFTKManifoldsManoptRATExt.jl
kellertuer: There is `AllocationEvaluation` (and there are usually also just these two) that would allocate a new tangent vector in every step. The signature is different and it usually a bit slower, but if a user just has that one, we should support it.

* `(M, X, p) -> X` is the default one stated there that computes a gradient in-place of `X`
* `(M, p) -> X` would allocate new memory for the gradient `X` and return that. Might be useful if you want to avoid side effects from shared memory.

ext/DFTKManifoldsManoptRATExt.jl
kellertuer: Manopt has exactly zero support for MPI itself.
I am not sure what would be required for that, I have never in my life worked with MPI.

ext/DFTKManifoldsManoptRATExt.jl
kellertuer: As far as I see in the code we need this in the conversion from Euclidean to Riemannian gradient in line 129 of the code.

ext/DFTKManifoldsManoptRATExt.jl
kellertuer: It would be nice to just have a constrictor `StopWhenDensityChangeLess(M, too)` because then a user does not have to think about the technical details that it internally has to store the `ρ`, even better, if a user would not provide a manifold but just a tolerance, we have a “stopping criterion factory” that can to that automatically later.

But it would require that given a manifold we can get at least the data type of `ρ` – and I do not understand / see if (and how) this would work.

So in short, if we can do that a user can just write `StopWhenDensityChangeLess(1e-9)` and we can do the rest.

src/terms/Hamiltonian.jl
kellertuer: I have zero preference here, for me that is just a function DFTK provides, that is not my package, so I have not much to sau on this. I personally do not understand what `ene_ops` stands for, it reads like som magic. If a name can help avoiding it sounding magic, I would always prefer that.

ext/DFTKManifoldsManoptRATExt.jl
kellertuer: I do not understand what you mean with “that is done via the basis” and what that implies for the code here, but yeah we should check that the code here is generic enough.

mfherbst:
About the return values, I think this would be a very good initial set:
- `ham`
- `basis`
- `energies` (the first result of `energy_hamiltonian`)
- `converged` (`true` or `false`)
- `ρ`,
- `occupation`
- `n_iter` (iteration count)
- `n_matvec` (number of hamiltonian application ... if it's easy to obtain, else don't bother)
- `ψ`
- `history_Δρ`, `history_Etot` (changes of `ρ` in each step, changes of cost function (total energy) in each step)
- `algorithm="StiefelDM"` (some short description o the algorithm that has been employed
- `runtime_ns` (The total run time in nanoseconds if you can easily record that)

ext/DFTKManifoldsManoptRATExt.jl
kellertuer: So I checked what I can understand about this, and I can get from `filled_occ` to `occupation` but not the other way around. So this has to be fixed by someone more clever than me.

kellertuer:
@mfherbst I think I have finished everything we discussed and I was able to do. Here is a few remaining things I am unsure about

* [ ] documentation: I am currently not sure what is documented where, since I have not yet understood the structure of the docs. running them locally does at least not throw errors that docs are missing, but it currently also does not seem to load/document extensions.
* [ ] the [wannier](https://github.com/JuliaMolSim/DFTK.jl/blob/master/docs/src/ecosystem/wannier.jl) example from the ecosystem errors when running the docs, it seems that is unrelated to what I did, since
* [ ] every now and then in the [examples/gross_pitaevskii.jl-L89](https://github.com/JuliaMolSim/DFTK.jl/blob/668a3be3f0fe1a5aa93d1be87ca4e14033767957/examples/gross_pitaevskii.jl#L89) fails and I have not yet found out why
* [ ] I could run the tests yesterday and they were fine; I merged master today and can no longer run them locally due to a version conflict on ForwardDiff
* [ ] currently `direct_minimization` can not yet report `result.converged` see the explanation [here](https://github.com/kellertuer/DFTK.jl/blob/67dc92f1fde8f2d5f3c269e9a831a03bca6d8018/ext/DFTKManifoldsManoptRATExt.jl#L496-L502)
* [ ] while most old examples still run, we could consider writing a separate example for this new solver, maybe using Grassmann?
* [ ] the internal `InsulatorEnergy` currently stores both `occupation` and `filled_occ`, since as I wrote, I am not able to change that with the knowledge I have.

I think this is how far I can provide code here. For a next step I would probably need help. Feedback on the code is of course also welcome.

kellertuer:
This is a careful bump after 2 months without any reaction on my last post with questions and/or a review.

If this is too much work or too much of a change to ask, we can also not do this.

mfherbst:
Thanks for the reminder and sorry for the delay. I'll try to follow up in the next days.

ext/DFTKManifoldsManoptRATExt.jl
elorannug: Hi @kellertuer and @mfherbst. If I understand the problem correctly, attatched is a notebook explaining how to get `filled_occ` from `occupation`. 

**TLDR:** 
_`occupation` is a matrix where all the elements equal `filled_occ`
Number of rows in `occupation` = `Nk`
Number of collumns in `occupation` = `n_bands`_

As I believe `occupation` is not modified after its creation in the optimization routine, this method should be safe. Here is the example: [occupation_to_filled_occ_example.html](https://github.com/user-attachments/files/22255568/occupation_to_filled_occ_example.html)

It has been a while since I've worked on this, so I would be very happy if you could point out if there's any thing wrong with the argument!😊

However, as `filled_occ` is simply an integer, I do not see how this improves the code-performance measurably. I'd appreciate an explanation on that front!😅

Also, @mfherbst, I see you've highlighted line 82-85. I assume you're not also suggesting we're storing `basis` supurflously? If so, I would love an explanation.😊

kellertuer: The simple explanation is that one only needs to store one thing instead of 2 (redundant) ones. But sure for a single integer this is not a memory problem, but maybe only the redundancy? Anything stored redundant has the danger of having inconsistent state.

I think my problem would be the other way around: If we need the number `filled_occ`, can we just take _any_ entry from `occupation`?
Besides that one has then of course to refactor the code accordingly, so that is why I wanted to wait for an explanation here.

elorannug: Thank you for the quick answer! In that case, assuming `occupation` is never modified, yes: any entry from `occupation` can be used to fetch the value of `filled_occ`. Additionally, we would not need to store `Nk` either, as the number of rows for `occupation` equals `Nk`.

From my reading of the code, `occupation` is never modified after its creation. However this claim definetly needs the authority of @mfherbst for its support or rebuttal!

mfherbst: Yes, that was pretty much my argument. `filled_occ` is not state, it's a property of the physics and `Nk` a property of the discretisation. So if you store `basis` and `occupation` that's sufficient.

Indeed `Nk = length(occupation)`. Moreover in the examples we considered so far `occupation` only has the elements `filled_occ` (or zero), but this may be different in the future. So thinking ahead I would directly use `occupation` in https://github.com/JuliaMolSim/DFTK.jl/pull/1105/files#diff-95ca3c357df674b3d1503d4b62a5b1294b5c57d0227c6f9446141105dd891b5eR125 instead of `filled_occ`

kellertuer: You have far more knowledge on the topic than me. But I still then struggle with the question I posed in both my comments here.

How?

`filled_occ` is a number. `occupation` is a matrix. so in that place I would have to choose a number from `occupation`? Which one? Random? The first? Prefer a zero? Prefer e nonzero one? 
That is what I am asking here since June 21.

Feel free to modify the code to anything else _or_ tell me what needs to be changed.
But as I wrote on July 4th, I am not able to modify this, because I simply do not know _how_.

examples/compare_solvers.jl
```diff
@@ -8,6 +8,7 @@ using AtomsBuilder
 using DFTK
 using LinearAlgebra
 using PseudoPotentialData
+using Manopt, Manifolds, RecursiveArrayTools
```
mfherbst, line 11: It's a little weird that one needs to `using` all there three packages explicitly. Can we somehow avoid that ?

mfherbst:
>  every now and then in the examples/gross_pitaevskii.jl-L89 fails

If it's just a small energetic glitch as you mention it, I would not be too worried. But if fails means it crashes, then I am.

>  while most old examples still run, we could consider writing a separate example for this new solver, maybe using Grassmann?

We can, but not extremely important from my point of view. If you have it handy, it does not hurt, though.

> the internal InsulatorEnergy currently stores both occupation and filled_occ, since as I wrote, I am not able to change that with the knowledge I have.

If it's clear now feel free to do it. If not, I'll do a cleanup pass and then I can take a look.

>  documentation: I am currently not sure what is documented where, since I have not yet understood the structure of the docs. running them locally does at least not throw errors that docs are missing, but it currently also does not seem to load/document extensions.

I don't know about the extensions. Basically what we do is that we re-use the examples also as documentation. Is that your confusion ? It may also be that the literate step causes trouble ?

ext/DFTKManifoldsManoptRATExt.jl
```diff
@@ -0,0 +1,503 @@
...
+* `retraction_method = `[`ProjectionRetraction`](@extref `Manifolds.ProjectionRetraction`) the retration to use on the manifold to more into a certain direction.
+* `vector_transport_method=`[`ProjectionTransport`](@extref `Manifolds.ProjectionTransport`)`()` the vector transport to use to move tangent vectors between tangent spaces.
+* `stepsize = `[`ArmijoLinesearch`](@extref `Manopt.ArmijoLinesearch`)`(; retraction_method=retraction_method)`
+   specify a step size rule.
+
+All other keyword arguments are passed to the solver state constructor as well as to
+both a decorate for the objective and the state.
+This allows both to set other setting for a solver but also to add `debug=` funtionality
+or even a `cache=` to the objective, though the default objective already caches parts
+that both a cost and a gradient at the same point would require.
```
mfherbst, line 392: I would add here a note that this interface is not yet stable and may change without warning, because it's quite low-level and I want to reserve ourselves the right to have breaking changes in the future.

ext/DFTKManifoldsManoptRATExt.jl
```diff
@@ -0,0 +1,503 @@
...
+
+* `maxiter=1000`: The maximum number of iterations for the optimization. If you set the `stopping_criterion=` directly, this keyword has no effect.
+* `ψ=nothing`: The initial guess for the wave functions. If not provided, random orbitals will be generated.
+* `ρ=guess_density(basis), # would be consistent with other scf solvers
+* `tol=1e-6`: stop when the change in the density is less than this tolerance. If you set the `stopping_criterion=` directly, this keyword has no effect.
+* `manifold=`[`Stiefel`](@extref `Manifolds.Stiefel`): the manifold the optimisation is running on.
+  The current default cost function allows to also use ∞`Grassmann`](@extref `Manifolds.Grassmann`),
+  both in their complex form.
+* `record=[Manopt.RecordCost(), DFTK.RecordDensityChange(ρ), Manopt.RecordTime(:total)]`:
+  specify what to record during the Iterations. If present, these are included in the returned named tuplpe
```
mfherbst: Typo: Tuplple

ext/DFTKManifoldsManoptRATExt.jl
```diff
@@ -0,0 +1,503 @@
...
+
+# Keyword Arguments
+
+* `maxiter=1000`: The maximum number of iterations for the optimization. If you set the `stopping_criterion=` directly, this keyword has no effect.
+* `ψ=nothing`: The initial guess for the wave functions. If not provided, random orbitals will be generated.
+* `ρ=guess_density(basis), # would be consistent with other scf solvers
+* `tol=1e-6`: stop when the change in the density is less than this tolerance. If you set the `stopping_criterion=` directly, this keyword has no effect.
+* `manifold=`[`Stiefel`](@extref `Manifolds.Stiefel`): the manifold the optimisation is running on.
+  The current default cost function allows to also use ∞`Grassmann`](@extref `Manifolds.Grassmann`),
+  both in their complex form.
```
mfherbst, line 324: Is there a way to avoid duplicating info between here and the low-level function below?
Perhaps we only provide docs here and the in the low-level version we don't document it, referring to this version instead ?

mfherbst (commented):
I think it's converging. Take a look what you can fix with the additional information from my end, else I'd suggest we have a call to clean things up and then I can take over the rest whatever is needed. Things we should clarify:

- [ ] How does the printing look like
- [ ] Anything else to be discussed on docs building ?
- [ ] Source code formatting: We try to keep line length around 92 (see the scarce style guide https://docs.dftk.org/stable/developer/style_guide/)

examples/compare_solvers.jl
kellertuer: Not really.
* Manopt is needed to load the solver(s) we use
* Manifolds is needed for Grassmann and the Product manifold
* The product manifold is basically a dummy until RecursiveArrayTools are loaded end the extension (in Manifolds) provides the efficient product manifold functions (baes on `ArrayPartition`)

so all 3 are needed.

kellertuer:
To answer a few already:

> > every now and then in the examples/gross_pitaevskii.jl-L89 fails
> 
> If it's just a small energetic glitch as you mention it, I would not be too worried. But if fails means it crashes, then I am.

to me it looks like small glitches.
 
> > while most old examples still run, we could consider writing a separate example for this new solver, maybe using Grassmann?
> 
> We can, but not extremely important from my point of view. If you have it handy, it does not hurt, though.

Mainly a matter of time, and I fear it would take me quite some time.

> 
> > the internal InsulatorEnergy currently stores both occupation and filled_occ, since as I wrote, I am not able to change that with the knowledge I have.
> 
> If it's clear now feel free to do it. If not, I'll do a cleanup pass and then I can take a look.

to me this is to 0% clear. 

> > documentation: I am currently not sure what is documented where, since I have not yet understood the structure of the docs. running them locally does at least not throw errors that docs are missing, but it currently also does not seem to load/document extensions.
> 
> I don't know about the extensions. Basically what we do is that we re-use the examples also as documentation. Is that your confusion ? It may also be that the literate step causes trouble ?

Well. To be precise: The new doc strings are defined in the extension.
They are not included in the documentation at all. But the documentation also does not error, like it usually does when doc strings are not included in the docs. The simple plain reason is, that the documentation is not aware the extension exists.
See https://github.com/JuliaManifolds/Manopt.jl/blob/e4a5872c097ec839b799c5d4cacd314d9a5837f6/docs/make.jl#L162-L169 how you have to add them (and of course also have to have loaded them beforehand).
I think this is unrelated to literate.

examples/compare_solvers.jl
mfherbst: I see, but that's quite inconvenient for a potential user. Could we avoid that by making a metapackage that depends on all three of them (like DFTKManopt.jl) ?

kellertuer: Uff. I would have no clue how to do that?
I also do not agree that that is inconvenient, it is what the extensions now bring – more modular packages without direct dependencies:

One can use Manopt with an own manifold (nand then not load Manifolds)
One can use Manifolds with an own implementation for the product structure (and then not load RecursiveArrayTools)

What would that package look like? Something with the Reexport all package? I am not sure what to do here.

ext/DFTKManifoldsManoptRATExt.jl
kellertuer: Sure. Did that.

ext/DFTKManifoldsManoptRATExt.jl
kellertuer: Hm, these two variants allow for a slightly different level of settings. 

This is the version that is meant for ease of use and follow a bit closer the DFTK scheme. That limits a bit what one can set up.
The other (maybe considered a bit more internal) already requires to pass a Manopt state, but then provides more access to “Manopt-stuff”. Sure we can leave that one undocumented, but then I will have forgotten half of what I wrote in about a year and would not be able to help (or help easily) if someone comes along with a problem.
That would not be my preferred way to go.

kellertuer:
Since we released `has_converged` yesterday, I readied the `converged=` field to the named tuple as well.

kellertuer:
I tried at least illustrating the documentation of things in the extension

1) load all packages to trigger loading of the extension
2) adding the extension to the modules in `makedocs`
3) collect all those doc strings on a separate page

When trying to do 3 I usually check which signatures fail to add them on that page – otherwise it is a bit hard to specify that by hand.

However, I can not render the docs locally, the Error I get is (starting snippet)

```
┌ Error: failed to run `@example` block in src/ecosystem/wannier.md:177-198
│ ```@example wannier
│ using wannier90_jll  # Needed to make run_wannier90 available
│ run_wannier90(scfres;
        [... skipped rest of code block that is run ... ]
│              );
│ nothing #hide
│ ```
│   exception =
│    UndefVarError: `wannier90` not defined in `wannier90_jll`
│    Suggestion: check for spelling errors or missing imports.
│    Stacktrace:
│      [1] (::DFTKWannier90Ext.var"#2#3"{String, String, String})()
```

I would assume that is unlrelated to this PR? However, the last commit is hence only a first step towards documenting the extension.

I checked the other extensions, and only one has a one-line (more-like-dummy) doc string. So maybe for other extensions this is not necessary. For the one here, I am unsure how to continue – also not for the other open points. 

Let me know what best to do then.

mfherbst:
> to me it looks like small glitches.

ok, good then.

> Mainly a matter of time, and I fear it would take me quite some time.

then let's not

> to me this is to 0% clear.

Ok then I do it. Is it ok if I push to your branch directly ?

> If you could make more precise what you mean here?

How do the output look like when running the direct minimisation with ManOpt.

> As written in the post above, currently 0 documentation of extensions seems to be in the documentation.

Ok, but then all it takes is to add the `Base.get_extension` calls, right ? 

> I would assume that is unlrelated to this PR? 

I would think so. Are you on a windows machine ? Because Wannier90 does not compile on windows (hence no binary in Yggdrasil).

PR #1110, opened 2025-06-16
Use seeds for RNGs to make consecutive SCF runs reproducible

Technici4n:
This PR makes SCFs reproducible (for a fixed Manifest.toml, possibly requiring no MPI as well) by using a fixed seed for the used pseudorandom number generators. If there is a concern over the random numbers being "helpful" in rare cases, I can look into allowing a seed or rng to be passed through `self_consistent_field` for expert usage.

### DIFF
--- a/src/eigen/diag.jl
+++ b/src/eigen/diag.jl
@@ -9,7 +9,8 @@ that really does the work, operating on a single ``k``-Block.
 function diagonalize_all_kblocks(eigensolver, ham::Hamiltonian, nev_per_kpoint::Int;
                                  ψguess=nothing,
                                  prec_type=PreconditionerTPA, interpolate_kpoints=true,
-                                 tol=1e-6, miniter=1, maxiter=100, n_conv_check=nothing)
+                                 tol=1e-6, miniter=1, maxiter=100, n_conv_check=nothing,
+                                 rng=default_rng())
     kpoints = ham.basis.kpoints
     results = Vector{Any}(undef, length(kpoints))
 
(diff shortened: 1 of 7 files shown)


mfherbst:
Thanks, in general I agree it would be nice to have these things more reproducible. But I'm not sure this is the right way. E.g. perhaps using orbital initial guesses from the pseudopotentials. Would be good to check what QE is doing here.

Generally, things I don't like about this PR:
  - Hard-codes `Xoshiro`: The default RNG has changed in the past and may change again
  - Not configurable.

Can you make this configurable (either using arguments or using preferences)? Ideally one does not need to pass the `TaskLocalRng` three times to disable this.

src/eigen/diag.jl
```diff
@@ -9,7 +9,8 @@ that really does the work, operating on a single ``k``-Block.
 function diagonalize_all_kblocks(eigensolver, ham::Hamiltonian, nev_per_kpoint::Int;
                                  ψguess=nothing,
                                  prec_type=PreconditionerTPA, interpolate_kpoints=true,
-                                 tol=1e-6, miniter=1, maxiter=100, n_conv_check=nothing)
+                                 tol=1e-6, miniter=1, maxiter=100, n_conv_check=nothing,
+                                 rng=default_rng())
```
mfherbst, line 13: This is a very confusing name. I would expect `default_rng() = Random.default_rng()`, which is not the case. Call it `DFTK.default_rng()` the very least.

mfherbst:
> OK, I sprinkled in some rng kwargs

well that's sort of what I wanted to avoid: The need to pass arguments super deep down to go back to the old behaviour ... I don't have a good other idea spontaneously, but I'm not sure I like this. Let's talk when I'm back.

kellertuer:
Hm, at least in my head it is as easy as:

* if you do not have a preferred random you introduce the keyword as `rng = Random.default_rng()`
* if you have a good random generator that works well (or do not want to break other peoples code) you do `rng= Xoshiro(15)`

That way you can fully keep the ones you used until now but make them transparently visible to the user and they can change it if needed.

PR #1118, opened 2025-06-26
Integrate PseudoFamily closer with Model

### LINKED ISSUE #1109: Make dependency on PseudoPotentialData hard?
I often open the DFTK project to try things out, however PseudoPotentialData is not available since it's only a weak dependency. So either I use the deprecated internal DFTK pseudo library, or I have to create another environment which is a hassle. Could we make it a strong dependency? I feel like it doesn't bring much to make it optional.

mfherbst:
Closes #1109.

TODO:
- [x] Automatically determine `Ecut` and `supersampling` in `PlaneWaveBasis` construction if possible.
- [x] Tests for automatic `Ecut` and `supersampling` determination in `PlaneWaveBasis` constructor
- [x] An example in the docs and some notes on this in the tutorial
- [x] Write pseudopotential family from model into json and jld2 output data

This makes it possible to construct a `PlaneWaveBasis` just as `PlaneWaveBasis(model)` provided that for the chosen pseudopotentials there are recommended cutoffs `Ecut` stored in the pseudopotential data. This works for the common case of the PseudoDojo pseudopotentials (e.g. the Tutorial, the `examples/silicon.jl`), but in many cases the users still needs to provide an `Ecut` argument, same as before.

@antoine-levitt Is this too much magic in the background or is this something we want ?

At least for quick calculations I often find myself looking up the recommended values anyway using `PseudPotentialData.recommended_cutoff` ... so it would be nice to just do this programmatically in the back.

### DIFF
--- a/test/elements.jl
+++ b/test/elements.jl
@@ -64,6 +64,7 @@ end
 
 @testitem "Check constructing ElementPsp from family" begin
     using DFTK
+    using DFTK: pseudofamily
     using PseudoPotentialData
     pd_lda_family = PseudoFamily("dojo.nc.sr.lda.v0_4_1.standard.upf")
 


src/Model.jl
```diff
@@ -48,10 +48,13 @@ struct Model{T <: Real, VT <: Real}
...
     # `atom_groups` contains the groups of indices into atoms and positions, which
     # point to identical atoms. It is computed automatically on Model construction and may
-    # be used to optimise the term instantiation.
+    # be used to optimise the term instantiation. The pseudofamily contains the common family
+    # of pseudopotentials used to determine the pseudopotentials of all atoms or nothing if
+    # such a common family cannot be found.
     atoms::Vector{Element}
     positions::Vector{Vec3{T}}  # positions[i] is the location of atoms[i] in fract. coords
     atom_groups::Vector{Vector{Int}}  # atoms[i] == atoms[j] for all i, j in atom_group[α]
+    pseudofamily::Union{PseudoFamily, Nothing}
```
antoine-levitt: That field feels wrong

src/Model.jl
```diff
@@ -131,6 +134,10 @@ function Model(lattice::AbstractMatrix{T},
     end
     isempty(terms) && error("Model without terms not supported.")
     atom_groups = [findall(Ref(pot) .== atoms) for pot in Set(atoms)]
+    pseudofamily = nothing
+    if length(Set(atom.family for atom in atoms if atom isa ElementPsp)) == 1
```
antoine-levitt: unique instead of set?

src/Model.jl
antoine-levitt: Just have a helper function if needed, it's not like this is expensive

src/Model.jl
```diff
@@ -337,6 +344,28 @@ spin_components(model::Model) = spin_components(model.spin_polarization)
...
+function PseudoPotentialData.recommended_cutoff(model::Model)
+    function get_maximum(property)
+        isnothing(model.pseudofamily) && return missing
+        maximum(model.atom_groups) do group
+            atom = model.atoms[first(group)]
+            getproperty(recommended_cutoff(model.pseudofamily, element_symbol(atom)), property)
+        end
+    end
+
+    (; :Ecut          => get_maximum(:Ecut),
```
antoine-levitt: Why not a standard named tuple?

antoine-levitt: Oh wait this does the same thing. But weird syntax.

src/PlaneWaveBasis.jl
```diff
@@ -293,6 +297,16 @@ Note, this disables certain symmetry features.
         kgrid_inner = MonkhorstPack(kgrid, kshift)
     end
 
+    if ismissing(supersampling)
+        supersampling = 2.0
```
antoine-levitt: take that definition in the recommended_cutoff() function

antoine-levitt:
The point is that DFTK sort of has like a "core" that doesn't really know about DFT and just wants to solve PDEs, and then on top of that is tools for doing DFT, and it'd be nice to keep those two things as separate as possible (when possible/convenient of course, we don't want to be too dogmatic about this)

mfherbst:
> I think it's nice, but I just feel weird about having this code in pwbasis and (especially) in model.

Agree. For the model this can be moved out, for the basis it's more tricky, but I also think it's less relevant.

PR #1120, opened 2025-07-09
Fix inconsistency between CPU and GPU in LOBPCG

abussy:
Remove the `real` conversion in the CPU version of `columnwise_dots`. It is inconsistent with the GPU version and the code prior to #1068. It is also unnecessary, since `real` is already explicitly taken when that matters.

Spotted by @Technici4n

### DIFF
--- a/src/eigen/linalg.jl
+++ b/src/eigen/linalg.jl
@@ -8,7 +8,7 @@ end
     [dot(A[:, i], B[:, i]) for i = 1:size(A, 2)]
 end
 
-# Returns a vector of real(dot(A[:, i], M, B[:, i])), for all columns of
+# Returns a vector of dot(A[:, i], M, B[:, i]), for all columns of
 # A, B, and matrix M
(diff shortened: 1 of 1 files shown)


src/eigen/linalg.jl
```diff
@@ -5,11 +5,11 @@ end
 
 # Returns a vector of dot(A[:, i], B[:, i]), for all columns of A, B
 @views function columnwise_dots(A::AbstractArray{T}, B::AbstractArray{T}) where {T}
-    [real(dot(A[:, i], B[:, i])) for i = 1:size(A, 2)]
+    [dot(A[:, i], B[:, i]) for i = 1:size(A, 2)]
 end
 
 # Returns a vector of real(dot(A[:, i], M, B[:, i])), for all columns of
```
Technici4n:
```suggestion
# Returns a vector of dot(A[:, i], M, B[:, i]), for all columns of
```

mfherbst:
Thanks.

PR #1125, opened 2025-07-29
Ultrasoft pseudopotentials

Technici4n:
The goal of this PR is to **make the ultrasoft SCF work**. Further PRs can implement US extensions for interatomic forces, DFPT, etc... This should also make implementing PAW easier...

Currently I get 8e-4 Ry agreement with QE 7.4.1 on 4x4x4 Silicon from pslibrary 1.0.0 (the SSSP Efficiency one). Maybe QE 7.5 with the integration changes will even improve this?

The flip side is that I had to hardcode the specific pseudopotential in a few places... Definitely needs a ton of cleanup. 😄

Some points to discuss:
- I hacked in reading the PP_AUGMENTATION data. PseudoPotentialIO 0.1 cannot read PP_AUGMENTATION data. And PseudoPotentialIO 0.2 seems to do more than just IO - some of its dependencies are not compatible with DFTK's. @azadoks
- How to store the overlap in the basis/hamiltonian?
- Is `compute_density` the right place to add the augmentation charge?
- How to deal with the Hamiltonian contribution of the augmentation charges?
- We need two FFT grids to get real speedup from US. The orbitals are a lot softer than the density hence they should use a smaller FFT. Matching two FFT grids of different max frequencies is easily done in Fourier space.
- We need some US pseudopotentials in PseudoPotentialData. Somewhat related: supersampling is awkward with US - specifying a density cutoff would be more natural.
- Performance is not great... yet! 😄

A few references for USPPs that were very helpful for me:
- CASTEP website as a starting point: https://www.tcm.phy.cam.ac.uk/castep/documentation/WebHelp/content/modules/castep/thcastepultrapseudo.htm
- The 1993 paper by Laasonen et al.
- The QE code.


mfherbst:
The secret is revealed :tada:

src/common/spherical_harmonics.jl
```diff
@@ -45,5 +45,57 @@ function ylm_real(l::Integer, m::Integer, rvec::AbstractVector{T}) where {T}
...
+function pack_lm(l::Integer, m::Integer)
+    @assert 0 ≤ l
+    @assert -l ≤ m ≤ l
+    return l^2 + l + m + 1
+end
+
+# TODO: check this
+function unpack_lm(lm::Integer)
+    @assert lm ≥ 1
+    l = floor(Int, sqrt(lm - 1))
```
antoine-levitt, line 72: `isqrt`?

src/common/spherical_harmonics.jl
```diff
@@ -45,5 +45,57 @@ function ylm_real(l::Integer, m::Integer, rvec::AbstractVector{T}) where {T}
...
+    m = lm - (l^2 + l + 1)
+    return l, m
+end
```
antoine-levitt, line 75: define in the same vein a nharm(l) function or something like that (to have at all the same place the details of the isomorphism from (l,m) to lm)

src/common/spherical_harmonics.jl
```diff
@@ -45,5 +45,57 @@ function ylm_real(l::Integer, m::Integer, rvec::AbstractVector{T}) where {T}
...
+
+function gaunt_coefficients(T::Type, l::Integer)
+    nharm = (2l+1)^2
+    # Sample points randomly on a sphere
+    rng = Xoshiro(42)
+    points = rand(rng, T, 3, nharm)
+    points ./= sqrt.(sum(points.^2, dims=1))
```
antoine-levitt, line 82: we have a columnwise_norm function now I think

src/common/spherical_harmonics.jl
```diff
@@ -45,5 +45,57 @@ function ylm_real(l::Integer, m::Integer, rvec::AbstractVector{T}) where {T}
...
+
+# TODO: check this
+function unpack_lm(lm::Integer)
+    @assert lm ≥ 1
+    l = floor(Int, sqrt(lm - 1))
+    m = lm - (l^2 + l + 1)
+    return l, m
+end
+
+function gaunt_coefficients(T::Type, l::Integer)
```
antoine-levitt, line 77: This algorithm is weird (but I know nothing about spherical harmonics), is there no deterministic way to get those? Add a ref at least.

src/common/spherical_harmonics.jl
```diff
@@ -45,5 +45,57 @@ function ylm_real(l::Integer, m::Integer, rvec::AbstractVector{T}) where {T}
...
+    points = rand(rng, T, 3, nharm)
+    points ./= sqrt.(sum(points.^2, dims=1))
+    # Evaluate spherical harmonics at these points
+    Y = zeros(T, nharm, nharm)
+    for j in 1:nharm
+        for i in 1:nharm
+            Y[i, j] = ylm_real(unpack_lm(j)..., points[:, i])
+        end
+    end
+    fact = factorize(Y)
```
antoine-levitt, line 90: I'd possibly be concerned about the possibility of degeneracy here (ie the potentially bad conditioning), but maybe it's very rare. Can you make sure the rng you use is not subject to changes in future versions of julia? Then you can just run this function with various seeds for various l, and pick a seed such that the condition number is mild.

(Otherwise you can maybe get something more robust by doing a rectangular method (ie fitting the gaunt coefficients least squares, with M observations for N unknowns, M >= N))

src/densities.jl
```diff
@@ -42,7 +42,51 @@ using an optional `occupation_threshold`. By default all occupation numbers are
     end
     ρ = sum(getfield.(storages, :ρ))
 
+    # Hacky augmentation charges
```
antoine-levitt, line 45: We need some kind of extra_density mechanism to handle core correction and ultrasoft in the same framework

antoine-levitt: (which could be part of the terms interface, so that densities.jl calls out to all the terms to ask them for their contribution to the density)

src/ultrasoft.jl
```diff
@@ -0,0 +1,59 @@
+function atom_local_grid(lattice::AbstractMatrix{T}, fft_grid::FFTGrid, position, radius) where {T}
+    # Find bounds on the points we have to check along each dimension
```
antoine-levitt, line 2: That looks like computations we have to do for ewald and/or fft_size?

antoine-levitt (commented):
Awesome, thanks a lot for tackling this! Some minor comments skimming through the code.

I thought US needed an overlap matrix, how are you comparing accuracy wrt QE without this?

> We need two FFT grids to get real speedup from US. The orbitals are a lot softer than the density hence they should use a smaller FFT. Matching two FFT grids of different max frequencies is easily done in Fourier space.

So this is another level of supersampling then? You're saying we need one basis set to represent orbitals, one FFT grid to do hamiltonian applies (potential-orbital), and one FFT grid to compute the density/potential? (right now the two FFT grids are the same)

src/eigen/diag.jl
```diff
@@ -13,6 +13,28 @@ function diagonalize_all_kblocks(eigensolver, ham::Hamiltonian, nev_per_kpoint::
...
+        model = ham.basis.model
+
+        # keep only pseudopotential atoms and positions
+        psp_groups = [group for group in model.atom_groups
+                        if model.atoms[first(group)] isa ElementPsp]
+        psps          = [model.atoms[first(group)].psp for group in psp_groups]
+        psp_positions = [model.positions[group] for group in psp_groups]
+
+        # We have an overlap term
+        B = map(ham.blocks) do block
```
Technici4n, line 28: @antoine-levitt this is my hacky implementation of the overlap term. It has the same structure as the nonlocal term (nonlocal is PDP'; overlap is I + PQP'), just with a different matrix in the center: replace D by Q.

(Right now this Q matrix comes straight from the PP_Q in the pseudopotential file, but we will likely want to integrate the discretized augmentation charges to ensure that we get the right number of electrons in the integrated density.)

Btw it's nice that LOBPCG "just works" with the overlap B. :)

src/densities.jl
Technici4n: Non-linear core correction (NLCC) is exclusive to the exchange-correlation term, so it's currently implemented in the Xc term only (and was already done like this before this PR). (In some papers you really see it as a separate density that is added only in the XC functional like $E_{xc}[\rho + \rho_{nlcc}]$.)

You can really see US as separating local "hard" contributions to the density; and "soft" contributions from the orbitals. So the real physical charge density (as used for hartree, local potentials, xc) needs both the soft contribution from the orbitals and the hard contribution from the augmentation charges. But we might want the two parts of the density somewhat separate in case it makes sense for preconditioning or other cases.

src/common/spherical_harmonics.jl
Technici4n: This is inspired by what QE does (it's probably the exact same thing?). The idea is that since spherical harmonics are a basis (of functions), randomly evaluating them should also give a basis (of finite dimensional vectors). But with finite dimensional vectors it's just a simple matrix inversion to recover the basis coefficients.

We could also compute the coefficients directly but it's more technical (two Clebsch-Gordan coefficients - can be computed with WignerSymbols.jl but still annoying :/), especially with real spherical harmonics for which the formulas are apparently slightly different.

For sure this needs a bit of thought wrt. condition numbers, maybe adding a few extra points, etc... And certainly worth some unit tests as well.

src/eigen/diag.jl
antoine-levitt: Sorry, somehow I missed that.

Technici4n:
Ah I see what you mean. Indeed I think we can do the (relatively expensive) Fourier interpolation at setup for each augmentation channel, and only store the real-space result in small boxes around the atoms. (The augmentation charge should be 0 elsewhere).

antoine-levitt:
I actually meant to just not optimize anything and store the full thing. Compute the fourier coefficients of the periodized augmentation charges using hankel, and ifft on the full grid to get the augmentation charges. Or are you concerned that storing basically the density Natom times is too memory-intensive?

mfherbst:
> density Natom times is too memory-intensive?

I think for larger-ish systems that can indeed become an issue. But that would be something I would argue we could optimise in a second pass and be simple and wasteful in the first implementation.

PR #1126, opened 2025-07-29
Fix Pdos

fsicignano:
This new version fixes the old pdos implementation. 
The old version missed a proper orthogonalization of all atomic orbitals, while the new one creates a matrix for each k-point containing all projectors from all atoms (listed as shown in the 'projector_labels' according the atom numbering and n, l, m quantum numbers) and orthogonalizes them. 
Löwdin orthogonalization is used to ensure that the 'identity' loss of the orbital projectors after orthogonalization is minimal.


src/postprocess/dos.jl
```diff
@@ -70,65 +71,157 @@ end
 """
 Projected density of states at energy ε for an atom with given i and l.
 """
-function compute_pdos(ε, basis, eigenvalues, ψ, i, l, psp, position;
-                      smearing=basis.model.smearing,
-                      temperature=basis.model.temperature)
+
+function compute_all_pdos(εs, bands;
```
mfherbst: Makes probably sense to have such a function. I see two things we should start thinking about, also in line with moving forward towards Hubbard. We should have some simple syntax to do things like "select the d manifold of the nickel atom 2, or all silicon p orbitals". This can then be used as a universal selector for both hubbard and here (and potentially other parts in the code). Along these lines we should have some standard convention how the atomic projectors are indexed / stored to make it easy to map from these selectors to the index ranges.

For the PDOS itself, I propose to do this in 3 stages:
1. Compude pdos at some range of `εs`, return some `struct` holding that information.
2. A function to get the raw pdos potting data when one uses above selectors to get the result from applying subsets of the data.
3. Custom `Plots` recipies that allow to use above selectors and the function of 2 on the struct of 1 to do reasonable plotting.

What I have in mind as an interface could be sth like:
```julia
pdos = compute_pdos(scfres)
plot(pdos, [(atom=:Si, l=1)])  # Plot PDOS of silicon p projectors
plot!(pdos, [(atom=:C)])   # Plot PDOS of projecting onto carbon only, but all orbitals
```
But I'm honestly unsure how the selectors should look like.

mfherbst: Another option could be that the step in 1. actually only computes the projectors, the function computes the pdos over an ε range. The advantage is that in combination with plotting (step 3) the actual ε range selection can be lazy and controlled by the user using the standard functions of the plotting library.

mfherbst (commented):
Some ideas where to go from here regarding the interface.

The integration with Plots is optional, but I don't think we get around spending some time how the projectors should look like.

src/common/ortho.jl
```diff
@@ -7,3 +7,17 @@
         return x[:, 1:size(φk, 2)]
     end
 end
+
+@timing function ortho_lwd(φk::ArrayType) where {ArrayType <: AbstractArray}
```
Technici4n: I would probably call this function `ortho_lowdin`.

src/postprocess/dos.jl
```diff
@@ -70,65 +71,157 @@ end
 """
 Projected density of states at energy ε for an atom with given i and l.
```
Technici4n: The documentation of the function should go here rather than inside the function. This also makes it accessible in the interactive help (when you type `?` in the Julia REPL).

src/postprocess/dos.jl
Technici4n: We could keep the name `compute_pdos`.

Technici4n: I would not take the full `bands` object but rather only the required parts, similarly to what was there before.
We can have a second overload of `compute_pdos` that takes the `bands` object and forwards the needed part to the "real" version of the function, similar to what existed before.

src/postprocess/dos.jl
```diff
@@ -70,65 +71,157 @@ end
...
+    positions = model.positions
+    natoms = length(positions)
+    psps = Vector{NormConservingPsp}([model.atoms[i].psp for i in 1:natoms])  # PSP for all atoms
+    
+    max_l = Vector{Any}(undef, natoms)                
+    max_l = [psps[iatom].lmax for iatom in 1:natoms]  # lmax for all atoms
+    max_n = Vector{Vector{Int}}(undef, natoms)        # Principal quantum number for all atoms
+    for iatom in 1:natoms
+        max_n[iatom] = [DFTK.count_n_pswfc_radial(psps[iatom], l) for l in 0:max_l[iatom]]
+    end
```
Technici4n: There's no need to precompute and store these - I would call the needed function in the loop below. For example:
```jl
for l in 0:psps[iatom].lmax
```

src/common/ortho.jl
```diff
@@ -7,3 +7,17 @@
...
+    S = φk' * φk
+    evals, evecs = eigen(S)
+    # Check for linear dependence: eigenvalues close to zero
+    tol = maximum(size(φk)) * eps(real(eltype(φk))) * maximum(abs, evals)
+    if any(abs.(evals) .< tol)
+        error("Input vectors are linearly dependent or nearly so (small eigenvalues detected).")
+    end
```
Technici4n, line 18: Having such a check is not a bad idea but:
- it's enough to check for the smallest eigenvalue (`evals[1]`)
- I am not sure why the `tol` is computed like that
- the error message should print the eigenvalue

src/common/ortho.jl
```diff
@@ -7,3 +7,17 @@
...
+    ihS = evecs * Diagonal(evals .^ (-0.5)) * evecs'
+    x = φk * ihS
+    return x
```
Technici4n, line 21: Just this is actually enough: (`return` is automatic for the last statement of the function)
```suggestion
    φk * ihS
```

src/postprocess/dos.jl
```diff
@@ -70,65 +71,157 @@ end
...
+    
+    projector_labels = Vector{NTuple{4, Int64}}(undef, 0)
+    for iatom in 1:natoms
+        for l in 0:max_l[iatom]
+            for n in 1:max_n[iatom][l+1]
+                for m in -l:l
+                    push!(projector_labels, (iatom, n, l, m))
```
Technici4n: We should also add the "human-readable" label from the psp file.

src/postprocess/dos.jl
```diff
@@ -70,65 +71,157 @@ end
...
+                      smearing=bands.basis.model.smearing, 
+                      temperature=bands.basis.model.temperature)
+# Compute the projected density of states (PDOS) for all atoms and orbitals
+# Input: 
+# - εs: vector of energies at which to compute the PDOS
+# - bands: Bands object containing the eigenvalues, wavefunction, basis and positions
+# Output:
+# - pdos: 3D array of PDOS, pdos[iε_idx, iproj, σ] = PDOS at energy εs[iε_idx] for projector iproj and spin σ
+# - projector_labels: vector of tuples (iatom, n, l, m) for each projector, that maps the iproj index to the 
+#                    corresponding atomic orbital (atom index, principal quantum number, angular momentum, magnetic quantum number)
```
Technici4n: I wouldn't necessarily call `n` the principal quantum number since this would mean we get 1s 1p and 1d orbitals which doesn't make sense. More reason to use the human-readable labels maybe.

src/postprocess/dos.jl
```diff
@@ -70,65 +71,157 @@ end
...
+                for m in -l:l
+                    push!(projector_labels, (iatom, n, l, m))
+                end 
+            end
+        end
+    end
+    nprojs = length(projector_labels) 
 
-    projs = compute_pdos_projs(basis, eigenvalues, ψ, i, l, psp, position)
+    println("---Building projectors for $(nprojs) projectors...")
```
Technici4n: I would remove these print calls.

src/postprocess/dos.jl
```diff
@@ -70,65 +71,157 @@ end
...
+
+        for (ik, G_plus_k) in enumerate(G_plus_k_all_cart)
+            for (ip, p) in enumerate(G_plus_k)
+                radials_p = radials[norm(p)]
+                # see "Fourier transforms of centered functions" in the docs for the formula
+                angular = (-im)^l * ylm_real(l, m, p)
+                form_factors[ik][ip, iproj] = radials_p * angular
+            end
+        end
+    end
```
Technici4n: This should probably be replaced by a call to `build_form_factors` such that we don't have to duplicate the code?

src/postprocess/dos.jl
```diff
@@ -70,65 +71,157 @@ end
...
-    projs = compute_pdos_projs(basis, eigenvalues, ψ, i, l, psp, position)
+    println("---Building projectors for $(nprojs) projectors...")
+    projectors = build_projections(basis, ψ, psps, positions, projector_labels)
 
-    D = zeros(typeof(ε[1]), length(ε), 2l+1)
-    for (i, iε) in enumerate(ε)
-        for (ik, projk) in enumerate(projs)
+    println("---Computing PDOS for $(length(εs)) energies and $(nprojs) projectors...")
+    D = zeros(typeof(εs[1]), length(εs), nprojs, model.n_spin_components)  
+    for (iε_idx, iε) in enumerate(εs)
```
Technici4n: I would call the index `iε` (the i typically stands for **i**ndex) and the energy itself `ε`.

src/postprocess/dos.jl
```diff
@@ -70,65 +71,157 @@ end
...
-function compute_pdos_projs(basis, eigenvalues, ψ, i, l, psp::NormConservingPsp, position)
-    # Precompute the form factors on all kpoints at once so we can better use the cache (memory-compute tradeoff).
-    # Revisit this (pass the cache around explicitly) if RAM becomes an issue.
+# TODO: function compute_single_pdos, or add optional arguments to the previous function to make it print only one pdos
+
+
+function build_projections(basis::PlaneWaveBasis{T}, ψ,
+                    psps::AbstractVector{<: NormConservingPsp}, 
+                    positions, 
+                    labels::Vector{Tuple{Int,Int,Int,Int}}) where {T}
```
Technici4n: It probably makes more sense to construct the labels directly inside of this function, rather than passing them?

src/postprocess/dos.jl
```diff
@@ -70,65 +71,157 @@ end
...
+# Build the projection matrices projsk for all k-points at the same time.
```
Technici4n: In Julia, function comments go as a string above the function. Typically `"""` is used for multiline comments.

src/postprocess/dos.jl
```diff
@@ -70,65 +71,157 @@ end
...
+        proj_vectors = hcat(proj_vectors...)    # Create a matrix of projections for this k-point
+        @assert size(proj_vectors, 2) == nprojs "Projection matrix size mismatch: $(size(proj_vectors)) != $nprojs"
+        proj_vectors = ortho_lwd(proj_vectors)  # Lowdin-orthogonal
+        
+        projs[ik] = abs2.(ψk' * proj_vectors)   # Contract on ψk to get the projections
+        @assert size(projs[ik]) == (size(ψk,2), nprojs) "Projection matrix size mismatch: $(size(projsk)) != $(length(ψk)), $nprojs"
     end
 
-    projs
+    return projs
```
Technici4n: You don't need the return here - the last thing in the function is returned in Julia.

src/postprocess/dos.jl
```diff
@@ -70,65 +71,157 @@ end
...
 end
 
+
```
Technici4n, line 232: Extra lines that don't need to be changed (there's quite a few of those).

src/postprocess/dos.jl
```diff
@@ -70,65 +71,157 @@ end
...
+    end
 
     projs = Vector{Matrix}(undef, length(basis.kpoints))
-    for (ik, ψk) in enumerate(ψ)
-        structure_factor = [cis2pi(-dot(position, p)) for p in G_plus_k_all[ik]]
-        # TODO orthogonalize pseudo-atomic wave functions?
-        proj_vectors = structure_factor .* form_factors_all[ik] ./ sqrt(basis.model.unit_cell_volume)
-        projs[ik] = abs2.(ψk' * proj_vectors) # contract on p -> projs[ik][iband,m]
+    for (ik, ψk) in enumerate(ψ) # The loop now iterates over k-points regardless of the spin component
+        proj_vectors = Vector{Vector{Any}}(undef, 0)
```
Technici4n: I would probably make this a Matrix directly. For example `Matrix{T}(undef, length(G_plus_k_all[ik]), len(labels))` and then write each column as `proj_vectors[:, iproj] = ...`.

src/postprocess/dos.jl
```diff
@@ -70,65 +71,157 @@ end
...
+        for (iproj, (iatom, n, l, m)) in enumerate(labels)
+            structure_factor = [cis2pi(-dot(positions[iatom], p)) for p in G_plus_k_all[ik]]
+            @assert length(structure_factor) == length(G_plus_k_all[ik]) "Structure factor length mismatch: $(length(structure_factor)) != $(length(G_plus_k))"
+            proj_vector = structure_factor .* form_factors[ik][:, iproj] ./ sqrt(basis.model.unit_cell_volume)
+            push!(proj_vectors, proj_vector)    # Collect all projection vectors for this k-point    
+        end
+
+        proj_vectors = hcat(proj_vectors...)    # Create a matrix of projections for this k-point
+        @assert size(proj_vectors, 2) == nprojs "Projection matrix size mismatch: $(size(proj_vectors)) != $nprojs"
+        proj_vectors = ortho_lwd(proj_vectors)  # Lowdin-orthogonal
```
Technici4n: I think this is worth 1-2 lines of comment explaining that ALL projectors are being orthogonalized and that mentioning that this matches `atomic-ortho` for Hubbard in QE.

Technici4n (changes requested):

src/postprocess/dos.jl
fsicignano: Maybe something like "first psp quantum number" could do as long as we don't implement the labels.

src/postprocess/dos.jl
fsicignano: At the moment build_form_factors works a little differently since it contains the loop on m internally. To use it I should change the label system or ask inside the loop to perform the task only once per (iatom, n, l) value, but this would make the code less readable.

PR #1129, opened 2025-08-06
Fix Anderson on GPU

abussy:
The PR fixes the Anderson acceleration on GPU. Errors were triggered because of `M = @view M[:, 1:end .!= idrop]`, where `1:end .!= idrop` produces an array of Bool (which is not appreciated as indexing by GPU arrays).

Note that the error was only obvious on AMD GPUs. However, it also sometimes happened on NVIDIA GPUs too. It is not clear to me why the issue arises more often on some hardware. It is possible to trigger the error with high probability by changing the value of  [maxcond](https://github.com/JuliaMolSim/DFTK.jl/blob/79d911cf40c08ff474e76f324bcadf83bbe874f3/src/scf/anderson.jl#L42), e.g. by setting it so `1e-2`.


antoine-levitt:
Lgtm (but give col_filter a more descriptive name like kept_cols)

PR #1131, opened 2025-08-08
Enable precompilation on AMD GPUs

abussy:
Saves ~20s at startup on a Mi200.

### DIFF
--- a/ext/DFTKAMDGPUExt.jl
+++ b/ext/DFTKAMDGPUExt.jl
@@ -1,7 +1,8 @@
 module DFTKAMDGPUExt
 using AMDGPU
+using PrecompileTools
 using LinearAlgebra
-import DFTK: GPU
+import DFTK: GPU, precompilation_workflow
 using DFTK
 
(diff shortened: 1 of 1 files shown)


ext/DFTKAMDGPUExt.jl
```diff
@@ -12,4 +13,23 @@ function LinearAlgebra.cholesky(A::Hermitian{T, <:AMDGPU.ROCArray}) where {T}
     LinearAlgebra.Cholesky(Acopy, A.uplo, info)
 end
```
mfherbst, line 15: Do you not need some fencing like https://github.com/JuliaMolSim/DFTK.jl/blob/master/ext/DFTKCUDAExt.jl#L20 here ?

abussy: In the CUDA case, access to Libxc GPU is necessary for pre-compilation. In the AMD case, Libxc runs on the CPU, and the density is moved back and forth between host and device. So I don't think this is a necessity.

mfherbst: Ok, true, thanks.

PR #1132, opened 2025-08-14
GPU| reduce bubbles and kernel overhead

abussy:
This PR introduces 2 minor changes when looping over columns of wave functions:

1) There is no need for device synchronization in the loop over the bands of $\Psi$. Each band is treated independently, and there is no risk of race condition. All it does is create bubbles in the GPU execution.

2) Applying `H.fourier_op.multiplier` band by band is not the most efficient on the GPU. There is overhead in launching many small kernels instead of a large one. Moreover, GPU parallelism is better exploited with large arrays.

I did not spot any performance impact on the CPU side. On NVIDIA GPUs, I observe a general speedup of ~10% on the SCF.

### DIFF
--- a/src/terms/Hamiltonian.jl
+++ b/src/terms/Hamiltonian.jl
@@ -168,9 +168,8 @@ end
         end
     end
 
-    @timing "local" begin
-        Hψ .+= H.fourier_op.multiplier .* ψ
-    end
+    # Kinetic term
+    Hψ .+= H.fourier_op.multiplier .* ψ
 
     # Apply the nonlocal operator.
     if !isnothing(H.nonlocal_op)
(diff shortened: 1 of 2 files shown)


antoine-levitt:
Great! Rename the timing to just local then (and don't have a timing for kinetic, it should be negligible)

src/terms/Hamiltonian.jl
```diff
@@ -167,8 +166,10 @@ end
         if Threads.threadid() == 1
             merge!(DFTK.timer, to; tree_point=[t.name for t in DFTK.timer.timer_stack])
         end
+    end
 
-        synchronize_device(H.basis.architecture)
+    @timing "local" begin
```
mfherbst: I think @antoine-levitt  wanted to not have *any* timing for this part, which sounds reasonable to me

abussy: Right, my bad

PR #1133, opened 2025-08-15
Include missing SCF metadata fields in forwarddiff_rules.jl

shivang57721:


antoine-levitt:
Thanks! Bit annoying that we have to update that list... Couldn't we just include all fields (assume they're non-differentiable by default) and just overload those that are differentiable? But I guess it's better to have something just error out rather than not propagate derivative information. What we can do though is to add a test that the derivative of the scfres has all the fields that the primal scfres does.

mfherbst:
@shivang57721  @niklasschmitz  Can one of you either add a test as described above or open an issue with a tests label to make sure @antoine-levitt's idea does not get forgotten ?

test/forwarddiff.jl
```diff
@@ -260,3 +260,60 @@ end
...
+    function construct_dual(psp::PspHgh{T}) where {T <: AbstractFloat}
+        PspHgh(psp.Zion,
+            to_dual(psp.rloc),
+            to_dual(psp.cloc),
+            psp.lmax,
+            to_dual(psp.rp),
+            [to_dual.(hl) for hl in psp.h],
+            psp.identifier,
+            psp.description)
+    end
```
niklasschmitz: Can we avoid these extra definitions if only used in this test?

test/forwarddiff.jl
```diff
@@ -260,3 +260,60 @@ end
...
+                    to_dual(model.positions);
+                    model.model_name,
+                    model.n_electrons,
+                    magnetic_moments=[],
+                    terms=model.term_types,
+                    temperature=to_dual(model.temperature),
+                    model.smearing,
+                    model.spin_polarization,
+                    model.symmetries,
+                    disable_electrostatics_check=true)
```
niklasschmitz: What should work is using the short-hand update constructor, something like
```
T = ForwardDiff.Tag(:mytag, Float64)
x_dual = ForwardDiff.Dual{T}(1.0, 1.0)
model_dual = Model(model; lattice=x_dual  * model.lattice)
```
Using `ForwardDiff.Tag` (like in https://github.com/JuliaDiff/ForwardDiff.jl/blob/d9afcbed61c2975b2b287877655b7b01b1635eff/src/derivative.jl#L12) should make it play correctly with other Duals, as an internal global tag-counter in ForwardDiff.jl will be increased (https://github.com/JuliaDiff/ForwardDiff.jl/blob/d9afcbed61c2975b2b287877655b7b01b1635eff/src/config.jl#L16)

test/forwarddiff.jl
```diff
@@ -260,3 +260,60 @@ end
...
+    # Construct the primal and dual basis
+    basis = PlaneWaveBasis(model; Ecut=5, kgrid=(1,1,1))
+    basis_dual = PlaneWaveBasis(model_dual; Ecut=to_dual(5.0), kgrid=(1,1,1))
+
+    # Compute scfres with primal and dual basis
+    scfres_dual = self_consistent_field(basis_dual; tol=1e-5)
+    scfres = self_consistent_field(basis; tol=1e-5)
+    
+    # Check that scfres_dual has the same parameters as scfres
+    @test isempty(setdiff(keys(scfres), keys(scfres_dual)))
```
niklasschmitz, line 404: Nice setdiff idea!

test/forwarddiff.jl
shivang57721: Ok done! 
It wasn't working with `T = ForwardDiff.Tag(:mytag, Float64)` (ForwardDiff couldn't distinguish between the different levels), but the test worked with `T = ForwardDiff.Tag{UnionAll, Int64}`. Not sure exactly why.

niklasschmitz: Thanks!
Ah yes my bad, the correct form is more complicated: I missed `typeof` which is present in the ForwardDiff.derivative implementation linked above. The form is then:
```
T = typeof(ForwardDiff.Tag(:mytag, Float64))
x_dual = ForwardDiff.Dual{T}(1.0, 1.0)
```
I just quickly added this in a commit here

PR #1139, opened 2025-08-28
Protect DFTK citation

Technici4n:
Else everyone copy-pasting this citation has to do this.


Technici4n:
Your arxiv reference is already there, no? https://docs.dftk.org/stable/publications/

mfherbst:
> EDIT: Or did you mean a .bib entry?

yes I meant the .bib file. The idea is that all papers dealing with algorithms of DFTK are in that file.

PR #1140, opened 2025-09-04
Update to PDOS and projector computations

fsicignano:
The new version of dos.jl computes the projectors as the fundamental object and the projections as a consequence. The documentation has been updated. 
The new get_pdos function allows the user to handle the result of the compute_pdos effectively for subsequent manipulations.
If just the plot is needed, then the plot_pdos will handle everything, following the same pattern as plot_dos.

### DIFF
--- a/src/postprocess/dos.jl
+++ b/src/postprocess/dos.jl
@@ -77,8 +77,9 @@ Overview of inputs:
 
 Overview of outputs:
 - `pdos`: 3D array of PDOS, pdos[iε_idx, iproj, σ] = PDOS at energy εs[iε_idx] for projector iproj and spin σ
-- `projector_labels` : Vector of tuples (iatom, species, label, n, l, m) for each projector, that maps the iproj index to the corresponding atomic orbital. 
-   For details see the documentation for `atomic_orbital_projectors`.
+- `projector_labels` : Vector of tuples (iatom, species, label, n, l, m) for each projector, 
+     that maps the iproj index to the corresponding atomic orbital. 
+     For details see the documentation for `atomic_orbital_projectors`.
 
 Notes: 
 - The atomic orbital projectors are taken from the pseudopotential files used to build the atoms. 
@@ -137,9 +138,9 @@ Can be called with an orbital NamedTuple and returns a boolean
 end
 function (s::OrbitalManifold)(orb)
     iatom_match    = isnothing(s.iatom)   || (s.iatom == orb.iatom)
-    species_match  = isnothing(s.species) || (s.species == orb.species) || (orb.species == s.species) #TODO: https://github.com/JuliaMolSim/AtomsBase.jl/issues/139
+    # See JuliaMolSim/AtomsBase.jl#139 why both species equalities are needed
+    species_match  = isnothing(s.species) || (s.species == orb.species) || (orb.species == s.species)
     label_match    = isnothing(s.label)   || (s.label == orb.label)
-
     iatom_match && species_match && label_match
 end
 
@@ -151,7 +152,8 @@ Build the matrices of projectors onto the pseudoatomic orbitals.
     projector[ik][iG, iproj] = 1/√Ω FT{ ϕperₙₗₘ(. - Rᵢ) }(k+G) + orthogonalization
 
  where Ω is the unit cell volume, ϕperₙₗₘ(. - Rᵢ) is the periodized pseudoatomic orbital (n, l, m) centered at Rᵢ
-  and iproj is the index corresponding to atom i and the quantum numbers (n, l, m). This correspondance is recorded in `labels`.
+  and iproj is the index corresponding to atom i and the quantum numbers (n, l, m). 
+  This correspondance is recorded in `labels`.
 
 The projectors are computed by decomposition into a form factor multiplied by a structure factor:
   FT{ ϕperₙₗₘ(. - Rᵢ) }(k+G) = Fourier transform of periodized atomic orbital ϕₙₗₘ (form factor)
@@ -159,16 +161,22 @@ The projectors are computed by decomposition into a form factor multiplied by a 
   
 Overview of inputs: 
 - `positions` : Positions of the atoms in the unit cell. Default is model.positions.
-- `isonmanifold` (opt) : (see notes below) OrbitalManifold struct to select only a subset of orbitals for the computation.
+- `isonmanifold` (opt) : (see notes below) OrbitalManifold struct to select only a subset of orbitals 
+     for the computation.
 
 Overview of outputs:
 - `projectors`: Vector of matrices of projectors
 - `labels`: Vector of NamedTuples containing iatom, species, n, l, m and orbital name for each projector
 
 Notes: 
-- The orbitals used for the projectors are all orthogonalized against each other. This corresponds to ortho-atomic projectors in Quantum Espresso.
-- 'n' in labels is not exactly the principal quantum number, but rather the index of the radial function in the pseudopotential. As an example, if the only S orbitals in the pseudopotential are 3S and 4S, then those are indexed as n=1, l=0 and n=2, l=0 respectively.
-- Use 'isonmanifold' kwarg with caution, since the resulting projectors would be orthonormalized only against the manifold basis. Most applications require the whole projectors basis to be orthonormal instead.
+- The orbitals used for the projectors are all orthogonalized against each other. 
+    This corresponds to ortho-atomic projectors in Quantum Espresso.
+- 'n' in labels is not exactly the principal quantum number, but rather the index of the radial function 
+    in the pseudopotential. As an example, if the only S orbitals in the pseudopotential are 3S and 4S, 
+    then those are indexed as n=1, l=0 and n=2, l=0 respectively.
+- Use 'isonmanifold' kwarg with caution, since the resulting projectors would be 
+    orthonormalized only against the manifold basis. 
+    Most applications require the whole projectors basis to be orthonormal instead.
 """
 function atomic_orbital_projectors(basis::PlaneWaveBasis{T};
                                    isonmanifold = l -> true,
@@ -192,7 +200,9 @@ function atomic_orbital_projectors(basis::PlaneWaveBasis{T};
             form_factors_l = build_form_factors(fun, l, G_plus_k_all_cart)
             for ik in 1:length(G_plus_k_all_cart)
                structure_factor = [cis2pi(-dot(positions[iatom], p)) for p in G_plus_k_all[ik]]
-               projectors[ik] = hcat(projectors[ik], form_factors_l[ik] .* structure_factor ./ sqrt(basis.model.unit_cell_volume))
+               projectors[ik] = hcat(projectors[ik], 
+                                     form_factors_l[ik] .* structure_factor 
+                                      ./ sqrt(basis.model.unit_cell_volume))
             end
             for m in -l:l
                 push!(labels, (; iatom, atom.species, n, l, m, label))
@@ -228,7 +238,8 @@ end
 """
     sum_pdos(pdos_res, manifolds)
 
-This function extracts and sums up all the PDOSes, directly from the output of the `compute_pdos` function, that match any of the manifolds.
+This function extracts and sums up all the PDOSes, directly from the output of the `compute_pdos` function, 
+  that match any of the manifolds.
 
 Overview of inputs:
 - `pdos_res`: Whole output from compute_pdos.


ext/DFTKPlotsExt.jl
```diff
@@ -141,56 +141,41 @@ function plot_ldos(basis, eigenvalues, ψ; εF=nothing, unit=u"hartree",
...
-                   temperature=basis.model.temperature,
-                   smearing=basis.model.smearing,
-                   εrange=default_band_εrange(eigenvalues; εF),
-                   n_points=1000, p=nothing, kwargs...)
+function plot_pdos(basis::PlaneWaveBasis{T}, eigenvalues, ψ; iatom, label=nothing,
+                      positions=basis.model.positions,
+                      smearing=basis.model.smearing, 
+                      temperature=basis.model.temperature,
+                      εF=nothing, unit=u"hartree",
+                      εrange=default_band_εrange(eigenvalues; εF),
```
Technici4n: Can this be reordered to match the `plot_dos` order? (`εF`, `unit`, `temperature`, `smearing`, `εrange`)

ext/DFTKPlotsExt.jl
```diff
@@ -141,56 +141,41 @@ function plot_ldos(basis, eigenvalues, ψ; εF=nothing, unit=u"hartree",
...
 
-function plot_pdos(basis, eigenvalues, ψ, i, l,
-                   psp, position, el::Symbol;
-                   εF=nothing, unit=u"hartree",
-                   temperature=basis.model.temperature,
-                   smearing=basis.model.smearing,
-                   εrange=default_band_εrange(eigenvalues; εF),
-                   n_points=1000, p=nothing, kwargs...)
+function plot_pdos(basis::PlaneWaveBasis{T}, eigenvalues, ψ; iatom, label=nothing,
+                      positions=basis.model.positions,
```
Technici4n: Remove the 3 leading spaces to ensure alignment between the lines.

ext/DFTKPlotsExt.jl
```diff
@@ -141,56 +141,41 @@ function plot_ldos(basis, eigenvalues, ψ; εF=nothing, unit=u"hartree",
...
-            for i = 1:DFTK.count_n_pswfc_radial(psp, l)
-                plot_pdos(basis, scfres_unfold.eigenvalues, scfres_unfold.ψ,
-                          i, l, psp, position, el; scfres.εF, p, kwargs...)
-            end
-        end
+    p = Plots.plot(; kwargs...)
+    spinlabels = spin_components(basis.model)
+    colors = [:blue, :red]
+    pdos = DFTK.get_pdos(compute_pdos(εs, basis, ψ, eigenvalues;
+                    positions, temperature, smearing); iatom, label)
```
Technici4n: Can you fix the alignment here? (`positions` at the same level as `εs`, and probably putting `iatom` and `label` on another line).

ext/DFTKPlotsExt.jl
```diff
@@ -141,56 +141,41 @@ function plot_ldos(basis, eigenvalues, ψ; εF=nothing, unit=u"hartree",
...
 plot_ldos(scfres; kwargs...) = plot_ldos(scfres.basis, scfres.eigenvalues, scfres.ψ; scfres.εF, kwargs...)
 
-function plot_pdos(basis, eigenvalues, ψ, i, l,
-                   psp, position, el::Symbol;
-                   εF=nothing, unit=u"hartree",
-                   temperature=basis.model.temperature,
-                   smearing=basis.model.smearing,
-                   εrange=default_band_εrange(eigenvalues; εF),
-                   n_points=1000, p=nothing, kwargs...)
+function plot_pdos(basis::PlaneWaveBasis{T}, eigenvalues, ψ; iatom, label=nothing,
```
Technici4n, line 144: Since the `iatom` is mandatory, I would suggest making it a positional argument rather than a keyword argument.

src/common/ortho.jl
```diff
@@ -7,3 +7,17 @@
         return x[:, 1:size(φk, 2)]
     end
 end
+
+@timing function ortho_lowdin(φk::ArrayType) where {ArrayType <: AbstractArray}
+    S = φk' * φk
+    evals, evecs = eigen(S)
+    # Check for linear dependence: eigenvalues close to zero
+    tol = maximum(size(φk)) * eps(real(eltype(φk))) * maximum(abs, evals)
```
Technici4n: Where are these tolerances coming from?

src/common/ortho.jl
```diff
@@ -7,3 +7,17 @@
...
+    if any(abs.(evals) .< tol)
+        error("Input vectors are linearly dependent or nearly so (small eigenvalues detected).")
```
Technici4n: If we keep this check, it should print the eigenvalues and the tolerances.

src/postprocess/dos.jl
```diff
@@ -16,7 +16,7 @@
 Total density of states at energy ε
 """
 function compute_dos(ε, basis, eigenvalues; smearing=basis.model.smearing,
-                     temperature=basis.model.temperature)
+                     temperature=basis.model.temperature)
```
Technici4n: whitespace change

src/postprocess/dos.jl
```diff
@@ -32,6 +32,7 @@ function compute_dos(ε, basis, eigenvalues; smearing=basis.model.smearing,
     end
     D = mpi_sum(D, basis.comm_kpts)
 end
+
```
Technici4n: whitespace change

src/postprocess/dos.jl
```diff
@@ -136,4 +272,4 @@ function plot_dos end
 
 function plot_ldos end
 
-function plot_pdos end
+function plot_pdos end
```
Technici4n: whitespace change (probably a line ending issue)

src/postprocess/dos.jl
```diff
@@ -68,65 +69,200 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+     - labels          : structure containing iatom, species, n, l, m and orbital name for each projector
+
+Notes: 
+
+- 'n' is not exactly the principal quantum number, but rather the index of the radial function in the pseudopotential. As an example, if the pseudopotential contains the 3S and 4S orbitals, then those are indexed as n=1, l=0 and n=2, l=0 respectively.
+- Use 'manifold' kwarg with caution, since the resulting projectors would be orthonormalized only against the manifold basis. Most applications require the whole projectors basis to be orthonormal instead.
+"""
+function build_projectors(basis::PlaneWaveBasis{T};
+                          manifold = nothing,  #Should we allow to take and orthogonalize only the manifold?
+                          positions = basis.model.positions
```
Technici4n: I would say that we can remove these two kwargs for now? Being able to give a manifold is a bit dangerous, and overriding the positions seems unnecessary.

src/postprocess/dos.jl
```diff
@@ -68,65 +69,200 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+    Output:
+     - projectors      : vector of matrices of projectors
+     - labels          : structure containing iatom, species, n, l, m and orbital name for each projector
+
+Notes: 
+
+- 'n' is not exactly the principal quantum number, but rather the index of the radial function in the pseudopotential. As an example, if the pseudopotential contains the 3S and 4S orbitals, then those are indexed as n=1, l=0 and n=2, l=0 respectively.
+- Use 'manifold' kwarg with caution, since the resulting projectors would be orthonormalized only against the manifold basis. Most applications require the whole projectors basis to be orthonormal instead.
+"""
+function build_projectors(basis::PlaneWaveBasis{T};
```
Technici4n: `build_projectors` is a bit too generic of a name, as this could be confused with the beta projectors (also in the pseudopotential). What about `build_atomic_orbital_projectors`?

src/postprocess/dos.jl
```diff
@@ -68,65 +69,200 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
     G_plus_k_all = [Gplusk_vectors(basis, basis.kpoints[ik])
                     for ik = 1:length(basis.kpoints)]
     G_plus_k_all_cart = [map(recip_vector_red_to_cart(basis.model), gpk) 
                          for gpk in G_plus_k_all]
 
-    # Build form factors of pseudo-wavefunctions centered at 0.
-    fun(p) = eval_psp_pswfc_fourier(psp, i, l, p)
-    # form_factors_all[ik][p,m]
-    form_factors_all = build_form_factors(fun, l, G_plus_k_all_cart)
+    psps = Vector{NormConservingPsp}(undef, length(basis.model.atoms))
```
Technici4n: This array doesn't seem to be necessary.

src/postprocess/dos.jl
```diff
@@ -68,65 +69,200 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+    for (iatom, atom) in enumerate(basis.model.atoms)
+        if !isnothing(manifold)
+            if manifold[1] != Symbol(atom.species) && manifold[1] != iatom
+               continue # Skip atoms that do not match the manifold species, if any is provided
+            end
+        end
+        psps[iatom] = atom.psp
+        for l in 0:psps[iatom].lmax
+            for n in 1:DFTK.count_n_pswfc_radial(psps[iatom], l)
+                label = psps[iatom].pswfc_labels[l+1][n]
```
Technici4n: This should not use the `pswfc_labels` field directly since it's not part of the `NormConservingPsp.jl` interface. Rather, a new method should be added, see `NormConservingPsp.jl`. For example `get_pswfc_label(psp, i::Integer, l::Integer)`. It should also be added to the documentation in NCPsp.jl

src/postprocess/dos.jl
```diff
@@ -68,65 +69,200 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
                     for ik = 1:length(basis.kpoints)]
     G_plus_k_all_cart = [map(recip_vector_red_to_cart(basis.model), gpk) 
                          for gpk in G_plus_k_all]
 
-    # Build form factors of pseudo-wavefunctions centered at 0.
-    fun(p) = eval_psp_pswfc_fourier(psp, i, l, p)
-    # form_factors_all[ik][p,m]
-    form_factors_all = build_form_factors(fun, l, G_plus_k_all_cart)
+    psps = Vector{NormConservingPsp}(undef, length(basis.model.atoms))
+    labels = []
```
Technici4n, line 191: Here you can already construct the projector matrices (without filling them) with `projectors = [Matrix{Complex{T}}(undef, length(G_plus_k), nprojs) for G_plus_k in G_plus_k_all]`.

src/postprocess/dos.jl
```diff
@@ -68,65 +69,200 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+            for n in 1:DFTK.count_n_pswfc_radial(psps[iatom], l)
+                label = psps[iatom].pswfc_labels[l+1][n]
+                if !isnothing(manifold) && lowercase(manifold[2]) != lowercase(label)
+                    continue # Skip atoms that do not match the manifold species, if any is provided
+                end
+                fun(p) = eval_psp_pswfc_fourier(psps[iatom], n, l, p)
+                form_factors_l = build_form_factors(fun, l, G_plus_k_all_cart)
+                for ik in 1:length(G_plus_k_all_cart)
+                   form_factors[ik] = hcat(form_factors[ik], form_factors_l[ik])  # Concatenate the form factors for this l
+                end
```
Technici4n: You can directly multiply the form factor with the structure factor here, and put it in the already allocated (see other comment) `projectors[ik][:,iproj:iproj+2m+1] .= form_factors[ik] .* structure_factors ./ sqrt(...)`. This will require keeping track of the `iproj` as well.

src/postprocess/dos.jl
```diff
@@ -68,65 +69,200 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+        end
+
+        @assert size(proj_vectors, 2) == nprojs "Projection matrix size mismatch: $(size(proj_vectors)) != $nprojs"
+        # At this point proj_vectors is a matrix containing all orbital projectors from all atoms. 
+        #   What we want is to have them all orthogonal, to avoid double counting in the Hubbard U term contribution.
+        #   We use Lowdin orthogonalization to minimize the "identity loss" of individual orbital projectors after the orthogonalization
+        proj_vectors = ortho_lowdin(proj_vectors)  
+        
+        projectors[ik] = proj_vectors  
+    end
```
Technici4n: With the changes mentioned above, only the lowdin orthogonalization should still be needed here. You could do it with a list comprehension.

src/postprocess/dos.jl
```diff
@@ -68,65 +69,200 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+
+    return (;projectors, labels)
```
Technici4n: Usually you'd write `(; projectors, labels)` with a space after the `;`.

src/postprocess/dos.jl
```diff
@@ -68,65 +69,200 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+  and ϕilm is the atomic orbital for atom i, quantum numbers (n,l,m)
+
+    Input:
+     -> basis  : PlaneWaveBasis
+     -> ψ      : Wavefunction
+    Output:
+     -> projs  : Vector of matrices of projections
+     -> labels : NamedTuple containing iatom, species, n, l, m and orbital name for each projector
+Note: 'n' is not the principal quantum number, but the index of the radial wavefunction in the pseudopotential.
+          So for Si, the 3s orbital would have n=1, l=0, m=0, and the 3p orbital would have n=1, l=1, m=-1,0,1.
```
Technici4n: This could be added to the docstring of `build_projectors`, and then you just link to it for an explanation of the labels.

src/postprocess/dos.jl
```diff
@@ -68,65 +69,200 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+                          ) where {T}
+    projectors, labels = build_projectors(basis; manifold=manifold, positions=positions)
     projs = Vector{Matrix}(undef, length(basis.kpoints))
     for (ik, ψk) in enumerate(ψ)
-        structure_factor = [cis2pi(-dot(position, p)) for p in G_plus_k_all[ik]]
-        # TODO orthogonalize pseudo-atomic wave functions?
-        proj_vectors = structure_factor .* form_factors_all[ik] ./ sqrt(basis.model.unit_cell_volume)
-        projs[ik] = abs2.(ψk' * proj_vectors) # contract on p -> projs[ik][iband,m]
+       projs[ik] = abs2.(ψk' * projectors[ik])
+    end
```
Technici4n, line 233: This could be
```jl
    projections = map(zip(ψ, projectors)) do (ψk, projectorsk)
        abs2.(ψk' * projectorsk)
    end
```

src/postprocess/dos.jl
```diff
@@ -68,65 +69,200 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+    projs[ik][iband, iproj] = projsk[iband, iproj] = |<ψnk, ϕilm>|^2
+
+ where ψnk is the wavefunction for band iband at k-point kpt,
+  and ϕilm is the atomic orbital for atom i, quantum numbers (n,l,m)
+
+    Input:
+     -> basis  : PlaneWaveBasis
+     -> ψ      : Wavefunction
+    Output:
+     -> projs  : Vector of matrices of projections
```
Technici4n: Let's call this `projections`, not `projs`. The latter can be confused with the projectors.

src/postprocess/dos.jl
```diff
@@ -68,65 +69,200 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+This function extracts the required pdos from the output of the `compute_pdos` function. 
+
+    Input:
+     -> res         : Whole output from compute_pdos.
+     -> label (opt) : String for the whole orbital label. 
+     -> iatom (opt) : Atom number in the model.atoms vector. It becomes important for magnetically or spatially non-equivalent atoms. 
+    Output:
+     -> pdos        : Vector containing the pdos(ε).
+"""
+function get_pdos(pdos_res, projs::AbstractVector)
```
Technici4n: Is there a need for this more complicated version of `get_pdos` (compared to `get_pdos(pdos_res; iatom=nothing, label=nothing)`)? The handling of the `projs` is a bit confusing.

src/postprocess/dos.jl
fsicignano: About the positions, wouldn't it be useful for the computation of forces?

src/postprocess/dos.jl
fsicignano: Unfortunately at this level we still don't know nprojs, so we can't construct the projector matrices, at least not in the way we do that now.

src/pseudo/NormConservingPsp.jl
```diff
@@ -218,3 +218,5 @@ count_n_pswfc(psp::NormConservingPsp, l) = count_n_pswfc_radial(psp, l) * (2l +
 function count_n_pswfc(psp::NormConservingPsp)
     sum(l -> count_n_pswfc(psp, l), 0:psp.lmax; init=0)::Int
 end
+
+get_pswfc_label(psp::NormConservingPsp, i, l) = psp.pswfc_labels[l+1][i]
```
Technici4n: This new method needs to be documented in `NormConservingPsp.jl` as well.

Project.toml
```diff
@@ -7,6 +7,7 @@ version = "0.7.16"
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
 Artifacts = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
 AtomsBase = "a963bdd2-2df7-4f54-a1ee-49d51e6be12a"
+AtomsBuilder = "f5cc8831-eeb7-4288-8d9f-d6c1ddb77004"
```
Technici4n: Not used anywhere I think?

ext/DFTKPlotsExt.jl
```diff
@@ -141,56 +141,42 @@ function plot_ldos(basis, eigenvalues, ψ; εF=nothing, unit=u"hartree",
...
-                plot_pdos(basis, scfres_unfold.eigenvalues, scfres_unfold.ψ,
-                          i, l, psp, position, el; scfres.εF, p, kwargs...)
-            end
-        end
+    p = Plots.plot(; kwargs...)
+    spinlabels = spin_components(basis.model)
+    colors = [:blue, :red]
+    pdos = DFTK.sum_pdos(compute_pdos(εs, basis, ψ, eigenvalues;
+                                      positions, temperature, smearing), 
+                         [DFTK.OrbitalManifold(;iatom=iatom, label=label)])
```
Technici4n: We generally add a space after `;` in kwargs or named tuples:
```suggestion
                         [DFTK.OrbitalManifold(; iatom=iatom, label=label)])
```

src/postprocess/dos.jl
```diff
@@ -68,65 +68,186 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+     -> basis            : PlaneWaveBasis object from bands computation
+     -> ψ                : Wavefunction from the bands
+     -> eigenvalues      : Eigenvalues from the bands
+    Output:
+     -> pdos             : 3D array of PDOS, pdos[iε_idx, iproj, σ] = PDOS at energy εs[iε_idx] for projector iproj and spin σ
+     -> projector_labels : Vector of tuples (iatom, n, l, m) for each projector, that maps the iproj index to the corresponding atomic orbital (atom index, principal quantum number, angular momentum, magnetic quantum number)
+Notes: 
+ - All the information about projectors is taken from the PseudoPotential files used to build the atoms. 
+    There may be cases where the relevant data are missing for the desired projectors, despite being apparently declared.
+    As an example, it may happen that the PseudoPotential file does not have all projectors up to the l_max declared in the psp.
```
Technici4n: This is because the l_max refers to the beta projectors and not the pseudized atomic orbitals, so this remark might be confusing.

src/postprocess/dos.jl
```diff
@@ -68,65 +68,186 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+ - The pdos matrix has different projectors for each atom, even if they are of the same atom type. 
+    As such, the sum of all iproj columns for each σ yields the total DOS at each energy εs[iε_idx].
+    This is different from Quantum ESPRESSO, where the pdos for atoms of the same type are summed together 
+      even though they are printed separately (i.e. summing over all QE pdos from all output files does not yield the DOS).
```
Technici4n: Some remarks:
- The first sentence is a bit confusing to me. The pdos matrix doesn't really have projectors, rather it contains projections.
- It doesn't quite sum to the DOS since the projector set is never complete. (In practice it almost does in particular below the Fermi level)
- I am not sure about the remark wrt. QE. Are we sure that the factor of 2 in QE comes from the atom types and not from the spin?

src/postprocess/dos.jl
```diff
@@ -68,65 +68,186 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+function compute_pdos(εs, basis::PlaneWaveBasis{T}, ψ, eigenvalues; 
+                      positions=basis.model.positions,
+                      smearing=basis.model.smearing, 
+                      temperature=basis.model.temperature) where {T}
     if (temperature == 0) || smearing isa Smearing.None
         error("compute_pdos only supports finite temperature")
     end
     filled_occ = filled_occupation(basis.model)
+    
+    projections, projector_labels = atomic_orbital_projections(basis, ψ; positions=positions)
```
Technici4n:
```suggestion
    projections, projector_labels = atomic_orbital_projections(basis, ψ; positions)
```

src/postprocess/dos.jl
```diff
@@ -68,65 +68,186 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+            projsk = projections[ik]  
             @views for (iband, εnk) in enumerate(eigenvalues[ik])
-                enred = (εnk - iε) / temperature
-                D[i, :] .-= (filled_occ .* basis.kweights[ik] .* projk[iband, :]
-                             ./ temperature
-                             .* Smearing.occupation_derivative(smearing, enred))       
+                enred = (εnk - ε) / temperature
+                for iproj in 1:size(projsk, 2)
+                    projk = projsk[:, iproj]
+                    D[iε, iproj,σ] -= (filled_occ * basis.kweights[ik] * projk[iband]
```
Technici4n:
```suggestion
                    D[iε, iproj,σ] -= (filled_occ * basis.kweights[ik] * projsk[iband, iproj]
```

src/postprocess/dos.jl
```diff
@@ -68,65 +68,186 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+                                             ./ temperature
+                                             .* Smearing.occupation_derivative(smearing, enred))
```
Technici4n: Should be aligned with `filled_occ` above.

src/postprocess/dos.jl
```diff
@@ -68,65 +68,186 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+
+    return (; pdos, projector_labels, εs)
+end
+
+function compute_pdos(εs, bands; kwargs...)
+    compute_pdos(εs, bands.basis, bands.ψ, bands.eigenvalues; kwargs...)
+end
+
+"""
+Structure for manifold choice and projectors estraction. Fields:
```
Technici4n:
```suggestion
Structure for manifold choice and projectors extraction. Fields:
```

Technici4n: Should mention that the fields are optional, and only the fields that are given will be used for selection.

src/postprocess/dos.jl
```diff
@@ -68,65 +68,186 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+    -> iatom   : Int64 corresponding to the atom position in the atoms array.
+    -> species : Symbol for the Chemical Element as in ElementPsp.
+    -> label   : String with the orbital name, i.e.: "3S".
```
Technici4n: I wouldn't mention the types here necessarily, but rather in the struct

Technici4n: "i.e." means "that is". Here you probably meant e.g. ("for example").

src/postprocess/dos.jl
```diff
@@ -68,65 +68,186 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+Implemented function for OrbitalManifold can be applied to an orbital NamedTuple and returns a boolean
+    stating whether the orbital belongs to the manifold.
+"""
+@kwdef struct OrbitalManifold
+    iatom   = nothing
```
Technici4n: For example the type can be written here as `iatom::Union{Int, Nothing} = nothing`

src/postprocess/dos.jl
```diff
@@ -68,65 +68,186 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+    compute_pdos(εs, bands.basis, bands.ψ, bands.eigenvalues; kwargs...)
+end
+
+"""
+Structure for manifold choice and projectors estraction. Fields:
+    -> iatom   : Int64 corresponding to the atom position in the atoms array.
+    -> species : Symbol for the Chemical Element as in ElementPsp.
+    -> label   : String with the orbital name, i.e.: "3S".
+Implemented function for OrbitalManifold can be applied to an orbital NamedTuple and returns a boolean
+    stating whether the orbital belongs to the manifold.
```
Technici4n: This sentence is a bit confusing for someone who doesn't know what is happening.

src/postprocess/dos.jl
```diff
@@ -68,65 +68,186 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+Build the projectors matrices projsk for all k-points at the same time.
+         
+    projector[ik][:, iproj] = |ϕinlm>(kpt)
+
+ where ϕinlm is the atomic orbital for atom i, quantum numbers (n,l,m)
+   and iproj is the corresponding column index. The mapping is recorded in 'labels'.
+  
+    Input: 
+     - basis           : PlaneWaveBasis
+     - manifold  (opt) : (see notes below) tuple of (Atom, Orbital) to select only a subset of orbitals for the computation. 'Atom' must be either a Symbol or an Int64, 'Orbital' must be a String with the orbital name in uppercase.
```
Technici4n: outdated

src/postprocess/dos.jl
```diff
@@ -68,65 +68,186 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+     - positions (opt) : positions of the atoms in the unit cell
```
Technici4n: `(opt)` is confusing: it gives the impression that passing no positions at all is valid; I would rather say that it defaults to the positions in the model

src/postprocess/dos.jl
```diff
@@ -68,65 +68,186 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
 
-# Build atomic orbitals projections projs[ik][iband,m] = |<ψnk, ϕ>|^2 for a single atom.
-# TODO optimization ? accept a range of positions, in case we want to compute the PDOS
-# for all atoms of the same type (and reuse the computation of the atomic orbitals)
-function compute_pdos_projs(basis, eigenvalues, ψ, i, l, psp::NormConservingPsp, position)
-    # Precompute the form factors on all kpoints at once so we can better use the cache (memory-compute tradeoff).
-    # Revisit this (pass the cache around explicitly) if RAM becomes an issue.
+Notes: 
+
+- 'n' is not exactly the principal quantum number, but rather the index of the radial function in the pseudopotential. As an example, if the pseudopotential contains the 3S and 4S orbitals, then those are indexed as n=1, l=0 and n=2, l=0 respectively.
```
Technici4n: This is not quite correct since the pseudopotential can in principle also contain a 2S orbital. What you really mean is that if the **only** S orbitals in the pseudopotential are 3S and 4S then they have n=1 and n=2 respectively.

src/postprocess/dos.jl
```diff
@@ -68,65 +68,186 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+- Use 'manifold' kwarg with caution, since the resulting projectors would be orthonormalized only against the manifold basis. Most applications require the whole projectors basis to be orthonormal instead.
+"""
+function atomic_orbital_projectors(basis::PlaneWaveBasis{T};
+                                   ismanifold = nothing,  #Should we allow to take and orthogonalize only the manifold?
```
Technici4n: Make it `ismanifold = l -> true,` by default. This also means you don't have to check for nothing below.

src/postprocess/dos.jl
```diff
@@ -68,65 +68,186 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+                                   positions = basis.model.positions
+                                   ) where {T}
```
Technici4n: usually we put the `)` on the previous line

src/postprocess/dos.jl
```diff
@@ -68,65 +68,186 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
-        projs[ik] = abs2.(ψk' * proj_vectors) # contract on p -> projs[ik][iband,m]
+    projectors = [Matrix{Complex{T}}(undef, length(G_plus_k), 0) for G_plus_k in G_plus_k_all]
+    form_factors = [Matrix{Complex{T}}(undef, length(G_plus_k), 0)  for G_plus_k in G_plus_k_all_cart]
+    labels = []
+    for (iatom, atom) in enumerate(basis.model.atoms)
+        psp = atom.psp
+        for l in 0:psp.lmax
+            for n in 1:DFTK.count_n_pswfc_radial(psp, l)
+                label = DFTK.get_pswfc_label(psp, n, l)
+                if !isnothing(ismanifold) && !ismanifold((;iatom=iatom, species=Symbol(atom.species), label=label))
```
Technici4n: - no need for the `isnothing` check if the default is a lambda that always returns true.
- missing space after `;`
- I would pass `l` as well

src/postprocess/dos.jl
```diff
@@ -68,65 +68,186 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
-    form_factors_all = build_form_factors(fun, l, G_plus_k_all_cart)
-
-    projs = Vector{Matrix}(undef, length(basis.kpoints))
-    for (ik, ψk) in enumerate(ψ)
-        structure_factor = [cis2pi(-dot(position, p)) for p in G_plus_k_all[ik]]
-        # TODO orthogonalize pseudo-atomic wave functions?
-        proj_vectors = structure_factor .* form_factors_all[ik] ./ sqrt(basis.model.unit_cell_volume)
-        projs[ik] = abs2.(ψk' * proj_vectors) # contract on p -> projs[ik][iband,m]
+    projectors = [Matrix{Complex{T}}(undef, length(G_plus_k), 0) for G_plus_k in G_plus_k_all]
+    form_factors = [Matrix{Complex{T}}(undef, length(G_plus_k), 0)  for G_plus_k in G_plus_k_all_cart]
```
Technici4n: this variable doesn't seem necessary

src/postprocess/dos.jl
```diff
@@ -68,65 +68,186 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+                if !isnothing(ismanifold) && !ismanifold((;iatom=iatom, species=Symbol(atom.species), label=label))
+                    continue
+                end
+                fun(p) = eval_psp_pswfc_fourier(psp, n, l, p)
+                form_factors_l = build_form_factors(fun, l, G_plus_k_all_cart)
+                iproj = length(labels) + 1
+                for ik in 1:length(G_plus_k_all_cart)
+                   form_factors[ik] = hcat(form_factors[ik], form_factors_l[ik])  # Concatenate the form factors for this l
+                   structure_factor = [cis2pi(-dot(positions[iatom], p)) for p in G_plus_k_all[ik]]
+                   @assert length(structure_factor) == length(G_plus_k_all[ik]) "Structure factor length mismatch: $(length(structure_factor)) != $(length(G_plus_k))"
```
Technici4n: This assert is trivially always true.

src/postprocess/dos.jl
```diff
@@ -68,65 +68,186 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+    end
+
+    projectors = ortho_lowdin.(projectors)
+
+    return (;projectors, labels)
+end
+
+"""
+Build the projection matrices projsk for all k-points at the same time.
+  and ϕilm is the atomic orbital for atom i, quantum numbers (n,l,m)
```
Technici4n: This second line seems to be coming out of nowhere.

src/postprocess/dos.jl
```diff
@@ -68,65 +68,186 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+
+See documentation for 'atomic_orbital_projectors'
+"""
+function atomic_orbital_projections(basis::PlaneWaveBasis{T}, ψ;
+                                    ismanifold=nothing,
+                                    positions = basis.model.positions           
+                                   ) where {T}
```
Technici4n: Move to end of previous line

src/postprocess/dos.jl
```diff
@@ -68,65 +68,186 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+    projectors, labels = atomic_orbital_projectors(basis; ismanifold=ismanifold, positions=positions)
```
Technici4n: No need to repeat the value variable name when it's the same as the keyword name
```suggestion
    projectors, labels = atomic_orbital_projectors(basis; ismanifold, positions)
```

Project.toml
fsicignano: I am a little puzzled, since I didn't change this file in the last commit. Also, I don't have a file Project.toml that contains this four package references in my local repository, so I don't know where this change comes from.

src/postprocess/dos.jl
fsicignano: I might change it with: "- All the information about projectors is taken from the PseudoPotential files used to build the atoms. 
    There may be cases where the relevant data are missing for the desired projectors, despite being apparently declared.
    As an example, it may happen that the PseudoPotential file does not have all atomic projectors up to the l_max declared in the psp, 
    since such l_max refers instead to the beta projectors."

src/postprocess/dos.jl
fsicignano: I am removing this comment entirely, since it is still not entirely clear to me what is going on here with the pdos/dos comparison within QE

src/postprocess/dos.jl
fsicignano: In fact it is not

src/postprocess/dos.jl
fsicignano: removed

src/postprocess/dos.jl
```diff
@@ -68,65 +68,180 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
 end
 
 """
-Projected density of states at energy ε for an atom with given i and l.
+Compute the projected density of states (PDOS) for all atoms and orbitals.
+
+    Input: 
+     -> εs               : Vector of energies at which the PDOS will be computed
+     -> basis            : PlaneWaveBasis object from bands computation
+     -> ψ                : Wavefunction from the bands
```
fsicignano: Here "from the bands" is an indication to get the most physical results but in principle also scfres is fine. Should we leave it like that?

src/postprocess/dos.jl
```diff
@@ -68,65 +68,180 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+     -> eigenvalues      : Eigenvalues from the bands
+    Output:
+     -> pdos             : 3D array of PDOS, pdos[iε_idx, iproj, σ] = PDOS at energy εs[iε_idx] for projector iproj and spin σ
+     -> projector_labels : Vector of tuples (iatom, n, l, m) for each projector, that maps the iproj index to the corresponding atomic orbital (atom index, principal quantum number, angular momentum, magnetic quantum number)
```
fsicignano: Here we are showing only a part of all the projector_labels elements (species and label are missing). Should we keep it brief for clarity, leaving the full explanation in the documentation for atomic_orbital_projectors, or put them all here as well? Maybe we could name them all and then point to the other documentation for a full explanation.

src/postprocess/dos.jl
fsicignano: Should I remove it? But then how does people know that they can use it as a function as well?

src/postprocess/dos.jl
fsicignano: But if I set for example the iatom to be an Int64 it gives me an error if I set it as nothing. Should I set it as either Int64 or Nothing?

src/postprocess/dos.jl
```diff
@@ -68,65 +68,180 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+    projections = map(zip(ψ, projectors)) do (ψk, projectorsk)
+        abs2.(ψk' * projectorsk)
     end
 
-    projs
+    return (; projections, labels)
+end
+
+"""
+This function extracts the required pdos from the output of the `compute_pdos` function.
```
fsicignano: Should we specify that it sums up all the pdos that correspond to the given manifolds? Or maybe it is clear from the name and final structure of the pdos object

src/postprocess/dos.jl
```diff
@@ -68,65 +68,180 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+end
+
+"""
+This function extracts the required pdos from the output of the `compute_pdos` function. 
+
+    Input:
+     -> pdos_res  : Whole output from compute_pdos.
+     -> manifolds : Vector of OrbitalManifolds to select the desired projectors pdos.
+    Output:
+     -> pdos      : Vector containing the pdos(ε).
```
fsicignano: Should we specify that it is actually two vectors, one for each spin?

examples/collinear_magnetism.jl
```diff
@@ -21,7 +21,7 @@ bulk(:Fe)
 
 Ecut  = 15         # kinetic energy cutoff in Hartree
 kgrid = [3, 3, 3]  # k-point grid (Regular Monkhorst-Pack grid)
-pseudopotentials = PseudoFamily("cp2k.nc.sr.lda.v0_1.largecore.gth")
+pseudopotentials = PseudoFamily("dojo.nc.sr.pbe.v0_4_1.standard.upf")
```
fsicignano, line 24: I had to make this change since the original pseudopotential doesn't contain atomic orbitals, so the pdos couldn't be computed

ext/DFTKPlotsExt.jl
```diff
@@ -141,56 +141,44 @@ function plot_ldos(basis, eigenvalues, ψ; εF=nothing, unit=u"hartree",
...
 plot_ldos(scfres; kwargs...) = plot_ldos(scfres.basis, scfres.eigenvalues, scfres.ψ; scfres.εF, kwargs...)
 
-function plot_pdos(basis, eigenvalues, ψ, i, l,
-                   psp, position, el::Symbol;
+function plot_pdos(basis::PlaneWaveBasis{T}, eigenvalues, ψ; iatom, label=nothing,
+                   positions=basis.model.positions,
                    εF=nothing, unit=u"hartree",
                    temperature=basis.model.temperature,
                    smearing=basis.model.smearing,
+                   colors = [:blue, :red],
```
fsicignano, line 149: I think it makes sense to have it here, differently to what was done before, because it could be very common to overlay different pdos plots together or with the dos and then one has to change the colors to make them differ.

ext/DFTKPlotsExt.jl
```diff
@@ -141,56 +141,44 @@ function plot_ldos(basis, eigenvalues, ψ; εF=nothing, unit=u"hartree",
...
                    εrange=default_band_εrange(eigenvalues; εF),
-                   n_points=1000, p=nothing, kwargs...)
+                   n_points=1000, p=nothing, kwargs...) where {T}
     eshift = something(εF, 0.0)
     εs = range(austrip.(εrange)..., length=n_points)
+    n_spin = basis.model.n_spin_components
+    isnothing(iatom) ? species = "all atoms" : species = basis.model.atoms[iatom].species
+    isnothing(label) ? orb_name = "all orbitals" : orb_name = label
```
Technici4n:
```suggestion
    species = isnothing(iatom) ? "all atoms" : basis.model.atoms[iatom].species
    orb_name = isnothing(label) ? "all orbitals" : label
```

src/postprocess/dos.jl
Technici4n: I think that we don't even need to document `basis`, `ψ`, `eigenvalues` (they are self-explanatory)

src/postprocess/dos.jl
Technici4n: Being approximate/slightly incorrect is worse than not writing anything, so I would at least give the full tuple, and link to `atomic_orbital_projectors` for a deeper explanation.

src/postprocess/dos.jl
```diff
@@ -68,65 +68,180 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+    Input: 
+     -> εs               : Vector of energies at which the PDOS will be computed
+     -> basis            : PlaneWaveBasis object from bands computation
+     -> ψ                : Wavefunction from the bands
+     -> eigenvalues      : Eigenvalues from the bands
+    Output:
+     -> pdos             : 3D array of PDOS, pdos[iε_idx, iproj, σ] = PDOS at energy εs[iε_idx] for projector iproj and spin σ
+     -> projector_labels : Vector of tuples (iatom, n, l, m) for each projector, that maps the iproj index to the corresponding atomic orbital (atom index, principal quantum number, angular momentum, magnetic quantum number)
+Notes: 
+ - All the information about projectors is taken from the PseudoPotential files used to build the atoms.
```
Technici4n:
```suggestion
 - The atomic orbital projectors are taken from the pseudopotential files used to build the atoms. 
```

src/postprocess/dos.jl
```diff
@@ -68,65 +68,180 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+     -> ψ                : Wavefunction from the bands
+     -> eigenvalues      : Eigenvalues from the bands
+    Output:
+     -> pdos             : 3D array of PDOS, pdos[iε_idx, iproj, σ] = PDOS at energy εs[iε_idx] for projector iproj and spin σ
+     -> projector_labels : Vector of tuples (iatom, n, l, m) for each projector, that maps the iproj index to the corresponding atomic orbital (atom index, principal quantum number, angular momentum, magnetic quantum number)
+Notes: 
+ - All the information about projectors is taken from the PseudoPotential files used to build the atoms. 
+    There may be cases where the relevant data are missing for the desired projectors, despite being apparently declared.
+    As an example, it may happen that the PseudoPotential file does not have all atomic projectors up to the l_max declared in the psp, 
+    since such l_max refers instead to the beta projectors.
```
Technici4n:
```suggestion
    In doubt, consult the pseudopotential file for the list of available atomic orbitals.
```

src/postprocess/dos.jl
```diff
@@ -68,65 +68,180 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
             @views for (iband, εnk) in enumerate(eigenvalues[ik])
-                enred = (εnk - iε) / temperature
-                D[i, :] .-= (filled_occ .* basis.kweights[ik] .* projk[iband, :]
-                             ./ temperature
-                             .* Smearing.occupation_derivative(smearing, enred))       
+                enred = (εnk - ε) / temperature
+                for iproj in 1:size(projsk, 2)
+                    D[iε, iproj, σ] -= (filled_occ * basis.kweights[ik] * projsk[iband, iproj]
+                                        ./ temperature
+                                        .* Smearing.occupation_derivative(smearing, enred))
```
Technici4n:
```suggestion
                                        / temperature
                                        * Smearing.occupation_derivative(smearing, enred))
```

src/postprocess/dos.jl
```diff
@@ -68,65 +68,180 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
-    # TODO do the symmetrization instead of unfolding    
-    scfres = unfold_bz(scfres)
-    compute_pdos(ε, scfres.basis, scfres.eigenvalues, scfres.ψ, i, l, psp, position; kwargs...)
+"""
+Structure for manifold choice and projectors extraction. Fields:
+    -> iatom   : Atom position in the atoms array.
+    -> species : Chemical Element as in ElementPsp.
+    -> label   : Orbital name, e.g.: "3S".
+All fields are optional, only the given ones will be used for selection.
+Implemented function for OrbitalManifold can be applied to an orbital NamedTuple and returns a boolean
```
Technici4n:
```suggestion
Can be called with an orbital NamedTuple and returns a boolean
```

src/postprocess/dos.jl
```diff
@@ -68,65 +68,180 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
-# Build atomic orbitals projections projs[ik][iband,m] = |<ψnk, ϕ>|^2 for a single atom.
-# TODO optimization ? accept a range of positions, in case we want to compute the PDOS
-# for all atoms of the same type (and reuse the computation of the atomic orbitals)
-function compute_pdos_projs(basis, eigenvalues, ψ, i, l, psp::NormConservingPsp, position)
-    # Precompute the form factors on all kpoints at once so we can better use the cache (memory-compute tradeoff).
-    # Revisit this (pass the cache around explicitly) if RAM becomes an issue.
+    iatom_match && species_match && label_match
+end
+"""
+Build the projectors matrices projsk for all k-points at the same time.
```
Technici4n: Add a note somewhere that the projectors are orthogonalized (corresponds to ortho-atomic in QE?)

src/postprocess/dos.jl
```diff
@@ -68,65 +68,180 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+     - projectors       : Vector of matrices of projectors
+     - labels           : Vector of NamedTuples containing iatom, species, n, l, m and orbital name for each projector
+
+Notes: 
+
+- 'n' in labels is not exactly the principal quantum number, but rather the index of the radial function in the pseudopotential. As an example, if the only S orbitals in the pseudopotential are 3S and 4S, then those are indexed as n=1, l=0 and n=2, l=0 respectively.
+- Use 'ismanifold' kwarg with caution, since the resulting projectors would be orthonormalized only against the manifold basis. Most applications require the whole projectors basis to be orthonormal instead.
+"""
+function atomic_orbital_projectors(basis::PlaneWaveBasis{T};
+                                   ismanifold = l -> true,  #Should we allow to take and orthogonalize only the manifold?
```
Technici4n:
```suggestion
                                   ismanifold = l -> true,
```

src/postprocess/dos.jl
```diff
@@ -68,65 +68,180 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+    for σ in 1:size(pdos_res.pdos, 3)
+        pdos_values = zeros(Float64, length(pdos_res.εs))
+        for ismanifold in manifolds
+            for (j, orb) in enumerate(pdos_res.projector_labels)
+                if ismanifold(orb)
+                    pdos_values += pdos_res.pdos[:, j, σ]
+                end
+            end
+        end
+        push!(pdos, pdos_values)
```
Technici4n:
```suggestion
        for (j, orb) in enumerate(pdos_res.projector_labels)
            if any(manifolds.(orb))
                pdos_values += pdos_res.pdos[:, j, σ]
            end
        end
        push!(pdos, pdos_values)
```

src/postprocess/dos.jl
```diff
@@ -68,65 +68,180 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+This function extracts the required pdos from the output of the `compute_pdos` function. 
+
+    Input:
+     -> pdos_res  : Whole output from compute_pdos.
+     -> manifolds : Vector of OrbitalManifolds to select the desired projectors pdos.
+    Output:
+     -> pdos      : Vector containing the pdos(ε).
+"""
+function sum_pdos(pdos_res, manifolds::AbstractVector)
+    pdos = []
```
Technici4n: Change from vector of vectors to matrix, e.g. `zeros(Float64, length(pdos_res.εs), nspin)`

src/pseudo/NormConservingPsp.jl
```diff
@@ -218,3 +218,11 @@ count_n_pswfc(psp::NormConservingPsp, l) = count_n_pswfc_radial(psp, l) * (2l +
...
     sum(l -> count_n_pswfc(psp, l), 0:psp.lmax; init=0)::Int
 end
+
+"""
+    get_pswfc_label(psp, i, l)
+
+Gives the orbital label (i.e.: "3S") taken from the PseudoPotential file, given the
+angular momentum `l` and radial function index `i` (not the principal quantum number) are provided.
+"""
+get_pswfc_label(psp::NormConservingPsp, i, l) = psp.pswfc_labels[l+1][i]
```
Technici4n: Move method for `psp::PspUpf` to `PspUpf.jl` file, and document method as a comment at the top of `NormConservingPsp.jl`

ext/DFTKPlotsExt.jl
```diff
@@ -141,56 +141,44 @@ function plot_ldos(basis, eigenvalues, ψ; εF=nothing, unit=u"hartree",
...
-        end
+    isnothing(p) && (p = Plots.plot(; kwargs...))
+    p = Plots.plot(p; kwargs...)
+    spinlabels = spin_components(basis.model)
+    pdos = DFTK.sum_pdos(compute_pdos(εs, basis, ψ, eigenvalues;
+                                      positions, temperature, smearing), 
+                         [DFTK.OrbitalManifold(;iatom, label)])
+    for σ = 1:n_spin
+        plot_label = n_spin > 1 ? "$(species) $(orb_name) $(spinlabels[σ]) spin" : "$(species) $(orb_name)"
+        Plots.plot!(p, (εs .- eshift) .* to_unit, pdos[:, σ]*2/n_spin; label=plot_label, color=colors[σ])
```
Technici4n:
```suggestion
        Plots.plot!(p, (εs .- eshift) .* to_unit, pdos[:, σ]; label=plot_label, color=colors[σ])
```

src/postprocess/dos.jl
```diff
@@ -68,65 +68,170 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
-        proj_vectors = structure_factor .* form_factors_all[ik] ./ sqrt(basis.model.unit_cell_volume)
-        projs[ik] = abs2.(ψk' * proj_vectors) # contract on p -> projs[ik][iband,m]
+    projectors = [Matrix{Complex{T}}(undef, length(G_plus_k), 0) for G_plus_k in G_plus_k_all]
+    labels = []
+    for (iatom, atom) in enumerate(basis.model.atoms)
+        psp = atom.psp
+        for l in 0:psp.lmax
+            for n in 1:DFTK.count_n_pswfc_radial(psp, l)
+                label = DFTK.get_pswfc_label(psp, n, l)
+                if !ismanifold((;iatom, species=Symbol(atom.species), label))
```
Technici4n: This will produce a ChemicalSpecies from AtomsBase but it's ok since it can be `==`-compared to a symbol:
```suggestion
                if !ismanifold((;iatom, atom.species, label))
```

src/postprocess/dos.jl
```diff
@@ -68,65 +68,170 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+    -> iatom   : Atom position in the atoms array.
+    -> species : Chemical Element as in ElementPsp.
+    -> label   : Orbital name, e.g.: "3S".
+All fields are optional, only the given ones will be used for selection.
+Can be called with an orbital NamedTuple and returns a boolean
+  stating whether the orbital belongs to the manifold.
+"""
+@kwdef struct OrbitalManifold
+    iatom   ::Union{Int64,  Nothing} = nothing
+    species ::Union{Symbol, Nothing} = nothing
```
Technici4n: Also accept ChemicalSpecies in the union?

src/pseudo/PspUpf.jl
```diff
@@ -173,6 +173,8 @@ end
 
 count_n_pswfc_radial(psp::PspUpf, l) = length(psp.r2_pswfcs[l+1])
 
+get_pswfc_label(psp::PspUpf, i, l) = psp.pswfc_labels[l+1][i]
```
mfherbst: I'd drop the `get_` prefix.

src/common/ortho.jl
```diff
@@ -7,3 +7,19 @@
         return x[:, 1:size(φk, 2)]
     end
 end
+
+@timing function ortho_lowdin(φk::ArrayType) where {ArrayType <: AbstractArray}
```
mfherbst: No `where`, you don't need the array type anywhere.

src/common/ortho.jl
```diff
@@ -7,3 +7,19 @@
...
+    S = φk' * φk
+    evals, evecs = eigen(S)
+    # Check for linear dependence: eigenvalues close to zero
+    tol = eps(real(eltype(φk))) * maximum(abs, evals)
+    if any(abs.(evals) .< tol)
+        small_evals = filter(eval -> eval < tol, evals)
+        error("Input vectors are linearly dependent or nearly so
```
mfherbst: This is a low-level routine. Such routines should not `error`. Use an `@assert` or throw an exception. This anyway should not happen.

src/postprocess/dos.jl
```diff
@@ -68,65 +68,170 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
 
 """
-Projected density of states at energy ε for an atom with given i and l.
+Compute the projected density of states (PDOS) for all atoms and orbitals.
+
+    Input: 
+     -> εs               : Vector of energies at which the PDOS will be computed
+    Output:
+     -> pdos             : 3D array of PDOS, pdos[iε_idx, iproj, σ] = PDOS at energy εs[iε_idx] for projector iproj and spin σ
+     -> projector_labels : Vector of tuples (iatom, species, label, n, l, m) for each projector, that maps the iproj index to the corresponding atomic orbital. For details see the documentation for `atomic_orbital_projectors`.
```
mfherbst: Formatting of the docs is a little unusual. Can you make it more similar to the other docstrings in the code ?

mfherbst (commented):
Looks generally good. I'll have a closer look before we merge. Keep an eye on line length and general formatting

src/postprocess/dos.jl
```diff
@@ -68,65 +68,187 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+                    continue
+                end
+                fun(p) = eval_psp_pswfc_fourier(psp, n, l, p)
+                form_factors_l = build_form_factors(fun, l, G_plus_k_all_cart)
+                for ik in 1:length(G_plus_k_all_cart)
+                   structure_factor = [cis2pi(-dot(positions[iatom], p)) for p in G_plus_k_all[ik]]
+                   projectors[ik] = hcat(projectors[ik], form_factors_l[ik] .* structure_factor ./ sqrt(basis.model.unit_cell_volume))
+                end
+                for m in -l:l
+                    push!(labels, (; iatom, species = Symbol(atom.species), n, l, m, label))
```
Technici4n:
```suggestion
                    push!(labels, (; iatom, atom.species, n, l, m, label))
```

src/common/ortho.jl
```diff
@@ -7,3 +7,15 @@
         return x[:, 1:size(φk, 2)]
     end
 end
+
+@timing function ortho_lowdin(φk::AbstractArray)
+    S = φk' * φk
+    evals, evecs = eigen(S)
```
mfherbst:
```suggestion
    evals, evecs = eigen(Hermitian(φk' * φk))
````

File: src/common/ortho.jl [hunk 70]
```diff
@@ -7,3 +7,15 @@
...
+    # Check for linear dependence: eigenvalues close to zero
+    tol = eps(real(eltype(φk))) * maximum(abs, evals)
+    @assert all(abs.(evals) .> tol) "Input vectors are linearly dependent or nearly so (small eigenvalues detected: $(evals) < $(tol) tolerance)."
```
mfherbst:
```suggestion
    @assert minimum(abs, evals) > eps(real(T)) * maximum(abs, evals)
```

src/common/ortho.jl
```diff
@@ -7,3 +7,15 @@
         return x[:, 1:size(φk, 2)]
     end
 end
+
+@timing function ortho_lowdin(φk::AbstractArray)
```
mfherbst:
```suggestion
@timing function ortho_lowdin(φk::AbstractArray{T}) where {T}
```

src/common/ortho.jl
```diff
@@ -7,3 +7,15 @@
...
+    S = φk' * φk
+    evals, evecs = eigen(S)
+    # Check for linear dependence: eigenvalues close to zero
+    tol = eps(real(eltype(φk))) * maximum(abs, evals)
+    @assert all(abs.(evals) .> tol) "Input vectors are linearly dependent or nearly so (small eigenvalues detected: $(evals) < $(tol) tolerance)."
+    ihS = evecs * Diagonal(evals .^ (-0.5)) * evecs'
+    x = φk * ihS
+    return x
```
mfherbst:
```suggestion
    φk * ihS
```

src/postprocess/dos.jl
```diff
@@ -68,65 +68,187 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+
+ where ψnk is the atomic wavefunction component for band iband and kpoint kpt.
+
+For more details, see documentation for 'atomic_orbital_projectors'.
+"""
+function atomic_orbital_projections(basis::PlaneWaveBasis{T}, ψ;
+                                    ismanifold = l -> true,
+                                    positions = basis.model.positions) where {T}
+    projectors, labels = atomic_orbital_projectors(basis; ismanifold, positions)
+    projections = map(zip(ψ, projectors)) do (ψk, projectorsk)
```
mfherbst: `map` needs no `zip`, just supply two arguments to `map`:
```suggestion
    projections = map(ψ, projectors) do ψk, projectorsk
```

src/postprocess/dos.jl
```diff
@@ -68,65 +68,187 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
-        structure_factor = [cis2pi(-dot(position, p)) for p in G_plus_k_all[ik]]
-        # TODO orthogonalize pseudo-atomic wave functions?
-        proj_vectors = structure_factor .* form_factors_all[ik] ./ sqrt(basis.model.unit_cell_volume)
-        projs[ik] = abs2.(ψk' * proj_vectors) # contract on p -> projs[ik][iband,m]
+    projectors = [Matrix{Complex{T}}(undef, length(G_plus_k), 0) for G_plus_k in G_plus_k_all]
+    labels = []
+    for (iatom, atom) in enumerate(basis.model.atoms)
+        psp = atom.psp
+        for l in 0:psp.lmax
+            for n in 1:DFTK.count_n_pswfc_radial(psp, l)
```
mfherbst: Put on one `for` to avoid deep nesting

src/postprocess/dos.jl
```diff
@@ -68,65 +68,187 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+                label = DFTK.pswfc_label(psp, n, l)
+                if !ismanifold((;iatom, atom.species, label))
```
mfherbst: space

src/postprocess/dos.jl
```diff
@@ -68,65 +68,187 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+- `projectors`: Vector of matrices of projectors
+- `labels`: Vector of NamedTuples containing iatom, species, n, l, m and orbital name for each projector
+
+Notes: 
+- The orbitals used for the projectors are all orthogonalized against each other. This corresponds to ortho-atomic projectors in Quantum Espresso.
+- 'n' in labels is not exactly the principal quantum number, but rather the index of the radial function in the pseudopotential. As an example, if the only S orbitals in the pseudopotential are 3S and 4S, then those are indexed as n=1, l=0 and n=2, l=0 respectively.
+- Use 'ismanifold' kwarg with caution, since the resulting projectors would be orthonormalized only against the manifold basis. Most applications require the whole projectors basis to be orthonormal instead.
+"""
+function atomic_orbital_projectors(basis::PlaneWaveBasis{T};
+                                   ismanifold = l -> true,
```
mfherbst: Maybe `isonmanifold` is a better name ?

src/postprocess/dos.jl
```diff
@@ -68,65 +68,187 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
         error("compute_pdos only supports finite temperature")
     end
     filled_occ = filled_occupation(basis.model)
+    
+    projections, projector_labels = atomic_orbital_projections(basis, ψ; positions)
 
-    projs = compute_pdos_projs(basis, eigenvalues, ψ, i, l, psp, position)
+    nprojs = length(projector_labels) 
 
-    D = zeros(typeof(ε[1]), length(ε), 2l+1)
```
mfherbst, line 83: This is a few too many empty lines for my taste.

To me there is always a balance: Too much space makes less code fit on a screen, which is equally bad to get an overview of what's going on to no space at all.

src/postprocess/dos.jl
```diff
@@ -68,65 +68,187 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+
+    return (; projectors, labels)
+end
+
+"""
+    atomic_orbital_projections(basis, ψ; [ismanifold, positions])
+
+Build the projection matrices projsk for all k-points at the same time.
+
+    projection[ik][iband, iproj] = <ψnk|*projector[ik][iband, iproj] =  <ψnk|ϕinlm>(kpt)
```
mfherbst: I am not sure I understand this. Let's try to find a better description.

src/postprocess/dos.jl
```diff
@@ -68,65 +68,187 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
-# for all atoms of the same type (and reuse the computation of the atomic orbitals)
-function compute_pdos_projs(basis, eigenvalues, ψ, i, l, psp::NormConservingPsp, position)
-    # Precompute the form factors on all kpoints at once so we can better use the cache (memory-compute tradeoff).
-    # Revisit this (pass the cache around explicitly) if RAM becomes an issue.
+"""
+    atomic_orbital_projectors(basis; [ismanifold, positions])   
+
+Build the projectors matrices projsk for all k-points at the same time.
+         
+    projector[ik][:, iproj] = |ϕinlm>(kpt)
```
mfherbst: Not sure I understand the tailling `(kpt)`

mfherbst (commented):
Still a few random space and lines too long issues, which I did not all flag, but generally looks very good. Let's finish this up next week !

src/postprocess/dos.jl
fsicignano: Maybe `projector[ik][iband, iproj] = |ϕinlm>(k + G)` could be better?

Technici4n:
@mfherbst I think the comments have been addressed.

src/postprocess/dos.jl
```diff
@@ -68,65 +68,185 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+  stating whether the orbital belongs to the manifold.
+"""
+@kwdef struct OrbitalManifold
+    iatom   ::Union{Int64,  Nothing} = nothing
+    species ::Union{Symbol, AtomsBase.ChemicalSpecies, Nothing} = nothing
+    label   ::Union{String, Nothing} = nothing
+end
+function (s::OrbitalManifold)(orb)
+    iatom_match    = isnothing(s.iatom)   || (s.iatom == orb.iatom)
+    species_match  = isnothing(s.species) || (s.species == orb.species) || (orb.species == s.species) #TODO: https://github.com/JuliaMolSim/AtomsBase.jl/issues/139
```
mfherbst: Space after  `#`, this should just be a comment on the line above like:
`# See https://github.com/JuliaMolSim/AtomsBase.jl/issues/139 why both species equalities are needed

src/postprocess/dos.jl
```diff
@@ -68,65 +68,185 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+        for l in 0:psp.lmax, n in 1:DFTK.count_n_pswfc_radial(psp, l)
+            label = DFTK.pswfc_label(psp, n, l)
+            if !isonmanifold((; iatom, atom.species, label))
+                continue
+            end
+            fun(p) = eval_psp_pswfc_fourier(psp, n, l, p)
+            form_factors_l = build_form_factors(fun, l, G_plus_k_all_cart)
+            for ik in 1:length(G_plus_k_all_cart)
+               structure_factor = [cis2pi(-dot(positions[iatom], p)) for p in G_plus_k_all[ik]]
+               projectors[ik] = hcat(projectors[ik], form_factors_l[ik] .* structure_factor ./ sqrt(basis.model.unit_cell_volume))
```
mfherbst: too long.

src/postprocess/dos.jl
```diff
@@ -68,65 +68,185 @@ function compute_ldos(scfres::NamedTuple; ε=scfres.εF, kwargs...)
...
+    compute_pdos(εs, basis, ψ, eigenvalues; [positions, smearing, temperature])
+
+Compute the projected density of states (PDOS) for all atoms and orbitals.
+
+Overview of inputs: 
+- `εs`: Vector of energies at which the PDOS will be computed
+
+Overview of outputs:
+- `pdos`: 3D array of PDOS, pdos[iε_idx, iproj, σ] = PDOS at energy εs[iε_idx] for projector iproj and spin σ
+- `projector_labels` : Vector of tuples (iatom, species, label, n, l, m) for each projector, that maps the iproj index to the corresponding atomic orbital.
```
mfherbst: I agree with everything, but some lines are still to long (this is just one example). The loose limit is 92 characters. If it helps readability, being a bit longer is fine (i.e. we are not super strict about it). But these two lines are both well over 100 characters ... that's not ok.

PR #1141, opened 2025-09-04
Make apply_kernel error when not implemented, instead of assuming zero

niklasschmitz:


src/terms/terms.jl
```diff
@@ -37,6 +42,7 @@ struct TermNoop <: Term end
 function ene_ops(term::TermNoop, basis::PlaneWaveBasis{T}, ψ, occupation; kwargs...) where {T}
     (; E=zero(eltype(T)), ops=[NoopOperator(basis, kpt) for kpt in basis.kpoints])
 end
+apply_kernel(term::TermNoop, basis::AbstractBasis, δρ; kwargs...) = nothing
```
mfherbst: Compute missing

niklasschmitz: Right, now I will make it a TermLinear as well instead

PR #1143, opened 2025-09-05
Refactoring of tests

### LINKED ISSUE #799: Split up tests
It's starting to become really annoying how long the windows and Mac tests take. We should select some subset of tests for those architectures, testing only the main user features (SCF, forces, response) and leave the detailed internal testing for the linux CI. The Mac/Windows CI then only needs to run if the linux one is successful.

abussy:
Fixes #799.

Change log:                                                                                          
                                                                                                     
- Extension of the `:minimal` test set to include SCFs, forces and stresses for various systems (functional, open-/close-shell, chemical formula), AD and I/O. This is meant to be a fast test to insure usability of DFTK.
- Removal of the `:core` test set. Seemingly outdated, not used in a consistent way. Also clashing with `:minimal`.
- Fixed bug that always led to running all tests. Previously, `Pkg.test("DFTK"; test_args=["minimal"])` would run all tests, included those with a `:minimal` tag (*i.e.* specifying a subset actually did nothing). Now, running this commend only launches tests from the `:minimal` subset
- Changed logic to prioritize sorting tests based on exclusion. Allows launching tests such as `Pkg.test("DFTK"; test_args=["forces", "noslow"])` for a fast check on forces only.
- Consolidate current logic of tagging non MPI compatible tests with `:dont_test_mpi`, as DFTK should run under MPI by default (and these situations should be the exception)
                                                                                                     
Note for the CI: we could use the `:minimal` for windows and macOS. Generally, it would also be possible to first run the fast tests on the `:minimal` subset, and then conditionally run the rest with `Pkg.test("DFTK";  
test_args=["nominimal"])`.

### DIFF
--- a/src/response/hessian.jl
+++ b/src/response/hessian.jl
@@ -293,14 +293,14 @@ Input parameters:
 
     # compute total δρ
     # TODO Can be smarter here, e.g. use mixing to come up with initial guess.
-    ε = DielectricAdjoint(ham, ρ, ψ, occupation, εF, eigenvalues, occupation_threshold,
-                          bandtolalg, maxiter_sternheimer, q)
+    ε_adj = DielectricAdjoint(ham, ρ, ψ, occupation, εF, eigenvalues, occupation_threshold,
+                              bandtolalg, maxiter_sternheimer, q)
     precon = FunctionPreconditioner() do Pδρ, δρ
         Pδρ .= vec(mix_density(mixing, basis, reshape(δρ, size(ρ));
                                ham, basis, ρin=ρ, εF, eigenvalues, ψ))
     end
     callback_inner(info) = callback(merge(info, (; runtime_ns=time_ns() - start_ns)))
-    info_gmres = inexact_gmres(ε, vec(δρ0);
+    info_gmres = inexact_gmres(ε_adj, vec(δρ0);
                                tol, precon, krylovdim, maxiter, s,
                                callback=callback_inner, kwargs...)
     δρ = reshape(info_gmres.x, size(ρ))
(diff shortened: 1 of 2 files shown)


test/helium_all_electron.jl
```diff
@@ -1,4 +1,4 @@
-@testitem "Helium all electron" tags=[:minimal, :core, :dont_test_mpi] begin
+@testitem "Helium all electron" tags=[:minimal, :dont_test_mpi, :coveragae] begin
```
mfherbst: coverage ... typo

docs/src/developer/testsystem.md
```diff
@@ -10,18 +10,25 @@ the tests.
...
+Pkg.test("DFTK"; test_args = ["forces", "example"])
 ```
-runs both `:core` and `:atomsbase` tests.
+will test forces and run the examples. It is also possible to disable certain tests:
+```julia
+using Pkg
+Pkg.test("DFTK"; test_args = ["noslow"])
+```
+will ignore any test tagged as `:slow`. Finally, parallel tests can be run by passing `"mpi"` or
+`"gpu"` to the `test_args` keyword argument.
```
mfherbst: I would not mention `:mpi` and `:gpu` in one sentence to avoid confusion. One is parallel testing (with MPI) and one is GPU testing.

mfherbst:
What flags up in the coverage is:
- No kinetic energy density computed
- No stress computed
- Dielectric mixing not used
- load_scfres / save_scfres not used ?
Seems weird, these should be used ?

mfherbst:
Ok 45min is still too long.

I'd suggest to only test MacOS and Windows on the master branch and only Linux in the PR.

abussy:
There is this message in the macOS CI output:
```
```

mfherbst:
That sounds very reasonable to me, thank you @abussy.

mfherbst:
Let's see what happens ...

mfherbst:
MacOS we're now down to 30mins, which is ok.

PR #1144, opened 2025-09-09
Symmetry correction for antiferromagnetic ordering

ClementineBarat:
Correction of symmetries in collinear spin:

Previously, 'symmetry_operations' returned the same symmetries for supercells with equal or opposite magnetic moments on different atoms, which did not allow antiferromagnetic solutions to exist. This behavior comes from 'get_symmetry_with_collinear_spin', which seems to consider only the norm of the magnetic moments.

This PR fixes the issue by discarding symmetries that involve a spin flip (identified by 'get_symmetry_with_site_tensors' but not by 'get_symmetry_with_collinear_spin'). This is a quick fix, these discarded symmetries could maybe be properly taken into account.

### DIFF
--- a/src/symmetry.jl
+++ b/src/symmetry.jl
@@ -92,8 +92,14 @@ on each of the atoms. The symmetries are determined using spglib.
         elseif spin_polarization == :collinear
             rotations, translations, spin_flips = Spglib.get_symmetry_with_site_tensors(
                 cell, tol_symmetry)
-            # Keep only the symmetries that don't flip the spins (see PR #1144).
+            # Keep only the symmetries that don't flip the spins
             rotations[spin_flips.==1], translations[spin_flips.==1]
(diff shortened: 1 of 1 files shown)


antoine-levitt:
Nice catch! Discussed offline with Clementine: optimizing here would mean detecting antiferromagnetic order (where the down orbitals are the up orbitals translated by half a cell), computing only the up channels, and building the density by deducing it by symmetry. So density has two components, but orbitals only one spin channel. Very annoying optimization just to cut the runtime by a factor of 2, let's not do it. @ClementineBarat can you add a comment in the code explaining this (or pointing to this PR)? Then let's merge.

mfherbst:
Can we just copy what Antoine wrote into the code? That makes it self-contained.

See #1145 for the flaky test.

mfherbst:
Thanks !

PR #1149, opened 2025-09-17
Update DielectricAdjoint internal docs comments

niklasschmitz:
I adjusted the comments, removing the inverse which corresponds to the linear solver call on an outer level.

A remaining minor inconsistency that I tripped over for a moment: in the math format we use $\varepsilon^\dagger$ which implies that $\varepsilon$ is the name for the dielectric function. However in code we use `ε::DielectricAdjoint` such that `mul_approximate(ε, δρ)` $\approx \varepsilon^\dagger \delta\rho$. What about writing `ε_adj` instead in code? Con: It's longer than 1 character, Pro: it avoids confusion.

### DIFF
--- a/src/response/hessian.jl
+++ b/src/response/hessian.jl
@@ -362,7 +362,8 @@ struct DielectricAdjoint{Tρ, Tψ, Toccupation, TεF, Teigenvalues, Tq}
 end
 
 @doc raw"""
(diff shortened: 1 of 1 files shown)


antoine-levitt:
Yes this is consistently annoying... eps_adj seems fine to me.

Issue #1155, opened 2025-09-23
ForwardDiff with temperature does not work


antoine-levitt:
Needs the following patch
```
diff --git a/src/Model.jl b/src/Model.jl
index cd4c5cedd..ddf93f374 100644
--- a/src/Model.jl
+++ b/src/Model.jl
@@ -110,6 +110,8 @@ function Model(lattice::AbstractMatrix{T},
                symmetries=default_symmetries(lattice, atoms, positions, magnetic_moments,
                                              spin_polarization, terms),
                ) where {T <: Real}
+    T2 = promote_type(T, typeof(temperature), eltype(magnetic_moments))
+
     # Validate εF and n_electrons
     if !isnothing(εF)  # fixed Fermi level
         if !isnothing(n_electrons)
@@ -133,7 +135,7 @@ function Model(lattice::AbstractMatrix{T},
     atom_groups = [findall(Ref(pot) .== atoms) for pot in Set(atoms)]
 
     # Special handling of 1D and 2D systems, and sanity checks
-    lattice = Mat3{T}(lattice)
+    lattice = Mat3{T2}(lattice)
     n_dim = count(!iszero, eachcol(lattice))
     n_dim > 0 || error("Check your lattice; we do not do 0D systems")
     for i = n_dim+1:3
@@ -159,7 +161,7 @@ function Model(lattice::AbstractMatrix{T},
     )
     n_spin = length(spin_components(spin_polarization))
 
-    temperature = T(austrip(temperature))
+    temperature = T2(austrip(temperature))
     temperature < 0 && error("temperature must be non-negative")
 
     if !allunique(string.(nameof.(typeof.(terms))))
@@ -175,7 +177,7 @@ function Model(lattice::AbstractMatrix{T},
     end
     @assert !isempty(symmetries)  # Identity has to be always present.
 
-    Model{T,value_type(T)}(model_name,
+    Model{T2,value_type(T2)}(model_name,
                            lattice, recip_lattice, n_dim, inv_lattice, inv_recip_lattice,
                            unit_cell_volume, recip_cell_volume,
                            n_electrons, εF, spin_polarization, n_spin, temperature, smearing,
diff --git a/src/workarounds/forwarddiff_rules.jl b/src/workarounds/forwarddiff_rules.jl
index 26f946db2..b9edebb15 100644
--- a/src/workarounds/forwarddiff_rules.jl
+++ b/src/workarounds/forwarddiff_rules.jl
[...]

PR #1156, opened 2025-09-23
Fix DFPT wrt temperature

antoine-levitt:
Fix https://github.com/JuliaMolSim/DFTK.jl/issues/1155

### DIFF
--- a/src/workarounds/forwarddiff_rules.jl
+++ b/src/workarounds/forwarddiff_rules.jl
@@ -327,10 +327,3 @@ function erfc(x::Complex{Dual{T,V,N}}) where {T,V,N}
     complex(Dual{T,V,N}(real(erfc(xx)), ForwardDiff.Partials{N,V}(tuple(real(dgamma)...))),
             Dual{T,V,N}(imag(erfc(xx)), ForwardDiff.Partials{N,V}(tuple(imag(dgamma)...))))
 end
-function Smearing.xlogx(x::Dual{T,V,N}) where {T,V,N}
-    # Hardcode the fact that the derivative at exactly zero is 0, rather than Inf.
-    # Technically this is wrong, but in any reasonable application (that yields a finite result in the end),
-    # that (logarithmic) Inf will be multiplied by a (linear) 0, which will win
-    xx = ForwardDiff.value(x)
-    (xx == 0) ? zero(x) : Dual{T}(Smearing.xlogx(xx), ntuple(j -> (1+log(xx)) * ForwardDiff.partials(x, j), N))
-end
--- a/test/forwarddiff.jl
+++ b/test/forwarddiff.jl
@@ -427,7 +427,7 @@ end
     end
     T0 = .01
     derivative_ε = let ε = 1e-5
-        (get(T+ε) - get(T0-ε)) / 2ε
+        (get(T0+ε) - get(T0-ε)) / 2ε
     end
     derivative_fd = ForwardDiff.derivative(get, T0)
     @test norm(derivative_ε - derivative_fd) < 1e-4
--- a/src/Smearing.jl
+++ b/src/Smearing.jl
@@ -83,7 +83,13 @@ function xlogx(x)
 end
 function entropy(S::FermiDirac, x)
     f = occupation(S, x)
-    - (xlogx(f) + xlogx(1 - f))
+    # protect against the occupation being exactly zero or one, which causes trouble with the derivative
+    # this check is a bit stupid, but if we just check for f == 0, the branch won't get picked up by forwarddiff
+    if abs(f) < eps(typeof(x))|| abs(1-f) < eps(typeof(x))
+        zero(x)
+    else
+        - (xlogx(f) + xlogx(1 - f))
+    end
 end
 function occupation_divided_difference(S::FermiDirac, x, y, εF, temp)
     temp == 0 && return occupation_divided_difference(None(), x, y, εF, temp)


Technici4n:
Are you sure about passing `δtemperature` to the final chi0? Not sure it makes sense if you think in terms of epsilon_adj^-1 (df/dtheta + chi0 * dH/dtheta). The inside of epsilon_adj (4P) should not be affected by δtemperature and the final chi0 comes partially from rewriting epsilon_adj (4P) using a matrix identity.

Technici4n:
What we did with @niklasschmitz on the blackboard is starting from $P_\ast = f(\theta, H(P_\ast, \theta))$, which eventually gives:
```math
\frac{\partial P_\ast}{\partial \theta} = \left(1 - \chi_0K\right)^{-1} \left(\frac{\partial f}{\partial \theta} + \chi_0\frac{\partial H}{\partial \theta}\right)
```
where here $\chi_0 = \frac{\partial f}{\partial H}$. So to me it looks like putting the $\delta T$ in the first $\chi_0$ application is reasonable, but I am not sure it should be in the final $\chi_0$ since that comes from rewriting $\left(1 - \chi_0K\right)^{-1}$ using some matrix identities.

antoine-levitt:
Had another read, no I think what I'm doing is correct. We have ψ = F(V, T), V = Vext + Vhxc(ρ(ψ)), so δψ = χ04P I δV + ∂F∂T δT, δV = K I* δψ, where I : δV->(δVψ) and I* : δψ -> δρ. So δψ = χ04P I K I* δψ + ∂F∂T δT. Multiply by I* to get δρ = χ0 K δρ + I* ∂F∂T δT. So we first compute this δρ, then, we compute δV = K δρ, and then we get δψ = χ04P I δV + ∂F∂T δT. So in both cases we need the +δT term. Does that make sense?

antoine-levitt:
The logic is a bit convoluted here because we want dpsi as an output, not just drho, but the Dyson equation is on drho, not dpsi. So we first compute drho, but then we need to go through it again to get dpsi.

antoine-levitt:
Errors should be fixed, but there's something about amdgpus I don't know how to debug

antoine-levitt:
> (Will look at the code again soon but I am fighting symmetries currently, and they are winning...)

Symmetries always win

Technici4n:
OK you are correct! With my approach the $\frac{\partial f}{\partial \theta}$ is part of the "rhs" and thus appears both in the first and in the last X0 application! So I agree with passing dtemperature to both. 😄

The math looks good to me. One thing though: If we go for switching the sign of the RHS we have to change it at every callsite of the Hessian solver, and there are quite a few:
image
(and many others...)
Might be easiest not to 🤷

antoine-levitt:
Duh, of course. I need to be replaced by an AI quick... I'll update it, it's not that large...

Technici4n (approved):
LGTM. Will probably need a minor version bump for the rhs sign change.

antoine-levitt:
Done. I also had to fix a tricky numerical issue with the derivative of xlogx at exactly zero. Let's hope this doesn't mess up other things...

test/forwarddiff.jl
```diff
@@ -402,4 +402,33 @@ end
...
+
+    function get(T)
+        model = model_DFT(lattice, atoms, positions; functionals=LDA(), temperature=T)
+        basis = PlaneWaveBasis(model; Ecut=10, kgrid=[1, 1, 1])
+        scfres = self_consistent_field(basis, tol=1e-12)
+        scfres.energies.total
+    end
+    T0 = .01
+    derivative_ε = let ε = 1e-5
+        (get(T+ε) - get(T0-ε)) / 2ε
```
Technici4n: First T should probably be T0?

src/workarounds/forwarddiff_rules.jl
```diff
@@ -327,3 +327,10 @@ function erfc(x::Complex{Dual{T,V,N}}) where {T,V,N}
     complex(Dual{T,V,N}(real(erfc(xx)), ForwardDiff.Partials{N,V}(tuple(real(dgamma)...))),
             Dual{T,V,N}(imag(erfc(xx)), ForwardDiff.Partials{N,V}(tuple(imag(dgamma)...))))
 end
+function Smearing.xlogx(x::Dual{T,V,N}) where {T,V,N}
```
Technici4n: This actually feels wrong. If you compute the derivative of the entire entropy don't you just get enkred?

test/forwarddiff.jl
antoine-levitt: Yes... I should stop using CI to run my tests but I always forget how to run just one locally

src/workarounds/forwarddiff_rules.jl
antoine-levitt: You're annoyingly correct, I knew I was too lazy but hoped to get away with it. That means it has to be fixed at the level of entropy, not of the xlogx function. I'll do that, thanks.

antoine-levitt: No wait you're wrong: the derivative of the function x -> xlogx(occ(x)) goes to zero for very negative x, so it's fine. But I can just check that with the partials so I'll do that.

niklasschmitz: I wonder if we can solve this again similarly like what we did in #1022  and just change the primal suitably, avoiding a custom rule?

antoine-levitt: Technically yes, but I'd have to write an explicit expression for xlogx(ffd(x)) and rearrange it properly, and I don't want to. What I pushed now should be fine.

Technici4n: So the entropy should be
```math
S(f) = -f \log f - (1-f) \log (1-f)
```
Its derivative is
```math
S'(f) = \log(\frac{1-f}{f})
```

Now if we plug in Fermi-Dirac $f_{FD}(x)$:
```math
\frac{\mathrm{d} S}{\mathrm{d} x} = S'(f_{FD}(x)) f_{FD}'(x) = \log(\frac{1-f_{FD}(x)}{f_{FD}(x)}) f_{FD}'(x) = \log(e^x) f_{FD}'(x) = x f_{FD}'(x) = - x e^x f_{FD}^2(x)
```
Does this look correct?

For very small x (occupation 1) this should give 0 and for very large x (occupation 0) this should also give 0?

antoine-levitt: That's right, hence my hack of returning zero for small and large x. To do it properly, you need to express S(f(x)) in a way that is free of numerical trouble, like we did for other things.

antoine-levitt: OK, chatgpt has the answer with a cosh trick: https://chatgpt.com/share/68e65ee9-111c-800d-a28d-86ae063c9637. I'll just do that then...

Technici4n: Ah yes. Might be good to add a few tests that compute the derivative of `entropy(S::FermiDirac, x)` directly at a few relevant x?

antoine-levitt: wait no this still need custom code in forwarddiff_rules for teaching forwarddiff the derivative of S. I'd rather keep the current version, hackish as it is.

test/forwarddiff.jl
```diff
@@ -402,4 +402,33 @@ end
...
+                       [1 1 0.]]
+    Si = ElementPsp(:Si, PseudoFamily("dojo.nc.sr.lda.v0_4_1.standard.upf"))
+    atoms     = [Si, Si]
+    positions = [ones(3)/8, -ones(3)/8]
+
+    function get(T)
+        model = model_DFT(lattice, atoms, positions; functionals=LDA(), temperature=T)
+        basis = PlaneWaveBasis(model; Ecut=10, kgrid=[1, 1, 1])
+        scfres = self_consistent_field(basis, tol=1e-12)
+        scfres.energies.total
```
Technici4n, line 426: This is not the best test because by HF any mistake in the DFPT will be multiplied by zero anyway.

antoine-levitt: Yet again annoyingly correct. (but I did test it locally with something else, in fact the energy is the most annoying because of the entropy) I'll change the test

PR #1159, opened 2025-09-30
Disable AMDGPU precompilation if !functional

Technici4n:
Hopefully this will fix the DFTKAMDGPUExt precompilation errors on CI, which seem to become fatal in Julia 1.12. (But are generally annoying in 1.10 already...)

@abussy looks good right?


abussy:
LGTM. This will considerably clean-up test outputs, thanks!

PR #1160, opened 2025-10-02
Update to PseudoPotentialIO 0.3

Technici4n:
It's not released yet, hence I had to commit the manifest.

Just testing the legacy wrapper first and then I'll migrate to the new API.


src/pseudo/PspUpf.jl
```diff
@@ -72,24 +72,25 @@ Does not support:
 - GIPAW reconstruction data
 """
 function PspUpf(path; identifier=path, rcut=nothing)
-    pseudo = load_upf(path)
+    pseudo = load_psp_file(path)
```
mfherbst: Can we assert this a Upf file here ... just to be safe.

Technici4n: Sure. Related question: should we auto-convert `Psp8File`s?

mfherbst: What does this bring ? I may not understand what you mean.

mfherbst: What does this bring ? I may not understand what you mean.

Technici4n: It would bring support for psp8 files "for free", should it ever be useful.

mfherbst: Ah I see, for that I would just have a PspPsp8 function that reads the file and returns a PspUpf for now, no ?

Technici4n: Ok I did something.

src/pseudo/load_psp.jl
```diff
@@ -76,6 +76,9 @@ function load_psp(key::AbstractString; kwargs...)
     elseif endswith(lowercase(key), ".upf")
         pseudo_type = PspUpf
         extension = ".upf"
+    elseif endswith(lowercase(key), ".psp8")
+        pseudo_type = PspUpf
```
mfherbst, line 85: I think this is a sufficient solution.

But best add a comment here, too, why this is done. Else one may mistake this as a typo in the future.

Actually one may also add a sentence in the docstring of `load_psp` which formats are supported.

PR #1161, opened 2025-10-07
Make SCF runs optionally reproducible by providing a seed

Technici4n:
We had a discussion wrt. randomness before, and the conclusion was that we should keep DFTK random by default but allow passing a seed for reproducibility[^1]. It seems that the best way to achieve this in Julia is to seed the `TaskLocalRNG` since any subtask will be seeded from it automatically. This takes care of the multithreading aspect and makes the required changes minimal.

Additionally, the scfres now contains a seed that can be used to reproduce the run[^1].

If this is satisfactory @mfherbst I am happy to implement the same approach for the other solvers (direct minimization etc) as well. For tests, it should be sufficient to just seed the task local RNG before each test; I haven't looked into it yet though.

[^1]: Assuming the exact same Julia and packages versions.

### DIFF
--- a/src/DFTK.jl
+++ b/src/DFTK.jl
@@ -45,6 +45,7 @@ include("common/hankel.jl")
 include("common/hydrogenic.jl")
 include("common/derivatives.jl")
 include("common/linalg.jl")
+include("common/random.jl")
 
 export PspHgh
 export PspUpf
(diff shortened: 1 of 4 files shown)


mfherbst:
Yes, I like this ! Please go ahead with the other solvers.

PR #1162, opened 2025-10-07
Add AD-DFPT elastic constants example

niklasschmitz:


niklasschmitz:
I agree it would be nice to have, but would do a separate PR. 

It can be implemented in several stages of increasing efficiency:

1. A first inefficient but general implementation, disabling all symmetries (and thus working in a fully unfolded kgrid) and just computing the 6x6 Jacobian by something like `ForwardDiff.jacobian(stress_from_strain, zeros(6))`.

2. A refined implementation including:
2.1 strain patterns adapted to any of the seven crystal systems determined by spglib (cubic, hexagonal, ..., triclinic), chosen on the fly (between 1 and 6 strains). I think the strain patterns can be just tabulated as a small hard-coded dict, like for example Table 1 and Table 3 in https://doi.org/10.1016/j.cpc.2009.11.017.
2.2 using different symmetry subgroups for each strain (which needs seperate calls to ForwardDiff since the primal is changing for each).
2.3 Starting from a single primal calculation of full symmetry, and then doing partial unfolds of the kgrid for each perturbation, and combining the results afterwards.

examples/elastic_constants.jl
```diff
@@ -0,0 +1,133 @@
...
+
+function stress_from_strain(model0, voigt_strain; symmetries, Ecut, kgrid, tol)
+    lattice = DFTK.voigt_strain_to_full(voigt_strain) * model0.lattice
+    model = Model(model0; lattice, symmetries)
+    basis = PlaneWaveBasis(model; Ecut, kgrid)
+    scfres = self_consistent_field(basis; tol)
+    DFTK.full_stress_to_voigt(compute_stresses_cart(scfres))
+end 
+
+stress_fn(voigt_strain) = stress_from_strain(model0, voigt_strain; symmetries=symmetries_strain, Ecut, kgrid, tol)
```
mfherbst: Just do a pass on line length when you are done.

docs/make.jl
```diff
@@ -47,6 +47,7 @@ PAGES = [
         "examples/polarizability.jl",
         "examples/forwarddiff.jl",
         "examples/phonons.jl",
+        "examples/elastic_constants.jl",
```
mfherbst: This is currently the most interesting example for most people. It should be the first.

examples/elastic_constants.jl
```diff
@@ -0,0 +1,133 @@
...
+stress, (dstress,) = value_and_pushforward(stress_fn, AutoForwardDiff(), zeros(6), (strain_pattern,));
+@show stress dstress;
```
mfherbst: If you want this to look nice on the website than don't use show. Rather just return it as the last entry in a cell.

You can split cells using `#-` or something like that ... or you just add a one-line comment as you did below.

niklasschmitz: Ok done

examples/elastic_constants.jl
niklasschmitz: done

PR #1163, opened 2025-10-07
Port AtomicLocal integration to GPU

abussy:
This PR ports the integration of the `AtomicLocal` form factors to the GPU.

In standard calculations, the instantiation of the `AtomicLocal` term in the `PlaneWaveBasis` is generally negligible. However, when ForwardDiff Duals are involved, this is an other story. In particular, for stress calculations on the GPU (not yet generally available, see https://github.com/JuliaMolSim/DftFunctionals.jl/pull/23), this step becomes vastly more expensive than anything else.

For efficiency, the form factors are only calculated for a set of `G` vector norms, rather than for each individual `G`. With ForwardDiff Duals, 2 `G` vectors with the same value might differ in their partials and have a different norm. As a result, a lot more integrations must take place.

This PR introduces a special case for the GPU, where a `map!` kernel is launched over the `G` vector norms, and each GPU thread takes care of an integration on the atomic grid. This only triggers in the special case of UPF pseudos on the GPU.

Notes:
- This PR will drastically speed up stress calculations on the GPU. While the new function will also be used for the instantiation of a standard `PlaneWaveBasis`, its impact there will be smaller
- The `AtomicLocal` term will remain expensive on the CPU. Since everything else also gets more expensive, it is less of an obvious bottleneck in that case. However, I think it would be trivial to parallelize it over Julia threads
- Due to the complex dispatching taking place, and the fact that the `ElementPsp{<:PspUpf}` type is far from being `isbits`, I did not manage to come up with a single code to run on both CPU and GPU.
- The instantiation of the `Xc` term suffers from a similar problem, to be treated in a separate PR

### DIFF
--- a/src/architecture.jl
+++ b/src/architecture.jl
@@ -50,4 +50,4 @@ end
 """
 Returns the architecture of the given array.
 """
-get_architecture(x::AbstractArray) = x isa AbstractGPUArray ? GPU{typeof(x)}() : CPU()
+architecture(x::AbstractArray) = x isa AbstractGPUArray ? GPU{typeof(x)}() : CPU()
--- a/src/common/quadrature.jl
+++ b/src/common/quadrature.jl
@@ -38,9 +38,9 @@ point `i` (not necessarily in order).
 """
 simpson
 @inbounds function simpson(integrand, x::AbstractVector)
-    n = length(x)
-    n <= 4 && return trapezoidal(integrand, x)
-    if (x[2] - x[1]) ≈ (x[3] - x[2])
+    if length(x) <= 4
+        trapezoidal(integrand, x)
+    elseif (x[2] - x[1]) ≈ (x[3] - x[2])
         simpson_uniform(integrand, x)
     else
         simpson_nonuniform(integrand, x)
@@ -116,12 +116,12 @@ end
 end
 
 """
-Return the approproate integration finction given a quadrature
+Return the approproate integration function given a PSP quadrature
 """
-function get_integration_function(x::AbstractArray)
-    n = length(x)
-    n <= 4 && return trapezoidal
-    if (x[2] - x[1]) ≈ (x[3] - x[2])
+function default_psp_quadrature(x::AbstractArray)
+    if length(x) <= 4
+        trapezoidal
+    elseif (x[2] - x[1]) ≈ (x[3] - x[2])
         simpson_uniform
     else
         simpson_nonuniform
--- a/src/elements.jl
+++ b/src/elements.jl
@@ -52,6 +52,12 @@ function core_charge_density_fourier(::Element, ::T)::T where {T <: Real}
     error("Abstract elements do not necesesarily provide core charge density.")
 end
 
+# Generic vectorized version of local_potential_fourier, GPU-safe
+function local_potential_fourier(el::Element, ps::AbstractVector{T}) where {T <: Real}
+    arch = architecture(ps)
+    to_device(arch, map(p -> local_potential_fourier(el, p), to_cpu(ps)))
+end
+
 Base.show(io::IO, el::Element) = print(io, "$(typeof(el))(:$(species(el)))")
 
 #
@@ -83,9 +89,7 @@ function local_potential_fourier(el::ElementCoulomb, p::T) where {T <: Real}
     Z = charge_nuclear(el)
     return -4T(π) * Z / p^2
 end
-function local_potential_fourier(el::ElementCoulomb, ps::AbstractVector{T}) where {T <: Real}
-    map(p -> local_potential_fourier(el, p), ps)
-end
+
 local_potential_real(el::ElementCoulomb, r::Real) = -charge_nuclear(el) / r
 
 
@@ -165,7 +169,7 @@ eval_psp_energy_correction(T, el::ElementPsp) = eval_psp_energy_correction(T, el
 function local_potential_fourier(el::ElementPsp, p::T) where {T <: Real}
     eval_psp_local_fourier(el.psp, p)
 end
-# Vectorized, GPU compatible version
+# Vectorized version of the above
 function local_potential_fourier(el::ElementPsp, ps::AbstractVector{T}) where {T <: Real}
     eval_psp_local_fourier(el.psp, ps)
 end
@@ -249,9 +253,6 @@ function local_potential_fourier(el::ElementCohenBergstresser, p::T) where {T <:
     psq_pi = Int(round(p^2 / (2π / el.lattice_constant)^2, digits=2))
     T(get(el.V_sym, psq_pi, 0.0))
 end
-function local_potential_fourier(el::ElementCohenBergstresser, ps::AbstractVector{T}) where {T <: Real}
-    map(p -> local_potential_fourier(el, p), ps)
-end
 # TODO Strictly speaking needs a eval_psp_energy_correction
 
 
@@ -284,9 +285,6 @@ function local_potential_real(el::ElementGaussian, r)
 end
 function local_potential_fourier(el::ElementGaussian, p::Real)
     -el.α * exp(- (p * el.L)^2 / 2)  # = ∫_ℝ³ V(x) exp(-ix⋅p) dx
-end
-function local_potential_fourier(el::ElementGaussian, ps::AbstractVector{T}) where {T <: Real}
-    map(p -> local_potential_fourier(el, p), ps)
 end
 # TODO Strictly speaking needs a eval_psp_energy_correction
 
--- a/src/pseudo/NormConservingPsp.jl
+++ b/src/pseudo/NormConservingPsp.jl
@@ -18,7 +18,7 @@ abstract type NormConservingPsp end
 # eval_psp_projector_fourier(psp, i, l, p::Real)
 # eval_psp_local_real(psp, r::Real)
 # eval_psp_local_fourier(psp, p::Real)
-# eval_psp_local_fourier(psp, ps::AbstractArray{Real})
+# eval_psp_local_fourier(psp, ps::AbstractArray{<Real})
 # eval_psp_energy_correction(T::Type, psp)
 
 #### Optional methods:
@@ -85,6 +85,12 @@ V_{\rm loc}(p) &= ∫_{ℝ^3} (V_{\rm loc}(r) - C(r)) e^{-ip·r} dr + F[C(r)] \\
 """
 eval_psp_local_fourier(psp::NormConservingPsp, p::AbstractVector) =
     eval_psp_local_fourier(psp, norm(p))
+
+# Fallback vectorized implementation for non GPU-optimized code.
+function eval_psp_local_fourier(psp::NormConservingPsp, ps::AbstractVector{T}) where {T <: Real}
+    arch = architecture(ps)
+    to_device(arch, map(p -> eval_psp_local_fourier(psp, p), to_cpu(ps)))
+end
 
 @doc raw"""
     eval_psp_energy_correction([T=Float64,] psp::NormConservingPsp)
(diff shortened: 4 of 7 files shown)


src/gpu/local.jl
```diff
@@ -0,0 +1,34 @@
...
+        else
+            # GPU compilation error if branching done within generic simpson() function
+            I = method(rgrid) do i, r
+                 r * (r * vloc[i] - -Zion * erf(r)) * sphericalbesselj_fast(0, p * r)
+            end
+            4T(π) * (I + -Zion / p^2 * exp(-p^2 / T(4)))
+        end
+    end
+
+    ints_cpu = to_cpu(ints)
```
mfherbst: This is weird. The only reason this is on the CPU is because we needed it for something. Now essentially you put `form_factors` on the CPU only to put it on the GPU once this function call is over, right ? Can this not directly be a GPU array in the GPU version of the code ?

mfherbst (commented):
Thanks for this suggestion. I think it becomes increasingly clear that `local_potential_fourier(element, p)` is not the right interface. I think it's better to fix that rather than adding yet another level of indirection with the `atomic_local_inner_loop!`.

What I have in mind could be an interface where one gets multiple `ps` at once or something similar, perhaps in combination with an in-place version, where one is supposed to place it into CPU or GPU memory. That allows specialisation directly at the level of the PspUpf implementation.

What I really do not like here is that gpu/local.jl now essentially contains bleeded-out details of how one is supposed to integrate a PspUpf. Such code should be directly associated to the PspUpf structure and nothing else.

What do you think @abussy ?

abussy:
> What I really do not like here is that gpu/local.jl now essentially contains bleeded-out details of how one is supposed to integrate a PspUpf. Such code should be directly associated to the PspUpf structure and nothing else.

I fully agree. The proposed solution happens to be the one with the least amount of change to the current code, but some change would allow a much smoother integration.

As you suggest, allowing `local_potential_fourier(element, p)` to treat multiple ps at once would pass the loop logic down to the specific pseudo implementation. If we remain general and pass an `AbstractArray` of ps, we might even be able to write architecture agnostic code. I'll look into it.

abussy:
> I wonder what exactly makes them so slow. In principle it should only be twice as slow.

There is a whole machinery to only perform integrations for unique values of `norm(G)` (see [here](https://github.com/JuliaMolSim/DFTK.jl/blob/fd901ade421e1810c4abadf6920ac45be04e7b71/src/terms/local.jl#L78-L83)). When duals are involved, the number of unique norms explodes. My interpretation is that G vectors with the same value can have different partials. So the overall norm is different.

abussy:
This last commit is a proof of concept and not a final product. It illustrates how one might implement the loop over `norm(G)` at the Psp level (here only implemented for `<:PspUpf`).

To make it general, one needs to:
- Modify all function signatures of `element.jl` so that arrays are expected, and make sure the underlying implementations match that
- Update the rest of the code to this new convention

Before proceeding, I'd like to make sure we agree on the way forward.

src/pseudo/PspHgh.jl
```diff
@@ -129,10 +129,15 @@ end
...
 function eval_psp_local_fourier(psp::PspHgh, p::T) where {T <: Real}
+    p == 0 && return zero(T)  # Compensating charge background
     t::T = p * psp.rloc
     psp_local_polynomial(T, psp, t) * exp(-t^2 / 2) / t^2
 end
 
+# Vectorized version of the above
+function eval_psp_local_fourier(psp::PspHgh, ps::AbstractVector{T}) where {T <: Real}
+    map(p -> eval_psp_local_fourier(psp, p), ps)
+end
```
mfherbst: I would make this the default implementation in `NormConservingPsp.jl`.

src/pseudo/NormConservingPsp.jl
```diff
@@ -18,6 +18,7 @@ abstract type NormConservingPsp end
 # eval_psp_projector_fourier(psp, i, l, p::Real)
 # eval_psp_local_real(psp, r::Real)
 # eval_psp_local_fourier(psp, p::Real)
+# eval_psp_local_fourier(psp, ps::AbstractArray{Real})
```
mfherbst: `ps::AbstractArray{<Real}` I'd say.

src/elements.jl
```diff
@@ -83,6 +83,9 @@ function local_potential_fourier(el::ElementCoulomb, p::T) where {T <: Real}
     Z = charge_nuclear(el)
     return -4T(π) * Z / p^2
 end
+function local_potential_fourier(el::ElementCoulomb, ps::AbstractVector{T}) where {T <: Real}
+    map(p -> local_potential_fourier(el, p), ps)
+end
```
mfherbst: You can avoid all of these with the fallback as mentioned above.

src/common/quadrature.jl
```diff
@@ -114,3 +114,16 @@ end
 
     return I
 end
+
+"""
+Return the approproate integration finction given a quadrature
+"""
+function get_integration_function(x::AbstractArray)
```
mfherbst: Not so happy with the name. Usually for these things we use something like `default_psp_quadrature(x::AbstractArray)`. This is also quite tailored to what you need to do for PSPs, that's why I think we should have this in the name and add a respective sentence in the docstring.

src/pseudo/PspUpf.jl
```diff
@@ -203,19 +203,42 @@ end
 
 eval_psp_local_real(psp::PspUpf, r::T) where {T<:Real} = psp.vloc_interp(r)
 
-function eval_psp_local_fourier(psp::PspUpf, p::T)::T where {T<:Real}
+# Low-level function for the local part of the pseudopotential in reciprocal space
+function eval_psp_local_fourier(rgrid, vloc, Zion, p::T;
+                                integration_function=simpson)::T where {T<:Real}
```
mfherbst: I would use the interface:
```suggestion
function eval_psp_local_fourier_(quadrature, rgrid, vloc, Zion, p::T)::T where {T<:Real}
```
Also add a comment that `quadrature` is expected to be one of the functions defined in `common/quadrature.jl`, e.g. `simpson` or `trapezoidal`.

Why do you need the `::T` type annotation in the return type ? This is weird and points to a type-instability.

src/terms/local.jl
```diff
@@ -81,17 +81,18 @@ function atomic_local_form_factors(basis::PlaneWaveBasis{T}; q=zero(Vec3{T})) wh
...
 
-    form_factors_cpu = zeros(T, length(norm_indices), length(basis.model.atom_groups))
-    for(p, ifnorm) in norm_indices
-        for (igroup, group) in enumerate(basis.model.atom_groups)
-            element = basis.model.atoms[first(group)]
-            form_factors_cpu[ifnorm, igroup] = local_potential_fourier(element, p)
-        end
+    ni_pairs = collect(pairs(norm_indices))
+    ps = to_device(basis.architecture, [p for (p, idx) in ni_pairs])
+    indices = to_device(basis.architecture, [idx for (p, idx) in ni_pairs])
```
mfherbst:
```suggestion
    ps      = to_device(basis.architecture, first.(pairs(norm_indices)))
    indices = to_device(basis.architecture, last.(pairs(norm_indices)))
```

src/common/quadrature.jl
```diff
@@ -114,3 +114,16 @@ end
...
+    n = length(x)
+    n <= 4 && return trapezoidal
+    if (x[2] - x[1]) ≈ (x[3] - x[2])
+        simpson_uniform
+    else
+        simpson_nonuniform
+    end
```
mfherbst: I think it's clearer to structure:
```suggestion
    if length(x) <= 4
        trapezoidal
    elseif (x[2] - x[1]) ≈ (x[3] - x[2])
        simpson_uniform
    else
        simpson_nonuniform
    end
```

src/pseudo/PspUpf.jl
mfherbst: I would use the interface:
```suggestion
function eval_psp_local_fourier_(quadrature, rgrid, vloc, Zion, p::T)::T where {T<:Real}
```
Also add a comment that `quadrature` is expected to be one of the functions defined in `common/quadrature.jl`, e.g. `simpson` or `trapezoidal`.

Why do you need the `::T` type annotation in the return type ? This is weird and points to a type-instability.

src/pseudo/PspUpf.jl
```diff
@@ -203,19 +203,42 @@ end
...
     vloc  = @view psp.vloc[1:psp.ircut]
-    I = simpson(rgrid) do i, r
-         r * (r * vloc[i] - -psp.Zion * erf(r)) * sphericalbesselj_fast(0, p * r)
+    eval_psp_local_fourier(rgrid, vloc, psp.Zion, p)
+end
+
+# Vectorized version of the above, GPU compatible
+function eval_psp_local_fourier(psp::PspUpf, ps::AbstractVector{T}) where {T<:Real}
+    x = @view psp.rgrid[1:3]
+    integration_function = get_integration_function(x)
```
mfherbst: This is a bug. Will always return `trapezoidal`, right ?

src/architecture.jl
```diff
@@ -46,3 +46,8 @@ Explanaition of key entries:
 function memory_usage(::CPU)
     (; max_rss=Sys.maxrss(), gc_bytes=Base.gc_live_bytes())
 end
+
+"""
+Returns the architecture of the given array.
+"""
+get_architecture(x::AbstractArray) = x isa AbstractGPUArray ? GPU{typeof(x)}() : CPU()
```
mfherbst: In julia it's more common to not have the `get_` prefix, so I'd call it just `architecture( )`

src/terms/local.jl
```diff
@@ -81,17 +81,18 @@ function atomic_local_form_factors(basis::PlaneWaveBasis{T}; q=zero(Vec3{T})) wh
...
+
+    form_factors = similar(ps, length(norm_indices), length(basis.model.atom_groups))
+    for (igroup, group) in enumerate(basis.model.atom_groups)
+        element = basis.model.atoms[first(group)]
+        @inbounds form_factors[indices, igroup] .= local_potential_fourier(element, ps)
```
mfherbst, line 93: This cries as if we should have a `local_potential_fourier!(form_factors[indices, igroup], element, ps)`. Is this hard to, respectively has little benefits ?

If yes, just make a comments this is possible, if no, maybe give it a try ?

mfherbst (commented):
Thanks

src/elements.jl
abussy: Yes, good idea.

src/common/quadrature.jl
abussy: Sure, I'll also modify the `simpson` function, where I burrowed this structure from.

src/pseudo/PspUpf.jl
abussy: > Why do you need the ::T type annotation in the return type ? This is weird and points to a type-instability.

Here, I simply maintain an existing pattern, see: https://github.com/JuliaMolSim/DFTK.jl/blob/58f92b61b5b15a89ec03b28a36c2108afdcdef93/src/pseudo/PspUpf.jl#L206

I am not sure why it is there in the first place, but most scalar functions in `PspUpf.jl` have this kind of pattern. Logically, I should keep this at the lowest level function (and remove it from other functions calling this).

src/pseudo/PspUpf.jl
abussy: Indeed, well spotted.

src/terms/local.jl
abussy: I think an in-place implementation would be quite tedious. Main reason is the non-contiguous/scrambled `indices` array, which would need to be passed down all the way down to the most specialized functions.

I also think performance benefits would be low, as there is only one allocation per atomic species.

abussy:
Implemented requested and suggested changes. When generalizing vectorized general fallbacks of the type:
```
function local_potential_fourier(el::Element, ps::AbstractVector{T}) where {T <: Real}
    arch = architecture(ps)
    to_device(arch, map(p -> local_potential_fourier(el, p), to_cpu(ps)))
end
```

I also made sure it was GPU safe. Some non UPF calculations would fail on the GPU.

Note: up to date with current master

Issue #884, opened 2023-09-20
Elastic constants


antoine-levitt:
cc @LaurentVidal95 

Elastic constants appear to work more or less out of the box (with one fix), we should probably fix and advertise them in an example. It does take ages to precompile the forwarddiff stuff though, I'm not sure what we can do about that...

```
@@ -192,10 +192,10 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Number, fft_size, variational,
         @assert isnothing(kweights)
         kcoords, kweights, symmetries = bzmesh_ir_wedge(kgrid, symmetries; kshift)
     else
-        # Manual kpoint set based on kcoords/kweights
-        @assert length(kcoords) == length(kweights)
-        all_kcoords = unfold_kcoords(kcoords, symmetries)
-        symmetries  = symmetries_preserving_kgrid(symmetries, all_kcoords)
+        # # Manual kpoint set based on kcoords/kweights
+        # @assert length(kcoords) == length(kweights)
+        # all_kcoords = unfold_kcoords(kcoords, symmetries)
+        # symmetries  = symmetries_preserving_kgrid(symmetries, all_kcoords)
     end
 
     # Init MPI, and store MPI-global values for reference
```

```julia
# Very basic setup, useful for testing
using DFTK
using ForwardDiff

a = 10.26  # Silicon lattice constant in Bohr

Si = ElementPsp(:Si, psp=load_psp("hgh/lda/Si-q4"))
atoms     = [Si, Si]
positions = [ones(3)/8, -ones(3)/8]


function E(a)
    lattice = a / 2 * [[0 1 1.];
                       [1 0 1.];
                       [1 1 0.]]
    model = model_LDA(lattice, atoms, positions, symmetries=false)
    basis = PlaneWaveBasis(model; Ecut=15, kgrid=[2, 2, 2])
    scfres = self_consistent_field(basis, tol=1e-10)
    scfres.energies.total
end

# as = range(10, 11, 10)
# plot(as, E.(as))


function dEda(a)
    lattice = a / 2 * [[0 1 1.];
                       [1 0 1.];
                       [1 1 0.]]
    model = model_LDA(lattice, atoms, positions, symmetries=false)
    basis = PlaneWaveBasis(model; Ecut=15, kgrid=[2, 2, 2])
    scfres = self_consistent_field(basis, tol=1e-10)
    function HF_energy(new_a)
        basis = scfres.basis
        lattice = new_a / 2 * [[0 1 1.];
[...]

PR #1166, opened 2025-10-08
Fix DivideAndConquer in Julia 1.12

Technici4n:
Julia 1.12.0 is now used by the "Julia latest" workflow in CI even though no announcement was made yet (see https://julialang-s3.julialang.org/bin/versions.json).

Turns out that `DivideAndConquer` is not exported, so we need to prefix it by `LinearAlgebra.`. The `>= v"1.12"` check is false for pre-releases, which is why this only started failing now.


niklasschmitz (approved):
Yes I saw this needed on #884 as well. Thanks!

PR #1168, opened 2025-10-10
Try adding downgrade CI

Technici4n:
I read this https://www.stochasticlifestyle.com/semantic-versioning-semver-is-flawed-and-downgrade-ci-is-required-to-fix-it/ and was curious to try it out. 😄


mfherbst:
Guess this is kind of stale and can be closed.

PR #1169, opened 2025-10-13
Fix symmetries, round 2: uniform scaling

Technici4n:
@niklasschmitz realized that symmetries were incorrectly being filtered out for a uniform scaling. The simplest fix would have been to replace `wcart = A0 * w` by `wcart = A*w`, but we took this chance to fix the filtering such that it doesn't need access to primal values, in preparation for higher-order implementation.

### DIFF
--- a/src/common/quadrature.jl
+++ b/src/common/quadrature.jl
@@ -64,11 +64,10 @@ simpson(y::AbstractVector, x::AbstractVector) = simpson((i, xi) -> y[i], x)
     # Note: We used @turbo here before, but actually the allocation overhead
     #       needed to get all the data into an array is worse than what one gains
     #       with LoopVectorization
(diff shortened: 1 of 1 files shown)


test/forwarddiff.jl
```diff
@@ -330,6 +330,14 @@ end
         @test length(symmetries_filtered) == 4
         check_symmetries(symmetries_filtered, [mirrory, mirroryz])
     end
+
+    @testset "Isotropic lattice scaling" begin
+        lattice_modified = lattice + ε * I
```
niklasschmitz: This should be `(1 + ε) * lattice` for isotropic scaling

niklasschmitz:
Merging this as it is just a bug fix; Let's do a follow-up with nested Duals & higher-order symmetry breaking later.

PR #1172, opened 2025-10-15
Automate `elastic_constants` by AD-DFPT as a postprocessing

niklasschmitz:
Following up on the ideas at #1162.

This PR aims to automate the clamped-ion elastic constants as a convenient postprocessing function `elastic_constants(scfres)` for 1) the generic case of not using symmetry (most expensive, 6xDFPT on a fully unfolded kgrid...), 2.1) the most common case of cubic crystals (least expensive, 1xDFPT on a partially unfolded kgrid). Optimizations for other the point groups could be implemented later following a similar pattern.

progress:
- [x] Implement generic case
- [x] Implement cubic case
- [x] use SCF results from scfres and partially unfold (#1308; Update: just warm-starting with the correct density and tightened diagonalization tolerance)
- [x] more testing
- [x] mention in elastic constants docs example


test/elastic.jl
```diff
@@ -0,0 +1,37 @@
...
+    kgrid = (4, 4, 4)
+    tol = 1e-8
+    basis = PlaneWaveBasis(model0; Ecut, kgrid)
+
+    scfres = self_consistent_field(basis; tol)
+    (; voigt_stress, C) = DFTK.elastic_constants(scfres)
+
+    @test size(voigt_stress) == (6,)
+    @test size(C) == (6, 6)
+    # @test norm(C - Cref) < 3e-6  # ~0.1 GPa
```
mfherbst: I think it's fair to also compare against a reference from QE or similar here, just to make sure we agree. If too much effort, we could even compare against ourselves, I'm pretty convinced we do the right thing. just to make sure that we don't regress on this.

test/elastic.jl
```diff
@@ -0,0 +1,37 @@
...
+    atoms = fill(ElementPsp(silicon.atnum, load_psp(silicon.psp_gth)), 2)
+    model0 = model_DFT(lattice, atoms, silicon.positions; functionals=PBE())
+
+    Ecut = 18
+    kgrid = (4, 4, 4)
+    tol = 1e-8
+    basis = PlaneWaveBasis(model0; Ecut, kgrid)
+
+    scfres = self_consistent_field(basis; tol)
+    (; voigt_stress, C) = DFTK.elastic_constants(scfres)
```
mfherbst: `DFTK.` not needed I guess ?

src/postprocess/elastic.jl
```diff
@@ -0,0 +1,59 @@
...
+    # TODO restart SCF from previous
+    lattice = DFTK.voigt_strain_to_full(voigt_strain) * model0.lattice
+    model = Model(model0; lattice, symmetries)
+    basis = PlaneWaveBasis(model; Ecut, kgrid)
+    scfres = self_consistent_field(basis; tol, kwargs...)
+    DFTK.full_stress_to_voigt(compute_stresses_cart(scfres))
+end
+
+"""Computes the *clamped-ion* elastic constants (without ionic relaxation)"""
+function elastic_constants(scfres::NamedTuple;
```
mfherbst: How about `elastic_tensor` because that's what it is. The constants are the individual entries.

src/postprocess/elastic.jl
```diff
@@ -0,0 +1,59 @@
...
+    lattice = DFTK.voigt_strain_to_full(voigt_strain) * model0.lattice
+    model = Model(model0; lattice, symmetries=true)
+    model.symmetries
+end
+
+function stress_from_strain(model0, voigt_strain; symmetries, Ecut, kgrid, tol, kwargs...)
+    # TODO restart SCF from previous
+    lattice = DFTK.voigt_strain_to_full(voigt_strain) * model0.lattice
+    model = Model(model0; lattice, symmetries)
+    basis = PlaneWaveBasis(model; Ecut, kgrid)
```
mfherbst: This is dangerous as basis of the SCF may be very different (e.g. fft_grid may be set, symmetries may be customised, architecture etc.). We have to find a better mechanics here. I would be in favour of having a special PlaneWaveBasis constructor that essentially just changes the model under the hood and otherwise tries to keep as many of the settings of the "old basis". Would also be useful for https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/postprocess/stresses.jl#L22.

src/postprocess/elastic.jl
```diff
@@ -0,0 +1,59 @@
+import DifferentiationInterface as DI
+
+
+function symmetries_from_strain(model0, voigt_strain)
+    lattice = DFTK.voigt_strain_to_full(voigt_strain) * model0.lattice
+    model = Model(model0; lattice, symmetries=true)
+    model.symmetries
```
mfherbst: Can't we directly invoke the symmetry detection here instead of constructing the full model and then throwing it away ?

test/elastic.jl
niklasschmitz: Regarding this, I now opted to compare against our own value as well as against tightly converged finite differences. With kinetic_blowup we can do this comparison consistently at very low Ecut and kgrid; To compare against standard results would require converged discretization which would make this quite a slow test for the CI. I thus just added comments on this to the test.

src/postprocess/elastic.jl
niklasschmitz: Fair, I renamed it and added a comment to the docstring on it being the 6x6 matrix in Voigt notation

src/postprocess/elastic.jl
niklasschmitz: Now I pass all old basis args through like in stresses.jl . A better PlaneWaveBasis update constructor for swapping the Model would indeed be more elegant. I could try this in a smaller follow-up PR if this first version is good enough as is?

niklasschmitz: Actually I just added the new `PlaneWaveBasis(basis, model)` now while we are here. That's the one you meant, right?

src/postprocess/elastic.jl
```diff
@@ -0,0 +1,67 @@
...
+automatic differentiation of the stress tensor with respect to strain.
+Returns a named tuple `(; voigt_stress, C)` where `C[i,j] = ∂σᵢ/∂ηⱼ` is
+the 6×6 elastic tensor in Voigt notation.
+
+For cubic systems the three independent constants (C11, C12, C44) are
+obtained from a single directional derivative; for other symmetries the
+full Jacobian is computed.
+"""
+function elastic_tensor(scfres::NamedTuple;
+                        magnetic_moments=[],
```
niklasschmitz: Needing the `magnetic_moments` here is a bit weird, we only need it to redetermine symmetries as the model doesn't remember them. Alternatively we could have the Model store magnetic_moments as a field

niklasschmitz: Or better perhaps, the Model just has to remember not the magnetic_moments but the pointgroup_symbol that spglib gives

src/postprocess/elastic.jl
mfherbst: Yes, essentially. Maybe make `model` a keyword argument to be in line with the other update constructors ?

src/postprocess/elastic.jl
mfherbst: Yes this is an issue we have at multiple places in the code and that is bad, I agree.

Indeed, a solution could be to not just save the symmetry operations but also some metadata. This could also be useful for band structures etc. where we have the same problem. I'm not sure if pointgroup_symbol is the thing to rely on. There is also some integer number which identifies the crystal symmetry group. I think we support this already for getting the symmetry operations themselves (`symmetry_from_database`). The advantage is that Brillouin also supports this (unlike the strings). Could that be used plus calling some function to the the point group symbol from there be a solution ?

mfherbst: We can also push this down the road, if you cannot be asked to do it now. In that case just open an issue.

niklasschmitz: opened #1307

niklasschmitz:
> Something that bothers me as well is that it's basically impossible to send any kind of parameters to the DFPT solver in your implementation. There is the `ResponseOptions` for this purpose. Maybe just taking one as a keyword argument from the elastic constant wrapper function would be simple solution ?

Agreed. I now added the more general solution to pass arbitrary `kwargs_scf...` to `elastic_tensor`.

> The displacement `0.01` should depend on the `SYMMETRY_TOLERANCE`

 Makese sense. I added tol_symmetry as keyword argument and displace by `100 * tol_symmetry` (the `1e-5` default `SYMMETRY_TOLERANCE` gets `1e-3` strain).

examples/elastic_constants.jl
```diff
@@ -152,6 +152,21 @@ println("C44 (FD): ", uconvert(u"GPa", dstress_fd[4] * u"hartree" / u"bohr"^3))
...
+
+basis = PlaneWaveBasis(model0; Ecut, kgrid)
+scfres = self_consistent_field(basis; tol)
+(; C) = elastic_tensor(scfres; tol)
+
+println("C11: ", uconvert(u"GPa", C[1, 1] * u"hartree" / u"bohr"^3))
+println("C12: ", uconvert(u"GPa", C[1, 2] * u"hartree" / u"bohr"^3))
+println("C44: ", uconvert(u"GPa", C[4, 4] * u"hartree" / u"bohr"^3))
+
 # ## Moving to meta-GGA functionals
```
mfherbst: Use your function here instead of the manual shenanigans and move up as well.

examples/elastic_constants.jl
```diff
@@ -152,6 +152,21 @@ println("C44 (FD): ", uconvert(u"GPa", dstress_fd[4] * u"hartree" / u"bohr"^3))
 # Note the discrepancy in c44, which is due to us not yet including
 # ionic relaxation in this example.
 
+# ## Using the `elastic_tensor` postprocessing function
```
mfherbst: If I was a practitioner, this is what I care about not any of the things written above (which is "specialist's knowledge".) So this should come first. In fact, there may even be a point to move out the detailed instructions (same as for the polarisability) and make this a separate thing ... but this can come later.

src/postprocess/elastic.jl
mfherbst: Ok, but this needs a todo here as well.

src/postprocess/elastic.jl
```diff
@@ -0,0 +1,81 @@
...
+    model = Model(model0; lattice, symmetries)
+    basis = PlaneWaveBasis(basis0; model)
+    scfres = self_consistent_field(basis; kwargs_scf...)
+    DFTK.full_stress_to_voigt(compute_stresses_cart(scfres))
+end
+
+"""
+    elastic_tensor(scfres; magnetic_moments=[],
+                   tol_symmetry=SYMMETRY_TOLERANCE,
+                   kwargs_scf...)
```
mfherbst: I am not so sure. This is a very surprising interface. Imagine you converge the scf to 1e-12. You expect the elastic tensor calculation to respect this accuracy, but it won't. We should build a function to automatically extract the relevant SCF kwargs from the `scfres` for convergence control (`is_converged`, `mixing`, `damping`, ...) and use that here.

Regarding the kwarg, since this is a response step that comes after the SCF step, the only reasonable kwarg in my opinion is `ResponseOptions` (which should become the standard way to pass kwargs to the implicit `solve_ΩplusK_split` executed when an SCF is differentiated).

We can also settle on another pattern, but I think now is the moment we should work towards one. Because if we don't it quickly gets super messy. If all functions take just kwargs and these get decomposed along the call stack, we will loose the overview which kwargs can be set from where.

niklasschmitz: Agreed, I put `ResponseOptions` first now as kwarg, the other options are just workarounds.

I now inlined a scavenging of the scfres for SCF-related options, a temporary fix with a TODO. A cleaner option could be to bundle SCF options together in an scfres struct.

Even the `tol_symmetry` kwarg for `elastic_tensor` one might discuss whether it can be removed - to me it would make sense to store tol_symmetry as part of the model next to the symmetry group information (i.e. the symmetry result can be bundled with its uncertainty) then we don't have to have it as kwarg here and can just read it out of `scfres.basis.model`.

examples/elastic_constants.jl
niklasschmitz: done

examples/elastic_constants.jl
niklasschmitz: done

src/postprocess/elastic.jl
```diff
@@ -2,39 +2,58 @@ import DifferentiationInterface as DI
...
 function _stress_from_strain(basis0::PlaneWaveBasis, voigt_strain;
-                             symmetries=true, kwargs_scf...)
-    # TODO restart SCF from previous
+                             symmetries=true, ρ=nothing, kwargs_scf...)
     model0 = basis0.model
     lattice = DFTK.voigt_strain_to_full(voigt_strain) * model0.lattice
     model = Model(model0; lattice, symmetries)
     basis = PlaneWaveBasis(basis0; model)
-    scfres = self_consistent_field(basis; kwargs_scf...)
+    scfres = self_consistent_field(basis; ρ=something(ρ, guess_density(basis)),
```
mfherbst: `guess_density` needs the magnetic moments. I think it's fine to have `ρ` be a mandatory kwarg in this internal function, what do you think ?

src/postprocess/elastic.jl
mfherbst: > store tol_symmetry as part of the model next to the symmetry group information

Yes, can you add this to the other issue ?

src/postprocess/elastic.jl
niklasschmitz: Yes I think mandatory \rho is fine here, changed it

src/postprocess/elastic.jl
niklasschmitz: done

src/input_output.jl
```diff
@@ -369,6 +370,11 @@ function scfres_to_dict!(dict, scfres::NamedTuple; save_ψ=true, save_ρ=true)
         if hasproperty(scfres, :mixing)
             dict["mixing"] = string(scfres.mixing)
         end
+        for sym in (:is_converged, :nbandsalg, :fermialg, :diagtolalg, :solver, :eigensolver)
+            if hasproperty(scfres, sym)
+                dict[string(sym)] = string(getproperty(scfres, sym))
+            end
+        end
```
mfherbst, line 377: This is code duplication and I'm not sure if the strings you get out of `solver` and `eigensolver` are very useful. E.g. eigensolver could just be a large bunch of numbers (which could blow up the size of the json). Have you checked that ?

niklasschmitz: It seems to work fine for the function arguments, it just prints the name. On `examples/silicon.jl` it looks like this:
```
julia> dict = Dict()
Dict{Any, Any}()

julia> for sym in (:is_converged, :nbandsalg, :fermialg, :diagtolalg, :solver, :eigensolver)
     if hasproperty(scfres, sym)
         dict[string(sym)] = string(getproperty(scfres, sym))
     end
 end

julia> dict
Dict{Any, Any} with 6 entries:
  "nbandsalg"    => "AdaptiveBands(4, 7, 1.0e-6, 0.01)"
  "eigensolver"  => "lobpcg_hyper"
  "is_converged" => "ScfConvergenceDensity(1.0e-8)"
  "diagtolalg"   => "AdaptiveDiagtol(0.2, nothing, 0.005, 0.03)"
  "solver"       => "anderson"
  "fermialg"     => "FermiTwoStage()"
```

mfherbst: Perfect, great, thanks for checking !

PR #1185, opened 2025-11-03
Increase GPU robustness of LOBPCG

abussy:
Inversion and Cholesky decomposition of ill-conditioned matrices on the GPU (both AMD and NVIDIA) sometime yield NaN elements. This happens silently, but gets caught by the various `@assert !any(isnan, Array)` checks.

This rarely happens, but likelihood seems to increase when pushing very tight SCF convergence thresholds.

This PR introduces a fall-back to the more robust CPU routines. Impact on performance is negligible (rare event), and stability is greatly improved.

### DIFF
--- a/src/occupation.jl
+++ b/src/occupation.jl
@@ -30,6 +30,11 @@ end
 function compute_occupation(basis::PlaneWaveBasis{T}, eigenvalues::AbstractVector, εF::Number;
                             temperature=basis.model.temperature,
                             smearing=basis.model.smearing) where {T}
+    # Check that eigenvalues are increasing monotonically, for contiguous occupations
+    if !all(all(diff(εk) .≥ -eps(T)) for εk in eigenvalues)
+        error("Eigenvalues should be monotonically increasing.")
+    end
+
     # This is needed to get the right behaviour for special floating-point types
     # such as intervals.
     inverse_temperature = iszero(temperature) ? T(Inf) : 1/temperature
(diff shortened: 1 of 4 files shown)


antoine-levitt:
Let me try some refactoring and I'll send it to you

src/gpu/gpu_arrays.jl
```diff
@@ -5,6 +5,14 @@ using Preferences
 # https://github.com/JuliaGPU/CUDA.jl/issues/1565
 LinearAlgebra.dot(x::AbstractGPUArray, D::Diagonal, y::AbstractGPUArray) = x' * (D * y)
 
+# Norm of Hermitian matrices. See https://github.com/JuliaGPU/AMDGPU.jl/issues/843
+function LinearAlgebra.norm(A::Hermitian{T, <:AbstractGPUArray}) where {T}
```
mfherbst, line 10: Do we also need this for CUDA ? Going by the same argument that this can have side effects (e.g. unexpected performance drop), maybe we should be more specific here and only do this for rocArrays ?

abussy: The issue is also present with CUDA. I initially discovered it on AMD, hence the message. This reminds to actually open an issue on CUDA.jl as well. I'll reference it here too.

mfherbst: Perfect, then I agree with the code.

PR #1187, opened 2025-11-07
Port stress/response calculations to the GPU

abussy:
This PR enables ForwardDiff calculations (stress and response) on the GPU. Main changes are:

1) Data transfer from/to the device where necessary
2) Various small changes to avoid GPU compiler confusion (e.g. see changes in `src/workarounds/forwarddiff_rules.jl`)
3) CPU fall-backs for all XC operations taking place in the `DftFunctionals.jl` package
4) Refactoring of the ForwardDiff tests, such that all tests can be run on various architectures (CPU, CUDA, AMDGPU)

With this PR, all ForwardDiff workflows currently tested on the CPU successfully run on both NVIDIA and AMD GPUs.

Future improvements will come with:
- PR #1185 for the tests to consistently finish on GPUs (right now, they regularly fail due to Cholesky instability on the GPU)
- PR https://github.com/JuliaMolSim/DftFunctionals.jl/pull/23 to pave the way for XC operations on the GPU
- PR #1163 for more efficient `PlaneWaveBasis` instantiation

### DIFF
--- a/docs/src/tricks/compute_clusters.md
+++ b/docs/src/tricks/compute_clusters.md
@@ -27,7 +27,7 @@ the `JULIA_DEPOT_PATH` to be a subdirectory of `/scratch`.
 **EPFL scitas.**
 On scitas the right thing to do is to insert
 ```
-export JULIA_DEPOT_PATH="$JULIA_DEPOT_PATH:/scratch/$USER/.julia"
+export JULIA_DEPOT_PATH="/scratch/$USER/.julia"
 ```
 into your `~/.bashrc`.
 
--- a/examples/hubbard.jl
+++ b/examples/hubbard.jl
@@ -58,7 +58,10 @@ U = 10u"eV"
 manifold = OrbitalManifold(Ni, "3D")
 
 # Run SCF with a DFT+U setup, notice the `extra_terms` keyword argument, setting up the Hubbard +U term.
-model = model_DFT(lattice, atoms, positions; extra_terms=[Hubbard(manifold, U)],
+# It is also possible to set up multiple manifolds with different U values by passing each pair
+# as a separate entry in the `Hubbard` constructor (i.e. `Hubbard(manifold1 => U1, manifold2 => U2, etc.)`)
+# or as two vectors (i.e. `Hubbard([manifold1, manifold2, etc.], [U1, U2, etc.])`).
+model = model_DFT(lattice, atoms, positions; extra_terms=[Hubbard(manifold => U)],
                   functionals=PBE(), temperature=5e-3, magnetic_moments)
 basis = PlaneWaveBasis(model; Ecut=20, kgrid=[2, 2, 2])
 scfres = self_consistent_field(basis; tol=1e-6, ρ=guess_density(basis, magnetic_moments));
--- a/src/terms/hubbard.jl
+++ b/src/terms/hubbard.jl
@@ -89,7 +89,7 @@ end
 
 function extract_manifold(manifold::ResolvedOrbitalManifold, projectors, labels)
     # We extract the labels only for orbitals belonging to the manifold
-    proj_indices = findall(orb -> (orb.iatom in manifold.iatoms
+    proj_indices = findall(orb -> (orb.iatom ∈  manifold.iatoms
                                 && orb.l     == manifold.l
                                 && orb.n     == manifold.i), labels)
     isempty(proj_indices) && @warn "Projector for $(manifold) not found."
@@ -107,26 +107,42 @@ Hubbard energy, following the Dudarev et al. (1998) rotationally invariant forma
 ```
 """
 struct Hubbard{T}
-    manifold::OrbitalManifold
-    U::T
-    function Hubbard(manifold::OrbitalManifold, U)
-        U = austrip(U)
-        new{typeof(U)}(manifold, U)
+    manifolds::Vector{OrbitalManifold}
+    U::Vector{T}
+    function Hubbard{T}(manifolds::Vector{OrbitalManifold}, U::Vector{T}) where {T}
+        length(manifolds) == length(U) ||
+            error("Number of U values ($(length(U))) must match "
+                  * "number of manifolds ($(length(manifolds))).")
+        new(manifolds, U)
     end
 end
-function (hubbard::Hubbard)(basis::AbstractBasis)
-    manifold = resolve_hubbard_manifold(hubbard.manifold, basis.model)
+function Hubbard(manifolds::Vector{OrbitalManifold}, U::Vector{T}) where {T}
+    U = austrip.(U)
+    Hubbard{eltype(U)}(manifolds, U)
+end
+function Hubbard(manifold_to_U::Vararg{<:Pair})
+    Hubbard([m[1] for m in manifold_to_U], [m[2] for m in manifold_to_U])
+end
 
+function (hubbard::Hubbard{T})(basis::AbstractBasis) where {T}
+    manifolds = [resolve_hubbard_manifold(manifold, basis.model) for manifold in hubbard.manifolds]
     projs, labels = atomic_orbital_projectors(basis)
-    labels, projectors_matrix = extract_manifold(manifold, projs, labels)
-    TermHubbard(manifold, hubbard.U, projectors_matrix, labels)
+    manifold_data = map(manifold -> extract_manifold(manifold, projs, labels), manifolds)
+    manifold_labels = [data.manifold_labels for data in manifold_data]
+    projs_matrices = [data.manifold_projectors for data in manifold_data]
+    # Concatenate projectors from all manifolds per each k-point
+    projs_per_kpt = map(1:length(basis.kpoints)) do ik
+        reduce(hcat, map(P -> P[ik], projs_matrices))
+    end
+    TermHubbard(manifolds, hubbard.U, projs_matrices, projs_per_kpt, manifold_labels)
 end
 
 struct TermHubbard{T, PT, L} <: Term
-    manifold::ResolvedOrbitalManifold
-    U::T   # U value
-    P::PT  # projectors
-    labels::L
+    manifolds::Vector{ResolvedOrbitalManifold}
+    U::Vector{T}  # U values
+    P::Vector{PT} # projectors per manifold
+    P_vec::Vector{Matrix{Complex{T}}} # projectors per k-point (all manifolds concatenated)
+    labels::Vector{L}
 end
 
 @timing "ene_ops: hubbard" function ene_ops(term::TermHubbard,
@@ -136,22 +152,32 @@ end
     if isnothing(hubbard_n)
        return (; E=zero(T), ops=[NoopOperator(basis, kpt) for kpt in basis.kpoints])
     end
-    proj = term.P
-
+   
     filled_occ = filled_occupation(basis.model)
-    natoms = length(term.manifold.iatoms)
     n_spin = basis.model.n_spin_components
-    nproj_atom = size(hubbard_n[1,1,1], 1) # This is the number of projectors per atom, namely 2l+1
-    # For the ops we have to reshape hubbard_n to match the NonlocalOperator structure, using a block diagonal form
-    nhub = [zeros(Complex{T}, nproj_atom*natoms, nproj_atom*natoms) for _ in 1:n_spin]
-    E = zero(T)
-    for σ in 1:n_spin, iatom in 1:natoms
-        proj_range = (1+nproj_atom*(iatom-1)):(nproj_atom*iatom)
-        nhub[σ][proj_range, proj_range] = hubbard_n[σ, iatom, iatom]
-        E += filled_occ * 1/T(2) * term.U *
-             real(tr(hubbard_n[σ, iatom, iatom] * (I - hubbard_n[σ, iatom, iatom])))
+    projs = term.P_vec
+    # For the ops we have to reshape hubbard_n to match the NonlocalOperator structure, 
+    # using a block diagonal form whit one block per atom per manifold
+    # 2l+1 is the number of projectors per atom, i.e. the dimension of the blocks
+    D_size = 0
+    for manifold in term.manifolds
+        D_size += (2*manifold.l + 1) * length(manifold.iatoms)
     end
-    ops = [NonlocalOperator(basis, kpt, proj[ik], 1/T(2) * term.U * (I - 2*nhub[kpt.spin])) 
+    D = [zeros(Complex{T}, D_size, D_size) for _ in 1:n_spin]
+
+    E = zero(T)
+    offset = 0 
+    for (iman, manifold) in enumerate(term.manifolds)
+        nproj_atom = 2*manifold.l + 1
+        for σ in 1:n_spin, iatom in 1:length(manifold.iatoms)
+            proj_range = (offset + (iatom-1)*nproj_atom + 1):(offset + iatom*nproj_atom)
+            D[σ][proj_range, proj_range] = 1/T(2) * term.U[iman] * (I - 2*hubbard_n[iman][σ, iatom, iatom])
+            E += filled_occ * 1/T(2) * term.U[iman] *
+                 real(tr(hubbard_n[iman][σ, iatom, iatom] * (I - hubbard_n[iman][σ, iatom, iatom])))
+        end
+        offset += nproj_atom*length(manifold.iatoms)
+    end
+    ops = [NonlocalOperator(basis, kpt, projs[ik], D[kpt.spin]) 
            for (ik,kpt) in enumerate(basis.kpoints)]
     return (; E, ops)
 end
@@ -171,8 +197,10 @@ and ``Pᵢₘ₁`` is the pseudoatomic orbital projector for atom i and orbital 
 (just the magnetic quantum number, since l is fixed, as is usual in the literature).
 For details on the projectors see `atomic_orbital_projectors`.
 """
-function compute_hubbard_n(manifold::ResolvedOrbitalManifold, projectors, labels,
-                           basis::PlaneWaveBasis{T}, ψ, occupation) where {T}
+function compute_hubbard_n(manifold::ResolvedOrbitalManifold,
+                           projectors, labels,
+                           basis::PlaneWaveBasis{T},
+                           ψ, occupation) where {T}
     filled_occ = filled_occupation(basis.model)
     n_spin = basis.model.n_spin_components
 
@@ -195,10 +223,11 @@ function compute_hubbard_n(manifold::ResolvedOrbitalManifold, projectors, labels
             hubbard_n[σ, idx, jdx] = mpi_sum(hubbard_n[σ, idx, jdx], basis.comm_kpts)
         end
     end
-    hubbard_n = symmetrize_hubbard_n(basis.model, manifold, hubbard_n; basis.symmetries)
+    symmetrize_hubbard_n(basis.model, manifold, hubbard_n; basis.symmetries)
 end
 function compute_hubbard_n(term::TermHubbard, basis::PlaneWaveBasis, ψ, occupation)
-    compute_hubbard_n(term.manifold, term.P, term.labels, basis, ψ, occupation)
+    [compute_hubbard_n(manifold, P, labels, basis, ψ, occupation)
+     for (manifold, P, labels) in zip(term.manifolds, term.P, term.labels)]
 end
 
 """
--- a/test/hamiltonian_consistency.jl
+++ b/test/hamiltonian_consistency.jl
@@ -108,9 +108,13 @@ end
     test_consistency_term(Hartree())
     let
         Si = ElementPsp(14, load_psp(testcase.psp_upf))
-        test_consistency_term(Hubbard(OrbitalManifold([1, 2], "3P"), 0.01), atom=Si)
-        test_consistency_term(Hubbard(OrbitalManifold([1, 2], "3P"), 0.01), atom=Si,
+        test_consistency_term(Hubbard(OrbitalManifold([1, 2], "3P") => 0.01), atom=Si)
+        test_consistency_term(Hubbard(OrbitalManifold([1, 2], "3P") => 0.01), atom=Si,
                               spin_polarization=:collinear)
+        test_consistency_term(Hubbard(OrbitalManifold([1, 2], "3S") => 0.01,
+                                      OrbitalManifold([1, 2], "3P") => 0.02), atom=Si)
+        test_consistency_term(Hubbard([OrbitalManifold(Si, "3S"), OrbitalManifold(Si, "3P")],
+                                      [0.01, 0.02]), atom=Si)
     end
     # Disabled since the energy is constant, and the test guards against 0 differences
     # test_consistency_term(Ewald())
--- a/test/hubbard.jl
+++ b/test/hubbard.jl
@@ -102,7 +102,7 @@ end
    manifold = OrbitalManifold(Ni, "3D")
    
    model = model_DFT(lattice, atoms, positions; 
-                     extra_terms=[Hubbard(manifold, U)],
+                     extra_terms=[Hubbard(manifold => U)],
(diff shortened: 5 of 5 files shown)


abussy:
Merged master. Adapted tests to the refactoring brought by #1182.

Additionally, removed this problematic bit of code in `ext/DFTKAMDGPUExt.jl`:
```
# Enable comparisons of Duals on AMD GPUs
_val(x) = x
_val(x::Dual) = _val(ForwardDiff.value(x))
function Base.:<(x::Dual{T,V,N},
                 y::Dual{T,V,N}) where {T,V,N}
    _val(x) < _val(y)
end
function Base.:>(x::Dual{T,V,N},
                 y::Dual{T,V,N}) where {T,V,N}
    _val(x) > _val(y)
end
```

It turns out that comparison of Duals does not take place on the GPU, as long as all XC operations are done on the CPU. This might become a concern again in the future, once `DftFunctionals.jl` is refactored.

ext/DFTKAMDGPUExt.jl
```diff
@@ -4,6 +4,8 @@ using PrecompileTools
 using LinearAlgebra
 import DFTK: GPU, precompilation_workflow
 using DFTK
+import ForwardDiff
+import ForwardDiff: Dual
```
mfherbst: Can also be removed, right ?

src/gpu/gpu_arrays.jl
```diff
@@ -23,3 +23,14 @@ for fun in (:potential_terms, :kernel_terms)
...
+
+# Make sure that computations done by DftFunctionals.jl are done on the CPU (until refactoring)
+for fun in (:potential_terms, :kernel_terms)
+    @eval function DftFunctionals.$fun(fun::DispatchFunctional, ρ::AT,
+                                       args...) where {AT <: AbstractGPUArray}
+        # Fallback implementation for the GPU: Transfer to the CPU and run computation there
+        cpuify(::Nothing) = nothing
+        cpuify(x::AbstractArray) = Array(x)
+        $fun(fun, Array(ρ), cpuify.(args)...)
+    end
```
mfherbst: If we have this, do we need the above version at all (which is simply more specific to `Float64`, right ?

src/workarounds/forwarddiff_rules.jl
```diff
@@ -241,28 +242,29 @@ end
...
     end
 
     # Convert and combine
-    DT = Dual{ForwardDiff.tagtype(T)}
     ψ = map(scfres.ψ, getfield.(δresults, :δψ)...) do ψk, δψk...
         map(ψk, δψk...) do ψnk, δψnk...
-            Complex(DT(real(ψnk), real.(δψnk)),
-                    DT(imag(ψnk), imag.(δψnk)))
+            Complex(Dual{T}(real(ψnk), real.(δψnk)),
+                    Dual{T}(imag(ψnk), imag.(δψnk)))
```
mfherbst: Is this the right thing to do ? Does this not loose the tags ?
@niklasschmitz @Technici4n  may have better insights here.

src/terms/xc.jl
```diff
@@ -529,11 +529,19 @@ _matify(data::AbstractArray) = reshape(data, size(data, 1), :)
...
 for fun in (:potential_terms, :kernel_terms)
     @eval begin
-        function DftFunctionals.$fun(xc::Functional, density::LibxcDensities)
+        function DftFunctionals.$fun(xc::DispatchFunctional, density::LibxcDensities)
             $fun(xc, _matify(density.ρ_real), _matify(density.σ_real),
                      _matify(density.τ_real), _matify(density.Δρ_real))
         end
 
+        # Ensure functionals from DftFunctionals are sent to the CPU, until DftFunctionals.jl is refactored
+        function DftFunctionals.$fun(fun::DftFunctionals.Functional, density::LibxcDensities)
```
mfherbst, line 539: This looks weird to me. Why is this needed on top of the above ? Should the types not somehow depend on a GPU type here ?

src/workarounds/forwarddiff_rules.jl
```diff
@@ -31,9 +31,9 @@ end
 function LinearAlgebra.mul!(y::AbstractArray{<:Union{Complex{<:Dual}}},
                             p::AbstractFFTs.Plan,
                             x::AbstractArray{<:Union{Complex{<:Dual}}})
-    copyto!(y, p*x)
+    copyto!(y, _mul(p, x))
 end
-function Base.:*(p::AbstractFFTs.Plan, x::AbstractArray{<:Complex{<:Dual{Tg}}}) where {Tg}
+function _mul(p::AbstractFFTs.Plan, x::AbstractArray{<:Complex{<:Dual{Tg}}}) where {Tg}
```
mfherbst: Again this feels strange and is surprising to me. Why did you need this ?

test/forwarddiff.jl
```diff
@@ -1,492 +1,66 @@
-# Wrappers around ForwardDiff to fix tags and reduce compilation time.
```
mfherbst, line 1: I like the general idea to make these tests more generic, but I don't like the splitup. Because this way it's one more level of indirection I have to follow to see the actual test code (from the file and line information printed in the test suite output).

I recall I wrote test functions elsewhere that actually internally create multiple `@testcase` instances. Like you call just one function for the CPU tests and it internally actually creates the "templated" testcases with CPU architecture. Would that be an option here ? If not, reducing code duplication clearly wins and this is likely a good option.

Also for the cases that do not depend on `architecture`, I'd avoid the indirection alltogether and just keep the test code here.

But I have to think about this a bit more, if I can think of ideas how to improve the cut between the multiple files.

src/gpu/gpu_arrays.jl
abussy: Right, we probably only need the most general workaround. The thought process was that, once DftFunctionals.jl is refactored to run on the GPU, we can simply remove the `AbstractGPUArray` definition.

It's probably best to keep the code in the cleanest state in the meantime though, I'll see to it.

src/workarounds/forwarddiff_rules.jl
Technici4n: Yes that seems to loose the tags

src/workarounds/forwarddiff_rules.jl
abussy: Without this workaround, the GPU compiler throws an `invalid LLVM IR` error during stress calculations.  I think there is confusion around which method of `Base.:*` to use, but I don't understand why.

src/workarounds/forwarddiff_rules.jl
abussy: But isn't the T of `Dual{T,V,N}}` the tag? I am confused here

src/terms/xc.jl
abussy: The tricky bit is to avoid ambiguity with the [internal definitions of DftFunctionals.jl](https://github.com/JuliaMolSim/DftFunctionals.jl/blob/45eaebec95ebe509e6007535db06c1f4a09a6465/src/interface.jl#L79-L85).

Adding the following to `src/gpu/gpu_array.jl` leads to ambiguity, because it does not specialize on the type of functional (:lda, :gga, or :mgga):
```
for fun in (:potential_terms, :kernel_terms)
    @eval function DftFunctionals.$fun(fun::DftFunctionals.Functional, ρ::AT,
                                       args...) where {AT <: AbstractGPUArray}
        # Fallback implementation for the GPU: Transfer to the CPU and run computation there
        cpuify(::Nothing) = nothing
        cpuify(x::AbstractArray) = Array(x)
        $fun(fun, Array(ρ), cpuify.(args)...)
    end
end
```

Either I write the above for each functional type (lot of code duplication), or I parametrize it with a second loop over functional types, e.g.:
```
for fun in (:potential_terms, :kernel_terms), ftype in (:lda, :gga, :mgga)
    @eval function DftFunctionals.$fun(fun::DftFunctionals.Functional{$(QuoteNode(ftype))},
                                       ρ::AT, args...) where {AT <: AbstractGPUArray}
        # Fallback implementation for the GPU: Transfer to the CPU and run computation there
        cpuify(::Nothing) = nothing
        cpuify(x::AbstractArray) = Array(x)
        $fun(fun, Array(ρ), cpuify.(args)...)
    end
end
```

I don't like any of these alternative very much. I think the current solution carries a very clear message: anything DftFunctionals related goes to the CPU.

src/workarounds/forwarddiff_rules.jl
Technici4n: Oh right `T` is the tag and not the value type. Seems fine then, this is just making the `T,V,N` explicit instead of letting the compiler infer them. I am surprised that this is required though. Would the root problem be the type-unstable nature of `construct_value(basis)`?

abussy: Would
```
Complex(Dual{T,V,N}(real(ψnk), real.(δψnk)),
        Dual{T,V,N}(imag(ψnk), imag.(δψnk)))
```
solve the issue?

Also, what about here? Don't we have a similar situation?
https://github.com/JuliaMolSim/DFTK.jl/blob/3551ad319a41f070361e65b8077624fb09f6eabe/src/workarounds/forwarddiff_rules.jl#L321-L326

Technici4n: Ah no, I mean that `Dual{T}` is fine. I misread it the first time. What is weird to me is why this is necessary, given that the gpu compiler isn't operating directly on this method? (Or is it?)

Looking at `@code_typed` might be quite insightful

abussy: Sorry, I didn't see @Technici4n last message when I wrote the above.

The GPU compilers requires this to be able to compile. I am not sure what the root cause is, but type instability is a likely candidate. Generally, the GPU compiler is rather bad at type inference.

src/terms/xc.jl
Technici4n: I agree, this is a bit weird

src/terms/local_nonlinearity.jl
```diff
@@ -9,11 +9,13 @@ struct TermLocalNonlinearity{TF} <: TermNonlinear
 end
 (L::LocalNonlinearity)(::AbstractBasis) = TermLocalNonlinearity(L.f)
 
+# FD on the GPU, when T<:Dual causes all sorts of troubles, at least on AMD. TODO: also on NVIDIA?
+# TODO: only transfer to CPU when T <: Dual ?, or only if ROCArray?
```
Technici4n: Move comment inside, otherwise good enough?

test/forwarddiff_gpu.jl
```diff
@@ -0,0 +1,165 @@
+### ForwardDiff tests on the GPU, closely following forwarddiff.jl tests
```
Technici4n: There's probably a less verbose way to do this

src/terms/xc.jl
```diff
@@ -529,11 +529,19 @@ _matify(data::AbstractArray) = reshape(data, size(data, 1), :)
 
 for fun in (:potential_terms, :kernel_terms)
     @eval begin
-        function DftFunctionals.$fun(xc::Functional, density::LibxcDensities)
+        function DftFunctionals.$fun(xc::DispatchFunctional, density::LibxcDensities)
             $fun(xc, _matify(density.ρ_real), _matify(density.σ_real),
                      _matify(density.τ_real), _matify(density.Δρ_real))
         end
 
+        # Ensure functionals from DftFunctionals are sent to the CPU, until DftFunctionals.jl is refactored
```
Technici4n: As discussed, change the comment to add TODO and mention that it can be removed once DftFunctionals is fixed.

src/terms/xc.jl
mfherbst: I am still not too happy about this, but ok for now.

src/workarounds/forwarddiff_rules.jl
```diff
@@ -219,10 +221,9 @@ function construct_value(basis::PlaneWaveBasis{T}) where {T <: Dual}
 end
 
 
-@timing "self_consistent_field ForwardDiff" function self_consistent_field(
-        basis_dual::PlaneWaveBasis{T};
-        response=ResponseOptions(),
-        kwargs...) where {T <: Dual}
+@timing "self_consistent_field ForwardDiff" function self_consistent_field(basis_dual::PlaneWaveBasis{<:Dual{Tg,V,N}};
+                               response=ResponseOptions(),
+                               kwargs...) where {Tg,V,N}
```
mfherbst: The reformetting looks unintentional.

src/workarounds/forwarddiff_rules.jl
mfherbst: Ok, this we need to understand.

@niklasschmitz I recall we anyway only needed this because on the AbstractFFT side this was not properly supported. Could it be that now it is and we can drop our type piracy workaround alltogether ?

src/workarounds/forwarddiff_rules.jl
```diff
@@ -241,28 +242,29 @@ end
...
     end
     eigenvalues = map(scfres.eigenvalues, getfield.(δresults, :δeigenvalues)...) do εk, δεk...
-        map((εnk, δεnk...) -> DT(εnk, δεnk), εk, δεk...)
+        map((εnk, δεnk...) -> Dual{Tg}(εnk, δεnk), εk, δεk...)
     end
     occupation = map(scfres.occupation, getfield.(δresults, :δoccupation)...) do occk, δocck...
-        map((occnk, δoccnk...) -> DT(occnk, δoccnk), occk, δocck...)
+        occk_cpu = to_cpu(occk)
+        to_device(basis_dual.architecture,
+                  map((occnk, δoccnk...) -> Dual{Tg}(occnk, δoccnk), occk_cpu, δocck...))
```
mfherbst: This should work:
```suggestion
        to_device(basis_dual.architecture, map(Dual{Tg}, occk_cpu, δocck...))
```

test/forwarddiff/generic.jl
```diff
@@ -0,0 +1,44 @@
+### Generic ForwardDiff tests, independent of architecture
```
mfherbst, line 1: Maybe rather place in a subfolder ? `test/forwarddiff/generic.jl`

I like the explanations at the top of the file.

test/forwarddiff/geometry.jl
```diff
@@ -0,0 +1,336 @@
...
+    using AMDGPU
+    if CUDA.has_cuda() && CUDA.has_cuda_gpu()
+        StrainSensitivity.run_test(DFTK.GPU(CuArray))
+    end
+    if AMDGPU.has_rocm_gpu()
+        StrainSensitivity.run_test(DFTK.GPU(ROCArray))
+    end
+end
+
+@testmodule FilterBrokenSymmetries begin
```
mfherbst, line 169: Do we need this test in every architecture or is just CPU enough ?

test/forwarddiff/parameters.jl
```diff
@@ -0,0 +1,268 @@
+### ForwardDiff tests related to perturbations of model parameters. Each test is defined
+### in its own explicitly named module, implementing a run_test(architecture) function. This
+### allows easy testing of the same feature on CPU and GPU. @testitem are located immediatly
+### below the @testmodule definition. Note that the order of the test modules in the @testitem 
+### setup matters for successful compilation. 
+@testmodule PspSensitivity begin
```
mfherbst, line 7: empty line before

test/forwarddiff/geometry.jl
```diff
@@ -0,0 +1,336 @@
+### TODO: make sure the ForwardDiffWrappers help timings in this way, by measuring const
+###       of all FD tests at once. Worry is that recompilation occurs for wach file
+
+### ForwardDiff tests related to goemetry and symmetry perturbations. Each test is defined
+### in its own explicitly named module, implementing a run_test(architecture) function. This
+### allows easy testing of the same feature on CPU and GPU. @testitem are located immediatly
+### below the @testmodule definition. Note that the order of the test modules in the @testitem 
+### setup matters for successful compilation.
+@testmodule ForceDerivatives begin
```
mfherbst, line 7: empty line before

test/forwarddiff/parameters.jl
```diff
@@ -0,0 +1,268 @@
...
+end
+
+@testitem "Test scfres dual has the same params as scfres primal" tags=[:dont_test_mpi, :minimal] #=
+    =#    setup=[TestCases, ScfresParameterConsistency] begin
+    using DFTK
+    ScfresParameterConsistency.run_test(DFTK.CPU())
+end
+
+@testitem "Test scfres dual has the same params as scfres primal (GPU)" tags=[:gpu] #=
+    =#    setup=[TestCases, ScfresParameterConsistency] begin
```
mfherbst, line 210: Why do you need the `TestCases` module in these test callers ?

test/forwarddiff/parameters.jl
```diff
@@ -0,0 +1,268 @@
...
+    using DFTK
+    using CUDA
+    using AMDGPU
+    if CUDA.has_cuda() && CUDA.has_cuda_gpu()
+        ScfresParameterConsistency.run_test(DFTK.GPU(CuArray))
+    end
+    if AMDGPU.has_rocm_gpu()
+        ScfresParameterConsistency.run_test(DFTK.GPU(ROCArray))
+    end
```
mfherbst, line 219: This is super mechanical and bloaty. I'm divided, but in principle one could write a function `run_test_on_gpu` that just gets a function like
```julia
function run_test_on_gpu(testfunction, args...; kwargs...)
    if has_cuda()
        testfunction(DFTK.GPU(CuArray), args...; kwargs...)
    end
    if has_rocm()
        testfunction(DFTK.GPU(ROCArray), args...; kwargs...)
    end
end
```
put it in a module (which loads the modules once) and then just keep calling that ?

mfherbst (commented):
Very nice refactoring !

src/workarounds/forwarddiff_rules.jl
mfherbst: I suppose this is resolved ?

src/workarounds/forwarddiff_rules.jl
abussy: > I made progress there by properly reading the error message, and it turns out that a more specific definition of Base.:* in cufft takes priority: https://github.com/JuliaGPU/CUDA.jl/blob/44cde93bf03812012da5c883b6532d80a5226268/lib/cufft/fft.jl#L359-L377. While I understand the problem now, I don't see a better way to deal with it than the current solution. Any help/suggestion is welcome.

CUDA.jl overloads `Base.:*` for more specific types than `AbstractFFTs.Plan` and `AbstractArray`, but it breaks with (complex) Duals.

src/workarounds/forwarddiff_rules.jl
abussy: Indeed.

test/forwarddiff/generic.jl
abussy: > Maybe rather place in a subfolder ? test/forwarddiff/generic.jl

Good idea, I think it will be cleaner down the line.

test/forwarddiff/geometry.jl
abussy: As far as I can tell, it's all CPU code. However, I'd say it does not hurt to keep a GPU test, and it might save time down the line if things break.

test/forwarddiff/parameters.jl
abussy: Without it, imports like `using ..TestCases: silicon` fail in a `@testmodule`

test/forwarddiff/parameters.jl
abussy: I personally prefer the current explicit logic, I think it is simpler to read and modify.

test/forwarddiff/parameters.jl
mfherbst: So you need to do this in every test item, such that you can do `using ..TestCases: silicon` in a `@testmodule`. This seems wrong and the very least is extremely unexpected. Either we solve this in the `@testmodule` itself or if this is not possible we should explicitly pass the `silicon` object into the `@testmodule`. Otherwise someone may end up removing the `setup=[TestCases]` and get very unexpected errors.

test/forwarddiff/geometry.jl
mfherbst: Well if that's an issue would not the next test case just break on the GPU ?

src/workarounds/forwarddiff_rules.jl
mfherbst: So it's a bug in CUDA.jl, effectively ? Their typing is too broad as it covers Duals, which they don't support ?

test/forwarddiff/parameters.jl
mfherbst: Fair enough. Test code is usually not that volatile.

test/forwarddiff/parameters.jl
abussy: Not only `TestCases` must be passed to the test setup, but the order matters too! If you define a `@testmodule` named `MyTest` that depends on `TestCases`, then `setup=[TestCases, MyTest]` works, but `setup=[MyTest, TestCases]` fails.

That being said, I think this goes beyond this PR. Other parts of the tests rely on the same construct, e.g. all the `silicon_*.jl`. I feel like this should be done as part of a larger refactoring of the tests.

src/workarounds/forwarddiff_rules.jl
abussy: That's my understanding, yes.

test/forwarddiff/geometry.jl
abussy: I think you're right, only the CPU test is necessary.

test/forwarddiff/parameters.jl
mfherbst: ok, can you add an issue with the `testing` label ?

Like something to refactor the testing for CPU / GPU branching and for making sure that this module reordering business is resolved.

src/workarounds/forwarddiff_rules.jl
```diff
@@ -251,28 +253,28 @@ end
...
         map(ψk, δψk...) do ψnk, δψnk...
-            Complex(DT(real(ψnk), real.(δψnk)),
-                    DT(imag(ψnk), imag.(δψnk)))
+            Complex(Dual{Tg}(real(ψnk), real.(δψnk)),
+                    Dual{Tg}(imag(ψnk), imag.(δψnk)))
         end
     end
     eigenvalues = map(scfres.eigenvalues, getfield.(δresults, :δeigenvalues)...) do εk, δεk...
-        map((εnk, δεnk...) -> DT(εnk, δεnk), εk, δεk...)
+        map((εnk, δεnk...) -> Dual{Tg}(εnk, δεnk), εk, δεk...)
```
mfherbst, line 274: Does the map dual trick not apply here too ?

src/workarounds/forwarddiff_rules.jl
mfherbst: Once there is an issue opened and referenced here (please  `@mention me`) this is fine.

mfherbst (commented):
Final nits before we merge.

src/workarounds/forwarddiff_rules.jl
```diff
@@ -33,7 +33,9 @@ function LinearAlgebra.mul!(y::AbstractArray{<:Union{Complex{<:Dual}}},
                             x::AbstractArray{<:Union{Complex{<:Dual}}})
     copyto!(y, p*x)
 end
-function Base.:*(p::AbstractFFTs.Plan, x::AbstractArray{<:Complex{<:Dual{Tg}}}) where {Tg}
+# Custom rule for FFTs used to overload the Base.:* operator. This is a function
+# of it's own because DFTKCUDAExt.jl uses it to implement a specific overload.
+function dual_fft(p::AbstractFFTs.Plan, x::AbstractArray{<:Complex{<:Dual{Tg}}}) where {Tg}
```
mfherbst: `dual_fft_mul` ?

abussy: Sure

PR #1188, opened 2025-11-11
Updates for isbits functionals

Technici4n:
- Updates DFTK following https://github.com/JuliaMolSim/DftFunctionals.jl/pull/24.
- Adds an abstract `TermType` type that terms in the model must subtype.

Currently including the manifest such that I can run CI.

### DIFF
--- a/src/terms/hubbard.jl
+++ b/src/terms/hubbard.jl
@@ -106,7 +106,7 @@ Hubbard energy, following the Dudarev et al. (1998) rotationally invariant forma
 1/2 Σ_{σI} U * Tr[hubbard_n[σ,I,I] * (1 - hubbard_n[σ,I,I])]
 ```
 """
-struct Hubbard{T}
+struct Hubbard{T} <: TermType
     manifold::OrbitalManifold
     U::T
     function Hubbard(manifold::OrbitalManifold, U)


src/workarounds/forwarddiff_rules.jl
```diff
@@ -203,6 +203,28 @@ function construct_value(psp::PspUpf{T,I}) where {T <: AbstractFloat, I <: Abstr
     psp
 end
 
+# term types don't have a supertype, so we use a different function name
```
mfherbst: That can be changed ...

src/terms/xc.jl
```diff
@@ -25,7 +25,14 @@ end
 
 function Base.show(io::IO, xc::Xc)
     fac = isone(xc.scaling_factor) ? "" : ", scaling_factor=$(xc.scaling_factor)"
-    fun = join(xc.functionals, ", ")
+    functional_identifiers = map(xc.functionals) do f
+        if f isa DispatchFunctional
+            string(f.inner.identifier)
+        else
+            string(f)
+        end
```
mfherbst: I don't like this because it implies that when explicitly using `DftFunctionals` to make a standard functional such as LDA exchang (as opposed to a `DispatchFunctional`) a different string results, right? Moreover I'm not sure the default returned `string(f)` is super nice (especially when we have dual type parameters) ... but perhaps that we should fix inside `DftFunctionals` by providing good show functions for all functionals ?

Technici4n: Here is an example:

Without duals:
```
PbeExchange{Float64, Float64}(0.804, 0.21951497276451704)
```

With duals:
```
PbeExchange{Float64, Dual{ForwardDiff.Tag{Main.var"##239".ForwardDiffWrappers.DerivativeTag, Float64}, Float64, 1}}(0.804, Dual{ForwardDiff.Tag{Main.var"##239".ForwardDiffWrappers.DerivativeTag, Float64}}(0.21951497276451704,0.0))
```

mfherbst: Case in point, that's not super useful. But we can fix this with a custom show in `DftFunctionals`. In fact what we could do is actually re-introduce the `identifier` function, that employs knowledge of the parameter values to return the appropriate libxc symbol ? Or at the very least something like `PbeExchange(β=0.804, γ=0.2195)` (note rounding).

Technici4n: Note that for a custom functional, we would previously report something like `pbe_x_custom`, which I don't find super useful, and loses part of the info. My thought process was thus:
- For exact reconstruction, JLD2 will be able to reconstruct the object exactly anyway, nothing special needs to be done.
- The identifier was mostly used to give some info to users, e.g. when printing a model or basis. So for the standard DispatchFunctional you'd see the identifiers. But if the user is calling DftFunctionals directly I expect them to be an "expert" and thus be able to deal with the raw display.

> that employs knowledge of the parameter values to return the appropriate libxc symbol

Could be done as well

mfherbst: > Could be done as well

Yes and should not be too hard :)

Technici4n: We could, but is it relevant if any non-standard parameter choice will just be printed as `pbe_x_custom` or similar?

src/workarounds/forwarddiff_rules.jl
```diff
@@ -203,6 +203,28 @@ function construct_value(psp::PspUpf{T,I}) where {T <: AbstractFloat, I <: Abstr
...
+        if f isa DispatchFunctional
+            f
+        else
+            construct_value(f)
+        end
+    end
+    Xc(functionals, ForwardDiff.value(x.scaling_factor), x.potential_threshold, x.use_nlcc)
+end
+# get the value of each field of the functional
+@generated function construct_value(x::F) where {F <: DftFunctionals.Functional}
```
Technici4n: @niklasschmitz pointed out that we could use https://fluxml.ai/Functors.jl/stable/

src/terms/xc.jl
mfherbst: If we have only a print function that for those we could print the actual parameter values. So you have something more specific where possible (just e.g. `gga_x_pbe`) and something less specific for the rest `PbeExchange(β=..., γ=....)`. I think that would be very useful.

src/workarounds/forwarddiff_rules.jl
mfherbst: If it's not too heavy as a dependency and sufficiently stable as a package, I'm happy wtih it. But in my experience there is few worse things that commiting on too many, too brittle helper packages, that we need to undo or fix later.

PR #1189, opened 2025-11-11
Ensure contiguous occupations

abussy:
This PR adds checks to ensure that occupations are contiguous. In various places in the code, it is simply assumed that occupied and empty orbitals are clearly separated (e.g. [here](https://github.com/JuliaMolSim/DFTK.jl/blob/e2672b6d4f89cb6be307fec7d5048cf9663762d0/src/scf/nbands_algorithm.jl#L88-L91) or [here](https://github.com/JuliaMolSim/DFTK.jl/blob/e2672b6d4f89cb6be307fec7d5048cf9663762d0/src/orbitals.jl#L8-L10)). This is currently not a safe assumption, as it is not enforced.

This PR adds a check on eigenvalues monotonocity when computing the occupations. Additionally, a helper function providing masks in the form of ranges (useful for taking views into GPU arrays) was added. Finally, the creation of supercells now enforces consistent reordering of the eigenvalues and wave-functions.

Note: range masks are not used in `densities.jl`, as the `compute_density()` function is sometime hijacked in other contexts, where the assumption on occupations does not hold (e.g. [here](https://github.com/JuliaMolSim/DFTK.jl/blob/e2672b6d4f89cb6be307fec7d5048cf9663762d0/src/postprocess/dos.jl#L61-L64))

This PR is a necessary fix for #1187 tests to pass.

### DIFF
--- a/src/occupation.jl
+++ b/src/occupation.jl
@@ -216,7 +216,7 @@ end
 """
 Return ranges of occupied elements based on a given occupation threshold
 """
-function range_masks(occupation, occupation_threshold)
+function occupied_empty_masks(occupation, occupation_threshold)
     n_occ = map(occupation) do occ
         n = count(occ_i -> abs(occ_i) > occupation_threshold, occ)
         # Check that all occupied elements are contiguous, otherwise range is wrong
--- a/src/orbitals.jl
+++ b/src/orbitals.jl
@@ -5,7 +5,7 @@ using Random  # Used to have a generic API for CPU and GPU computations alike: s
 # threshold is a parameter to distinguish between states we want to keep and the
 # others when using temperature. It is set to 0.0 by default, to treat with insulators.
 function select_occupied_orbitals(basis, ψ, occupation; threshold=0.0)
-    mask_occ = range_masks(occupation, threshold).mask_occ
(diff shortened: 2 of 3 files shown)


src/occupation.jl
```diff
@@ -207,3 +212,18 @@ function check_full_occupation(basis::PlaneWaveBasis, occupation)
         all(occ_k .== filled_occ) || error("Only full occupation is supported, but $occ_k has partial occupation.")
     end
 end
+
+"""
+Return ranges of occupied elements based on a given occupation threshold
+"""
+function range_masks(occupation, occupation_threshold)
```
mfherbst: `range_masks` is a little generic. How about `occupied_empty_masks` ?

mfherbst:
Be careful @abussy. I think @antoine-levitt is right (and I unfortunately missed it). The important difference is
```julia
 n = count(occ_i -> abs(occ_i) > occupation_threshold, occ)
```
versus something like
```julia
N = [something(findlast(x -> abs(x) > threshold, occk), 0) for occk in occupation]
```
which has been used so far. Whereas the former fails for Methfessel-Paxton and Marzari-Vanderbilt, the latter will still work.

BTW we should have a test for this kind of edge case. At least not the first time I make this mistake.

PR #1190, opened 2025-11-12
Fix occupation masks and add tests

abussy:
Fixes the issue raised by @antoine-levitt in #1189.

A new test item was also added to `/test/occupation.jl`, checking the sanity of occupation masks over various smearing methods and temperatures.

An additional test specifically designed to fail with the current state of DFTK was added as well.


test/occupation.jl
```diff
@@ -248,3 +248,73 @@ end
...
+
+    # Special test specifically triggering case of occupied oribitals with occupations
+    # below threshold (oscillations of MethfesselPaxton smearing). If occupations are
+    # not carefully calculated, i.e. the last element with abs(occ) > threshold marks the
+    # boundary, some occupied orbitals can be classified as empty.
+    model = model_DFT(silicon.lattice, silicon.atoms, silicon.positions;
+                      functionals=LDA(), temperature=0.8,
+                      smearing=DFTK.Smearing.MethfesselPaxton(3))
+    basis = PlaneWaveBasis(model; Ecut=15, kgrid=[1, 1, 1])
+    scfres = self_consistent_field(basis; tol=1e-6)
```
mfherbst:
```suggestion
    basis = PlaneWaveBasis(model; Ecut=10, kgrid=[2, 2, 2])
    scfres = self_consistent_field(basis; tol=1e-2)
```
converges faster (at least for me) and also has the problem.

mfherbst (commented):
LGTM, thanks for the fix.

@antoine-levitt Do you agree ?

PR #1191, opened 2025-11-17
Don't require a psp to construct an OrbitalManifold

Technici4n:
This makes the construction of a Hubbard term a bit nicer, especially in cases where the psp is not so easily accessed (e.g. when interfacing via AtomsBase or potentially AiiDA).

### DIFF
--- a/src/terms/hubbard.jl
+++ b/src/terms/hubbard.jl
@@ -79,6 +79,8 @@ function resolve_hubbard_manifold(manifold::OrbitalManifold, model::Model)
     if manifold.projectors isa AbstractString
         l, i = find_pswfc(psp, manifold.projectors)
     else
+        # guaranteed by union
+        @assert manifold.projectors isa @NamedTuple{l::Int, i::Int}
         (; l, i) = manifold.projectors
     end
 


src/terms/hubbard.jl
```diff
@@ -8,41 +8,58 @@ Structure for Hubbard manifold choice and projectors extraction.
...
+The manifold contains two pieces of information:
+- The atomic sites at which the correction is applied.
+  These can either be specified as a chemical species for convenience,
+  or as a list of atom indices for maximum flexibility.
+- The pseudoatomic orbitals to use for the projectors.
+  These can either be specified by their label (e.g. `"3D"` for PseudoDojo),
+  or by their angular momentum number `l` + the corresponding index `i`.
 """
 struct OrbitalManifold
+    atoms::Union{ChemicalSpecies, Vector{Int64}}
```
mfherbst: If you use strange type unions like this either use type variables or don't type at all. Since this is not performance critical (unlike instantiated manifold), both should be fine, so maybe have no types ?

src/terms/hubbard.jl
```diff
@@ -8,41 +8,58 @@ Structure for Hubbard manifold choice and projectors extraction.
...
+    projectors::Union{AbstractString, @NamedTuple{l::Int64, i::Int64}}
+end
+function OrbitalManifold(atom::ElementPsp, projectors)
+    OrbitalManifold(atom.species, projectors)
+end
+function OrbitalManifold(atom::Symbol, projectors)
```
mfherbst, line 26: No need to do that chemical species can be compared to symbols

src/terms/hubbard.jl
```diff
@@ -8,41 +8,58 @@ Structure for Hubbard manifold choice and projectors extraction.
...
-end
-function OrbitalManifold(psp::NormConservingPsp, iatoms::Vector{Int64}, label::AbstractString)
-    (; l, i) = find_pswfc(psp, label)
-    OrbitalManifold(psp, iatoms, l, i)
-end
 
-function check_hubbard_manifold(manifold::OrbitalManifold, model::Model)
-    for atom in model.atoms[manifold.iatoms]
+function resolve_hubbard_manifold(manifold::OrbitalManifold, model::Model)
+    if manifold.atoms isa ChemicalSpecies
```
mfherbst, line 45: Do the reverse and check for abstractvector of integers (and be general in the checked type)

examples/hubbard.jl
```diff
@@ -44,13 +44,13 @@ plot_pdos(bands; p, iatom=1, label="3D", colors=[:yellow, :orange], εrange)
 # Note that "manifold" is the standard term used in the literature for the set of atomic orbitals
 # used to compute the Hubbard correction, but it is not meant in the mathematical sense.
 U = 10u"eV"
-manifold = OrbitalManifold(atoms, Ni, "3D")
+manifold = OrbitalManifold(Ni, "3D")
```
mfherbst, line 58: Add a comment that this also works with a list of integers and that this version can be used when constructing a model from an AtomsBase system.

mfherbst (commented):
Thanjs

examples/hubbard.jl
mfherbst: And mention the symbol option, too ... or even use it in the example.

(Many people will just look at this and never read the docs)

src/terms/hubbard.jl
mfherbst: Actually having to types is better, will also easily give more flexibility what to put inside these fields.

Technici4n: Well, the type annotation mostly serves as a check that users are passing something reasonable. Otherwise passing say a string like `"Ni"` will result in `findall` returning an empty list, which will give a somewhat unclear error like "Orbital manifold has no atoms."

mfherbst: Well but on the other hand the place where you actually assume the types is in `resolve_hubbard_manifold` and not in the struct. So better put the type checks there fully. You anyway already check some types. Just make the `else` an `elseif` and put an `else` with an `error( ... )` message.

My argument is like this the struct looks strangely complicated and prevents future extensions in an unnecessary way. E.g. I have to change two different pieces of code if I ever want to support other things (e.g. also an AtomsBase atoms instead of just a `Symbol` or `ChemicalSpecies`.

src/terms/hubbard.jl
```diff
@@ -57,10 +76,16 @@ function check_hubbard_manifold(manifold::OrbitalManifold, model::Model)
         end
     end
 
-    nothing
+    if manifold.projectors isa AbstractString
+        l, i = find_pswfc(psp, manifold.projectors)
+    else
+        (; l, i) = manifold.projectors
```
mfherbst, line 84: Have an assert here, too, please.

Technici4n: Ah yes oops

PR #1192, opened 2025-11-23
Multi U

fsicignano:
Hubbard update which allows for multiple manifold management and calculation through vectorization of the struct Hubbard fields.


examples/hubbard.jl
```diff
@@ -58,7 +58,7 @@ U = 10u"eV"
 manifold = OrbitalManifold(Ni, "3D")
 
 # Run SCF with a DFT+U setup, notice the `extra_terms` keyword argument, setting up the Hubbard +U term.
```
Technici4n, line 60: It would be good to explain in the comment here (which turns into https://docs.dftk.org/dev/examples/hubbard/, remember) how to add multiple manifolds. I think we don't need to use multiple for this example, but we should make the reader aware of it.

src/terms/hubbard.jl
```diff
@@ -107,51 +107,68 @@ Hubbard energy, following the Dudarev et al. (1998) rotationally invariant forma
...
+    function Hubbard(manifolds::Vector{OrbitalManifold}, U)
+        U = austrip.(U)
+        new{typeof(U[1])}(manifolds, U)
+    end
+    function Hubbard(params::Dict)
+        Hubbard(collect(keys(params)), collect(values(params)))
+    end
+    function Hubbard(params::Pair)
+        Hubbard([params[1]], [params[2]])
     end
```
Technici4n: I wouldn't use a `Dict`. To accept multiple pairs you should be able to use `params::Vararg{<:Pair}`.

For example, this works:
```jl
julia> function accept_pairs(manifold_to_U::Vararg{<:Pair,})
           map(p -> p[1], manifold_to_U), map(p -> p[2], manifold_to_U)
       end
accept_pairs (generic function with 1 method)

julia> accept_pairs("a" => 1, "b" => 2, "c" => 3)
(("a", "b", "c"), (1, 2, 3))
```

src/terms/hubbard.jl
```diff
@@ -107,51 +107,68 @@ Hubbard energy, following the Dudarev et al. (1998) rotationally invariant forma
...
+    manifolds = [resolve_hubbard_manifold(manifold, basis.model) for manifold in hubbard.manifolds]
     projs, labels = atomic_orbital_projectors(basis)
-    labels, projectors_matrix = extract_manifold(manifold, projs, labels)
-    TermHubbard(manifold, hubbard.U, projectors_matrix, labels)
+    manifold_labels, projs_matrices = [], []
+    for manifold in manifolds
+        push!(manifold_labels, extract_manifold(manifold, projs, labels).manifold_labels)
+        push!(projs_matrices, extract_manifold(manifold, projs, labels).manifold_projectors)
+    end
+    TermHubbard(manifolds, hubbard.U, projs_matrices, manifold_labels)
```
Technici4n: Don't call `extract_manifold` twice. It might be simpler to use a list of pairs instead of two lists for Hubbard:
- Careful with `[]` because it creates a `Vector{Any}`, which will lead to dynamic dispatch and thus bad performance.
- Might make it easier to `map` over things.

src/terms/hubbard.jl
```diff
@@ -107,51 +107,68 @@ Hubbard energy, following the Dudarev et al. (1998) rotationally invariant forma
...
-                                            basis::PlaneWaveBasis{T},
+@timing "ene_ops: hubbard" function ene_ops(term::TermHubbard, 
+                                            basis::PlaneWaveBasis{T}, 
                                             ψ, occupation; hubbard_n=nothing,
                                             kwargs...) where {T}
     if isnothing(hubbard_n)
        return (; E=zero(T), ops=[NoopOperator(basis, kpt) for kpt in basis.kpoints])
     end
-    proj = term.P
+    # Concatenate projectors from all manifolds per k-point as well as U value for each projector
```
Technici4n: Can this be done when the term is instantiated?

src/terms/hubbard.jl
```diff
@@ -195,10 +214,11 @@ function compute_hubbard_n(manifold::ResolvedOrbitalManifold, projectors, labels
             hubbard_n[σ, idx, jdx] = mpi_sum(hubbard_n[σ, idx, jdx], basis.comm_kpts)
         end
     end
-    hubbard_n = symmetrize_hubbard_n(basis.model, manifold, hubbard_n; basis.symmetries)
+    symmetrize_hubbard_n(basis.model, manifold, hubbard_n; basis.symmetries)
 end
-function compute_hubbard_n(term::TermHubbard, basis::PlaneWaveBasis, ψ, occupation)
-    compute_hubbard_n(term.manifold, term.P, term.labels, basis, ψ, occupation)
+function compute_hubbard_ns(term::TermHubbard, basis::PlaneWaveBasis, ψ, occupation)
```
Technici4n: We could keep the name for consistency with usage of `hubbard_n` elsewhere in the code.

src/terms/operators.jl
```diff
@@ -153,7 +153,47 @@ function apply!(Hψ, op::DivAgradOperator, ψ;
         Hψ.fourier .-= im .* G_plus_k[α] .* A∇ψ ./ 2
     end
 end
-# TODO Implement  Base.Matrix(op::DivAgradOperator)
+# TODO Implement  Matrix(op::DivAgrad)
```
Technici4n: This all looks like a git merge accident. 😉

test/hamiltonian_consistency.jl
```diff
@@ -108,8 +108,8 @@ end
     test_consistency_term(Hartree())
     let
         Si = ElementPsp(14, load_psp(testcase.psp_upf))
-        test_consistency_term(Hubbard(OrbitalManifold([1, 2], "3P"), 0.01), atom=Si)
-        test_consistency_term(Hubbard(OrbitalManifold([1, 2], "3P"), 0.01), atom=Si,
+        test_consistency_term(Hubbard(OrbitalManifold([1, 2], "3P") => 0.01), atom=Si)
+        test_consistency_term(Hubbard(OrbitalManifold([1, 2], "3P") => 0.01), atom=Si,
```
Technici4n, line 112: Can you add a test for two manifolds as well?

src/terms/hubbard.jl
```diff
@@ -107,51 +107,68 @@ Hubbard energy, following the Dudarev et al. (1998) rotationally invariant forma
...
+                                            basis::PlaneWaveBasis{T}, 
                                             ψ, occupation; hubbard_n=nothing,
                                             kwargs...) where {T}
     if isnothing(hubbard_n)
        return (; E=zero(T), ops=[NoopOperator(basis, kpt) for kpt in basis.kpoints])
     end
-    proj = term.P
+    # Concatenate projectors from all manifolds per k-point as well as U value for each projector
+    projs = [hcat((P_manifold[ik] for P_manifold in term.P)...) for ik in 1:length(term.P[1])]
+    U = [term.U[iman] for iman in 1:length(term.manifolds) for _ in 1:size(term.P[iman][1],2)]
```
Technici4n: This would probably be cleaner if done in the loop below, a bit hard to read as is.

src/terms/hubbard.jl
```diff
@@ -107,51 +107,68 @@ Hubbard energy, following the Dudarev et al. (1998) rotationally invariant forma
...
-    natoms = length(term.manifold.iatoms)
     n_spin = basis.model.n_spin_components
-    nproj_atom = size(hubbard_n[1,1,1], 1) # This is the number of projectors per atom, namely 2l+1
-    # For the ops we have to reshape hubbard_n to match the NonlocalOperator structure, using a block diagonal form
-    nhub = [zeros(Complex{T}, nproj_atom*natoms, nproj_atom*natoms) for _ in 1:n_spin]
+    # For the ops we have to reshape hubbard_n to match the NonlocalOperator structure, 
+    # using a block diagonal form whit one block per atom per manifold
+    # 2l+1 is the number of projectors per atom, i.e. the dimension of the blocks
+    nhub_size = sum((2*manifold.l+1)*length(manifold.iatoms) for manifold in term.manifolds)
+    nhub = [zeros(Complex{T}, nhub_size, nhub_size) for _ in 1:n_spin]
```
Technici4n: Can we maybe use `D_size` and `D` for the variables? To make it clear that this is all about building the diagonal matrix for the nonlocal operator.

src/terms/hubbard.jl
```diff
@@ -107,51 +107,68 @@ Hubbard energy, following the Dudarev et al. (1998) rotationally invariant forma
...
+        for σ in 1:n_spin, iatom in 1:length(manifold.iatoms)
+            proj_range = (offset + (iatom-1)*nproj_atom + 1):(offset + iatom*nproj_atom)
+            nhub[σ][proj_range, proj_range] = hubbard_n[iman][σ, iatom, iatom]
+            E += filled_occ * 1/T(2) * term.U[iman] *
+                 real(tr(hubbard_n[iman][σ, iatom, iatom] * (I - hubbard_n[iman][σ, iatom, iatom])))
+        end
+        offset += nproj_atom*length(manifold.iatoms)
     end
-    ops = [NonlocalOperator(basis, kpt, proj[ik], 1/T(2) * term.U * (I - 2*nhub[kpt.spin])) 
+    ops = [NonlocalOperator(basis, kpt, projs[ik], 1/T(2) * U .* (I - 2*nhub[kpt.spin]))
```
Technici4n: I would give it `D[kpt.spin]` and move the `1/2 * U * (I - 2n)` inside the for loop above.

src/terms/hubbard.jl
fsicignano: However this gives back a tuple instead of a Vector, so I had to "convert" to vectors manually. Not sure if this optimal.

src/terms/hubbard.jl
fsicignano: This operation mixes different manifolds and destroys the vectorization. We could do add it to the term as another field, but most of the times the old one will be used.

src/terms/hubbard.jl
fsicignano: Redefining the D as you said makes the U vector unnecessary, same for the projs if we store them in this form as well when the term is instantiated - so I will delete this from here entirely.

examples/hubbard.jl
fsicignano: Could the following work? 
"
It is also possible to set up multiple manifolds with different U values by passing each pair
 as a separate entry in the `Hubbard` constructor (i.e. `Hubbard(manifold1 => U1, manifold2 => U2, etc.)`)
 or as two vectors (i.e. `Hubbard([manifold1, manifold2, etc.], [U1, U2, etc.])`).
"

src/terms/hubbard.jl
fsicignano: I created a tuple with both elements for each manifold and then split it in two vectors, but maybe there's a more elegant way to go

examples/hubbard.jl
mfherbst: I like this, but make sure that you have a test for both versions, i.e. something that tests that `Hubbard(manifold1 => U1, manifold2 => U2, etc.)` and that `Hubbard([manifold1, manifold2, etc.], [U1, U2, etc.])` works as expected.

src/terms/hubbard.jl
```diff
@@ -107,51 +107,75 @@ Hubbard energy, following the Dudarev et al. (1998) rotationally invariant forma
...
+    U::Vector{T}
+    function Hubbard(manifolds::Vector{OrbitalManifold}, U)
+        U = austrip.(U)
+        new{typeof(U[1])}(manifolds, U)
+    end
+    function Hubbard(manifold_to_U::Vararg{<:Pair})
+        manifolds = map(p -> p[1], manifold_to_U)
+        Us = map(p -> p[2], manifold_to_U)
+        Hubbard([manifold for manifold in manifolds], [U for U in Us])
+    end
```
mfherbst: Make this a stand-alone function, also
```suggestion
Hubbard(manifold_to_U::Pair...) = Hubbard(first.(manifold_to_U), last.(manifold_to_U))
```
should do the same thing, but is a bit nicer.

src/terms/hubbard.jl
```diff
@@ -107,51 +107,75 @@ Hubbard energy, following the Dudarev et al. (1998) rotationally invariant forma
...
+    function Hubbard(params::Pair)
+        Hubbard([params[1]], [params[2]])
     end
```
mfherbst:
```suggestion
```
Should be obsolete with the vararg version above.

src/terms/hubbard.jl
```diff
@@ -107,51 +107,75 @@ Hubbard energy, following the Dudarev et al. (1998) rotationally invariant forma
...
-    TermHubbard(manifold, hubbard.U, projectors_matrix, labels)
+    manifold_data = map(manifold -> extract_manifold(manifold, projs, labels), manifolds)
+    manifold_labels = [data.manifold_labels for data in manifold_data]
+    projs_matrices = [data.manifold_projectors for data in manifold_data]
+    # Concatenate projectors from all manifolds per each k-point
+    projs_per_kpt = Vector{Matrix{Complex{T}}}(undef, length(basis.kpoints))
+    for iman in 1:length(manifolds), ik in 1:length(basis.kpoints)
+        projs_per_kpt[ik] = (iman == 1) ? projs_matrices[iman][ik] : 
+                                          hcat(projs_per_kpt[ik], projs_matrices[iman][ik])
+    end
```
mfherbst: This is weird. I think a
```julia
map(1:length(basis.kpoints)) do ik
    reduce(hcat, map(iman -> projs_matrices[iman][ik], 1:length(manifolds)))
end
```
or
```julia
map(1:length(basis.kpoints)) do ik
    reduce(hcat, getindex.(projs_matrices, Ref(ik)))
end
```
does the same thing, but is a little nicer to read ? Maybe still needs a comment, though.

src/terms/hubbard.jl
```diff
@@ -107,51 +107,75 @@ Hubbard energy, following the Dudarev et al. (1998) rotationally invariant forma
...
-    natoms = length(term.manifold.iatoms)
     n_spin = basis.model.n_spin_components
-    nproj_atom = size(hubbard_n[1,1,1], 1) # This is the number of projectors per atom, namely 2l+1
-    # For the ops we have to reshape hubbard_n to match the NonlocalOperator structure, using a block diagonal form
-    nhub = [zeros(Complex{T}, nproj_atom*natoms, nproj_atom*natoms) for _ in 1:n_spin]
+    projs = term.P_vec
+    # For the ops we have to reshape hubbard_n to match the NonlocalOperator structure, 
+    # using a block diagonal form whit one block per atom per manifold
+    # 2l+1 is the number of projectors per atom, i.e. the dimension of the blocks
+    D_size = sum((2*manifold.l+1)*length(manifold.iatoms) for manifold in term.manifolds)
```
mfherbst: Some better spacing (and maybe the `do ... end` form) would help readability here

mfherbst (commented):
Thanks @fsicignano for keeping on working on this. Some comments from a first pass.

src/terms/hubbard.jl
fsicignano: Do you mean something like this?
```
D_size = 0
for manifold in term.manifolds
  D_size += ...
end
```

src/terms/hubbard.jl
fsicignano: I'm not sure if it is my problem but this doesn't work on my computer, it doesn't accept the usual (manifold => U) input.

mfherbst: Interesting, this works fine for me (in a fresh shell):
```julia
struct Hubbard
    x
    y
end
Hubbard(manifold_to_U::Pair...) = Hubbard(first.(manifold_to_U), last.(manifold_to_U))

# then:
Hubbard("a" => 1)
Hubbard("a" => 1, "b" => 2)
```

src/terms/hubbard.jl
mfherbst: I mean:
```julia
    D_size = sum(term.manifolds) do manifold
        (2 * manifold.l + 1) * length(manifold.iatoms)
    end
```

src/terms/hubbard.jl
```diff
@@ -107,26 +107,44 @@ Hubbard energy, following the Dudarev et al. (1998) rotationally invariant forma
...
+        new{eltype(U)}(manifolds, U)
     end
 end
-function (hubbard::Hubbard)(basis::AbstractBasis)
-    manifold = resolve_hubbard_manifold(hubbard.manifold, basis.model)
+# Match the outer ctor generated by Julia if we had not defined our own inner ctor.
+# See https://docs.julialang.org/en/v1.12/manual/constructors/#Parametric-Constructors
+function Hubbard(manifolds::Vector{OrbitalManifold}, U::Vector{T}) where {T}
+    Hubbard{T}(manifolds, U)
+end
```
Technici4n, line 122: This is a bit subtle, I hope it's OK that I changed it like this. We have to strip the atomic units in the inner constructor otherwise it's possible to accidentally put them in the `Hubbard` (e.g. this happened when calling `Hubbard([OrbitalManifold(:Si, "3S")], [1u"eV"])`, but not anymore after my changes).

Technici4n (approved):
Looks great to me, many thanks! 🚀 I pushed a few stylistic fixes, I hope that's OK.

src/terms/hubbard.jl
fsicignano: Thank you! I had been trying to solve this same issue for some time but came out with a rather cumbersome solution...

mfherbst: Would be a solution to avoid the internal constructor and rather have a specialisation on Vector{T} with {T <: Quantity} ?

Technici4n: That would probably work as well since that would be a more specific constructor. Should I change to that? We'd lose the validation that `length(manifolds) == length(U)` but that's not a huge deal.

PR #1204, opened 2025-12-12
Fix memory usage estimate

Technici4n:
Currently `n_kpoints` is always 1 😅

### DIFF
--- a/src/memory_usage.jl
+++ b/src/memory_usage.jl
@@ -46,7 +46,7 @@ function estimate_memory_usage(basis::PlaneWaveBasis{T}, model=basis.model) wher
     isnothing(model.εF) || error("Cannot estimate memory usage for models with a fixed "
                                  * "Fermi level, since the number of bands is not known.")
 
-    n_kpoints = maximum(krange -> length(krange[1]), basis.krange_allprocs)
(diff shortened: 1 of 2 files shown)


src/memory_usage.jl
```diff
@@ -46,7 +46,7 @@ function estimate_memory_usage(basis::PlaneWaveBasis{T}, model=basis.model) wher
     isnothing(model.εF) || error("Cannot estimate memory usage for models with a fixed "
                                  * "Fermi level, since the number of bands is not known.")
 
-    n_kpoints = maximum(length, basis.krange_allprocs)
+    n_kpoints = maximum(krange -> length(krange[1]), basis.krange_allprocs)
```
mfherbst: I don't agree with this either, because this ignores the spin component. I think it should be
```julia
    n_kpoints = maximum(kranges_of_proc -> sum(length, kranges_of_proc), basis.krange_allprocs)
```

Maybe we should have a small set of unit tests, checking our expectations of this function ?

Technici4n: Indeed, thanks. I also fixed `ρ_bytes` which did not account for spin components. Both were found with a simple unit test.

PR #1205, opened 2025-12-12
Add more API crosslinks to the Docs

niklasschmitz:


docs/src/developer/gpu_computations.md
```diff
@@ -30,61 +30,6 @@ PlaneWaveBasis(model; Ecut, kgrid, architecture = DFTK.GPU(CuArray))
     Julia ecosystem concerning distributed architectures.
 
 Not all terms can be used when doing GPU computations. As of January 2023 this
-concerns `Anyonic`, `Magnetic` and `TermPairwisePotential`. Similarly GPU features are
+concerns [`Anyonic`](@ref), [`Magnetic`](@ref) and [`TermPairwisePotential`](@ref). Similarly GPU features are
 not yet exhaustively tested, and it is likely that some aspects of the code such as
 automatic differentiation or stresses will not work.
-
-## Pitfalls
```
mfherbst, line 37: I'm not in favour to completely deleting this as I don't want to loose this information for an outsider completely.

@abussy How much work would it be for you to make a quick pass on this and update it into something useful ?

src/Model.jl
```diff
@@ -65,32 +65,64 @@ end
...
+The `Model` represents the model system through its geometry (lattice, atomic positions),
+electronic structure (number of electrons, spin polarization), and physical terms
+(kinetic energy, external potentials, Hartree, exchange-correlation, etc.). 
+It does not include discretization parameters; those are set separately when constructing
+a [`PlaneWaveBasis`](@ref) from a `Model`.
 
-`spin_polarization` is :none by default (paired electrons)
-unless any of the elements has a non-zero initial magnetic moment.
-In this case the spin_polarization will be :collinear.
+Arguments
```
mfherbst: `## Arguments` ?

src/Model.jl
```diff
@@ -65,32 +65,64 @@ end
...
+- `lattice::AbstractMatrix`: Lattice vectors given column-wise (atomic units).
```
mfherbst: atomic units is not true. Any unitful length can be used. It will be converted.

src/Model.jl
```diff
@@ -65,32 +65,64 @@ end
...
+
+Notes
+- `magnetic_moments` are used only to determine `spin_polarization` and symmetries,
+    but are otherwise not stored on the `Model`.
+- The constructor validates consistency between `εF` and `n_electrons` and will error if
+    both are supplied.
+- Lattice conditioning is checked and a warning is issued for poorly conditioned lattices
+    (high matrix condition number).
+
+See also
```
mfherbst: `## See also` ?

src/Model.jl
```diff
@@ -65,32 +65,64 @@ end
 _is_well_conditioned(A; tol=1e5) = (cond(A) <= tol)
 
 """
-    Model(lattice, atoms, positions; n_electrons, magnetic_moments, terms, temperature,
-          smearing, spin_polarization, symmetries)
+    Model(lattice, atoms, positions; model_name, εF, n_electrons,
+                disable_electrostatics_check, magnetic_moments, terms,
+                temperature, smearing, spin_polarization, symmetries)
```
mfherbst, line 70: The `Model(system; kwargs...)` constructor is at least equally important to this version and should be mentioned prominently (even in the first paragraph I think). We should not give the detailed explanation of all `kwarg`s twice, but the two constructor docs should refer to each other.

src/Model.jl
```diff
@@ -65,32 +65,64 @@ end
...
-the automatic determination of lattice symmetries or (b) to pass an explicit list
-of symmetry operations to use for lowering the computational effort.
-The default behaviour is equal to `true`, namely that the code checks the
-specified model in form of the Hamiltonian `terms`, `lattice`, `atoms` and `magnetic_moments`
-parameters and from these automatically determines a set of symmetries it can safely use.
-If you want to pass custom symmetry operations (e.g. a reduced or extended set) use the
-`symmetry_operations` function. Notice that this may lead to wrong results if e.g. the
-external potential breaks some of the passed symmetries. Use `false` to turn off
-symmetries completely.
+## Keyword arguments
```
mfherbst, line 89: I would put the expert arguments `model_name`, `εF`, `disable_electrostatics_check` last and annotate these are expert.

mfherbst: Some of this information is fairly stable, I agree. But I would like to keep duplication to a minimum. Is there any way with https://github.com/JuliaDocs/DocumenterTools.jl to get the type and defaults automatically from the function definition ?

mfherbst: Regarding the annotation of types in particular: How useful is this ? Like how much do I learn as a user from `temperature::Real` as opposed to just `temperature`. My argument here is that all information that we do not give, cannot become outdated and thus confusing in the future.

src/Model.jl
```diff
@@ -65,32 +65,64 @@ end
...
+- `temperature::Real` (default: `0`): Electronic temperature.
+    If `> 0`, fractional occupations are enabled and `smearing` defaults to
+    [`Smearing.FermiDirac`](@ref)().
+- `smearing::Smearing.SmearingFunction` (default: `FermiDirac()` when `temperature>0`,
+    else `None()`):
+    Function used to convert eigenvalue differences to occupations.
+- `spin_polarization::Symbol` (default: `determine_spin_polarization(magnetic_moments)`):
+    Controls spin treatment; allowed values are `:none`, `:collinear`, `:spinless`.
+    (`:full` exists but is not supported and will error.)
+- `symmetries::Union{Bool,Vector{SymOp}}` (default: `true`):
```
mfherbst: Defaults and type is sort of true, but not really. I understand why you did that, but I do wonder if it would not be better to be less specific and refer to the function (mentioning that usually the result is `true`) and then indeed explaining the three cases as you do, but without annotating a type.

src/PlaneWaveBasis.jl
```diff
@@ -261,25 +261,76 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Real, fft_size::Tuple{Int, Int, I
...
+is then used by the Hamiltonian and SCF machinery.
+
+The constructor attempts to choose sensible defaults from the `model` (pseudopotentials,
+recommended cutoffs and symmetries) but allows explicit control of numerical
+parameters through keyword arguments described below.
+
+See [Performing a convergence study](@ref) for some practical
+recommendations how to tune these numerical discretization parameters.
+
+Arguments
```
mfherbst: Same thing, perhaps better to use `## ` sections

src/PlaneWaveBasis.jl
```diff
@@ -261,25 +261,76 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Real, fft_size::Tuple{Int, Int, I
...
+- `model::Model`: Physical model to be discretized. See [`Model`](@ref).
+
+Keyword arguments
```
mfherbst: Again I'd put expert arguments last.

src/PlaneWaveBasis.jl
```diff
@@ -261,25 +261,76 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Real, fft_size::Tuple{Int, Int, I
...
+    operations are used to reduce the explicit k-point set to the irreducible Brillouin
+    zone (saves work); set to `false` for debugging or to disable k-point reduction.
+- `comm_kpts` (default: `MPI.COMM_WORLD`): MPI communicator used to distribute k-points
+    across processes. Must be an initialized MPI communicator; the constructor calls
+    `MPI.Init()` if necessary.
+- `architecture` (default: `CPU()`): Hardware architecture descriptor (e.g. `CPU()` or
+    `GPU(...)`) that controls FFT/backend selection and device placement.
+
+Notes
+- If `Ecut` is left missing (explicit `missing`), an `ArgumentError` is thrown.
```
mfherbst: This is cryptic. Mention this in the context of the `recommended_cutoff` business when explaining the kwarg.

src/PlaneWaveBasis.jl
```diff
@@ -261,25 +261,76 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Real, fft_size::Tuple{Int, Int, I
...
+    Accepts a [`MonkhorstPack`](@ref) object, a vector of three integers,
+    a [`KgridSpacing`](@ref) object, or any object accepted by `build_kgrid`.
+    The default corresponds to a heuristic spacing.
+- `kshift` (deprecated): Legacy k-point shift. Prefer `kgrid=MonkhorstPack(kgrid, kshift)`.
+- `variational::Bool` (default: `true`): Whether the k-point-specific orbital basis is
+    variationally consistent with the density/potential grid. Non-variational mode is
+    experimental and may disable features.
+- `fft_size::Union{Nothing,Tuple{Int,Int,Int}}` (default: `nothing`): Explicit FFT grid
+    size. If `nothing`, an automatic FFT grid compatible with `Ecut` and `supersampling`
+    is computed. Specifying `fft_size` may disable certain symmetry optimizations.
```
mfherbst: Not optimisations, actually, more consistency aspects related to `symmetries_respect_rgrid`. I left it vague, because I believe that's quite an expert detail, which is moreover likely to change (symmetry is really difficult), so I'm not sure how good it is to be precise here and risk being wrong in the future when the code changes.

src/PlaneWaveBasis.jl
```diff
@@ -261,25 +261,76 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Real, fft_size::Tuple{Int, Int, I
...
+- `symmetries_respect_rgrid::Bool` (default: `isnothing(fft_size)`): When `true` the
+    automatically chosen FFT grid will be adjusted to be compatible with common fractional
+    translation denominators (2,3,4,6) to preserve symmetries; set to `false` to skip
```
mfherbst: Also not exactly correct. Again I would vouch to be more vague as people can look at the code for details.

src/PlaneWaveBasis.jl
```diff
@@ -261,25 +261,76 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Real, fft_size::Tuple{Int, Int, I
...
+    for Bloch orbitals (units of energy in atomic units). If DFTK cannot infer a
+    recommended cutoff from the pseudopotentials this argument must be supplied.
+- `supersampling::Real` (default: `recommended_cutoff(model).supersampling`): Factor
+    determining the density/potential cutoff relative to `Ecut`. The density cutoff is
+    `supersampling^2 * Ecut` (default supersampling = 2.0, i.e. density cutoff = 4 * Ecut).
+- `kgrid` (default: `KgridSpacing(2π * 0.022)`): Specifies Brillouin-zone sampling.
+    Accepts a [`MonkhorstPack`](@ref) object, a vector of three integers,
+    a [`KgridSpacing`](@ref) object, or any object accepted by `build_kgrid`.
+    The default corresponds to a heuristic spacing.
+- `kshift` (deprecated): Legacy k-point shift. Prefer `kgrid=MonkhorstPack(kgrid, kshift)`.
```
mfherbst: Don't document deprecated features.

src/PlaneWaveBasis.jl
```diff
@@ -261,25 +261,76 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Real, fft_size::Tuple{Int, Int, I
...
+- `variational::Bool` (default: `true`): Whether the k-point-specific orbital basis is
+    variationally consistent with the density/potential grid. Non-variational mode is
+    experimental and may disable features.
```
mfherbst: I'm not sure I would document this at all.

src/PlaneWaveBasis.jl
```diff
@@ -261,25 +261,76 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Real, fft_size::Tuple{Int, Int, I
...
+
+Examples
+```julia
+# Use recommended cutoffs from pseudopotentials if available
+basis = PlaneWaveBasis(model)
+
+# Provide explicit Ecut and k-grid
+basis = PlaneWaveBasis(model; Ecut=10.0, kgrid=[4,4,4])
+
+# Custom FFT grid (note: may disable symmetry-driven optimizations)
```
mfherbst: again, note this comment is not true, just drop it here.

docs/src/developer/gpu_computations.md
abussy: I agree, it should not go away. I also think I can contribute new tricks/gotcha I learned along the way. I can probably do that by Christmas.

This PR can go ahead without this updated though.

mfherbst: Ok, so you have it on your list and we should delete the section or you prefer to have the section as something to start from ?

abussy: I would keep it, it is still relevant information.

niklasschmitz: I was a bit too quick, sounds fair - I have restored the previous version now

src/Model.jl
niklasschmitz: Might be obvious but I still find it useful. I agree with the general sentiment for other functions, but as we said previously said is that for a small set of most important user-facing functions (`Model`, `PlaneWaveBasis`, `self_consistent_field`,...) it should be considered a bug if the docstring gets outdated.

niklasschmitz: Regarding automatic types+default docstrings, there might be more general ways of doing this with https://docstringextensions.juliadocs.org/stable/, but likely requires some effort to set up correctly

src/Model.jl
niklasschmitz: Ok, I removed the type annotation and link to the `default_symmetries` function

src/Model.jl
niklasschmitz: Added it prominently to the first paragraph now

src/PlaneWaveBasis.jl
niklasschmitz: Ok removed it

src/PlaneWaveBasis.jl
niklasschmitz: removed it

src/PlaneWaveBasis.jl
niklasschmitz: Ok I just removed that sentence

src/PlaneWaveBasis.jl
niklasschmitz: Removed it

src/PlaneWaveBasis.jl
niklasschmitz: Agreed, removed it here

src/PlaneWaveBasis.jl
niklasschmitz: done

src/Model.jl
niklasschmitz: The `austrip` conversion happens on an outer level before this constructor is called. Not sure how to document this better

mfherbst: From my point of view you don't document a method, you document a function. If someone calls this with a unit, the other method is called and so it still works. I think it's confusing to document only this innermost method and nothing else. Also I think it's too bloaty to document all methods, including the ones that are only around to convert units.

I don't know how this is rendered in the API list and the interactive help in the shell / Pluto, but given the above perhaps it makes sense to make this  fact that you document the function explicit by
```julia
"""
my crazy docstring
"""
function Model end
```
or something like that (check what it looks like, I don't know).

src/Model.jl
```diff
@@ -65,32 +65,69 @@ end
...
+- `spin_polarization::Symbol` (default: `determine_spin_polarization(magnetic_moments)`):
+    Controls spin treatment; allowed values are `:none`, `:collinear`, `:spinless`.
+    (`:full` exists but is not supported and will error.)
+- `symmetries` (default: `true`):
+    - `true`: run automatic symmetry detection with [`default_symmetries`](@ref).
+    - `false`: disable all symmetries.
+    - `::Vector{SymOp}`: use the provided explicit symmetry operations. Passing incorrect
+        symmetries may produce wrong results.
+
+## Keyword arguments part 2 (expert-level)
```
mfherbst:
```suggestion
## Keyword arguments (Expert level)
```

src/Model.jl
```diff
@@ -65,32 +65,69 @@ end
...
+    Hamiltonian.
+- `temperature::Real` (default: `0`): Electronic temperature.
+    If `> 0`, fractional occupations are enabled and `smearing` defaults to
+    [`Smearing.FermiDirac`](@ref)().
+- `smearing::Smearing.SmearingFunction` (default: `FermiDirac()` when `temperature>0`,
+    else `None()`):
+    Function used to convert eigenvalue differences to occupations.
+- `spin_polarization::Symbol` (default: `determine_spin_polarization(magnetic_moments)`):
+    Controls spin treatment; allowed values are `:none`, `:collinear`, `:spinless`.
+    (`:full` exists but is not supported and will error.)
```
mfherbst:
```suggestion
```
Future music ...

src/Model.jl
```diff
@@ -65,32 +65,69 @@ end
...
-The default behaviour is equal to `true`, namely that the code checks the
-specified model in form of the Hamiltonian `terms`, `lattice`, `atoms` and `magnetic_moments`
-parameters and from these automatically determines a set of symmetries it can safely use.
-If you want to pass custom symmetry operations (e.g. a reduced or extended set) use the
-`symmetry_operations` function. Notice that this may lead to wrong results if e.g. the
-external potential breaks some of the passed symmetries. Use `false` to turn off
-symmetries completely.
+## Keyword arguments
+- `n_electrons::Union{Int,Nothing}` (default: inferred from `atoms`): Total electron count.
+    If `εF !== nothing` then `n_electrons` must be `nothing`.
```
mfherbst:
```suggestion
```
Expert usage and mutually exclusive said below (documentation for `εF`)

src/Model.jl
```diff
@@ -65,32 +65,69 @@ end
...
-of symmetry operations to use for lowering the computational effort.
-The default behaviour is equal to `true`, namely that the code checks the
-specified model in form of the Hamiltonian `terms`, `lattice`, `atoms` and `magnetic_moments`
-parameters and from these automatically determines a set of symmetries it can safely use.
-If you want to pass custom symmetry operations (e.g. a reduced or extended set) use the
-`symmetry_operations` function. Notice that this may lead to wrong results if e.g. the
-external potential breaks some of the passed symmetries. Use `false` to turn off
-symmetries completely.
+## Keyword arguments
+- `n_electrons::Union{Int,Nothing}` (default: inferred from `atoms`): Total electron count.
```
mfherbst: I would just do:
```suggestion
- `n_electrons::Int` (default: inferred from `atoms`): Total electron count.
```

src/Model.jl
mfherbst: We already use DocStringExtensions (with some setup I just stole from somewhere else) and I think the time invested to set this up properly greatly outweighs the time you loose by getting mislead from wrong docstrings.

src/Model.jl
```diff
@@ -65,32 +65,69 @@ end
...
+    `spin_polarization` and to determine symmetries. These values influence model setup
+    but are otherwise not stored on the `Model`.
+- `terms::Vector` (default: `[Kinetic()]`): List of energy term types to include in the
+    Hamiltonian.
+- `temperature::Real` (default: `0`): Electronic temperature.
+    If `> 0`, fractional occupations are enabled and `smearing` defaults to
+    [`Smearing.FermiDirac`](@ref)().
+- `smearing::Smearing.SmearingFunction` (default: `FermiDirac()` when `temperature>0`,
+    else `None()`):
+    Function used to convert eigenvalue differences to occupations.
```
mfherbst: Not eigenvalue differences:
```suggestion
    Smearing function used to compute occupations from Kohn-Sham eigenvalues.
```

src/Model.jl
```diff
@@ -65,32 +65,69 @@ end
...
+        symmetries may produce wrong results.
+
+## Keyword arguments part 2 (expert-level)
+- `model_name::String` (default: "custom"): Human-readable model name (e.g. "LDA", "PBE").
+- `εF` (default: `nothing`): Fixed Fermi level option. If set, the model is defined at fixed
+    Fermi level instead of fixed electron count. Mutually exclusive with `n_electrons`.
+- `disable_electrostatics_check::Bool` (default: `all(iszero, charge_ionic.(atoms))`): When
+    `false` (the default safety path), constructing a non-neutral model will raise an error
+    unless explicitly allowed. Set to `true` to bypass neutrality/electrostatics
+    checks (experimental).
```
mfherbst:
```suggestion
```
Don't document this, this is just a hack and will silently disappear one day, hopefully.

src/PlaneWaveBasis.jl
```diff
@@ -261,25 +261,67 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Real, fft_size::Tuple{Int, Int, I
...
-cutoff for the Bloch waves and a `supersampling` of 2.0 for the density. This is equal
-to a density cutoff of ``\text{supersampling}^2 * \text{Ecut}`` or `4 * Ecut`.
-A fully custom FFT grid size can also be chosen by specifying the `fft_size` parameter.
-Note, this disables certain symmetry features.
+Creates a plane-wave discretization of a given [`Model`](@ref).
+
+`PlaneWaveBasis` defines the reciprocal-space basis used for Bloch orbitals and the
+real-space FFT grid used for densities and potentials. It bundles the kinetic-energy
+cutoff `Ecut`, k-point sampling into a single object which
+is then used by the Hamiltonian and SCF machinery.
```
mfherbst:
```suggestion
real-space FFT grid used for densities and potentials. It bundles the model, the
kinetic-energy cutoff `Ecut`, k-point sampling into a single object which
is then used by the Hamiltonian and SCF machinery.
```

src/PlaneWaveBasis.jl
```diff
@@ -261,25 +261,67 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Real, fft_size::Tuple{Int, Int, I
...
+
+The constructor attempts to choose sensible defaults from the `model` (pseudopotentials,
+recommended cutoffs and symmetries) but allows explicit control of numerical
+parameters through keyword arguments described below.
+
+See [Performing a convergence study](@ref) for some practical
+recommendations how to tune these numerical discretization parameters.
```
mfherbst:
```suggestion
See [Performing a convergence study](@ref) for some practical
recommendations for choosing plane-wave discretization parameters.
If parameters are not given explicitly as a kwarg, the constructor chooses sensible defaults from inspecting the `model`.
```

src/PlaneWaveBasis.jl
```diff
@@ -261,25 +261,67 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Real, fft_size::Tuple{Int, Int, I
...
+- `Ecut::Number` (default: `recommended_cutoff(model).Ecut`): Kinetic-energy cutoff
+    for Bloch orbitals (units of energy in atomic units). If DFTK cannot infer a
+    recommended cutoff from the pseudopotentials this argument must be supplied.
+- `supersampling::Real` (default: `recommended_cutoff(model).supersampling`): Factor
+    determining the density/potential cutoff relative to `Ecut`. The density cutoff is
+    `supersampling^2 * Ecut` (default supersampling = 2.0, i.e. density cutoff = 4 * Ecut).
+- `kgrid` (default: `KgridSpacing(2π * 0.022)`): Specifies Brillouin-zone sampling.
+    Accepts a [`MonkhorstPack`](@ref) object, a vector of three integers,
+    a [`KgridSpacing`](@ref) object, or any object accepted by `build_kgrid`.
+    The default corresponds to a heuristic spacing.
```
mfherbst:
```suggestion
```
Don't think this helps.

src/PlaneWaveBasis.jl
```diff
@@ -261,25 +261,67 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Real, fft_size::Tuple{Int, Int, I
...
+    zone (saves work); set to `false` for debugging or to disable k-point reduction.
+- `comm_kpts` (default: `MPI.COMM_WORLD`): MPI communicator used to distribute k-points
+    across processes. Must be an initialized MPI communicator; the constructor calls
+    `MPI.Init()` if necessary.
+- `architecture` (default: `CPU()`): Hardware architecture descriptor (e.g. `CPU()` or
+    `GPU(...)`) that controls FFT/backend selection and device placement.
+
+## Notes
+- When `fft_size` is provided the constructor validates it for compatibility with the
+    internal FFT routines (use `next_working_fft_size` if needed).
```
mfherbst: The error message says this, no need to put it here.
```suggestion
```

src/PlaneWaveBasis.jl
```diff
@@ -261,25 +261,67 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Real, fft_size::Tuple{Int, Int, I
...
+    recommended cutoff from the pseudopotentials this argument must be supplied.
+- `supersampling::Real` (default: `recommended_cutoff(model).supersampling`): Factor
+    determining the density/potential cutoff relative to `Ecut`. The density cutoff is
+    `supersampling^2 * Ecut` (default supersampling = 2.0, i.e. density cutoff = 4 * Ecut).
+- `kgrid` (default: `KgridSpacing(2π * 0.022)`): Specifies Brillouin-zone sampling.
+    Accepts a [`MonkhorstPack`](@ref) object, a vector of three integers,
+    a [`KgridSpacing`](@ref) object, or any object accepted by `build_kgrid`.
+    The default corresponds to a heuristic spacing.
+- `fft_size::Union{Nothing,Tuple{Int,Int,Int}}` (default: `nothing`): Explicit FFT grid
+    size. If `nothing`, an automatic FFT grid compatible with `Ecut` and `supersampling`
```
mfherbst:
```suggestion
    size. If `nothing` (recommended), an automatic FFT grid compatible with `Ecut` and `supersampling`
```

src/PlaneWaveBasis.jl
```diff
@@ -261,25 +261,67 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Real, fft_size::Tuple{Int, Int, I
...
+    is computed.
+- `use_symmetries_for_kpoint_reduction::Bool` (default: `true`): If `true`, symmetry
+    operations are used to reduce the explicit k-point set to the irreducible Brillouin
+    zone (saves work); set to `false` for debugging or to disable k-point reduction.
```
mfherbst:
```suggestion
    zone (saves work, recommended).
```

src/PlaneWaveBasis.jl
```diff
@@ -261,25 +261,67 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Real, fft_size::Tuple{Int, Int, I
...
+- `comm_kpts` (default: `MPI.COMM_WORLD`): MPI communicator used to distribute k-points
+    across processes. Must be an initialized MPI communicator; the constructor calls
+    `MPI.Init()` if necessary.
```
mfherbst:
```suggestion
    across processes.
```
not sure this is useful.

mfherbst (commented):
Thanks

src/Model.jl
niklasschmitz: Ok I added a sentence on Unitful being allowed too. Querying `julia>?Model` shows all docstrings concatenated, which seems fair

src/PlaneWaveBasis.jl
niklasschmitz: done

src/Model.jl
niklasschmitz: Ok I put it on the hackathon tasks list for next time

src/PlaneWaveBasis.jl
```diff
@@ -261,25 +261,64 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Real, fft_size::Tuple{Int, Int, I
...
+- `Ecut::Number` (default: `recommended_cutoff(model).Ecut`): Kinetic-energy cutoff
+    for Bloch orbitals (units of energy in atomic units). If DFTK cannot infer a
+    recommended cutoff from the pseudopotentials this argument must be supplied.
+- `kgrid` (default: `KgridSpacing(2π * 0.022)`): Specifies Brillouin-zone sampling.
+    Accepts a [`MonkhorstPack`](@ref) object, a vector of three integers,
+    a [`KgridSpacing`](@ref) object, or any object accepted by `build_kgrid`.
+- `architecture` (default: `CPU()`): Hardware architecture descriptor (e.g. `CPU()` or
+    `GPU(...)`) that controls backend selection and device placement.
+
+## Keyword arguments part 2 (expert-level)
```
mfherbst:
```suggestion
## Keyword arguments (Expert level)
```

PR #1215, opened 2025-12-19
Optimize symmetries_preserving_kgrid for MonkhorstPack

Technici4n:
The previous logic was quadratic in the number of reducible k-points, meaning that we could easily waste **5 minutes** with large k-grids like `[34, 34, 34]`. The new logic is linear in the number of reducible k-points, using the same idea as `symmetries_preserving_rgrid`. 😉

(PS: this should also correctly handle k-grids of size 1 in some directions since the `_check` will trivially pass for them)

### DIFF
--- a/src/symmetry.jl
+++ b/src/symmetry.jl
@@ -159,10 +159,10 @@ _is_approx_in(x::AbstractArray{T}, X) where {T} = any(y -> isapprox(x, y; atol=s
 """
 Filter out the symmetry operations that don't respect the symmetries of the discrete BZ grid
 """
-function symmetries_preserving_kgrid(symmetries, kgrid::ExplicitKpoints)
+function symmetries_preserving_kgrid(symmetries, kgrid::AbstractKgrid)
     # First apply symmetries as the provided k-points can be arbitrary
     # (e.g. only along a line or similar)
-    all_kcoords = unfold_kcoords(kgrid.kcoords, symmetries)
+    all_kcoords = unfold_kcoords(reducible_kcoords(kgrid).kcoords, symmetries)
     kcoords_normalized = normalize_kpoint_coordinate.(all_kcoords)
     function preserves_grid(symop)
         all(_is_approx_in(normalize_kpoint_coordinate(symop.S * k), kcoords_normalized)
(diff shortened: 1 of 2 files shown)


mfherbst:
If you bump the DFTK version, too, then we release a new version after this goes in. Seems reasonable to wait for this PR, though.

src/symmetry.jl
```diff
@@ -159,34 +159,31 @@ _is_approx_in(x::AbstractArray{T}, X) where {T} = any(y -> isapprox(x, y; atol=s
 """
 Filter out the symmetry operations that don't respect the symmetries of the discrete BZ grid
 """
-function symmetries_preserving_kgrid(symmetries, kcoords)
-    kcoords_normalized = normalize_kpoint_coordinate.(kcoords)
+function symmetries_preserving_kgrid(symmetries, kgrid::ExplicitKpoints)
+    # First apply symmetries as the provided k-points can be arbitrary
+    # (e.g. only along a line or similar)
+    all_kcoords = unfold_kcoords(kgrid.kcoords, symmetries)
```
mfherbst:
```suggestion
    all_kcoords = unfold_kcoords(reducible_kcoords(kgrid).kcoords, symmetries)
```
Then you can make `kgrid` even an `AbstractKgrid`

src/symmetry.jl
```diff
@@ -159,34 +159,31 @@ _is_approx_in(x::AbstractArray{T}, X) where {T} = any(y -> isapprox(x, y; atol=s
...
-        # TODO This can certainly be improved by knowing this is an MP grid,
-        #      see symmetries_preserving_rgrid below for ideas
-        symmetries_preserving_kgrid(symmetries, reducible_kcoords(kgrid).kcoords)
+    # we have to check if S * k is in the grid for all k
+    # by linearity, checking the origin and each unit vector is enough
+    kcoords_normalized = normalize_kpoint_coordinate.(reducible_kcoords(kgrid).kcoords)
+
+    function preserves_kgrid(symop)
+        function _check(dx, dy, dz)
+            k = (kgrid.kshift .+ Vec3(dx, dy, dz)) .// kgrid.kgrid_size
```
mfherbst, line 183: `k` should be a `Vector{Rational}` so I don't think you need the `is_approx_in` at all. Here the arithmetic should be exact, no ?

src/symmetry.jl
```diff
@@ -159,34 +159,31 @@ _is_approx_in(x::AbstractArray{T}, X) where {T} = any(y -> isapprox(x, y; atol=s
...
-        # if k' = Rk
-        # then
-        #    R' = diag(kgrid) R diag(kgrid)^-1
-        # should be integer where
-
-        # TODO This can certainly be improved by knowing this is an MP grid,
-        #      see symmetries_preserving_rgrid below for ideas
-        symmetries_preserving_kgrid(symmetries, reducible_kcoords(kgrid).kcoords)
+    # we have to check if S * k is in the grid for all k
+    # by linearity, checking the origin and each unit vector is enough
```
mfherbst: It's not really a unit vector. I would make it explicit and say that by linearity checking that origin and (1/kgrid_size[1], 0, 0), (0, 1/kgrid_size[2], 0), (0, 0, 1/kgrid_size[3]) is sufficient.

mfherbst:
> (PS: this should also correctly handle k-grids of size 1 in some directions since the _check will trivially pass for them)

This sounds like a good thing to check in a unit test.

PR #1217, opened 2026-01-06
mGGA optimizations and GPU porting

abussy:
I recently realized that mGGA was, in fact, not ported to the GPU. This PR fixes this, and adds a relevant GPU test.

Various optimizations where also done to reduce allocations, and reuse pre-computed data. Most of these optimizations where already mentioned in a TODO. Note that pre-computing `G_plus_k` has the biggest impact, particularly on the GPU.

With these modifications, the cost of applying `divAgrad` is almost exactly 3x that of applying `local` (as good as it gets). Previously, it was about 3.5x on the CPU.

### DIFF
--- a/test/external/atomsbase.jl
+++ b/test/external/atomsbase.jl
@@ -243,7 +243,7 @@ end
     pos1 = [0.0, 0.0, 0.0]
     pos2 = [1.35, 1.35, 1.35]
     cell = [[5.4, 0.0, 0.0], [0.0, 5.4, 0.0], [0.0, 0.0, 5.4]]
-    psp = PseudoFamily("cp2k.nc.sr.pbe.v0_1.largecore.gth")
+    psp = PseudoFamily("dojo.nc.sr.pbe.v0_4_1.standard.upf")
 
     # Float64 positions, Float64 lattice
     system_ff = periodic_system(


src/terms/Hamiltonian.jl
```diff
@@ -140,6 +140,10 @@ end
     n_bands = size(ψ, 2)
     iszero(n_bands) && return Hψ  # Nothing to do if ψ empty
     have_divAgrad = !isnothing(H.divAgrad_op)
+    if have_divAgrad
+        # Precompute G_plus_k for DivAgradOperator
+        G_plus_k = [map(p -> p[α], Gplusk_vectors_cart(H.basis, H.kpoint)) for α = 1:3]
```
mfherbst, line 148: Some codes actually compute this once and for all and store it somewhere.

Is this something worth exploring ? E.g. store it in the `Hamiltonian` object like a `\psi_scratch`, store it in one of the types (e.g. the xc type needs this also when building the potential) ?

I'm not saying we should do this, I just wanted to pose the question and see if this could be worth it. If it's unclear as of now, we should probably make a respective TODO to explore this later.

src/terms/operators.jl
```diff
@@ -143,14 +143,19 @@ struct DivAgradOperator{T <: Real, AT} <: RealFourierOperator
...
-    # TODO: Performance improvements: Unscaled plans, avoid remaining allocations
-    #       (which are only on the small k-point-specific Fourier grid
-    G_plus_k = [[p[α] for p in Gplusk_vectors_cart(op.basis, op.kpoint)] for α = 1:3]
+                ψ_real=zeros_like(ψ.fourier, complex(eltype(op.basis)), op.basis.fft_size...),
+                G_plus_k=nothing)
+    if isnothing(G_plus_k)
+        G_plus_k = [map(p -> p[α], Gplusk_vectors_cart(op.basis, op.kpoint)) for α = 1:3]
+    end
+    ψ_recip = similar(ψ.fourier)   # pre-allocate large array
+    norm = prod(op.basis.fft_size) # use unnormalized FFT plans for speed
```
mfherbst: I don't think this is the normalisation constant, actually. The normalisation constant is stored in `basis.fft_grid` if I don't misremember.

src/terms/operators.jl
```diff
@@ -143,14 +143,19 @@ struct DivAgradOperator{T <: Real, AT} <: RealFourierOperator
     A::AT
 end
 function apply!(Hψ, op::DivAgradOperator, ψ;
-                ψ_scratch=zeros(complex(eltype(op.basis)), op.basis.fft_size...))
-    # TODO: Performance improvements: Unscaled plans, avoid remaining allocations
-    #       (which are only on the small k-point-specific Fourier grid
-    G_plus_k = [[p[α] for p in Gplusk_vectors_cart(op.basis, op.kpoint)] for α = 1:3]
+                ψ_real=zeros_like(ψ.fourier, complex(eltype(op.basis)), op.basis.fft_size...),
+                G_plus_k=nothing)
```
mfherbst, line 147: Can you add an explicit comment that `\psi_real` is a pre-allocated scratch array to hold a real-space representation of the \psi, which will be overwritten by the function.

src/terms/Hamiltonian.jl
```diff
@@ -159,7 +163,8 @@ end
                 apply!((; fourier=Hψ[:, iband], real=nothing),
                        H.divAgrad_op,
                        (; fourier=ψ[:, iband], real=nothing);
-                       ψ_scratch=ψ_real)
+                       ψ_real=ψ_real,
+                       G_plus_k=G_plus_k)
```
mfherbst:
```suggestion
                       ψ_real, G_plus_k)  # overwrites ψ_real

File: src/terms/operators.jl [hunk 5]
```diff
@@ -143,14 +143,19 @@ struct DivAgradOperator{T <: Real, AT} <: RealFourierOperator
...
+    if isnothing(G_plus_k)
+        G_plus_k = [map(p -> p[α], Gplusk_vectors_cart(op.basis, op.kpoint)) for α = 1:3]
+    end
+    ψ_recip = similar(ψ.fourier)   # pre-allocate large array
```
mfherbst, line 157: Add a comment that one could move this even further out into the Hamiltonian scratch to avoid any allocations in here (but I would not do this for now as it complicates code quite a bit).

test/gpu.jl
```diff
@@ -103,3 +103,38 @@ end
...
 end
+
+@testitem "GPU magnesium mGGA test" tags=[:gpu] setup=[TestCases] begin
+    using DFTK
+    using CUDA
+    using AMDGPU
+    using LinearAlgebra
+    magnesium = TestCases.magnesium
+
+    function run_problem(; architecture)
```
mfherbst, line 114: In comparison to the ForwardDiff tests done elsewhere we should agree on either passing `architecture` as a positional argument or as a kwarg and be consistent about it. Don't care which one it is.

mfherbst (commented):
Nice, thank you !

src/terms/Hamiltonian.jl
abussy: From my measurements, the cost of computing `G+k` at each Hamiltonian application is small compared to the application of the `DivAgradOperator` (~3%). Given that memory usage is already a concern, I don't think storing an extra `n_Gkx3` array per k-point is the best idea right now. I'll leave a TODO.

test/gpu.jl
abussy: Sure, I think I prefer it as a kwarg. I'll change the ForwardDiff PR accordingly.

src/terms/operators.jl
abussy: I copied from `DftHamiltonian multiplication`. It's the same as the product of `fft_normalization` and `ifft_normalization` from the FFTGrid though. I'll make it more explicit.

PR #1223, opened 2026-01-12
Screened Coulomb and range-separated hybrid functionals

toschaefer:
I am opening this to discuss a new attempt for exact exchange in DFTK.  The code is not extensively tested but I want to see if this is the right direction and style.

Changes:
* implementation of exact exchange based on the exact-exchange branch by @mfherbst.
* includes various strategies to sample the Coulomb potential: spherical truncation, Wigner-Seitz truncation, probe charge Ewald.
* adding the option to display the convergence of the LOBPCG eigensolver.

Next steps:
* speed up application of Fock through adaptively compressed exchange (ACE) technique
* parallelization (maybe GPU?)

Comments:
* currently Gamma-only and non-spin polarized.
* maybe we have to look at the solvers. Currently small isolated molecules in large boxes fail to converge.
* enable hybrid functionals like PBE0 (as already implemented but I somehow messed it up)

Example:
```julia
model  = model_HF(lattice, atoms, positions; coulomb_model=WignerSeitzTruncated())
basis  = PlaneWaveBasis(model; Ecut=Ecut, kgrid=[1, 1, 1])
scfres_hf = self_consistent_field(
                basis;
                tol=1e-7,
                solver=DFTK.scf_damping_solver(damping=0.5), # isolated molecules converge better with 0.5
                ρ=scfres_pbe.ρ, 
                ψ=scfres_pbe.ψ, 
                maxiter=100, 
                diagtolalg=DFTK.AdaptiveDiagtol(; ratio_ρdiff=1e-5)
```

### DIFF
--- a/src/coulomb.jl
+++ b/src/coulomb.jl
@@ -54,6 +54,8 @@ function eval_kernel_fourier(k::ShortRangeCoulomb, Gsq::T) where {T}
     -(4T(π) / Gsq) * expm1(-Gsq / (4 * T(k.μ)^2))
 end
 function _compute_kernel_fourier(k::ShortRangeCoulomb, basis, qpt, q)
+    # Use ReplaceSingularity regularisation to explicitly set as the G==0
+    # component the exact limit of the kernel for G->0, namely π/μ^2
     _compute_kernel_fourier(k, ReplaceSingularity(π/k.μ^2), basis, qpt, q)
 end
 
@@ -104,6 +106,9 @@ function _compute_kernel_fourier(k::SphericallyTruncatedCoulomb, basis, qpt, q)
     Ω = basis.model.unit_cell_volume  
     Rcut = @something k.Rcut cbrt(3Ω/(4π))
     kRcut = SphericallyTruncatedCoulomb(Rcut)
+
+    # Use ReplaceSingularity regularisation to explicitly set as the G==0
+    # component the exact limit of the kernel for G->0
     _compute_kernel_fourier(kRcut, ReplaceSingularity(2π*Rcut^2), basis, qpt, q)
 end
 @doc raw"""
--- a/test/coulomb.jl
+++ b/test/coulomb.jl
@@ -50,9 +50,16 @@
     end
 
     @testset "ShortRangeCoulomb plus LongRangeCoulomb is coulomb" begin
-        k_lr = compute_kernel_fourier(LongRangeCoulomb(0.1, ProbeCharge()), basis)
-        k_sr = compute_kernel_fourier(ShortRangeCoulomb(0.1), basis)
-        @test maximum(abs, k_lr + k_sr - k_probe) < 1e-12
+        k_lr  = compute_kernel_fourier(LongRangeCoulomb(0.1, ProbeCharge()), basis)
+        k_sr  = compute_kernel_fourier(ShortRangeCoulomb(0.1), basis)
+        k_sum = k_lr + k_sr
+
+        # Note: The G=0 component does not match up, because in short-range Coulomb
+        # we can take the G->0 limit exactly, but in long-range we cannot and instead
+        # use the ProbeCharge regularisation. However, for finite k-points the ProbeCharge
+        # regularisation for LongRangeCoulomb and Coulomb does not agree, thus we need
+        # to exclude the G=0 component from the test below.
+        @test maximum(abs, (k_sum - k_probe)[2:end]) < 1e-12
     end
 
     @testset "SphericallyTruncatedCoulomb" begin
@@ -97,7 +104,7 @@ end
             numerical_integral, _ = quadgk(0, qmax) do q
                 4π * q^2 * DFTK.eval_kernel_fourier(kernel, q^2) * exp(-α * q^2)
             end
-            analytical_integral = DFTK.eval_probe_charge_integral(kernel, α, Ω)
+            analytical_integral = DFTK.eval_probe_charge_integral(kernel, α)
             @test abs(numerical_integral - analytical_integral) < 1e-12
         end
     end
--- a/test/silicon_hse.jl
+++ b/test/silicon_hse.jl
@@ -1,3 +1,5 @@
+# TODO: If this is too expensive, mark as :slow, features should be well-tested
+#       with other tests, too
 @testitem "Silicon HSE" tags=[:exx, :dont_test_mpi] setup=[RunSCF] begin
     using DFTK
     using PseudoPotentialData


toschaefer:
I take back what I said about molecules. With damping=0.5 it also works for my CO molecule test quite well. 

Regarding the solvers the following puzzles me.
I use a 8 atomic conventional cell of diamond and find different HF results
```
PBE: Energies(total = -47.578301022739254)
HF from direct_minimization starting from PBE:        Energies(total = -43.06725442190505)
HF from scf_damping_solver  starting from PBE:        Energies(total = -43.056402357690274)
HF from direct_minimization starting from stochastic: Energies(total = -43.05655090988068)
```
Find the code and the full output below.

Converging HF is always much more difficult than PBE (multiple minima etc), but this is a simple system. I don't see such behavior in VASP when I compare the conjugate gradient algorithm with (say) the Davidson algorithm. 

It is also suspicious that the direct_minimization stops at 4 iterations when starting with PBE orbitals. Do I use direct_minimization correctly?
Perhaps the symmetrization I implemented in the last commit is incorrect.

<details> <summary><b>Click to view code</b></summary>

```julia
pd_pbe_family = PseudoFamily("dojo.nc.sr.pbe.v0_5.stringent.upf") # pseudopotentials

C = ElementPsp(:C, pd_pbe_family)
atoms = [C, C, C, C, C, C, C, C]
box_length = 6.7406526013
lattice = [[ box_length  0.000000  0.00000];
           [ 0.00000  box_length  0.00000];
           [ 0.00000  0.000000  box_length]]
positions = [[0.00000, 0.00000, 0.00000],
             [0.50000, 0.50000, 0.00000],
             [0.00000, 0.50000, 0.50000],
             [0.50000, 0.00000, 0.50000],
             [0.25000, 0.25000, 0.25000],
             [0.75000, 0.75000, 0.25000],
             [0.25000, 0.75000, 0.75000],
             [0.75000, 0.25000, 0.75000]]
Ecut = pseudometa(C)["cutoffs_normal"]["Ecut"]

model  = model_PBE(lattice, atoms, positions)
basis  = PlaneWaveBasis(model; Ecut=Ecut, kgrid=[1, 1, 1])
scfres_pbe = self_consistent_field(basis; tol=1e-7, maxiter=100)

# HF with direct minimization starting from PBE
model  = model_HF(lattice, atoms, positions; coulomb_model=WignerSeitzTruncated(), temperature=0.0)
basis  = PlaneWaveBasis(model; Ecut=Ecut, kgrid=[1, 1, 1])
ψocc, occupation_occ = DFTK.select_occupied_orbitals(basis, scfres_pbe.ψ, scfres_pbe.occupation; threshold=1e-8)
scfres_hf1 = direct_minimization(
                 basis;
                 ψ=ψocc,
                 is_converged=ScfConvergenceEnergy(1e-7),
                 tol=1e-7,
                 maxiter=100,
             )

# HF with damped solver starting from PBE
scfres_hf2 = self_consistent_field(
                 basis;
                 tol=1e-7,
                 solver=DFTK.scf_damping_solver(damping=0.5),
                 ρ=scfres_pbe.ρ,
                 ψ=scfres_pbe.ψ,
                 maxiter=100,
                 diagtolalg=DFTK.AdaptiveDiagtol(; ratio_ρdiff=1e-5)
             )

# HF with direct minimization starting from stochastic
scfres_hf3 = direct_minimization(
                 basis;
                 is_converged=ScfConvergenceEnergy(1e-7),
                 tol=1e-7,
                 maxiter=500,
             )

println("PBE: $(scfres_pbe.energies)")
println("HF from direct_minimization starting from PBE:        $(scfres_hf1.energies)")
println("HF from scf_damping_solver  starting from PBE:        $(scfres_hf2.energies)")
println("HF from direct_minimization starting from stochastic: $(scfres_hf3.energies)")
```

</details>

<details> <summary><b>Click to view full output</b></summary>

```
n     Energy            log10(ΔE)   log10(Δρ)   Diag   Δtime                                                       
---   ---------------   ---------   ---------   ----   ------                                                      
  1   -47.53660160837                   -0.34   14.0    5.80s                                                      
  2   -47.57675100009       -1.40       -1.25    2.0    2.20s                                                      
  3   -47.57820957440       -2.84       -1.92    4.0    609ms                                                      
  4   -47.57826237572       -4.28       -2.47    2.0    243ms                                                      
  5   -47.57829902403       -4.44       -2.98    4.0    264ms                                                      
  6   -47.57830078680       -5.75       -3.41    4.0    536ms                                                      
  7   -47.57830098566       -6.70       -3.88    4.0    236ms                                                      
  8   -47.57830102006       -7.46       -4.40    4.0    265ms                                                      
  9   -47.57830102235       -8.64       -4.78    6.0    558ms                                                      
 10   -47.57830102266       -9.51       -5.14    3.0    260ms                                                      
 11   -47.57830102273      -10.13       -5.71    2.0    238ms                                                      
 12   -47.57830102274      -11.28       -6.08    4.0    264ms                                                      
 13   -47.57830102274      -12.31       -6.46    2.0    514ms                                                      
 14   -47.57830102274      -12.97       -7.03    3.0    231ms                                                      
┌ Warning: Exact exchange in DFTK is hardly optimised and not yet production-ready.                                
└ @ DFTK ~/src/DFTK_exx/src/standard_models.jl:156                                                                 
┌ Warning: x_tol is deprecated. Use x_abstol or x_reltol instead. The provided value (-1) will be used as x_abstol.
└ @ Optim ~/.julia/packages/Optim/gmigl/src/types.jl:110                                                           
┌ Warning: f_tol is deprecated. Use f_abstol or f_reltol instead. The provided value (-1) will be used as f_reltol.
└ @ Optim ~/.julia/packages/Optim/gmigl/src/types.jl:120                                                           
n     Energy            log10(ΔE)   log10(Δρ)   Δtime                                                              
---   ---------------   ---------   ---------   ------                                                             
  1   -43.06686101526                   -0.37    11.6s                                                             
  2   -43.06725442191       -3.41       -1.31    2.82s                                                             
  3   -43.06725442191   +    -Inf       -1.55    11.3s                                                             
  4   -43.06725442191   +    -Inf       -1.46    11.0s                                                             
┌ Warning: DM not converged.                                                                                       
└ @ DFTK ~/src/DFTK_exx/src/scf/scf_callbacks.jl:60                                                                
n     Energy            log10(ΔE)   log10(Δρ)   Diag   Δtime                                                       
---   ---------------   ---------   ---------   ----   ------                                                      
  1   -42.54894924055                   -0.47    3.0    1.67s                                                      
  2   -42.97495333136       -0.37       -1.13    2.0    6.67s                                                      
  3   -43.05142283455       -1.12       -1.36   11.0    15.0s                                                      
  4   -43.06326236368       -1.93       -1.49    9.0    13.8s                                                      
  5   -43.06226185930   +   -3.00       -1.74   10.0    13.5s                                                      
  6   -43.05963178797   +   -2.58       -2.05   10.0    13.5s                                                      
  7   -43.05783799284   +   -2.75       -2.37   11.0    14.8s                                                      
  8   -43.05694756553   +   -3.05       -2.67   11.0    14.0s                                                      
  9   -43.05658292209   +   -3.44       -2.94   11.0    14.0s                                                      
 10   -43.05645565818   +   -3.90       -3.21   10.0    13.5s                                                      
 11   -43.05641778492   +   -4.42       -3.48   10.0    13.5s                                                      
 12   -43.05640817761   +   -5.02       -3.75   10.0    13.5s                                                      
 13   -43.05640581157   +   -5.63       -4.02   10.0    13.5s                                                      
 14   -43.05640482911   +   -6.01       -4.29   10.0    13.5s
 15   -43.05640406696   +   -6.12       -4.55   10.0    13.5s
 16   -43.05640344719   +   -6.21       -4.82   10.0    13.4s
 17   -43.05640300281   +   -6.35       -5.07   10.0    13.5s
 18   -43.05640271859   +   -6.55       -5.33   10.0    13.4s
 19   -43.05640255135   +   -6.78       -5.58   10.0    13.5s
 20   -43.05640245841   +   -7.03       -5.84   12.0    15.4s
 21   -43.05640240859   +   -7.30       -6.09   11.0    14.0s
 22   -43.05640238238   +   -7.58       -6.33   10.0    13.5s
 23   -43.05640236868   +   -7.86       -6.58   11.0    13.9s
 24   -43.05640236149   +   -8.14       -6.82   11.0    14.0s
 25   -43.05640235769   +   -8.42       -7.07   10.0    13.5s
┌ Warning: x_tol is deprecated. Use x_abstol or x_reltol instead. The provided value (-1) will be used as x_abstol.
└ @ Optim ~/.julia/packages/Optim/gmigl/src/types.jl:110
┌ Warning: f_tol is deprecated. Use f_abstol or f_reltol instead. The provided value (-1) will be used as f_reltol.
└ @ Optim ~/.julia/packages/Optim/gmigl/src/types.jl:120
n     Energy            log10(ΔE)   log10(Δρ)   Δtime
---   ---------------   ---------   ---------   ------
  1   +435.6557749540                   -1.57    5.57s
  2   +274.0079872812        2.21       -0.51    3.84s
  3   +168.0393745353        2.03       -0.23    4.25s
  4   +24.68010225464        2.16       -0.11    4.24s
  5   -15.83463100157        1.61       -0.07    3.12s
  6   -24.77102566812        0.95       -0.05    3.09s
  7   -33.96755562803        0.96        0.04    3.12s
  8   -39.09896912481        0.71       -0.31    2.71s
  9   -39.48779184853       -0.41       -0.61    2.30s
 10   -40.00281144082       -0.29       -0.85    2.32s
 11   -41.62590672859        0.21       -0.55    2.71s
 12   -42.28320377688       -0.18       -0.83    2.72s
 13   -42.42076146682       -0.86       -1.08    2.71s
 14   -42.44933183177       -1.54       -1.30    2.00s
 15   -42.60020677568       -0.82       -1.34    2.30s
 16   -42.66636024871       -1.18       -1.24    2.70s
 17   -42.69650750528       -1.52       -1.42    2.26s
 18   -42.75960447911       -1.20       -1.43    2.00s
 19   -42.80244209861       -1.37       -1.38    2.30s
 20   -42.89322337676       -1.04       -1.33    2.30s
 21   -42.92239458658       -1.54       -1.21    2.01s
 22   -42.96836232724       -1.34       -1.45    2.29s
 23   -43.02516397480       -1.25       -1.42    2.72s
 24   -43.04970518063       -1.61       -1.50    2.72s
 25   -43.05480969023       -2.29       -1.81    2.71s
 26   -43.05517344066       -3.44       -1.94    2.71s
 27   -43.05541117986       -3.62       -2.16    2.71s
 28   -43.05571836483       -3.51       -2.23    2.71s
 29   -43.05585528473       -3.86       -2.30    2.71s
 30   -43.05591541213       -4.22       -2.44    2.72s
 31   -43.05625778790       -3.47       -2.51    2.00s
 32   -43.05655090988       -3.53       -2.86    2.29s
 33   -43.05655090988      -13.67       -2.76    12.8s
PBE: Energies(total = -47.578301022739254)
HF from direct_minimization starting from PBE:        Energies(total = -43.06725442190505)
HF from scf_damping_solver  starting from PBE:        Energies(total = -43.056402357690274)
HF from direct_minimization starting from stochastic: Energies(total = -43.05655090988068)
```

</details>

toschaefer:
Ok, thanks. 

I will undo the explicit symmetrization inside the apply! function of the exact exchange. 
Instead I would suggest to implement a function `symmetrize_ψ` in src/symmetry.jl which finds the Nocc eigenvectors of
```math
\gamma_\text{sym} = \frac 1 {N_\text{sym}} \sum_S S \left( \sum_i |\psi_i\rangle \langle \psi_i | \right) S^\dagger
```
in the space of the currently considered bands of the solver.

The resulting orbitals should lead to both a symmetrized density and a symmetrized Fock exchange operator for the next iteration.  I think this should agree with the default in VASP (ISYM=3) for HF calculations (see https://www.vasp.at/wiki/ISYM)

Similarly like the densities are symmetrized within scf solvers, this function should be called in the solvers when we deal with a model containing exact exchange. In this sense, all terms would maintain a consistent symmetry.

Does that make sense and is it consistent with the DFTK philosophy?

antoine-levitt:
https://github.com/JuliaMolSim/DFTK.jl/issues/1226

Project.toml
```diff
@@ -32,6 +32,7 @@ Polynomials = "f27b6e38-b328-58d1-80ce-0feddd5e7a45"
 PrecompileTools = "aea7be01-6a6a-4083-8856-8a6e6704d82a"
 Preferences = "21216c6a-2e73-6563-6e65-726566657250"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
+ProgressMeter = "92933f4c-e287-5a05-a399-4b506db050ca"
```
mfherbst: I don't think the changes in this file are all intentional. Let's discuss later.

src/postprocess/band_structure.jl
```diff
@@ -56,7 +56,7 @@ All kwargs not specified below are passed to [`diagonalize_all_kblocks`](@ref):
     #      types subtype. In a first version the ScfResult could just contain
     #      the currently used named tuple and forward all operations to it.
     (; basis=bs_basis, ψ=eigres.X, eigenvalues=eigres.λ, ρ, εF, occupation,
-     diagonalization=[eigres], seed)
+     diagonalization=[eigres], n_bands_converge=n_bands)
```
mfherbst: This looks unintentional.

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -44,6 +44,7 @@
 ## TODO debug orthogonalizations when A=I
 
 using LinearAlgebra
+using Printf
```
mfherbst: No need, already used in `src/DFTK.jl`

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -396,6 +397,12 @@ end
     full_BX = BX
     full_λs = λs
 
+    if display_progress
```
mfherbst: If we start implementing complex printing as such, we should move towards a callback infrastructure like in the other solvers to avoid blurring clarify of the algorithm by bloaty printing code.

src/standard_models.jl
```diff
@@ -123,7 +123,39 @@ function _model_DFT(functionals::AbstractVector, args...; kwargs...)
...
+                 model_name, kwargs...)
+end
+
+
+
+"""
+Build an Hartree-Fock model from the specified atoms.
+"""
+function model_HF(system::AbstractSystem; pseudopotentials, 
+                  coulomb_model::CoulombModel=ProbeCharge(), kwargs...)
```
mfherbst: If you don't capture the `coulomb_model` explicitly but implicitly by the `kwargs` you save some code here. Same in the next function.

src/standard_models.jl
```diff
@@ -123,7 +123,39 @@ function _model_DFT(functionals::AbstractVector, args...; kwargs...)
...
+    parsed = parse_system(system, pseudopotentials)
+    _model_HF(parsed.lattice, parsed.atoms, parsed.positions; 
+              coulomb_model=coulomb_model, parsed.magnetic_moments, kwargs...)
+end
+function model_HF(lattice::AbstractMatrix, atoms::Vector{<:Element},
+                  positions::Vector{<:AbstractVector}; 
+                  coulomb_model::CoulombModel=ProbeCharge(), kwargs...)
+    _model_HF(lattice, atoms, positions; coulomb_model=coulomb_model, kwargs...)
+end
+function _model_HF(args...; extra_terms=[], coulomb_model::CoulombModel=ProbeCharge(), kwargs...)
```
mfherbst: I would not annotate a type for `coulomb_model` unless needed.

src/standard_models.jl
```diff
@@ -123,7 +123,39 @@ function _model_DFT(functionals::AbstractVector, args...; kwargs...)
...
+    @warn "Exact exchange in DFTK is hardly optimised and not yet production-ready."
+    model_atomic(args...; extra_terms=[Hartree(), ExactExchange(coulomb_model=coulomb_model), extra_terms...],
```
mfherbst:
```suggestion
    model_atomic(args...; extra_terms=[Hartree(), ExactExchange(; coulomb_model), extra_terms...],
```
(no need to duplicate name of keyword argument and its variable if the variable has the same name)

src/standard_models.jl
```diff
@@ -123,7 +123,39 @@ function _model_DFT(functionals::AbstractVector, args...; kwargs...)
 end
 function _model_DFT(xc::Xc, args...; extra_terms=[], kwargs...)
```
mfherbst: We should think a little more how to do this. Splitting HF and DFT like this leads to potential  code duplication (e.g. the `coulomb_model` stuff needs to be dealt with in two places in an identical way).

Adding a `coulomb_model` keyword argument to `model_DFT` sounds fair. I'm not yet sure whether we should have a `model_HF`. We decided against a reduced Hartree-Fock model in the past since it's just DFT without any DFT functional. Same here: HF is like hybrid DFT without xc term.

src/terms/exact_exchange.jl
```diff
@@ -0,0 +1,62 @@
...
+"""
+struct ExactExchange
+    scaling_factor::Real  # to scale the term
+    coulomb_model::CoulombModel
+end
+ExactExchange(; scaling_factor=1, coulomb_model=ProbeCharge()) = ExactExchange(scaling_factor, coulomb_model)
+(exchange::ExactExchange)(basis) = TermExactExchange(basis, exchange.scaling_factor, exchange.coulomb_model)
+function Base.show(io::IO, exchange::ExactExchange)
+    fac = isone(exchange.scaling_factor) ? "" : "scaling_factor=$(exchange.scaling_factor), "
+    print(io, "ExactExchange($faccoulomb_model=$(exchange.coulomb_model))")
```
mfherbst: I think there is a mistake here as `faccoulomb_model` is not a variable name.

src/coulomb.jl
```diff
@@ -0,0 +1,252 @@
...
+- `q=zero(Vec3)`: Momentum transfer vector in fractional coordinates
+- `coulomb_model::CoulombModel=ProbeCharge()`: Method for treating singularity
+
+# Returns
+Vector of Coulomb kernel values for each G-vector in the spherical cutoff.
+"""
+function compute_coulomb_kernel(basis::PlaneWaveBasis{T};
+                                scaling_factor=one(T),
+                                q=zero(Vec3{T}),
+                                coulomb_model::CoulombModel=ProbeCharge()) where {T}
```
mfherbst: This determines the model or algorithm. This should be positional argument for me (like the `fermialg` in `compute_occupation`).

src/coulomb.jl
```diff
@@ -0,0 +1,252 @@
...
+- `basis::PlaneWaveBasis`: Plane-wave basis defining the grid
+- `scaling_factor=1`: Global scaling factor applied to the result
+- `q=zero(Vec3)`: Momentum transfer vector in fractional coordinates
+- `coulomb_model::CoulombModel=ProbeCharge()`: Method for treating singularity
+
+# Returns
+Vector of Coulomb kernel values for each G-vector in the spherical cutoff.
+"""
+function compute_coulomb_kernel(basis::PlaneWaveBasis{T};
+                                scaling_factor=one(T),
```
mfherbst: If the sole purpose of this is to plainly scale the values after the computation, why pass it ? Better just compute and scale outside to reduce complexity.

src/coulomb.jl
```diff
@@ -0,0 +1,252 @@
...
+                                q=zero(Vec3{T}),
```
mfherbst: Would we not later want to pass a list of `q`-points here ? Maybe just wait with designing this interface until we get there (until of course you know exactly what it needs to look like)

src/coulomb.jl
```diff
@@ -0,0 +1,252 @@
...
+                                coulomb_model::CoulombModel=ProbeCharge()) where {T}
+    # currently only works for Gamma-only (need correct q-point otherwise)
+    qpt = basis.kpoints[1] 
+    coulomb_kernel =  _compute_coulomb_kernel(basis, qpt, q, coulomb_model)
+
+    if iszero(q) # Symmetrize Fourier coeffs to have real iFFT.
+        #enforce_real!(coulomb_kernel, basis) # enforce_real doesn't work as we don't use the full cubic G-grid
+    end
```
mfherbst: Make a TODO and remove the code ?

src/coulomb.jl
```diff
@@ -0,0 +1,252 @@
...
+struct ProbeCharge <: CoulombModel end
+function _compute_coulomb_kernel(basis::PlaneWaveBasis{T},
+                                 qpt::Kpoint,
+                                 q::Vec3{T},
+                                 coulomb_model::ProbeCharge) where {T}
+    model = basis.model
+    NG = length(qpt.G_vectors)
+    coulomb_kernel = zeros(T, NG)
+    # default Coulomb potential using probe charge method for singularity
+    α = π^2 / basis.Ecut # well tested in VASP such that e^(-α*G²) is localized charge with full support on G grid
```
mfherbst: Two general style remarks:
- We try to keep lines below 92 characters (this is a soft limit)
- We use two spaces before starting comments

src/coulomb.jl
```diff
@@ -0,0 +1,252 @@
...
+    probe_charge_sum = zero(T)
+    for (iG, G) in enumerate(to_cpu(qpt.G_vectors))
+        G_cart = model.recip_lattice * (G+q)
+        Gnorm2 = sum(abs2, G_cart)
+        found_singularity = (iG==1 && iszero(q))
+        if !found_singularity
+            coulomb_kernel[iG] = 4T(π) / Gnorm2
+            probe_charge_sum += coulomb_kernel[iG] * exp(-α*Gnorm2)
+        end
+        if iG == NG # in the last cycle probe_charge_sum is complete
```
mfherbst: Set it outside the loop ?

src/coulomb.jl
```diff
@@ -0,0 +1,252 @@
...
+                                 q::Vec3{T},
+                                 coulomb_model::ProbeCharge) where {T}
+    model = basis.model
+    NG = length(qpt.G_vectors)
+    coulomb_kernel = zeros(T, NG)
+    # default Coulomb potential using probe charge method for singularity
+    α = π^2 / basis.Ecut # well tested in VASP such that e^(-α*G²) is localized charge with full support on G grid
+    probe_charge_sum = zero(T)
+    for (iG, G) in enumerate(to_cpu(qpt.G_vectors))
+        G_cart = model.recip_lattice * (G+q)
```
mfherbst: There is a `G_vectors_cart(basis, kpt)`

mfherbst: We try to do expensive operations (like transfer between GPU and CPU, MPI communication) in separate line in order to avoid missing them when reading the code.

src/coulomb.jl
```diff
@@ -0,0 +1,252 @@
...
+Probe charge Ewald method for treating the Coulomb singularity.
+
+Uses a Gaussian probe charge with width parameter α = π²/Ecut to regularize
+the G+q=0 component. Well-tested in VASP for production calculations.
+
+# Reference
+Phys. Rev. B 48, 5058 (1993), doi.org/10.1103/PhysRevB.48.5058
+"""
+struct ProbeCharge <: CoulombModel end
+function _compute_coulomb_kernel(basis::PlaneWaveBasis{T},
```
mfherbst: If you use these type as keyword arguments, then there is no need for the underscore function here.

src/coulomb.jl
```diff
@@ -0,0 +1,252 @@
...
+            coulomb_kernel[iG] = 4T(π) / Gnorm2
+            probe_charge_sum += coulomb_kernel[iG] * exp(-α*Gnorm2)
+        end
+        if iG == NG # in the last cycle probe_charge_sum is complete
+            Ω = model.unit_cell_volume  # volume of cell 
+            probe_charge_integral = 8*π^2*sqrt(π/α) * Ω/(2π)^3  #  = Ω/(2π)^3 ∫ 4π/q² ρ(q) dq  with  ρ(q)=e^(-αq²)
+            coulomb_kernel[1] = probe_charge_integral - probe_charge_sum
+        end
+    end
+    return coulomb_kernel
```
mfherbst: No explicit return, just write the variable.

src/coulomb.jl
```diff
@@ -0,0 +1,252 @@
...
+                                 qpt::Kpoint,
+                                 q::Vec3{T},
+                                 coulomb_model::ProbeCharge) where {T}
+    model = basis.model
+    NG = length(qpt.G_vectors)
+    coulomb_kernel = zeros(T, NG)
+    # default Coulomb potential using probe charge method for singularity
+    α = π^2 / basis.Ecut # well tested in VASP such that e^(-α*G²) is localized charge with full support on G grid
+    probe_charge_sum = zero(T)
+    for (iG, G) in enumerate(to_cpu(qpt.G_vectors))
```
mfherbst: These look to me as if a map would be better. For GPU support you will likely anyway need one. I would compute the probe charge sum in a second step (again better on the GPU anyway as it avoids branching). 

Instead of testing for `iG == 1` you can test for `G+q` being zero (or below a threshold), for example. Or even better for GPU support, only act on a range excluding the first G.

src/coulomb.jl
```diff
@@ -0,0 +1,252 @@
...
+If Rcut < 0 (default), uses Rcut = ∛(3V/(4π)) => V = 4/3π*Rcut^3
+where V is the BvK cell volume.
+
+# Reference
+Phys. Rev. B 77, 193110 (2008), doi.org/10.1103/PhysRevB.77.193110
+"""
+struct SphericallyTruncated <: CoulombModel 
+    Rcut::Float64
+end
+SphericallyTruncated(; Rcut=-1.0) = SphericallyTruncated(Rcut)
```
mfherbst: I'd rather use `nothing` to indicate that a value has not been assigned. Rather than a "special" value. The compiler internally treats `nothing` special, so `Union{Nothing,Float64}` types are basically as fast as having a `Float64` field.

src/coulomb.jl
```diff
@@ -0,0 +1,252 @@
...
+                                 coulomb_model::SphericallyTruncated) where {T}
+    model = basis.model
+    NG = length(qpt.G_vectors)
+    coulomb_kernel = zeros(T, NG)
+    if coulomb_model.Rcut < 0.0 
+        # use default value: V_BvK = 4/3πRcut³
+        Rcut = cbrt(basis.model.unit_cell_volume*3/4/π) # TODO: multiply V with number of k-points
+    else
+        Rcut = coulomb_model.Rcut
+    end
```
mfherbst: if you use `nothing` types, here you can use the `@something` macro.

src/postprocess/band_structure.jl
toschaefer: fixed. was a leftover from the cc4s branch

src/eigen/lobpcg_hyper_impl.jl
toschaefer: done

src/terms/exact_exchange.jl
toschaefer: fixed

src/coulomb.jl
toschaefer: done

src/coulomb.jl
toschaefer: done

src/coulomb.jl
toschaefer: done

src/coulomb.jl
toschaefer: thought that some stuff with q-points is computed in compute_coulomb_kernel and the actual computation is done in the underscore function... so I leave it for the moment as is.

src/coulomb.jl
toschaefer: done

src/coulomb.jl
toschaefer: done

src/coulomb.jl
toschaefer: done

Project.toml
toschaefer: fixed. was a leftover from the cc4s branch

src/eigen/lobpcg_hyper_impl.jl
toschaefer: ~~ok. I leave it because it's quite useful for testing implementations of the adaptively compressed Fock exchange.~~

src/standard_models.jl
toschaefer: since we have not yet finalized the strategy for the interfaces, I leave all interface related reviews untouched.

src/coulomb.jl
toschaefer: I don't quite understand what scaling_factor is for? Scaling the entire exx energy? Like to define hybrid functionals? I've taken it out of compute_coulomb_kernel for now.

src/coulomb.jl
toschaefer: Prefer to wait with further designing. Leave it untouched.

src/eigen/lobpcg_hyper_impl.jl
toschaefer: I replaced the complex printing with a simple callback mechanism in [7407859](https://github.com/JuliaMolSim/DFTK.jl/pull/1223/commits/740785920b331c81bf13507ee782001f3c0818fb)

For debugging I also provide `make_LOBPCG_debug_callback()` as a simple callback function. 

Note: I leave the keyword `display_progress` as it was already present, but I added a `# TODO` that it can principally be removed.

src/terms/operators.jl
```diff
@@ -177,3 +183,41 @@ function optimize_operators(ops)
...
+    kpoint::Kpoint{T}
+    W::TW
+    B::TB
+end
+
+function apply!(Hψ, op::ACExchangeOperator, ψ)
+    coeffs = op.W' * ψ.fourier
+    coeffs = op.B * coeffs
+    mul!(Hψ.fourier, op.W, coeffs, -1, 1)
+end
```
mfherbst: This is just a `NonLocalOperator`. No need to duplicate this code.

src/terms/exact_exchange.jl
```diff
@@ -0,0 +1,158 @@
+@doc raw"""
+Exact exchange term: the Hartree-Exact exchange energy of the orbitals
+```math
+-1/2 ∑_{nm} f_n f_m ∫∫ \frac{ψ_n^*(r)ψ_m^*(r')ψ_n(r')ψ_m(r)}{|r - r'|} dr dr'
+```
+"""
+
+abstract type EXXstrategy end
```
mfherbst: ExxAlgorithm ?

src/terms/exact_exchange.jl
```diff
@@ -0,0 +1,158 @@
...
+    # calculate occupation if missing
+    if isnothing(occupation) && !isnothing(get(kwargs, :eigenvalues, nothing))
+        eigenvalues = kwargs[:eigenvalues]
+        occupation, _ = DFTK.compute_occupation(basis, eigenvalues)
+    end
+    # check for the inconsistency first (ψ given but no occupation)
+    if !isnothing(ψ) && isnothing(occupation)
+        @warn "ψ provided but occupation is missing; cannot set up exact exchange. \
+               Provide eigenvalues or occupation to solve this problem."
+    end
```
mfherbst: The idea for the terms is to be low-level. So if something is missing they crash. It's the responsibility of the caller to make sure everything that is needed is available.

The way I would solve this here in practice is to have the SCF check that if one does an EXX calculation than one actually passes a \psi and occupation to the `self_consistent_field` function (see how we deal with hubbard_n or the τ). In practice one would anyway want to first do a crude DFT calculation, get some orbitals, and then start the HF or hybrid DFT. This automatically enforces this good habit.

src/terms/exact_exchange.jl
```diff
@@ -0,0 +1,158 @@
...
+    # If either is missing, we cannot proceed
+    if isnothing(ψ) || isnothing(occupation)
+        return (; E=zero(T), ops=NoopOperator.(basis, basis.kpoints))
+    end
```
mfherbst: Throw an argument error in such cases asking the user to pass a \psi / occuptaion explicitly (see xc.jl for cases where meta-gga is requested but no τ is given)

src/terms/exact_exchange.jl
```diff
@@ -0,0 +1,158 @@
...
+
+    @assert length(basis.kpoints) == basis.model.n_spin_components # no k-points, only spin
+
+    # TODO Occupation threshold
+    ψ, occupation = select_occupied_orbitals(basis, ψ, occupation; threshold=1e-8)
```
mfherbst: This is an unusual way to deal with this. I would do this directly in the loops over orbitals, implicitly skipping the ones that are numerically unoccupied. See `compute_density` how we do it there. The SCF knows the threshold. It may already be passed down via the kwargs. If not it's easy to add.

src/terms/exact_exchange.jl
```diff
@@ -0,0 +1,158 @@
...
+    compute_exx_ene_ops(term.exx_strategy, term.coulomb_kernel, basis, ψ, occupation)
+end
+
+
+"""
+    CanonicalEXX
+
+Canonical Fock exchange implementation.
+"""
+struct CanonicalEXX <: EXXstrategy end
```
mfherbst: Canonical ... hmm this term has little meaning. It's somehow the "unapproximated" verison or "naive" version. How about `NaiveExx` ?

src/terms/exact_exchange.jl
```diff
@@ -0,0 +1,158 @@
...
+function compute_exx_ene_ops(::CanonicalEXX,
```
mfherbst: I would just have a new method `ene_ops(::NaiveExx,  ...  )` instead of invent a new function name.

Also it seems that more code sharing with the ACE should be possible. Everything until the double loop over n,m is identical in both cases. (No need to pre-allocate the operators. These are super thin wrappers)

src/terms/exact_exchange.jl
```diff
@@ -0,0 +1,158 @@
...
+                ρmn_real = conj.(ψmk_real) .* ψnk_real
+                ρmn_fourier = fft(basis, kpt, ρmn_real) # actually we need a q-point here
+
+                Vmn_fourier = ρmn_fourier .* coulomb_kernel
+                Vmn_real = ifft(basis, kpt, Vmn_fourier)
+                fac_mk = occk[m] / (3-basis.model.n_spin_components)
+                Wnk_real .+= fac_mk .* ψmk_real .* Vmn_real
+                
+                fac_mn = occk[n] * occk[m] 
+                fac_mn /= T(3-basis.model.n_spin_components) # divide 2 (spin-paired) or 1 (spin-polarized)
```
mfherbst: This looks weird. Should this not be the `filled_occupation` ?

src/terms/exact_exchange.jl
```diff
@@ -0,0 +1,158 @@
...
+
+                E -= 0.5 * fac_mn * real(dot(ρmn_fourier .* coulomb_kernel, ρmn_fourier)) 
+            end
+            Wk[:, n] .= fft(basis, kpt, Wnk_real)
+        end
+        M = ψk' * Wk
+        M = Hermitian(0.5*(M+M'))
```
mfherbst: Why do we need this ?

src/terms/exact_exchange.jl
```diff
@@ -0,0 +1,158 @@
...
+        B = inv(M)
```
mfherbst: Explicit inversion is almost always bad. Don't we better want to compute a factorisation and use a backslash.

src/standard_models.jl
mfherbst: I would say we defer this for now and just target HF and no hybrid-DFT yet.

src/terms/exact_exchange.jl
toschaefer: Good point. Rename the following types/keywords:
```
EXXstrategy  -> ExxAlgorithm
CanonicalEXX -> TextbookExx
ACEXX        -> AceExx
exx_strategy -> exx_algorithm
```

src/terms/exact_exchange.jl
toschaefer: Yes, thanks.

src/terms/exact_exchange.jl
toschaefer: We don't need it, I just don't want half the matrix to feel unprivileged ;)
I replace it with
`M = Hermitian(ψk' * Wk)`

src/terms/exact_exchange.jl
toschaefer: I agree. But since M should be positive-definite, what about that?
```julia
M = Hermitian(ψk' * Wk)
B = inv(cholesky(M))
```
Should be safe, right?

I would like to avoid `M \ coeffs` for every application of the Fock exchange...

src/terms/exact_exchange.jl
toschaefer: Yes

src/terms/operators.jl
toschaefer: nice!

src/terms/exact_exchange.jl
mfherbst: Why ? It literally does a backward substitution, so scales quadratically in the matrix size, which is the same as matrix-vector product with the inverse. Actually using the inverse is worse in computational time as it makes the matrix dense ... on top of being numerically not such a good idea.

src/terms/exact_exchange.jl
mfherbst: Thanks. I'm a little unsure about `TextbookExx`.

I'd argue ACE is nowadays so standard, that it's also textbook :). Also the meaning of "textbook" may change over time. Argh. Naming is hard.

src/terms/exact_exchange.jl
toschaefer: Ok, but how to combine that with NonLocalOperator? Like introducing a wrapper that overloads multiplication `M*x` with `M\x`? Maybe it already exists in DFTK, but I couldn't find it.

src/terms/exact_exchange.jl
toschaefer: result of brainstorming with colleagues:
`PureExx`
`PedestrianExx`
`VanillaExx`
`RawExx`
`PlainExx`
`ReferenceExx`
`FullExx`
`DefaultExx`
`OriginalExx`
`Exx1935`
`TraditionalExx`
additional to the ones we already discussed here
`TextbookExx`
`NaiveExx`
`CanonicalExx`

I currently prefer `VanillaExx`, because it avoids any ambiguous reference but is still clear.

src/terms/exact_exchange.jl
toschaefer: All right, I replaced the code with a simple warning:
```julia
if isnothing(ψ) || isnothing(occupation)
    @warn "Exact exchange requires orbitals and occupation, return NoopOperator."
    return (; E=zero(T), ops=NoopOperator.(basis, basis.kpoints))
end
```
I think it is still good to allow HF even without starting orbitals. For some systems, DFT can be qualitatively so different from HF, so that DFT orbitals would even worsen HF convergence (wrong magnetization or metal instead of insulator etc...).

src/terms/exact_exchange.jl
toschaefer: I dump warning, see https://github.com/JuliaMolSim/DFTK.jl/pull/1223#discussion_r2815339162
Let me know if you would rather see an error here. As mentioned above, I find it quite useful to run HF even without initial orbitals.

src/terms/exact_exchange.jl
mfherbst: Yes. There is also the LinearMaps package that does this, but I'm not sure we want to use this dependency for such a small feature. Take a look at FunctionPreconditioner. That does something similar.

src/terms/exact_exchange.jl
toschaefer: done

src/terms/exact_exchange.jl
toschaefer: done (both ene_ops and code sharing)

src/terms/exact_exchange.jl
toschaefer: Ok, I simply introduced a little helper struct:
```julia
# helper struct to let M*x act as M\x
struct InverseMultiplier{T}
    M::T
end
Base.:*(Op::InverseMultiplier, x) = Op.M \ x
```

src/terms/exact_exchange.jl
mfherbst: Ok, let's keep that for now. From past experience it ended up being a good idea to make this a hard error at some point, so I would not be surprised this happens here at some point as well.

src/coulomb.jl
```diff
@@ -0,0 +1,384 @@
...
+- `q=zero(Vec3)`: Momentum transfer vector in fractional coordinates
+- `coulomb_kernel_model::CoulombKernelModel=ProbeCharge()`: Method for treating singularity
+
+# Returns
+Vector of Coulomb kernel values for each G-vector in the spherical cutoff.
+"""
+function compute_coulomb_kernel(basis::PlaneWaveBasis{T};
+                                q=zero(Vec3{T}),
+                                coulomb_kernel_model::CoulombKernelModel=ProbeCharge()) where {T}
+    if length(basis.kpoints) > 1 || !iszero(basis.kpoints[1].coordinate)
```
toschaefer: This also throws an error for spin-polarization. For instance "AFM H chain Hartree-Fock energy" in test/hartree_fock,jl fails. Is there a default way to deal with that in DFTK?

mfherbst: Ah this is my bad. We don't really have a restriction to Gamma-only anywhere but this code. The right thing to check would be `length(basis.kpoints) > basis.model.n_spin_components || !iszero(basis.kpoints[1].coordinate)`

src/coulomb.jl
```diff
@@ -0,0 +1,384 @@
...
+    else
+        @error "Cannot determine kgrid_size for VoxelAveraged Coulomb model."
+    end
+
+    # Define Voxels as reciprocal cell deivided by k-mesh
+    voxel_basis = model.recip_lattice * Diagonal(1 ./ Vec3{T}(kgrid_size))
+    voxel_vol = abs(det(voxel_basis))
+    
+    # Hard coded: 12-point Gauss-Legendre nodes and weights
+    nodes, weights = gauss_legendre_nodes_weights(T, 12)
```
toschaefer: At this point I would actually need the function gausslegendre from FastGaussQuadrature.jl
Currently I hard-coded 12 Gauss-Legendre quadrature points, but it turns out that very small unit cells (hence large reciprocal cells) require more (say 20) to reach sufficient precision. In particular for the voxel containing G+q=0. 
Would you mind adding this dependency to DFTK?

This Coulomb model has been used successfully, particularly for metals and anisotropic cells. James Shepherd's group (Michigan State University) is currently working with this Coulomb model, and I have already convinced them to try our Exx implementation in DFTK instead of VASP, since my implementation here is numerically much more stable than my initial VASP version. We currently try it on the uniform electron gas, and the DFTK version works like a charm...

mfherbst: I think for now I would just add it. 

Otherwise, since we only need it for one function, one could make it a [weak dependency (extmodule)](https://pkgdocs.julialang.org/v1/creating-packages/#Weak-dependencies). That means users need to explicitly load the package if they want to use this functionality, but by default DFTK users don't need to install it.

mfherbst: But honestly, if this is research code, should we not better add it in a follow-up PR ? I mean the entire `VoxelAveraged` ?

I don't think this PR will take long any. I just want to understand a bit better the CoulombModels and refactor away the shared code, hopefully tomorrow.

toschaefer: I wouldn't call it research code, as it is necessary for us for metals and surface slabs (anisotropic cells) and is already a kind of standard in our workflow. 
In addition to the original publication, it is already used as standard in these three publications:
* [CO adsorption on Pt(111) studied by periodic coupled cluster theory](https://doi.org/10.1039/D4FD00085D)
* [Ground States for Metals from Converged Coupled Cluster Calculations](https://doi.org/10.1021/acs.jpclett.4c03134)
* [Finite-Size Effects in Periodic EOM-CCSD for Ionization Energies and Electron Affinities](https://doi.org/10.1021/acs.jctc.4c01451)

I've added it now. Because I'll probably open follow-up PRs anyway. Next up is k-point support (I think it's now or never). 
But you have the final say. If you want, I'll check how to make it a weak dependency.

mfherbst: I see, understood.

My point was more that I would prefer to first have a minimal implementation in master, this is what is done in #1261. These methods should now also work on the GPU (or they will with basically no effort).

The advanced methods can come after. The current Wigner-Seitz implementation still looks too complicate to me. But today I will not have time to figure this out.

toschaefer:
> By the way: I added a few comments to the `ProbleCharge` kernel from the way I understood it. It may be good to get your review @toschaefer to see I understood it correctly.

Thanks for extending the comments. I am a bit unsure about "Brillouin zone diameter" in the comment. The definition contains Ecut but the Brillouin zone is independent of Ecut.
 I think we want a localized charge which is still representable in the plane wave basis. Therefore we chose α to be as broad as possible in reciprocal space, i.e. exp(-α*Gmax^2) < 0.01. Assuming Ecut is roughly given by Gmax^2/2 we find α=π/Ecut.
 
One could also mention, that the G+q=0 component is only problematic for monopole-like terms (i.e. the i=j case). For this case, ProbeCharge is a mathematical trick to analytically subtract these non-decaying monopole interaction (the singularity) from the discrete numerical sum, leaving a smooth function that can be safely integrated over the grid.

toschaefer:
Regarding the splitting between the Coulomb interaction model and the regularization of the singularity, what do you think about the following approach:
```julia
abstract type InteractionKernel end
struct Coulomb           <: InteractionKernel
struct TruncatedCoulomb  <: InteractionKernel  # Spherically and WignerSeitz
struct ShortRangeCoulomb <: InteractionKernel  # erfc(αr)/r

abstract type TruncationStrategy end
struct Spherically <: TruncationStrategy
struct WignerSeitz <: TruncationStrategy

abstract type KernelRegularization end
struct ProbeCharge <: KernelRegularization
struct VoxelAveraged <: KernelRegularization

# === Internal dispatch logic for InteractionKernel in _compute_kernel ===
# Default Coulomb & long-range Coulomb require regularization 
_compute_kernel(basis, qpt, q, ik::CoulombKernel)    = _compute_kernel(basis, qpt, q, ik, ik.regularization)
_compute_kernel(basis, qpt, q, ik::LongRangeCoulomb) = _compute_kernel(basis, qpt, q, ik, ik.regularization)

# Truncated kernels pass the trunaction strategy 
_compute_kernel(basis, qpt, q, ik::TruncatedCoulomb) = _compute_kernel(basis, qpt, q, ik.truncation_strategy)

# analytic forms of some interaction kernels
evaluate_kernel( ::CoulombKernel,     Gnorm2) =  4π / Gnorm2
evaluate_kernel(k::LongRangeCoulomb,  Gnorm2) = (4π / Gnorm2) * exp(-Gnorm2 / (4 * k.α^2))
evaluate_kernel(k::ShortRangeCoulomb, Gnorm2) = (4π / Gnorm2) * (1 - exp(-Gnorm2 / (4 * k.α^2)))


# === Actual implementations here ===
function _compute_kernel(basis, qpt, q, ik::InteractionKernel, ::ProbeCharge) end
function _compute_kernel(basis, qpt, q, ik::InteractionKernel, ::VoxelAveraged) end
function _compute_kernel(basis, qpt, q, ik::InteractionKernel, ::NeglectSingularity) end
function _compute_kernel(basis, qpt, q, ik::InteractionKernel, ::NoRegularization) end

function _compute_kernel(basis, qpt, q, ::Spherically) end
function _compute_kernel(basis, qpt, q, ::WignerSeitz) end

function _compute_kernel(basis, qpt, q, ik::ShortRangeCoulomb) end
```

toschaefer:
Refactored coulomb.jl in [97cafaf](https://github.com/JuliaMolSim/DFTK.jl/pull/1223/commits/97cafaf0d62689b26f2bc02b309b1e21386efc21) so that we can write
```julia
interaction_model = Coulomb(ProbeCharge())
interaction_model = Coulomb(NeglectSingularity())
interaction_model = Coulomb(VoxelAveraged())
interaction_model = TruncatedCoulomb(Spherically())
interaction_model = TruncatedCoulomb(WignerSeitz())
```
Will now quickly implement
```julia
interaction_model = ErfShortRangeCoulomb(α=0.2)
```
which allows us to define the HSE06 hybrid functional.

src/standard_models.jl
```diff
@@ -245,8 +245,7 @@ Possible keyword arguments are those accepted by [`Xc`](@ref).
 r2SCAN(; kwargs...) = Xc([:mgga_x_r2scan, :mgga_c_r2scan]; kwargs...)
 
 """
-Specify a PBE0 hybrid functional in conjunction with [`model_DFT`](@ref)
-<https://doi.org/10.1063/1.478522>.
```
mfherbst, line 249: Why do you remove the mention of PBE0 and the doi here ?

src/standard_models.jl
```diff
@@ -256,20 +255,32 @@ custom exact exchange fraction.
          Note further that at this stage (Feb 2026) there are still known performance bottle
          necks in the code.
 """
-PBE0(; kwargs...) = HybridFunctional([:hyb_gga_xc_pbeh]; kwargs...)
+# PBE0 (https://doi.org/10.1063/1.478522)
+PBE0(; kwargs...)  = HybridFunctional([:hyb_gga_xc_pbeh]; kwargs...)
 
+# Heyd-Scuseria-Ernzerhof (HSE03: https://doi.org/10.1063/1.2204597
+# range separation parameter ω chosen to match VASP (no QuantumESPRESSO implementaion)
+HSE03(; kwargs...) = HybridFunctional([:hyb_gga_xc_hse03];
```
mfherbst: Is HSE03 really so important that it derserves a shorthand ?

src/DFTK.jl
```diff
@@ -101,7 +101,11 @@ include("supercell.jl")
 export Energies
 include("Energies.jl")
 
-export NeglectSingularity, ProbeCharge, SphericallyTruncated
+export Coulomb, TruncatedCoulomb
+export ErfShortRangeCoulomb, ErfLongRangeCoulomb
+export Spherically, WignerSeitz
+export ProbeCharge, VoxelAveraged
+export ReplaceSingularity, ReplaceSingularity
```
mfherbst: duplicate.

src/DFTK.jl
```diff
@@ -101,7 +101,11 @@ include("supercell.jl")
 export Energies
 include("Energies.jl")
 
-export NeglectSingularity, ProbeCharge, SphericallyTruncated
+export Coulomb, TruncatedCoulomb
+export ErfShortRangeCoulomb, ErfLongRangeCoulomb
+export Spherically, WignerSeitz
```
mfherbst: Let's see how the interface discussion pans out, but `Spherically` might be too generic as a name to be exported. Not sure yet.

src/coulomb.jl
```diff
@@ -1,15 +1,18 @@
...
+- [`Coulomb`](@ref): 1/r
+- [`ErfShortRangeCoulomb`](@ref): erfc(ωr)/r
+- [`ErfLongRangeCoulomb`](@ref): erf(ωr)/r
+- [`TruncatedCoulomb`](@ref): spatially truncated 1/r
 
-See also: [`compute_coulomb_kernel`](@ref)
+See also: [`compute_interaction_kernel`](@ref)
 """
-abstract type CoulombSingulartyTreatment end
+abstract type InteractionModel end
```
mfherbst: `InteractionKernel` ? That's after all what it is.

src/coulomb.jl
```diff
@@ -44,32 +47,26 @@ function compute_coulomb_kernel(basis::PlaneWaveBasis{T}; q=zero(Vec3{T}),
 
     # currently only works for Gamma-only (need correct q-point otherwise)
     qpt = basis.kpoints[1] 
-    coulomb_kernel =  _compute_coulomb_kernel(singularity_treatment, basis, qpt)
+    fourier_kernel =  _compute_kernel(basis, qpt, q, interaction_model)
```
mfherbst: It does not matter so much, but here my intuition would be to have the `interaction_kernel` as the first argument.

src/coulomb.jl
```diff
@@ -44,32 +47,26 @@ function compute_coulomb_kernel(basis::PlaneWaveBasis{T}; q=zero(Vec3{T}),
...
 
     # TODO: if q=0, symmetrize Fourier coeffs to have real iFFT 
 
-    coulomb_kernel = to_device(basis.architecture, coulomb_kernel)
+    fourier_kernel = to_device(basis.architecture, fourier_kernel)
```
mfherbst: Actually, I'd just write
```suggestion
    to_device(basis.architecture, fourier_kernel)
```

src/coulomb.jl
```diff
@@ -44,32 +47,26 @@ function compute_coulomb_kernel(basis::PlaneWaveBasis{T}; q=zero(Vec3{T}),
...
-"""
-struct NeglectSingularity <: CoulombSingulartyTreatment end
-function _compute_coulomb_kernel(::NeglectSingularity, basis::PlaneWaveBasis{T}, qpt::Kpoint) where {T}
-    # Compute truncated Coulomb kernel without special-casing singularity at G+q=0 
-    coulomb_kernel = map(Gplusk_vectors_cart(basis, qpt)) do Gpq
-        4T(π) / sum(abs2, Gpq)
-    end
+@doc raw"""
+Abstract type for different strategies to regularize the
+diverging G+q=0 fourier componentn of interaction models.
```
mfherbst: of **Coulomb** interaction models ? There may be other kernels at some point where this component does not diverge.

src/coulomb.jl
```diff
@@ -90,61 +87,387 @@ charges, which can be computed using an Ewald sum.
...
     # Default value well-tested in VASP; ensures that e^(-α*G²) is localized 
     # charge with full support on G grid
-    α = @something kernel.α T(π)^2 / basis.Ecut
+    α = @something regularization.α T(π)^2 / basis.Ecut
 
     Ω = basis.model.unit_cell_volume  # volume of unit cell 
     Gpq = map(Gpq -> sum(abs2, Gpq), Gplusk_vectors_cart(basis, qpt))
 
-    coulomb_kernel = 4T(π) ./ Gpq  # Note: q+G = 0 component is not special-cased, i.e. wrong here
+    interaction_kernel = evaluate_kernel.(Ref(interaction_model), Gpq) # Note: q+G = 0 component is not special-cased, i.e. wrong here
```
mfherbst: line too long. Move comment above.

src/coulomb.jl
```diff
@@ -90,61 +87,387 @@ charges, which can be computed using an Ewald sum.
...
+@kwdef struct ProbeCharge <: KernelRegularization 
     α::Union{Float64, Nothing} = nothing  # Width of the probe charge
 end
-@views function _compute_coulomb_kernel(kernel::ProbeCharge, basis::PlaneWaveBasis{T},
-                                        qpt::Kpoint) where {T}
+@views function _compute_kernel(
+    basis::PlaneWaveBasis{T}, 
+    qpt, 
+    q, 
+    interaction_model::InteractionModel,
```
mfherbst: move to first argument. I'd not type this. This is an internal function and if one injects the wrong type then `evaluate_kernel` will break.

src/coulomb.jl
mfherbst: The `Ref` is weird here. This is how you get rid of it: https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/PlaneWaveBasis.jl#L101

src/coulomb.jl
```diff
@@ -90,61 +87,387 @@ charges, which can be computed using an Ewald sum.
...
 end
-@views function _compute_coulomb_kernel(kernel::ProbeCharge, basis::PlaneWaveBasis{T},
-                                        qpt::Kpoint) where {T}
+@views function _compute_kernel(
+    basis::PlaneWaveBasis{T}, 
+    qpt, 
+    q, 
+    interaction_model::InteractionModel, 
+    regularization::ProbeCharge
+) where {T}
```
mfherbst: This is not how we format such functions. We would align with the opening bracket, see other code.

src/coulomb.jl
```diff
@@ -90,61 +87,387 @@ charges, which can be computed using an Ewald sum.
...
+
+    if iszero(qpt.coordinate)  # Neglect the singularity
+        GPUArraysCore.@allowscalar interaction_kernel[1] = 
+            T(regularization.Gpq_zero_value)
+    end
+    interaction_kernel
 end
 
 
+"""
```
mfherbst: Sorry to be a bit picky here, but can we include `VoxelAveraged` and `WignerSeitz` in a follow-up PR ? It's just a bit too much in one go with the other refactoring for my brain to process.

src/coulomb.jl
```diff
@@ -90,61 +87,387 @@ charges, which can be computed using an Ewald sum.
...
+Long-range Coulomb interaction via error function: erf(ωr)/r
+"""
+@kwdef struct ErfLongRangeCoulomb{KR <: KernelRegularization} <: InteractionModel 
+    ω = 0.106 # ≈ 0.2 / Angstrom
+    regularization::KR = ProbeCharge()
+end
+evaluate_kernel(m::ErfLongRangeCoulomb, Gsq::T) where {T} = (4T(π) / Gsq) * exp(-Gsq / (4 * m.ω^2))
+evaluate_probe_charge_integral(m::ErfLongRangeCoulomb, α, Ω) = 8π^2 * sqrt(π / (α + 1/(4 * m.α^2))) * Ω / (2π)^3
+_compute_kernel(basis, qpt, q, m::ErfLongRangeCoulomb) = _compute_kernel(basis, qpt, q, m, m.regularization)
```
mfherbst: I would move these before the `compute_interaction_kernel` functions. Also some of these lines are too long (92 is the rough length).

src/coulomb.jl
```diff
@@ -90,61 +87,387 @@ charges, which can be computed using an Ewald sum.
...
+@doc raw"""
+Abstract type for truncation strategies of interaction models.
+
+Available models:
+- [`Spherically`](@ref): Spherical truncation at radius Rcut
+- [`WignerSeitz`](@ref): Wigner-Seitz cell truncation
+
+See also: [`InteractionModel`](@ref)
+"""
+abstract type TruncationStrategy end
```
mfherbst: Why do you even need this abstract type ? (in the sense of why not just have the child types as standalone types and remove the abstract one completely)

src/coulomb.jl
```diff
@@ -44,32 +47,26 @@ function compute_coulomb_kernel(basis::PlaneWaveBasis{T}; q=zero(Vec3{T}),
...
-    coulomb_kernel
-end
+Available models:
+- [`ProbeCharge`](@ref): Gygi-Baldereschi probe charge method
+- [`ReplaceSingularity`](@ref): Set G+q=0 component to given value (default is zero)
+- [`VoxelAveraged`](@ref): Replacing 1/(G+q)^2 by its average over the voxel
+
+See also: [`InteractionModel`](@ref)
+"""
+abstract type KernelRegularization end
```
mfherbst: Why do you even need this abstract type ? (in the sense of why not just have the child types as standalone types and remove the abstract one completely)

src/coulomb.jl
```diff
@@ -90,61 +87,387 @@ charges, which can be computed using an Ewald sum.
...
+
+
+"""
+Coulomb interaction 1/r requires regularization
+"""
+@kwdef struct Coulomb{KR <: KernelRegularization} <: InteractionModel 
+    regularization::KR = ProbeCharge()
+end
+evaluate_kernel(::Coulomb, Gsq::T) where {T} = 4T(π) / Gsq
+evaluate_probe_charge_integral(::Coulomb, α, Ω) = 8π^2 * sqrt(π / α) * Ω / (2π)^3
```
mfherbst: You should test these two functions agree by performing numerical quadrature against a proble charge with `QuadGK` in a unit test.

src/coulomb.jl
```diff
@@ -90,61 +87,387 @@ charges, which can be computed using an Ewald sum.
...
+_compute_kernel(basis, qpt, q, m::TruncatedCoulomb) = _compute_kernel(basis, qpt, q, m.truncation)
+
+
+"""
+Short-range Coulomb interaction via error function: erfc(ωr)/r
+"""
+@kwdef struct ErfShortRangeCoulomb <: InteractionModel 
+    ω = 0.106  # ≈ 0.2 / Angstrom
+end
+evaluate_kernel(m::ErfShortRangeCoulomb, Gsq::T) where {T} = -(4T(π) / Gsq) * expm1(-Gsq / (4 * m.ω^2))
```
mfherbst: I would call this function `eval_kernel_fourier` (following a similar naming convention to  `eval_psp_local_fourier` for the internal functions in PSP evaluation)

src/coulomb.jl
```diff
@@ -24,17 +27,17 @@ evaluated only on the spherical cutoff |G+q|² < 2Ecut (not the full cubic FFT g
...
-- `singularity_treatment::CoulombSingulartyTreatment`: Method for treating singularity
+- `interaction_model::InteractionModel`: Method for treating singularity
 
 ## Returns
 Vector of Coulomb kernel values for each G-vector in the spherical cutoff.
 """
-function compute_coulomb_kernel(basis::PlaneWaveBasis{T}; q=zero(Vec3{T}),
-                                singularity_treatment::CoulombSingulartyTreatment=ProbeCharge()) where {T}
+function compute_interaction_kernel(basis::PlaneWaveBasis{T}; q=zero(Vec3{T}),
+                                interaction_model::InteractionModel=ProbeCharge()) where {T}
```
mfherbst: indention.

src/coulomb.jl
mfherbst: Suggestion: Could also give the real-space version of these kernels if you want (may be useful for plotting etc.) and again check agreement between real-space and fourier-space by a numerical quadrature. We do this in the pseudo tests.

src/terms/exact_exchange.jl
```diff
@@ -10,28 +10,28 @@ abstract type ExxAlgorithm end
 """
 Unscreened exact exchange (exx) term. Keyword arguments:
 - `scaling_factor::Real`: Bulk scaling of the entire term
-- `singularity_treatment::CoulombSingulartyTreatment`: Method for treating Coulomb singularity
+- `interaction_model::InteractionModel`: Method for treating Coulomb singularity
 - `exx_algorithm::ExxAlgorithm`: Algorithm for evaluating the exx contribution
```
mfherbst: The docstrings are now largely out of date. By the way this is the reason why we tend to put sparse docstrings. We think that out-of-date docstrings are better than short docstrings that stay valid and encourage you to look at the code (which we also put emphasis on keeping short and clean and thus hopefully accessible).

But more broadly this opens the question if `ExactExchange` is still the right name for this term. Because if one puts a short-ranged coulomb (for example) it's no longer exact exchange, right ? How do people deal with this in the literature usually ?

src/standard_models.jl
toschaefer: oops, I wanted to adapt that for all hybrids and forgot to finish... reconciled.

src/standard_models.jl
toschaefer: That's a matter of taste. I remember a paper where we found it better than HSE06. At the end it's just a name for a fixed set of parameters... it's just another one in the zoo of functionalities.

src/DFTK.jl
toschaefer: Since I removed the type TruncatedCoulomb, they are now called SphericallyTruncatedCoulomb (and WignerSeitzTruncatedCoulomb for a follow up PR).

src/coulomb.jl
toschaefer: That was my initial naming, but then the keyword `interaction_kernel=Coulomb()` would be identical to the variable containing the actual  values `interaction_kernel[iG]`. So I renamed it to `InteractionModel` and `interaction_model=Coulomb()`. 

To summarize, currently we have:
```julia
function compute_interaction_kernel(basis::PlaneWaveBasis{T}; q=zero(Vec3{T}),
                                    interaction_model::InteractionModel=Coulomb()) where {T}
    # ...
    interaction_kernel
end
```

Will rename to the following:
```julia
function compute_fourier_kernel(basis::PlaneWaveBasis{T}; q=zero(Vec3{T}),
                                interaction_kernel::InteractionKernel=Coulomb()) where {T}
    # ...
    fourier_kernel
end
```

src/coulomb.jl
toschaefer: In order to clearly distinguish between the interaction model and the way we regularize a possible singularity.

Enables to set
`interaction_kernel = Coulomb(regularization=VoxelAveraged())` or
`interaction_kernel = ErfLongRangeCoulomb(regularization=ProbeCharge())` or whatever you like...

src/coulomb.jl
toschaefer: don't quite understand what you mean here.

src/terms/exact_exchange.jl
toschaefer: I see what you mean, but I think "exact" refers to the fact that the exchange-like contribution it is not a DFT approximation. Switching the interaction kernel from Coulomb to something else does not really change this philosophy. 

On the other hand, from an educational point of view, the word "exact" is pretty misleading. There are also exchange-like many-electron correlation effects...

One could call it `FockExchange`?

toschaefer: However, the term exact exchange is standard in the literature, and before we change everything and then still have “exact exchange” in the comments, I would rather leave it as it is...

src/coulomb.jl
mfherbst: We typically postfix the real versus fourier, so kernel_fourier would be better.

src/coulomb.jl
mfherbst: `evaluate_kernel_real(::Coulomb, r::T) = one(T) / r`  basically. That's easy to write and allows you to check in tests that the fourier-space expression agrees with the real-space expression you have in mind, by doing a numerical integral to compute the hankel transform.

src/terms/exact_exchange.jl
mfherbst: > However, the term exact exchange is standard in the literature, and before we change everything and then still have “exact exchange” in the comments, I would rather leave it as it is...

Well, my main intention was to spark the discussion and reach an educated decision. Now it's still easy to change compared to how it will be in 6 months.

I am also divided. I know that `ExactExchange` or `FockExchange` is what people use, but in both cases as soon as you put screening it's a bit of a misnomer.

We could be more vague and just call it `Exchange` ? Term names are not super specific in the code, also in other places. E.g. the `Kinetic` term is not the exact kinetic energy if you use Ecut smearing, `AtomicLocal` can be used to model non-atomic interactions. If someone constructs a term like `Exchange(; kernel=Coulomb())` it's pretty clear that this is what people call exact exchange, while something like `Exchange(; kernel=ShortRangeCoulomb())` is clear, too. Could be a good compromise ?

toschaefer: Even if Fock or Hartree only used the full Coulomb kernel in their publications, a different kernel would not change the fundamental idea behind their approach.  Therefore, I would not consider `ExactExchange` or `FockExchange`  in combination with another kernel to be a misnomer.

A Possible price one pays for being more vague: when a user searches for "exchange" in the docs/code, it might collide with exchange from xc functionals.

src/coulomb.jl
mfherbst: > Enables to set

But for that you don't need the abstract type. You just need the two types `VoxelAveraged` or `ProbeCharge` or whatever.

In Julia one often works "by convention" in interfaces rather than the object-oriented thing of defining abstract classes that enforce which functions are implemented or not. This also means that one does not always provide a very well-defined hierarchy of abstract types if this not needed. Here I would argue you don't need it ... which gives users that want to implement their own regularisation more flexibility.

src/terms/exact_exchange.jl
mfherbst: > A Possible price one pays for being more vague: when a user searches for "exchange" in the docs/code, it might collide with exchange from xc functionals.

Ah yes, that's true. In that case `ExactExchange` or `FockExchange` is probably better.

I don't know what's better ... you know the field more what people expect. I'm leaning to `FockExchange`  right now, but this is just an infinitesimal tendency.

src/standard_models.jl
mfherbst: I agree and that's the reason we have been very conservative in defining these shorthands ... to avoid them becoming a zoo. Basically the guideline is "things one would want to have handy for some quick testing". I would be in disfavour making both available and since I've never heard or used HSE03, I'd make HSE06 available, but not HSE03. However, if you tell me that HSE03 should be preferred, that's fine for me, too.

toschaefer: That sounds reasonable to me. Then let's stick with HSE (which is simply HSE06).

src/terms/exact_exchange.jl
toschaefer: Both is fine. The question is how to rename `exx_coefficient`, `exx_algorithm`, `exx_operator`, ... if we rename to `FockExchange`
Principally  `fock_exchange_coefficient`, `fock_exchange_algorithm`, `fock_exchange_operator` is possible, but a bit long and clumsy.

src/coulomb.jl
toschaefer: I see, but when I look at the Hankel transform of a real space function $v(r)$
```math
\int_0^R v(r) \cdot \frac{\sin(Gr)}{Gr}\cdot 4\pi r^2 \; \textrm d r
```
it does not converge with $R\rightarrow\infty$ for cases like $v(r) = 1/r$ or $v(r) = \textrm{erf}(μr)/r$.
So this would only work for the case of ErfShortRangeCoulomb with $v(r) = \textrm{erfc}(μr)/r$. But in this case the Fourier kernel is known and any mistake would be revealed in the agreement test with Quantum Espresso for HSE.

toschaefer: Or were you referring mainly to the integral with the probe charge?

src/coulomb.jl
toschaefer: Makes sense. Removed the abstract type `KernelRegularization` in [3baa0f7](https://github.com/JuliaMolSim/DFTK.jl/pull/1223/commits/3baa0f796bf23623b559d5d5766092e12efcec72)

src/terms/exact_exchange.jl
toschaefer: After all, I am leaning towards staying with `ExactExchange`.

src/coulomb.jl
toschaefer: Add a test for the probe charge formulas in [4749dcc](https://github.com/JuliaMolSim/DFTK.jl/pull/1223/commits/4749dcc33fdffd01e262f8cd84904d5c4fb30f6c)

PR #1236, opened 2026-01-21
Fix kernel self-adjointness tests to use k-weighted inner product

niklasschmitz:
This corrects the self-adjointness test to account for the appropriate inner product in the presence of symmetries: $\langle \phi, \psi\rangle = \sum_{k \in BZ} \langle \phi_k, \psi_k\rangle = \sum_{k \in IBZ} w_k \langle \phi_k, \psi_k\rangle$ with weights $w_k$ on the irreducible Brillouin zone $IBZ$. In particular the four-point kernel `apply_K` is the culprit where this matters, as it is not diagonal with respect to k (due to the density) and so does not commute with k-weight scaling.

With this, I also tighten here the tolerances on self-adjointness tests of `apply_K` and `apply_Ω` since they do not involve an inner  solver.

### DIFF
--- a/src/pseudo/PspUpf.jl
+++ b/src/pseudo/PspUpf.jl
@@ -15,8 +15,8 @@ struct PspUpf{T,I} <: NormConservingPsp
     # Kleinman-Bylander energies. Stored per AM channel `h[l+1][i,j]`.
     # UPF: `PP_DIJ`
     h::Vector{Matrix{T}}
-    # (UNUSED) Pseudo-wavefunctions on the radial grid. Can be used for wavefunction
-    # initialization and as projectors for PDOS and DFT+U(+V).
+    # Pseudo-wavefunctions on the radial grid. Used as projectors for PDOS
+    # and DFT+U(+V), could be used for wavefunction initialization as well.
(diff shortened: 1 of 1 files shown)


mfherbst:
Nice, thanks !

I wonder if this is not an issue also in the Newton and direct minimisation implementation ?

@antoine-levitt @gkemlin

niklasschmitz:
The non-split `solve_ΩplusK` uses `cg` which should rely on symmetry in principle? I wonder whether we can re-scale J = ΩplusK accordingly before passing to cg and whether this makes a difference on the convergence

mfherbst:
I'll go ahead and merge this as this PR is clearly fine and dandy, but the broader issue of should the weights appear somewhere is still to be discussed.

antoine-levitt:
> The non-split solve_ΩplusK uses cg which should rely on symmetry in principle? I wonder whether we can re-scale J = ΩplusK accordingly before passing to cg and whether this makes a difference on the convergence

Ah yes you're totally right. Indeed we should do that (although note that cg is surprisingly robust to non-self-adjointness)

PR #1237, opened 2026-01-22
Optimize response for GPUs

abussy:
This PR optimizes the response code for GPUs. This is already very optimized, but I'd like to raise a couple of questions and collect feedback before I clean it up. My questions will mostly be relevant to @mfherbst , @niklasschmitz and @gkemlin (I think).

Here is a performance teaser, based on a band gap optimization script (see below). These are timings of the `self_consistent_field ForwardDiff` function, reported in seconds:

|  | master | this PR |
|----------|----------|----------|
| CPU  | 4640  | 4104  |
| H100  | 780  | 69  |
| Mi250  |850  | 125 |

The main aspects of this PR are:
- Vectorization of the CG solver: currently, a different CG solver is applied to each band. This is not ideal for GPUs, which perform much better on larger arrays. This PR generalizes the CG solver for matrix input (also helps with BLAS threading)
- Usage of buffers in the projections. Currently the [projectors](https://github.com/JuliaMolSim/DFTK.jl/blob/af5f56a0909ffadc655c17ba5924246acdb57092/src/response/chi0.jl#L127-L135) allocate a lot of memory, especially because they are nested in each other, and called many times over. This PR introduces in-place projectors  that avoid memory allocations
- Optimization of Sternheimer solver setup. Right before the solver is called, nbands^2 column-wise operations [take place](https://github.com/JuliaMolSim/DFTK.jl/blob/af5f56a0909ffadc655c17ba5924246acdb57092/src/response/chi0.jl#L362-L375). This is very slow on the GPU, and this PR replaces it with a single linear algebra call

There are no fundamental changes, but I have a few questions:

- [ ] There are 2 functions that seem to do the same thing: `solve_ΩplusK` and `solve_ΩplusK_split`. I focused on the latter since it is the one used in response.  Should both exist? Should I also optimize the former?
- [ ] I duplicated the CG solver and left the current one that takes a `LinearMap`. Does keeping it make any sense, or should I adapt the existing code calling the CG solver to use the vectorized version?
- [ ] In `chi0.jl`, a [`R_ldiv!`](https://github.com/JuliaMolSim/DFTK.jl/blob/af5f56a0909ffadc655c17ba5924246acdb57092/src/response/chi0.jl#L180-L182) function is defined. It involves 2 nested projections R: `x .= R(precon \ R(y))`. I have a feeling the projection of `y` is redundant. In fact, I pass all the tests when removing it. Is there a good theoretical reason it should stay?
- [ ] In the [current code](https://github.com/JuliaMolSim/DFTK.jl/blob/af5f56a0909ffadc655c17ba5924246acdb57092/src/response/chi0.jl#L378-L379), there is one `tol` per band. Is that necessary?
- [ ] The current CG solver has MPI support, is that necessary? 

Benchmark script for above table:
```
using DFTK, PseudoPotentialData
using ForwardDiff, Optimization # version 4.8.0 required
using CUDA

function gaas_band_gap(a)
    @show a
    lattice = a * [1 0 0; 0 1 0; 0 0 1]
    pseudopot = PseudoFamily("dojo.nc.sr.pbe.v0_4_1.stringent.upf")
    Ga = ElementPsp(:Ga, pseudopot)
    As = ElementPsp(:As, pseudopot)
    atoms = [Ga, Ga, Ga, Ga, As, As, As, As]
    positions = [[0.00, 0.00, 0.00],
                 [0.00, 0.50, 0.50],
                 [0.50, 0.00, 0.50],
                 [0.50, 0.50, 0.00],
                 [0.25, 0.75, 0.75],
                 [0.25, 0.25, 0.25],
                 [0.75, 0.75, 0.25],
                 [0.75, 0.25, 0.75],
]
    model = model_DFT(lattice, atoms, positions; functionals=PBE(), temperature=1e-3)
    basis = PlaneWaveBasis(model; Ecut=32, kgrid=(3, 3, 3), architecture=DFTK.GPU(CuArray))
    @show basis.kcoords_global[1]
    scfres = self_consistent_field(basis; tol=1e-6)
    εг = scfres.eigenvalues[1] # Extract eigenvalues at П-point 
    vbm = maximum(εг[εг.≤ scfres.εF])
    cbm = minimum(εг[εг.> scfres.εF])
    @show cbm  - vbm
    band_gap = cbm  - vbm
end

a_original = 10.55 # GaAs lattice constant in bohrs
Eg_target = 0.015 # Target band gap in Hartree
band_gap_loss(a, p) = (Eg_target - gaas_band_gap(a[1]))^2
optf = OptimizationFunction(band_gap_loss, AutoForwardDiff())
prob = OptimizationProblem(optf, [a_original])
#warm-up
sol = solve(prob, Optimization.LBFGS(); maxiters=1, x_tol=1e-3, allow_f_increases=true, show_trace=true)
DFTK.reset_timer!(DFTK.timer)
t = @time begin
    sol = solve(prob, Optimization.LBFGS(); maxiters=10, x_tol=1e-3, allow_f_increases=true, show_trace=true)
end
println("Optimized design lattice constant: ", sol.u[1], "in ", t, "seconds") # 10.549
@show DFTK.timer
```

### DIFF
--- a/src/response/chi0.jl
+++ b/src/response/chi0.jl
@@ -415,8 +415,8 @@ function compute_δψ!(δψ, basis::PlaneWaveBasis{T}, H, ψ, εF, ε, δHψ, ε
         mul!(δψk, ψk, dot_prods, 1, 1)
 
         # Sternheimer contribution, for all columns of δψk at once.
-        εk_minus_q_gpu = to_device(basis.architecture, εk_minus_q),
-        res = sternheimer_solver(Hk, ψk, εk_minus_q_gpu, δHψ[ik];
+        εk_minus_q_device = to_device(basis.architecture, εk_minus_q)
+        res = sternheimer_solver(Hk, ψk, εk_minus_q_device, δHψ[ik];
                                  ψk_extra, εk_extra, Hψk_extra,
                                  tol=tolk_minus_q, kwargs_sternheimer...)
         !res.converged && @warn("Sternheimer CG not converged", res.tol, res.residual_norms)


src/response/cg.jl
```diff
@@ -7,63 +7,133 @@ end
...
+    catch MethodError
+        op_supports_active = false
+    end
+    function apply_op!(Ax, x; active=nothing)
+        if op_supports_active
+            A!(Ax, x; active)
+        else
+            A!(Ax, x)
+        end
+    end
```
abussy: I would appreciate feedback on this. To enable locking, the range of active columns might need to be passed to the `A!` input operator. I did not want to make calling CG in simple cases complicated by forcing the implementation of the `active` kwarg, so I came up with this. I do not know if there is a better way.

niklasschmitz: One downside of this is wasting one large matvec cost just to test the kwarg presence. Possibly this is a use case here for [Base.hasmethod](https://docs.julialang.org/en/v1/base/base/#Base.hasmethod) like
```
op_supports_active = hasmethod(
    A!,
    Tuple{typeof(x), typeof(x)},
    (:active,)
)
```

abussy: Thanks, I'll have a look

abussy: I used your suggestion: it looks cleaner, had has a measurable impact on performance for systems with numerous k-points (i.e. many individual calls to CG).

niklasschmitz: nice ! :) it probably helped the compiler, since the new version should be completely static in terms of types (not sure how easy try-catch is for compiler optimizations to happen, might be more sensitive to dynamic language semantics)

src/response/cg.jl
```diff
@@ -7,63 +7,131 @@ end
...
+             tol=1e-10*ones(T, size(b, 2)), maxiter=100, miniter=1,
+             my_dots=nothing) where {T}
+
+    # Columnwise dot products and norms, possibly custom
+    if isnothing(my_dots)
+        my_dots = columnwise_dots
+        my_norms = columnwise_norms
+    else
+        my_norms(x) = sqrt.(real.(my_dots(x, x)))
+    end
```
niklasschmitz: I think with `my_dots` as a kwargs instead of a positional arg, the Julia compiler cannot specialize on the type of the function; Possibly a function barrier (https://docs.julialang.org/en/v1/manual/performance-tips/#kernel-functions) helps here? (At the cost of an added indirection, so it depends whether it affects performance here)

src/response/cg.jl
```diff
@@ -7,63 +7,131 @@ end
...
-    my_norm(x) = sqrt(real(my_dot(x, x)))
+function cg!(x::AbstractArray{T}, A!, b::AbstractArray{T};
+             precon=I, proj! =copy!, callback=identity,
+             tol=1e-10*ones(T, size(b, 2)), maxiter=100, miniter=1,
+             my_dots=nothing) where {T}
+
+    # Columnwise dot products and norms, possibly custom
+    if isnothing(my_dots)
+        my_dots = columnwise_dots
+        my_norms = columnwise_norms
```
niklasschmitz: I found the old naming `my_dot` and `my_norm` (singular) more intuitive: Mathematically there is still just one user-defined choice (inner product) and an induced norm; Just that it is applied on every column in case there are multiple columns.

src/response/cg.jl
abussy: I'll check performance implications. I started developing with the hard-coded `columunwise_dots`, and I didn't observe any performance drop when generalizing. I have nothing against having it as a kwarg.

abussy: I'm still not 100% comfortable with these things. In the end, I made sure that all results from dot products and norms are written in existing arrays, such that specialization can take place with the results. I have seen small performance improvements, but it's not huge (I cannot rule out statistical variations of timings).

Do you thing this approach would do?

src/response/cg.jl
abussy: In the end I doubled down and call the `my_columnwise_dots` and `my_columnwise_norms`, because I think it must be clear to the caller that these should return an array.

src/response/cg.jl
```diff
@@ -7,63 +7,130 @@ end
 
 """
 Implementation of the conjugate gradient method which allows for preconditioning
-and projection operations along iterations.
+and projection operations along iterations. The solver is optimized for generic
+input arrays with multiple columns, although vector inputs are also supported.
+To reduce allocations, operator A! and projector proj! are expected to be in-place
+functions. A basic locking meachnism for converged columns is implemented.
```
mfherbst: This is not clear. Can you explain explicitly what the structure of the input arrays is expected to be. E.g. it solves simultaneously linear systems `A x[:, i] = b[:, i]` etc.

src/response/cg.jl
```diff
@@ -7,63 +7,130 @@ end
...
+
+    # Utility to accept operators that do not implement active ranges for locking. In such
+    # cases, all columns are always considered. Allows to use the same cg! implementation
+    # across the board, without extra complication. For performance, when x and b are not
+    # single column vectors, active ranges should be implemented.
+    op_supports_active = hasmethod(
+        A!,
+        Tuple{typeof(x), typeof(x)},
+        (:active,)
+    )
```
mfherbst: Why do we need this at all ? If you use views to narrow down to the columns which are already locked, should you not be just fine ?

src/response/cg.jl
```diff
@@ -7,63 +7,130 @@ end
 
 """
 Implementation of the conjugate gradient method which allows for preconditioning
-and projection operations along iterations.
+and projection operations along iterations. The solver is optimized for generic
+input arrays with multiple columns, although vector inputs are also supported.
+To reduce allocations, operator A! and projector proj! are expected to be in-place
+functions. A basic locking meachnism for converged columns is implemented.
 """
-function cg!(x::AbstractVector{T}, A::LinearMap{T}, b::AbstractVector{T};
```
mfherbst, line 13: There are versions of `LinearMap` that support in-place operations, I believe. In case you have not looked at this, this may be an option to consider. Forcing the usage of an `A!` is ok, but certainly makes the code less intuitive overall. I find it nicer to write `mul!(c, A, x)` rather than `A!(c, x)`.

src/response/cg.jl
```diff
@@ -7,63 +7,130 @@ end
...
-             my_dot=dot) where {T}
-    my_norm(x) = sqrt(real(my_dot(x, x)))
+function cg!(x::AbstractArray{T}, A!, b::AbstractArray{T};
+             precon=I, proj! =copy!, callback=identity,
+             tol=1e-10*ones(real(T), size(b, 2)), maxiter=100, miniter=1,
+             my_columnwise_dots=columnwise_dots) where {T}
+
+    # Columnwise dot products and norms, possibly custom
+    # Write results of dot products and norms into existing array to allow type specialization
+    my_columnwise_norms(x) = sqrt.(real.(my_columnwise_dots(x, x)))
```
mfherbst, line 36: I don't understand these comments.

src/response/cg.jl
```diff
@@ -7,63 +7,130 @@ end
...
     # preconditioned conjugate gradient
     while n_iter < maxiter
         # output
-        info = (; A, b, n_iter, x, r, residual_norm, converged, stage=:iterate)
+        info = (; A!, b=full_b, n_iter, x=full_x, r=full_r, residual_norms=full_residuals,
+                  converged, stage=:iterate)
         callback(info)
         n_iter += 1
-        if (n_iter ≥ miniter) && residual_norm ≤ tol
+        converged_cols .= full_residuals .<= tol
```
mfherbst, line 82: If it helps, we could assume that convergence is in order (i.e. column 2 always converges after column 1). For most CG use cases we have this is should be true (as the order says something about the distance of the eigenvalues from the fermi level and the closer, the harder to converge).

src/response/cg.jl
mfherbst: Also you only use this function in-place. So you could define
`my_columnwise_norms!(y, x) = map!(y, xi -> sqrt(real(xi)), my_columnwise_dots(x, x))`
not sure that helps or makes things more readable, but just a suggestion to maybe save a line of code below.

src/response/cg.jl
```diff
@@ -7,63 +7,130 @@ end
...
+        tmp = zeros_like(γ)
+        tmp .= my_columnwise_dots(p, c)
+        α = γ ./ tmp
 
         # update iterate and residual while ensuring they stay in Ran(proj)
-        x .= proj(x .+ α .* p)
-        r .= proj(r .- α .* c)
-        residual_norm = my_norm(r)
+        proj_buffer .= x .+ p .* α'
+        proj!(x, proj_buffer)
```
mfherbst: For these sort of things, where in-place operations obfuscate the actual algorithm a bit, it can be helpful to add a comment that writes out the "nicer" form of this algorithm, e.g.  `# x .= proj(x .+ α .* p)`

src/response/cg.jl
```diff
@@ -7,63 +7,130 @@ end
...
+        proj!(r, proj_buffer)
+        residual_norms .= to_cpu(my_columnwise_norms(r))
 
-        # apply preconditioner and prepare next iteration
-        ldiv!(c, precon, r)
-        γ_prev, γ = γ, my_dot(r, c)
-        β = γ / γ_prev
-        p .= proj(c .+ β .* p)
+        # apply preconditioner and prepare next iteration. Preconditioner applied to full arrays,
+        # because the FunctionPreconditioner type makes it hard to change the active set (cheap enough).
```
mfherbst: make a note this updates `c` as we are using views.

src/response/chi0.jl
```diff
@@ -109,12 +109,9 @@ precondprep!(P::FunctionPreconditioner, ::Any) = P
 # included).
```
mfherbst, line 114: I think some comment that explains that this is a simultaneous solver for multiple right-hand sides should be made (here and in other places in the call stack).

src/response/chi0.jl
```diff
@@ -152,8 +161,9 @@ precondprep!(P::FunctionPreconditioner, ::Any) = P
     # ψk_extra are not converged but have been Rayleigh-Ritzed (they are NOT
     # eigenvectors of H) so H(ψk_extra) = ψk_extra' (Hk-ε) ψk_extra should be a
     # real diagonal matrix.
-    H(ϕ) = Hk * ϕ - ε * ϕ
-    ψk_exHψk_ex = Diagonal(real.(εk_extra .- ε))
+    # Allow for application on an active range only
+    H(ϕ; active=1:size(ϕ, 2)) = Hk * ϕ - ϕ .* ε[active]'
```
mfherbst: Is this not just `H(ϕ; active=1:size(ϕ, 2)) = Hk * ϕ - ϕ * Diagonal(ε[active])` ?

src/response/chi0.jl
```diff
@@ -131,8 +128,20 @@ precondprep!(P::FunctionPreconditioner, ::Any) = P
...
+        Rϕ = ψk * (ψk' * ϕ) # P(ϕ)
+        mul!(Rϕ, ψk_extra, ψk_extra' * ϕ, -1, -1) # -P(ϕ) - P_extra(ϕ)
+        Rϕ .+= ϕ # R = ϕ -P(ϕ) - P_extra(ϕ)
+    end
+
+    function R!(Rϕ, ϕ)
+        mul!(Rϕ, ψk, ψk' * ϕ)
+        mul!(Rϕ, ψk_extra, ψk_extra' * ϕ, -1, -1)
+        Rϕ .+= ϕ
+    end
```
mfherbst: Why do we need the two versions ?

src/response/chi0.jl
```diff
@@ -370,22 +381,23 @@ function compute_δψ!(δψ, basis::PlaneWaveBasis{T}, H, ψ, εF, ε, δHψ, ε
...
-            !res.converged && @warn("Sternheimer CG not converged", ik, n, res.n_iter,
-                                    res.tol, res.residual_norm)
+        !res.converged && @warn("Sternheimer CG not converged", res.tol, res.residual_norms)
 
-            δψk[:, n] .+= res.δψkn
-            push!(residual_norms[ik], res.residual_norm)
-            push!(n_iter[ik], res.n_iter)
-            converged = converged && res.converged
-        end
+        δψk .+= res.δψkn
```
mfherbst: If the `δψkn` is all bands than the variable name should not have an `n` suffix.

src/response/chi0.jl
```diff
@@ -109,12 +109,9 @@ precondprep!(P::FunctionPreconditioner, ::Any) = P
 # included).
 @timing function sternheimer_solver(Hk, ψk, ε, rhs;
```
mfherbst: Here the in-place business and the "multiple rhs at once" business increases the complexity quite a bit. It's worth doing, but a few more comments and adapted changes of the variable names could help. We may also need another iteration to see if there is maybe some more clever linear algebra constructs we can follow to hide some of the complexity better.

src/response/chi0.jl
mfherbst: I see, this is why you need the `active` flag.

Essentially the underlying structure is that you have a stack of operators `H - εn I` and a stack of vectors `ϕn` and as you drop some of the `ϕn` due to locking you want to equivalently drop the respective `H - εn`, i.e. a [Sylvester equation](https://en.wikipedia.org/wiki/Sylvester_equation) $AX + XB = C$ with $A$ the Hamiltonian, $B$ equal to minus the diagonal matrix of eigenvalues, $X$ the stack of wavefunctions, $C$ the stack of RHS. I would suggest to make this more explicit.

src/response/cg.jl
abussy: My answer here covers multiple, related comments. I chose not to use a `LinearMap` because the operator `A!` does not simply behave like a `mul!(c, A, x)`. The main reason is the `active` range imposed by the locking mechanism cannot simply be passed to `x`, but also to `A`.

This is also the reason why `A!` arguments are inspected at the start of the function. There should be a fallback for simple calls to `cg!`, without imposing a burden on the caller to implement an `active` range (unless performance really matters).

If that makes things more readable, I think one could implement a small helper function along the lines of
```
mul!(c, A, x) = A!(c, x)
```

src/response/cg.jl
abussy: The comments refer to an earlier thread with @niklasschmitz, where there was a worry that the compiler would not specialize on the return type of `my_columnwise_dots`, if the function is passed as a kwarg.

I'll look into having an in-place version, which would remove the need for the unclear comment.

src/response/cg.jl
abussy: From my observations, this is not always the case here. However, this is nonetheless the assumption I make, because having non-contiguous masks is not GPU-compatible. See a few lines below:

```
        # Lock columns that are already converged. Because we can only take views with
        # contiguous ranges in GPU arrays, we lock the first and last columns.
        active = 1:size(b, 2)
        if op_supports_active
            locked_lb = findfirst(!, converged_cols)
            locked_ub = findlast(!, converged_cols)
            active = locked_lb : locked_ub
        end
```

src/response/chi0.jl
abussy: Ideally, we would only use the in-place version. However, this is not obviously done in

```
    @timing function R_ldiv!(x, y)
        R!(x, precon \ R(y))
    end
```
Which is called many times over. A allocation light  `R` version is therefore useful.

mfherbst: Why not
```julia
    @timing function R_ldiv!(x, y)
        R!(x, ldiv!(precon, R!(similar(y), y))
    end
```
(which of course needs some comment what actually happens otherwise people will be confused)

mfherbst: or even
```julia
    @timing function R_ldiv!(x, y)
        R!(x, ldiv!(precon, R!(x, y))
    end
```
as we are able to use `x` any way we please.
 
**Edit:** Just noted this does not work because then the last `R!` gets twice the same memory.

src/response/cg.jl
mfherbst: I think it's fine to make this assumption as it helps with the GPU and should usually be (at least roughly) true nevertheless. Just add a comment that you make this assumption/restriction in the code and why.

src/response/chi0.jl
abussy: I am not convinced this is better or clearer than having two implementations for `R` and `R!`.

src/response/cg.jl
abussy: Here I am not convinced that an in-place function brings much, especially because there are instances where this would be very awkward. Generally, any time there is CPU-GPU data transfer. e.g.:
```
residual_norms .= to_cpu(my_columnwise_norms(r))
```

Additionally, cost of allocation is small, because array are of length `n_columns`, which is always << `n_Gvectors`.

src/response/chi0.jl
mfherbst: Another idea could be to just have the in-place version and then additionally:
```julia
R(ϕ) = R!(similar(ϕ), ϕ)
```

src/response/chi0.jl
```diff
@@ -152,47 +163,51 @@ precondprep!(P::FunctionPreconditioner, ::Any) = P
     # ψk_extra are not converged but have been Rayleigh-Ritzed (they are NOT
     # eigenvectors of H) so H(ψk_extra) = ψk_extra' (Hk-ε) ψk_extra should be a
     # real diagonal matrix.
-    H(ϕ) = Hk * ϕ - ε * ϕ
-    ψk_exHψk_ex = Diagonal(real.(εk_extra .- ε))
+    inv_ψk_exHψk_ex = 1 ./(real.(εk_extra) .- ε')
+
+    # Apply H - εI to a given vector as a Sylvester equation:
+    # c = Ax - xB, where A is the Hamilitonian, and B a diagonal matrix of eigenvalues.
```
mfherbst: what is `c` ?

src/response/chi0.jl
```diff
@@ -131,14 +130,26 @@ precondprep!(P::FunctionPreconditioner, ::Any) = P
...
+        mul!(Rϕ, ψk, ψk' * ϕ)
+        mul!(Rϕ, ψk_extra, ψk_extra' * ϕ, -1, -1)
+        Rϕ .+= ϕ
+    end
 
     # We put things into the form
-    # δψkn = ψk_extra * αkn + δψknᴿ ∈ Ran(Q)
-    # where δψknᴿ ∈ Ran(R).
+    # δψk = ψk_extra * αk + δψkᴿ ∈ Ran(Q)
+    # where δψkᴿ ∈ Ran(R).
```
mfherbst: (A) Ok this is actually wrong without the `n`s (because than it's a matrix and not a vector, so not in the vector space `Ran(Q)`. It may actually not be worth fixing this in the comments (as this may complicate the presentation). Maybe it's easier to actually present everything in the original formulation and then add a paragraph where we add that we solve it in the Sylvester-equation sense. That may be less confusing.

src/response/chi0.jl
```diff
@@ -152,47 +163,51 @@ precondprep!(P::FunctionPreconditioner, ::Any) = P
...
     # ----------------------------
-    # writing αkn as a function of δψknᴿ, we get that δψknᴿ
+    # writing αk as a function of δψkᴿ, we get that δψkᴿ
     # solves the system (in Ran(1-P_computed))
     #
-    # R * (H - ε) * (1 - M * (H - ε)) * R * δψknᴿ = R * (1 - M) * b
+    # R * (H - ε) * (1 - M * (H - ε)) * R * δψkᴿ = R * (1 - M) * b
     #
     # where M = ψk_extra * (ψk_extra' (H-ε) ψk_extra)^{-1} * ψk_extra'
     # is defined above and b is the projection of -rhs onto Ran(Q).
```
mfherbst, line 181: This also needs some adaptation of the comments, depending on how we will deal with (A).

src/response/chi0.jl
```diff
@@ -152,47 +163,51 @@ precondprep!(P::FunctionPreconditioner, ::Any) = P
...
     #
     b = -Q(rhs)
-    bb = R(b -  Hψk_extra * (ψk_exHψk_ex \ ψk_extra'b))
-    @timing function RAR(ϕ)
-        Rϕ = R(ϕ)
+    bb = R(b -  Hψk_extra * (ψk_extra'b .* inv_ψk_exHψk_ex))
```
mfherbst: This is correct in computation, but obfuscates the maths (as one would apply the inverse to the result of ψk_extra'b). The issue here is that the right object for `εk_extra - ε` now becomes a stack of diagonal matrices, one per `ε`. Maybe we could make this more explicit ?

src/response/chi0.jl
```diff
@@ -152,47 +163,51 @@ precondprep!(P::FunctionPreconditioner, ::Any) = P
...
+    @timing function RAR!(RARϕ, ϕ; active=1:size(ϕ, 2))
```
mfherbst: I'm still a little divided on this interface. I think for the CG it would make it simpler if one made a small custom type of this object, where one implements a `mul_masked!` function (similar to the the `mul_approximate` function and the dielectric operator). This function we could then use in the CG and provide appropriate fallbacks for the case where one has a plain matrix (where masking is also possible) or a general function (where masking is in general not possible). I think this makes this a little cleaner.

What do you think ?

src/response/hessian.jl
```diff
@@ -155,21 +157,21 @@ that is return δψ where (Ω+K) δψ = -δHextψ.
...
         δψy = unsafe_unpack(y)
         # real(dot) here because we work in R^2N rather than C^N
-        weighted_ksum(basis, [real(dot(δψx[ik], δψy[ik])) for ik in 1:length(basis.kpoints)])
+        [weighted_ksum(basis, [real(dot(δψx[ik], δψy[ik])) for ik in 1:length(basis.kpoints)])]
     end
-    res = cg(J, -δHextψ_pack; precon=FunctionPreconditioner(f_ldiv!), proj, tol,
-             callback=info -> callback(merge(info, (; basis=basis))), my_dot=weighted_dot)
-    (; δψ=unpack(res.x), res.converged, res.tol, res.residual_norm,
+    res = cg(J, -δHextψ_pack; precon=FunctionPreconditioner(f_ldiv!), proj!, tol=tol,
+             callback=info -> callback(merge(info, (; basis=basis))), my_columnwise_dots=weighted_dots)
```
mfherbst: That's a bit long, I'd suggest to break.

src/response/hessian.jl
```diff
@@ -100,7 +100,9 @@ function (cb::ResponseCallback)(info)
     runtime_ns = current_time - cb.prev_time[]
     cb.prev_time[] = current_time
 
-    resnorm = @sprintf "%20.2f" log10(info.residual_norm)
+    # expect a scalar from solve_ΩplusK
+    @assert length(info.residual_norms) == 1
+    resnorm = @sprintf "%20.2f" log10(info.residual_norms[1])
```
mfherbst: For such things `only` is quite useful. I'd just do `log10(only(info.residual_norms))`, which will also break loudly if there happens to be more than one entry in the array.

src/response/cg.jl
```diff
@@ -6,65 +6,125 @@ function default_cg_print(info)
     @printf("%3d\t%1.2e\n", info.n_iter, info.residual_norm)
 end
 
+"""
+Matrix-vector product used in the CG solver. The CG locking mechanism is such that the first and last
+converged columns of x in Ax = b are ignored by considering a contiguous active set of columns,defined
+by first(!converged_columns):last(!converged_columns). The mask must be a contiguous range for GPU
+compatibility. This is the default implementation for matrix-like A. Custom opertators must implement
+their own method for mul_masked!.
```
mfherbst: Can you break around 95 chars ?

src/response/cg.jl
```diff
@@ -6,65 +6,125 @@ function default_cg_print(info)
...
+"""
+@timing mul_masked!(Ax, A, x; mask::UnitRange) = @views mul!(Ax, A, x[:, mask])
```
mfherbst: Should it not also be `mul!(Ax[: ,mask], A, x[:, mask])` ?
I would suggest to have a small test in the `test/cg.jl` just for matrices that covers this (and much of the locking logic in the CG).

Also, should we type-guard the `A`, i.e. say `A::AbstractArray`. Pro is this forces people to implement this function for custom types, which I guess is better than a silent wrong result ? Cons is one could also argue that most custom types that are just linear operators will implement a `mul!` and then this default should indeed work fine, so such a type constraint is too constraining. I'm ok with both.

src/response/cg.jl
```diff
@@ -6,65 +6,125 @@ function default_cg_print(info)
...
 
     # save one matrix-vector product
     if !iszero(x)
-        mul!(c, A, x)
+        mul_masked!(c, A, x; mask=1:size(x, 2))
         r .-= c
     end
     ldiv!(c, precon, r)
-    γ = my_dot(r, c)
+    γ = similar(b, T, size(b, 2)) # explicit type for stability
```
mfherbst: Two spaces before `#`, please :)
... at a couple of places in this file.

src/response/cg.jl
```diff
@@ -6,65 +6,125 @@ function default_cg_print(info)
...
+        proj!(r, proj_buffer) # r .= proj(r .- c .* α')
+        residual_norms .= to_cpu(my_columnwise_norms(r))
 
-        # apply preconditioner and prepare next iteration
-        ldiv!(c, precon, r)
-        γ_prev, γ = γ, my_dot(r, c)
-        β = γ / γ_prev
-        p .= proj(c .+ β .* p)
+        # apply preconditioner and prepare next iteration. Preconditioner applied to full arrays,
+        # because the FunctionPreconditioner type makes it hard to change the active set (cheap enough).
```
mfherbst: Lines long.

Please mark as a TODO, I can imagine we may want to change this (especially if we develop more sophisticated preconditioners)

src/response/cg.jl
```diff
@@ -6,65 +6,125 @@ function default_cg_print(info)
...
+        proj_buffer .= c .+ p .* β'
+        proj!(p, proj_buffer) # p .= proj(c .+ p .* β')/
     end
-    info = (; x, converged, tol, residual_norm, n_iter, maxiter, stage=:finalize)
+    info = (; x=full_x, converged, tol, residual_norms=full_residuals, n_iter, maxiter, stage=:finalize)
     callback(info)
     info
 end
-cg!(x::AbstractVector, A::AbstractMatrix, b::AbstractVector; kwargs...) = cg!(x, LinearMap(A), b; kwargs...)
-cg(A::LinearMap, b::AbstractVector; kwargs...) = cg!(zero(b), A, b; kwargs...)
```
mfherbst, line 69: Do we still need `LinearMap` objects anywhere in the code. If no, we could perhaps remove this dependency.

src/response/chi0.jl
```diff
@@ -103,21 +103,29 @@ LinearAlgebra.ldiv!(y::T, P::FunctionPreconditioner, x) where {T} = P.preconditi
 LinearAlgebra.ldiv!(P::FunctionPreconditioner, x) = (x .= P.precondition!(similar(x), x))
 precondprep!(P::FunctionPreconditioner, ::Any) = P
 
-# Solves (1-P) (H-ε) (1-P) δψn = - (1-P) rhs
+struct MaskedOperator
+    masked_product!::Function
+end
```
mfherbst: Please use the construction of the `FunctionPreconditioner` (i.e. with the type variable) as `Function` is an abstract type, so using `masked_product!` becomes a generic function call.

src/response/chi0.jl
```diff
@@ -103,21 +103,29 @@ LinearAlgebra.ldiv!(y::T, P::FunctionPreconditioner, x) where {T} = P.preconditi
...
+@timing mul_masked!(Ax, M::MaskedOperator, x; mask) = M.masked_product!(Ax, x, mask)
+
+# Solves (1-P) (H-ε) (1-P) δψ = - (1-P) rhs
 # where 1-P is the projector on the orthogonal of ψk
+# The solver simultaneously solves for multiple right-hand sides, i.e.:
+# (1-P) (H-ε) (1-P) δψ[:, i] = - (1-P) rhs[:, i] for all columns i.
 # /!\ It is assumed (and not checked) that ψk'Hk*ψk = Diagonal(εk) (extra states
 # included).
-@timing function sternheimer_solver(Hk, ψk, ε, rhs;
```
mfherbst, line 110: Why remove the timing ?

src/response/chi0.jl
```diff
@@ -103,21 +103,29 @@ LinearAlgebra.ldiv!(y::T, P::FunctionPreconditioner, x) where {T} = P.preconditi
...
+                                    ψk_extra=zeros_like(ψk), εk_extra=zeros_like(ε),
+                                    Hψk_extra=zeros_like(ψk), tol=1e-9,
                                     miniter=1, maxiter=100)
-    # TODO This whole projector business allocates a lot, which is rather slow.
-    #      We should optimise this a bit to avoid allocations where possible.
-
     basis  = Hk.basis
     kpoint = Hk.kpoint
 
+    # Note: to maintain clearer mathematical formulas, all commements assue the problem
```
mfherbst: assue -> assume

src/response/chi0.jl
```diff
@@ -103,21 +103,29 @@ LinearAlgebra.ldiv!(y::T, P::FunctionPreconditioner, x) where {T} = P.preconditi
...
+    #       is solved band by band, with ψkn = ψk[:, n]. In practice, the problem is solved
+    #       for all bands simultaneously for performance reasosn.
```
mfherbst: typo.

src/response/chi0.jl
```diff
@@ -103,21 +103,29 @@ LinearAlgebra.ldiv!(y::T, P::FunctionPreconditioner, x) where {T} = P.preconditi
...
-# Solves (1-P) (H-ε) (1-P) δψn = - (1-P) rhs
+struct MaskedOperator
+    masked_product!::Function
+end
+@timing mul_masked!(Ax, M::MaskedOperator, x; mask) = M.masked_product!(Ax, x, mask)
+
+# Solves (1-P) (H-ε) (1-P) δψ = - (1-P) rhs
 # where 1-P is the projector on the orthogonal of ψk
+# The solver simultaneously solves for multiple right-hand sides, i.e.:
+# (1-P) (H-ε) (1-P) δψ[:, i] = - (1-P) rhs[:, i] for all columns i.
```
mfherbst: Also `ε[i]`, right ? For bands we typically use indices `n` and `m` instead of `i`. Would be good to stick to this here as well.

src/response/chi0.jl
```diff
@@ -131,8 +139,15 @@ precondprep!(P::FunctionPreconditioner, ::Any) = P
...
+
     # R = 1-P_computed is the projector onto the orthogonal of computed states
-    R(ϕ) = ϕ - P_computed(ϕ)
+    # Implement allocation light, in-place version for performance
+    function R!(Rϕ, ϕ)
+        mul!(Rϕ, ψk, ψk' * ϕ) # P(ϕ)
+        mul!(Rϕ, ψk_extra, ψk_extra' * ϕ, -1, -1) # -P(ϕ) - P_extra(ϕ)
+        Rϕ .+= ϕ # R = ϕ -P(ϕ) - P_extra(ϕ)
+    end
+    R(ϕ) = R!(similar(ϕ), ϕ) #TODO: ok, is Rϕ actually returned?
```
mfherbst: Nice ... and yes, this should work.

src/response/chi0.jl
```diff
@@ -131,8 +139,15 @@ precondprep!(P::FunctionPreconditioner, ::Any) = P
     P_computed(ϕ) = P(ϕ) + P_extra(ϕ)
     # Q = 1-P is the projector onto the orthogonal of converged states
     Q(ϕ) = ϕ - P(ϕ)
+
     # R = 1-P_computed is the projector onto the orthogonal of computed states
-    R(ϕ) = ϕ - P_computed(ϕ)
+    # Implement allocation light, in-place version for performance
+    function R!(Rϕ, ϕ)
+        mul!(Rϕ, ψk, ψk' * ϕ) # P(ϕ)
```
mfherbst: Same point about two spaces before `#` :)

src/response/chi0.jl
```diff
@@ -149,50 +164,67 @@ precondprep!(P::FunctionPreconditioner, ::Any) = P
...
+    #given active range of columns of x, define by a mask.
+    function H(ϕ; mask=1:size(ϕ, 2))
+        Hϕ = Hk * ϕ
+        mul!(Hϕ, ϕ, Diagonal(ε[mask]), -1, 1) # Hϕ - ϕ * Diagonal(ε)
+    end
 
     # 1) solve for δψknᴿ
     # ----------------------------
-    # writing αkn as a function of δψknᴿ, we get that δψknᴿ
+    # writing αk as a function of δψkᴿ, we get that δψkᴿ
```
mfherbst: This looks like it's been missed

src/response/chi0.jl
```diff
@@ -149,50 +164,67 @@ precondprep!(P::FunctionPreconditioner, ::Any) = P
...
+
+    # Implementation of: R * (H - ε) * (1 - M * (H - ε)) * R * ϕ
+    @views function RAR!(RARϕ, ϕ, mask)
+        R!(RARϕ, ϕ[:, mask])
+        HRϕ = H(RARϕ; mask)
         # Schur complement of (1-P) (H-ε) (1-P)
         # with the splitting Ran(1-P) = Ran(P_extra) ⊕ Ran(R)
-        R(H(Rϕ) - Hψk_extra * (ψk_exHψk_ex \ Hψk_extra'Rϕ))
+        R!(RARϕ, mul!(HRϕ, Hψk_extra, inv_ψk_exHψk_ex[:, mask] .* Hψk_extra'RARϕ, -1, 1))
     end
```
mfherbst: Very nice, But I think for clarity I would do:
```suggestion
    @views function RAR!(RARϕ, ϕ, mask)
        Rϕ = R!(RARϕ, ϕ[:, mask])  # Use RARϕ as temporary memory
        HRϕ = H(Rϕ; mask)
        # Schur complement of (1-P) (H-ε) (1-P)
        # with the splitting Ran(1-P) = Ran(P_extra) ⊕ Ran(R)
        # HRϕ[:, n] -= Hψk_extra[:, n] * [1/(εk_extra_n - ε_l)] .* Hψk_extra[:, n]' Rϕ[:, l]
        mul!(HRϕ, Hψk_extra, inv_ψk_exHψk_ex[:, mask] .* Hψk_extra'Rϕ, -1, 1)
        R!(RARϕ, HRϕ)
    end
```

src/response/chi0.jl
```diff
@@ -103,21 +103,29 @@ LinearAlgebra.ldiv!(y::T, P::FunctionPreconditioner, x) where {T} = P.preconditi
 LinearAlgebra.ldiv!(P::FunctionPreconditioner, x) = (x .= P.precondition!(similar(x), x))
 precondprep!(P::FunctionPreconditioner, ::Any) = P
 
-# Solves (1-P) (H-ε) (1-P) δψn = - (1-P) rhs
+struct MaskedOperator
+    masked_product!::Function
+end
+@timing mul_masked!(Ax, M::MaskedOperator, x; mask) = M.masked_product!(Ax, x, mask)
```
mfherbst: `mask` also a `kwarg` in the `masked_product!` ?

src/response/chi0.jl
```diff
@@ -149,50 +164,67 @@ precondprep!(P::FunctionPreconditioner, ::Any) = P
...
     end
-    J = LinearMap{eltype(ψk)}(RAR, size(Hk, 1))
-    cg_res = cg(J, bb; precon=FunctionPreconditioner(R_ldiv!), tol, proj=R,
+    cg_res = cg(A, bb; precon=FunctionPreconditioner(R_ldiv!), tol, proj! =R!,
                 callback=info -> callback(merge(info, (; basis, kpoint))),
                 miniter, maxiter)
-    δψknᴿ = cg_res.x
+    δψkᴿ = cg_res.x
 
     # 2) solve for αkn now that we know δψknᴿ
```
mfherbst: Add:
```
        # We again do this for all right-hand sides at once, such that
        # αk[m, l] = 1/(εk_extra_n - ε_l) δnm (ψk_extra[:, n]' * (b[:, l] - (H - ε_l) * δψkᴿ[:, l])
```
also remove the `n` in the line below.

src/response/chi0.jl
```diff
@@ -370,22 +403,24 @@ function compute_δψ!(δψ, basis::PlaneWaveBasis{T}, H, ψ, εF, ε, δHψ, ε
...
+        # Array operations for GPU efficiency
+        dot_prods = ψk' * δHψ[ik]
+        dot_prods .*= to_device(basis.architecture, α)
+        mul!(δψk, ψk, dot_prods, 1, 1)
 
-            # Sternheimer contribution
-            res = sternheimer_solver(Hk, ψk, εk_minus_q[n], δHψ[ik][:, n]; ψk_extra,
-                                     εk_extra, Hψk_extra, tol=tolk_minus_q[n], kwargs_sternheimer...)
+        # Sternheimer contribution, for all columns of δψk at once.
+        res = sternheimer_solver(Hk, ψk, to_device(basis.architecture, εk_minus_q), δHψ[ik]; ψk_extra,
```
mfherbst: Can you put the `to_device` into a separate line (otherwise one may miss it too easily)

src/response/cg.jl
abussy: > Should it not also be mul!(Ax[: ,mask], A, x[:, mask]) ?

It it more logical, yes. I'll make it work that way (will lead to small modifications in the `RAR!` implementation).

> Also, should we type-guard the A, i.e. say A::AbstractArray. 

There I specifically made the choice not to type-guard the default `mul_masked!`. My reasoning is that the CG solver is mostly called with `LinearMap` throughout the code, and that most developers/users would have that habit. That being said, we could easily add 2 defaults: one for `AbstractArray{T}` and one for `LinearMap{T}`.

mfherbst: or one for `Union{AbstractArray{T}, LinearMap{T}` :smile:

src/response/chi0.jl
abussy: That was accidental, great that you spotted it. Reason is: `@timing` sometimes messes up the stack trace and makes debugging harder.

src/response/cg.jl
abussy: I think it's only used when setting up the CG solver, I kept it around for continuity. One could get rid of it and simplify the dependency tree (great). To me, the only down side is that a `mul_masked!` method must be implemented whenever the CG solver is called with something that is not a simple matrix-vector product, e.g. `ΩpK(x)` in `respons/hessiang.jl`. I'd be in favor of removing it, though.

src/response/chi0.jl
mfherbst: You can disable timings in your preferences for debugging, see the box above https://docs.dftk.org/stable/tricks/parallelization/#Rough-timing-estimates.

src/response/chi0.jl
abussy: Oops, that was supposed to be gone :sweat_smile:

src/response/cg.jl
mfherbst: Let's keep it for now. But it's on the list of things to remove.

PR #1241, opened 2026-01-23
Consistent used of basis.comm_kpts

abussy:
Consistently use the MPI communicator from the `PlaneWaveBasis` rather than `MPI.COMM_WORLD`.

Context: I was trying to dispatch many small DFTK calculations over MPI subgroups, but didn't get any speedup. The reason is `MPI.Bcast` calls with the `MPI.COMM_WORLD` in the SCF, and `MPI.Barrier` in JLD2 data dumping.

### DIFF
--- a/ext/DFTKJLD2Ext.jl
+++ b/ext/DFTKJLD2Ext.jl
@@ -66,7 +66,7 @@ function DFTK.load_scfres(::Val{:jld2}, filename::AbstractString, basis=nothing;
     DFTK.mpi_barrier(comm)
     scfres
 end
-function load_scfres_jld2(jld, basis; skip_hamiltonian, strict, comm=MPI.COMM_WORLD)
+function load_scfres_jld2(jld, basis; skip_hamiltonian, strict, comm)
     consistent_kpts = true
     if isnothing(basis)
         basis = load_basis(jld; comm)


ext/DFTKJLD2Ext.jl
```diff
@@ -25,7 +25,7 @@ function save_jld2(to_dict_function!, file::AbstractString, scfres::NamedTuple;
         dummy = Dict{String,Any}()
         to_dict_function!(dummy, scfres; save_ψ)
     end
-    MPI.Barrier(MPI.COMM_WORLD)
+    MPI.Barrier(scfres.basis.comm_kpts)
```
antoine-levitt: wrap this for consistency

ext/DFTKJLD2Ext.jl
```diff
@@ -47,29 +47,30 @@ function load_basis(jld)
     else
         basis_args = nothing
     end
-    basis_args = MPI.bcast(basis_args, 0, MPI.COMM_WORLD)
-    PlaneWaveBasis(basis_args..., MPI.COMM_WORLD, DFTK.CPU())
+    basis_args = MPI.bcast(basis_args, 0, comm)
```
antoine-levitt: also this

src/common/mpi.jl
```diff
@@ -16,6 +17,10 @@ mpi_max!( arr, comm::MPI.Comm) = MPI.Allreduce!(arr, max, comm)
 mpi_mean( arr, comm::MPI.Comm) = mpi_sum(arr, comm) ./ mpi_nprocs(comm)
 mpi_mean!(arr, comm::MPI.Comm) = (mpi_sum!(arr, comm); arr ./= mpi_nprocs(comm))
 
+mpi_bcast(arr, comm::MPI.Comm; root=0) = MPI.bcast(arr, root, comm)
+mpi_bcast(arr, root, comm::MPI.Comm) = MPI.bcast(arr, root, comm)
```
mfherbst: To avoid mixing things up, maybe type-annotate `root` as `Int` ?

src/common/mpi.jl
```diff
@@ -4,9 +4,10 @@ import MPI
 """
 Number of processors used in MPI. Can be called without ensuring initialization.
 """
-mpi_nprocs(comm=MPI.COMM_WORLD) = (MPI.Init(); MPI.Comm_size(comm))
-mpi_master(comm=MPI.COMM_WORLD) = (MPI.Init(); MPI.Comm_rank(comm) == 0)
+mpi_nprocs(comm) = (MPI.Init(); MPI.Comm_size(comm))
+mpi_master(comm) = (MPI.Init(); MPI.Comm_rank(comm) == 0)
```
mfherbst: I am not so sure about this change. It's convenient especially for printing guards to do just `if mpi_master()`. Do you think this will cause a lot of people to fall into a trap ?

Maybe you're right for the `mpi_nprocs` though 🤔

src/response/hessian.jl
```diff
@@ -76,7 +76,7 @@ function ResponseCallback()
     ResponseCallback(Ref(zero(UInt64)))
 end
 function (cb::ResponseCallback)(info)
-    mpi_master() || return info  # Only print on master
+    mpi_master(info.comm) && return info  # Only print on master
```
mfherbst: Interesting ... this is not necessarily what we want, but maybe the best we can do (certainly better than what was there before).

The case I have in mind is one where master is not in the `info.comm`. In that case do we really want to print nothing ?

src/common/mpi.jl
mfherbst: Having read on, I see why you make the change for `mpi_master`, ok, I agree ... but this is a breaking change (as we decided to export `mpi_master`). I'd rather not have this breaking change now and instead simply deprecate the usage of `mpi_master` and `mpi_nprocs` without a communicator. Look at the `@deprecate` macro.

src/response/hessian.jl
```diff
@@ -189,7 +189,7 @@ function (cb::OmegaPlusKDefaultCallback)(info)
         avgCG = mpi_mean(avgCG, first(info.Axinfos).basis.comm_kpts)
     end
 
-    !mpi_master() && return info  # Rest is printing => only do on master
+    !mpi_master(info.comm) && return info  # Rest is printing => only do on master
```
mfherbst: I'd explicitly use the `basis.comm_kpts` here.

src/response/inexact_gmres.jl
```diff
@@ -63,7 +64,8 @@ Parameters specific to inexact GMRES:
 @timing function inexact_gmres!(x, A, b::AbstractVector{T};
                                 precon=I, maxiter=100, krylovdim=20, tol=1e-6, s=1.0,
                                 callback=identity,
-                                orth::Orthogonalizer=KrylovKit.ModifiedGramSchmidt2()) where {T}
+                                orth::Orthogonalizer=KrylovKit.ModifiedGramSchmidt2(),
+                                comm::MPI.Comm=MPI.COMM_WORLD) where {T}
```
mfherbst: It's weird to have the `comm` here. It's not really used for anything but the callback.

src/response/hessian.jl
```diff
@@ -162,7 +162,7 @@ that is return δψ where (Ω+K) δψ = -δHextψ.
         weighted_ksum(basis, [real(dot(δψx[ik], δψy[ik])) for ik in 1:length(basis.kpoints)])
     end
     res = cg(J, -δHextψ_pack; precon=FunctionPreconditioner(f_ldiv!), proj, tol,
-             callback, my_dot=weighted_dot)
+             callback=info -> callback(merge(info, (; comm=basis.comm_kpts))), my_dot=weighted_dot)
```
mfherbst: That's one way. But perhaps it makes sense to give the `info` more metadata, e.g. the `basis` and extract the `comm` from there in the callback.

src/response/hessian.jl
```diff
@@ -312,7 +313,8 @@ Input parameters:
     callback_inner(info) = callback(merge(info, (; runtime_ns=time_ns() - start_ns)))
     info_gmres = inexact_gmres(ε_adj, vec(δρ0);
                                tol, precon, krylovdim, maxiter, s,
-                               callback=callback_inner, kwargs...)
+                               callback=callback_inner, comm=ham.basis.comm_kpts,
+                               kwargs...)
```
mfherbst: I'm not sure here. My standing is that the `gmres` should know nothing about MPI and the interface should be similar to the `cg`. Having the `comm` appear explicitly is weird. And actually you don't need it if you use the same convention as `cg`, namely that the default callback just prints on every processor.

src/scf/potential_mixing.jl
```diff
@@ -244,7 +245,7 @@ Simple SCF algorithm using potential mixing. Parameters are largely the same as
         info_next = info
         while n_backtrack ≤ max_backtracks
             diagtol = determine_diagtol(diagtolalg, info_next)
-            mpi_master() && @debug "Iteration $n_iter linesearch step $n_backtrack   α=$α diagtol=$diagtol"
+            mpi_master(basis.comm_kpts) && @debug "Iteration $n_iter linesearch step $n_backtrack   α=$α diagtol=$diagtol"
```
mfherbst: This is too long now. Use `if` block

src/response/inexact_gmres.jl
```diff
@@ -12,7 +13,7 @@ define a method for this function to provide faster approximate versions for `A*
 mul_approximate(A, X; rtol=0.0) = (; Ax=A * X, info=(; rtol))
 
 function default_gmres_print(info)
-    !mpi_master() && return info  # Rest is printing => only do on master
+    !mpi_master(info.comm) && return info  # Rest is printing => only do on master
```
mfherbst: Remove and make a comment that the default callback just prints on every processor, same for the default `cg` callback.

test/timings
```diff
@@ -0,0 +1,2 @@
+Old timings: 537 seconds
```
mfherbst: unintentional commit ?

src/standard_models.jl.save
```diff
@@ -0,0 +1,182 @@
+# High-level convenience functions to make standard models
```
mfherbst: Unintentional commit ?

src/common/mpi.jl
```diff
@@ -16,6 +17,10 @@ mpi_max!( arr, comm::MPI.Comm) = MPI.Allreduce!(arr, max, comm)
 mpi_mean( arr, comm::MPI.Comm) = mpi_sum(arr, comm) ./ mpi_nprocs(comm)
 mpi_mean!(arr, comm::MPI.Comm) = (mpi_sum!(arr, comm); arr ./= mpi_nprocs(comm))
 
+mpi_bcast(arr, comm::MPI.Comm; root=0) = MPI.bcast(arr, root, comm)
+mpi_bcast(arr, root, comm::MPI.Comm) = MPI.bcast(arr, root, comm)
+mpi_barrier(comm::MPI.Comm) = MPI.Barrier(comm)
```
mfherbst, line 26: Is this really useful to have ? I mean what do you gain ?

src/response/hessian.jl
abussy: AFAIK, all MPI communicators come with a master rank, i.e. a rank such that `MPI.Comm_rank(comm) == 0`.

In case of sub-communicators, the ranks within the sub-communicators are relabeled starting from 0. For example, if `MPI.COMM_WOLRD` has 4 ranks, and we split it into 2 sub-communicators of 2 ranks:

```
using MPI                                                                                            
                                                                                                     
MPI.Init()                                                                                           
                                                                                                     
comm = MPI.COMM_WORLD                                                                                
rank = MPI.Comm_rank(comm)                                                                           
size = MPI.Comm_size(comm)
@assert(size == 4)                             
                                                                                                     
# split into 2 groups                                                                                
color = rank < size ÷ 2 ? 0 : 1                                                                      
                                                                                                     
# ranks within subgroup ordered according to key value (but starting from 0)                         
key = 4 - rank                                                                                       
                                                                                                     
subcomm = MPI.Comm_split(comm, color, key)                                                           
                                                                                                     
subrank = MPI.Comm_rank(subcomm)                                                                     
subsize = MPI.Comm_size(subcomm)                                                                     
                                                                                                     
println("Global rank: $rank, Subcomm # $color with subrank $subrank, key: $key")                     
                                                                                                     
MPI.Finalize()
```

We get the output:
```
Global rank: 0, Subcomm # 0 with subrank 1, key: 4
Global rank: 1, Subcomm # 0 with subrank 0, key: 3
Global rank: 2, Subcomm # 1 with subrank 1, key: 2
Global rank: 3, Subcomm # 1 with subrank 0, key: 1
```

So there will be logging from 1 rank from each subgroup. The danger of not passing the communicator explicitly is if the master rank of `MPI.COMM_WORLD` is not within the sub-communicator, in which case there would be no logging.

src/common/mpi.jl
abussy: Not much is gained. It's about wrapping MPI calls consistently, as suggested by @antoine-levitt.

src/response/hessian.jl
abussy: Problem is that the basis is not necessarily available in `info`. It only is  if `haskey(info, :Axinfos)`, as tested in the code block above. One could always assume the basis is available, and pass it when setting up the callback, I guess.

src/response/hessian.jl
abussy: I agree that passing the `comm` explicitly to `inexact_gmres` is less than ideal. I am happy to change the default callback to print on every rank.

However, `inexact_gmres` is mostly (always ?) called with the `OmegaPlusKDefaultCallback`, which we do not want to print on each rank. Somehow, the communicator must be passed to `info`. This comes back to a previous comment by me:

> Problem is that the basis is not necessarily available in info. It only is if haskey(info, :Axinfos), as tested in the code block above. One could always assume the basis is available, and pass it when setting up the callback, I guess.

Shall we always assume `haskey(info, :Axinfos)`, and use the communicator from the basis?

src/standard_models.jl.save
abussy: Must be, my bad.

src/common/mpi.jl
mfherbst: I see. Does not matter so much, but for sure don't export it, so we can change later if we want.

src/response/hessian.jl
mfherbst: Ok, all good then !

src/response/hessian.jl
abussy: I went with passing `basis` to the callback, and extracting the communicator from it. If `basis` is not available, it defaults to `MPI.COMM_WORLD`.

src/common/mpi.jl
```diff
@@ -4,9 +4,16 @@ import MPI
...
-mpi_nprocs(comm=MPI.COMM_WORLD) = (MPI.Init(); MPI.Comm_size(comm))
-mpi_master(comm=MPI.COMM_WORLD) = (MPI.Init(); MPI.Comm_rank(comm) == 0)
+mpi_nprocs(comm::MPI.Comm) = (MPI.Init(); MPI.Comm_size(comm))
+mpi_master(comm::MPI.Comm) = (MPI.Init(); MPI.Comm_rank(comm) == 0)
+mpi_nprocs() = (MPI.Init(); MPI.Comm_size(MPI.COMM_WORLD))
+mpi_master() = (MPI.Init(); MPI.Comm_rank(MPI.COMM_WORLD) == 0)
 
+# Calling MPI without explicit communcator is deprecated
+@deprecate mpi_nprocs() mpi_nprocs(comm::MPI.Comm)
+@deprecate mpi_master() mpi_master(comm::MPI.Comm)
```
mfherbst: I think the syntax would be:
```suggestion
@deprecate mpi_nprocs() mpi_nprocs(MPI.COMM_WORLD)
@deprecate mpi_master() mpi_master(MPI.COMM_WORLD)
```

src/response/inexact_gmres.jl
```diff
@@ -1,6 +1,7 @@
 using LinearMaps
 using KrylovKit: OrthonormalBasis, Orthogonalizer
 using LinearAlgebra: Givens
+using MPI
```
mfherbst: Not needed any more.

ext/DFTKJLD2Ext.jl
```diff
@@ -47,29 +48,30 @@ function load_basis(jld)
...
-    PlaneWaveBasis(basis_args..., MPI.COMM_WORLD, DFTK.CPU())
+    basis_args = DFTK.mpi_bcast(basis_args, 0, comm)
+    PlaneWaveBasis(basis_args..., comm, DFTK.CPU())
 end
 
 
-function DFTK.load_scfres(::Val{:jld2}, filename::AbstractString, basis=nothing; kwargs...)
-    if mpi_master()
+function DFTK.load_scfres(::Val{:jld2}, filename::AbstractString, basis=nothing;
+                          comm=MPI.COMM_WORLD, kwargs...)
```
mfherbst: This should use the `comm` from the basis if one is given.

ext/DFTKJLD2Ext.jl
```diff
@@ -47,29 +48,30 @@ function load_basis(jld)
...
     else
-        scfres = load_scfres_jld2(nothing, basis; kwargs...)
+        scfres = load_scfres_jld2(nothing, basis; comm, kwargs...)
     end
-    MPI.Barrier(MPI.COMM_WORLD)
+    DFTK.mpi_barrier(comm)
     scfres
 end
-function load_scfres_jld2(jld, basis; skip_hamiltonian, strict)
+function load_scfres_jld2(jld, basis; skip_hamiltonian, strict, comm=MPI.COMM_WORLD)
```
mfherbst: Use comm from the basis, no need for kwarg as the basis can be expected to have the right comm internally.

src/response/chi0.jl
```diff
@@ -182,7 +182,7 @@ precondprep!(P::FunctionPreconditioner, ::Any) = P
     end
     J = LinearMap{eltype(ψk)}(RAR, size(Hk, 1))
     cg_res = cg(J, bb; precon=FunctionPreconditioner(R_ldiv!), tol, proj=R,
-                callback=info -> callback(merge(info, (; basis, kpoint))),
+                callback=info -> callback(merge(info, (; basis, kpoint, comm=basis.comm_kpts))),
```
mfherbst: Do we still need this ?

src/response/hessian.jl
```diff
@@ -338,7 +343,8 @@ Input parameters:
                            maxiter=maxiter_sternheimer, tol=tol * factor_final,
                            bandtolalg, occupation_threshold, q, kwargs...)
     callback((; stage=:final, runtime_ns=time_ns() - start_ns,
-                Axinfos=[(; basis, tol=tol*factor_final, resfinal...)]))
+                Axinfos=[(; tol=tol*factor_final, resfinal...)],
+                basis=basis))
```
mfherbst: A bit more compact:
```suggestion
    callback((; stage=:final, runtime_ns=time_ns() - start_ns, basis,
                Axinfos=[(; tol=tol*factor_final, resfinal...)]))
```

src/response/hessian.jl
```diff
@@ -296,7 +300,8 @@ Input parameters:
                            bandtolalg, occupation_threshold,
                            q, kwargs...)  # = χ04P * δHext
         callback((; stage=:noninteracting, runtime_ns=time_ns() - start_ns,
-                    Axinfos=[(; basis, tol=tol*factor_initial, res0...)]))
+                    Axinfos=[(; tol=tol*factor_initial, res0...)],
+                    basis=basis))
```
mfherbst: more compact:
```suggestion
        callback((; stage=:noninteracting, runtime_ns=time_ns() - start_ns, basis,
                    Axinfos=[(; tol=tol*factor_initial, res0...)]))
```

mfherbst (commented):
Thanks,  I agree with this. Just a few smaller nits.

src/response/hessian.jl
mfherbst: I like your current solution.

ext/DFTKJLD2Ext.jl
abussy: But there is a good chance that `isnothing(basis) == true`, in which case a communicator must be provided. I think that defaulting to `basis.comm_kpts` if the basis is provided one level up in `DFTK.load_scfres` is enough.

mfherbst: True, but given that I would actually not assign a default to the `kwarg` and in this way force the caller to provide a communcator.

PR #1259, opened 2026-02-19
Use AD to compute xc kernels

Technici4n:
Key insight: Kernel application is just a Hessian-vector product, which is typically done with forward-over-reverse. We already write the gradient by hand, so to get the kernel we don't need to use reverse mode and we can just do a push-forward with ForwardDiff. This already just works for DftFunctionals, and this PR also implements the necessary rules for Libxc lda and gga.

In my testing this is 1-2x slower than writing the kernel by hand, but it does require less code, and will extend trivially to mgga!

TODO:
- [x] Does this work on the GPU?
- [x] Nothing is testing that the derivative of `e` (energy per unit volume) is correct.


antoine-levitt:
Yay, nice one!

Technici4n:
The `Anisotropic strain sensitivity using ForwardDiff (GPU)` test passes with this PR so this seems to work on the GPU as well (although it likely dispatches to the CPU version of Libxc). 😉

mfherbst:
> Quick benchmarking run:

Yeah ok, this is not relevant, I agree. The SCF takes one order of time longer than that.

Technici4n:
> CI does not run on the GPU, so this is likely more a typo than an indicator that it works on the GPU.

I mean that it does run on odonatum (with a real GPU ;))

src/DispatchFunctional.jl
```diff
@@ -109,31 +109,95 @@ function DftFunctionals.potential_terms(func::LibxcFunctional{:mggal}, ρ::Abstr
     (; e, Vρ, Vσ, Vτ, Vl)
 end
 
-function DftFunctionals.kernel_terms(func::LibxcFunctional{:lda}, ρ::AbstractMatrix{Float64})
+# Kernel support via automatic differentiation
+
+import ForwardDiff: Dual
```
mfherbst: Move to top of file or in forwarddiff workarounds

src/DispatchFunctional.jl
```diff
@@ -109,31 +109,95 @@ function DftFunctionals.potential_terms(func::LibxcFunctional{:mggal}, ρ::Abstr
...
+
+map_to_dual(T, x, δxs...) = map(x, δxs...) do xi, δxi...
+    Dual{T}(xi, δxi...)
+end
+combine_spins(xs...) = vcat(transpose.(xs)...)
+
+@views function DftFunctionals.potential_terms(func::LibxcFunctional{:lda},
+                                               ρ_δρ::AbstractMatrix{DT}) where {N,T,DT<:Dual{T,Float64,N}}
+    ρ = ForwardDiff.value.(ρ_δρ)
     s_ρ, n_p = size(ρ)
```
mfherbst, line 136: Call `s_ρ` `n_spins`, which is what we conventionally use for this.

src/DispatchFunctional.jl
```diff
@@ -109,31 +109,95 @@ function DftFunctionals.potential_terms(func::LibxcFunctional{:mggal}, ρ::Abstr
...
+    end
+    δVσ = ntuple(Val(N)) do n
+        δρ = ForwardDiff.partials.(ρ_δρ, n)
+        δσ = ForwardDiff.partials.(σ_δσ, n)
+        if s_σ == 1
+            terms.v2rhosigma .* δρ .+ terms.v2sigma2 .* δσ
+        else
+            combine_spins(
+                terms.v2rhosigma[1,:] .* δρ[1,:] .+ terms.v2rhosigma[4,:] .* δρ[2,:]
+                .+ terms.v2sigma2[1,:] .* δσ[1,:] .+ terms.v2sigma2[2,:] .* δσ[2,:] .+ terms.v2sigma2[3,:] .* δσ[3,:],
```
mfherbst: You can format the spacing nicer, to make it easier to see what happens.

src/DispatchFunctional.jl
```diff
@@ -109,31 +109,95 @@ function DftFunctionals.potential_terms(func::LibxcFunctional{:mggal}, ρ::Abstr
...
+    end
+    δVρ = ntuple(Val(N)) do n
+        δρ = ForwardDiff.partials.(ρ_δρ, n)
+        δσ = ForwardDiff.partials.(σ_δσ, n)
+        if s_ρ == 1
+            terms.v2rho2 .* δρ .+ terms.v2rhosigma .* δσ
+        else
+            combine_spins(
+                terms.v2rho2[1,:] .* δρ[1,:] .+ terms.v2rho2[2,:] .* δρ[2,:]
+                .+ terms.v2rhosigma[1,:] .* δσ[1,:] .+ terms.v2rhosigma[2,:] .* δσ[2,:] .+ terms.v2rhosigma[3,:] .* δσ[3,:],
```
mfherbst: Pretty sure that exceeds the usual line limitations. I would move this to another function to make sure one keeps the overview what happens in this beast.

src/terms/xc.jl
```diff
@@ -361,39 +361,36 @@ function LibxcDensities(basis, max_derivative::Integer, ρ, τ)
...
-    fac = term.scaling_factor
+    # For LDA the Kernel is known to be diagonal, so we can get away
+    # with a single push-forward (two for spin-polarized case)
     if n_spin == 1
-        Diagonal(vec(fac .* kernel))
+        f_spinless(ε) = xc_potential_real(term, basis, nothing, nothing; ρ=ρ.+ε).potential
+        δpotential = ForwardDiff.derivative(f_spinless, zero(T))
+        Diagonal(vec(δpotential))
     else
-        # Blocks in the kernel matrix mapping (ρα, ρβ) ↦ (Vα, Vβ)
```
mfherbst, line 382: Don't remove this comment. Still true.

src/DispatchFunctional.jl
```diff
@@ -109,31 +109,95 @@ function DftFunctionals.potential_terms(func::LibxcFunctional{:mggal}, ρ::Abstr
...
 
-function DftFunctionals.kernel_terms(func::LibxcFunctional{:lda}, ρ::AbstractMatrix{Float64})
+# Kernel support via automatic differentiation
+
+import ForwardDiff: Dual
+
+map_to_dual(T, x, δxs...) = map(x, δxs...) do xi, δxi...
+    Dual{T}(xi, δxi...)
+end
+combine_spins(xs...) = vcat(transpose.(xs)...)
```
mfherbst: `vcat(array...)` can lead to stack overflows. Always use `reduce(vcat, array)`.

Add a comment why these are around. The latter looks like you need it to make libxc happy, so call it  `libxc_combine_spins` ?

src/DispatchFunctional.jl
```diff
@@ -109,31 +109,95 @@ function DftFunctionals.potential_terms(func::LibxcFunctional{:mggal}, ρ::Abstr
...
 end
 
-function DftFunctionals.kernel_terms(func::LibxcFunctional{:lda}, ρ::AbstractMatrix{Float64})
+# Kernel support via automatic differentiation
+
+import ForwardDiff: Dual
+
+map_to_dual(T, x, δxs...) = map(x, δxs...) do xi, δxi...
+    Dual{T}(xi, δxi...)
+end
```
mfherbst: Why do you need to define this function? This is the same thing as the one-liner
```julia
map(Dual{T}, x, δxs...)
```

src/DispatchFunctional.jl
Technici4n: I kept it consistent with the primal function above. It's also used e.g. for `σ` where `s_σ` becomes less obvious (either 1 or 3 components). What should that then be changed to?

mfherbst: Ah, true. In that case I would actually add a comment that `s_ρ` is here the number of spin pairs (αα, αβ or ββ) and refer to the primal code.

src/DispatchFunctional.jl
Technici4n: Ah yes, nice

Technici4n:
@mfherbst I added a comment explaining the overall strategy, hopefully that clears up the spin handling. I'm not sure it can be broken up without making it less readable. It's quite verbose but also very mechanical (and thus readable) once one understands why it needs to be done.

src/DispatchFunctional.jl
```diff
@@ -177,12 +182,14 @@ end
...
+            # For both ρ spin components: one line for ∂²/∂ρ², one line for ∂²/∂ρ∂σ
             libxc_combine_spins(
-                # For all 2 ρ components: one line for ∂²/∂ρ², one line for ∂²/∂ρ∂σ
-                terms.v2rho2[1,:] .* δρ[1,:] .+ terms.v2rho2[2,:] .* δρ[2,:]
-                .+ terms.v2rhosigma[1,:] .* δσ[1,:] .+ terms.v2rhosigma[2,:] .* δσ[2,:] .+ terms.v2rhosigma[3,:] .* δσ[3,:],
-                terms.v2rho2[2,:] .* δρ[1,:] .+ terms.v2rho2[3,:] .* δρ[2,:]
-                .+ terms.v2rhosigma[4,:] .* δσ[1,:] .+ terms.v2rhosigma[5,:] .* δσ[2,:] .+ terms.v2rhosigma[6,:] .* δσ[3,:],
+                (@. terms.v2rho2[1,:]     * δρ[1,:] + terms.v2rho2[2,:]     * δρ[2,:]
+                 +  terms.v2rhosigma[1,:] * δσ[1,:] + terms.v2rhosigma[2,:] * δσ[2,:]
+                 +  terms.v2rhosigma[3,:] * δσ[3,:]),
```
Technici4n: I don't mind it, but IMO this is less readable since one of the HVP components is now on a different line. Makes it harder to see the Hessian matrix.

mfherbst: Ah that we can fix

src/DispatchFunctional.jl
```diff
@@ -171,45 +170,44 @@ end
     e  = libxc_energy(terms, ρ)
     Vρ = reshape(terms.vrho,   s_ρ, n_p)
     Vσ = reshape(terms.vsigma, s_σ, n_p)
+    Vρρ = terms.v2rho2
```
Technici4n, line 173: Nice!

PR #1261, opened 2026-02-20
First minimal exact exchange and Hartree-Fock

mfherbst:
Split-off PR from #1223 to first get a minimal version tested and merged into master.


mfherbst:
I am happy with this code as it stands for a first HF implementation.

Pinging @antoine-levitt and @toschaefer if you have additional thoughts. If you both are fine and tests are green, I'll merge.

PR #1262, opened 2026-02-20
Optimizing XC instantiation for GPUs

abussy:
This PR optimizes the instantiation of the XC term on the GPU, with a particularly large impact when Duals are involved (stress and response calculations). It follows the same design patterns as #1163.

Particular care was given to the `hankel` function in `workarounds/forwarddiff_rules.jl`, in order to make it GPU compatible. A side effect is the suppression of allocations, which also marginally (few %) speeds-up Xc instantiation on the CPU.

For illustration, here are XC instantiation timings for the input below (stress of a 3x3x3 rattled aluminium supercell):

| | master | this PR |
|----------|----------|----------|
| A100 | 65 s | 2 s |
| Mi250 | 66 s | 2 s |

```
using DFTK
using PseudoPotentialData
using CUDA
setup_threading()

arch = DFTK.GPU(CuArray)

Ecut = 50.0
kgrid = [1, 1, 1]
maxiter = 10
tol = 1.0e-10
temperature = 0.01

factor = 3
a = 3.8267
                      [1.0 0.0 1.0];
                      [1.0 1.0 0.0]]
end

#rattle a few atoms                                                                                  
positions[5] = positions[5] + [0.01, 0.0, -0.01]
positions[10] = positions[10] + [0.01, 0.0, -0.01]
positions[15] = positions[15] + [0.01, 0.0, -0.01]
positions[20] = positions[20] + [0.01, 0.0, -0.01]
positions[25] = positions[25] + [0.01, 0.0, -0.01]

model = model_DFT(lattice, atoms, positions; temperature=temperature,
                  
# warmup
scfres = self_consistent_field(basis; maxiter=3, tol=tol);
forces = compute_forces(scfres)
stress = compute_stresses_cart(scfres)

#actual calculations
DFTK.reset_timer!(DFTK.timer)
scfres = self_consistent_field(basis; maxiter=maxiter, is_converged=ScfConvergenceEnergy(tol))
forces = compute_forces(scfres)
stress = compute_stresses_cart(scfres)
@show DFTK.timer
```


src/pseudo/PspUpf.jl
```diff
@@ -259,6 +259,19 @@ function eval_psp_density_core_fourier(psp::PspUpf, p::T) where {T<:Real}
     return hankel(rgrid, r2_ρcore, 0, p)
 end
 
+# Vectorized version of the above, GPU compatible
+function eval_psp_density_core_fourier(psp::PspUpf, ps::AbstractVector{T}) where {T<:Real}
+    quadrature = default_psp_quadrature(psp.rgrid)
+    arch = architecture(ps)
+    rgrid = to_device(arch, @view psp.rgrid[1:psp.ircut])
+    r2_ρcore = to_device(arch, @view psp.r2_ρcore[1:psp.ircut])
```
mfherbst, line 289: I agree with what you do here. But I think it's worth having a small discussion as this does now emerge as a more general pattern. Some questions:

- Right now we implement the "single-`p`-version" in all pseudos and then make vectorised versions for the cases where we want extra speedup on the GPU. Would it make more sense to go the other way and implement the vectorised version and then provide the "single-`p`-version" as a fallback, i.e. just calling the vector-`p` version with `[p]` ? My thinking is that it probably makes sense to do exactly the same thing (prefer the vectorised version for the GPU) for the HGH pseudos and for possible UPF pseudos we implement in the future. So it's probably the better pattern to make the vector version the thing you actually have to implement as a developer.
- What about the other functions that also take a single `p` right now. Should we not move those to vector versions as well ? The very least for consistency ?
- In line with that, does it make sense to put `rgrid` and `r2_ρcore` onto the GPU from the start ?

src/workarounds/forwarddiff_rules.jl
```diff
@@ -312,20 +319,17 @@ function hankel(r::AbstractVector, r2_f::AbstractVector, l::Integer, p::TT) wher
...
-    jl = sphericalbesselj_fast.(l, pv .* r)
-    value = 4T(π) * simpson((i, r) -> r2_f[i] * jl[i], r)
-
-    if iszero(pv)
-        return TT(value, zero(T) * ForwardDiff.partials(p))
-    end
-    derivative = 4T(π) * simpson(r) do i, r
-        (r2_f[i] * (l * jl[i] / pv - r * sphericalbesselj_fast(l+1, pv * r)))
+    # To reduce allocations, compute value and derivative simultaneously, using a SVector as integrand
+    res = 4V(π) * quadrature(r) do i, r
```
mfherbst: You don't want to unpack directly here ? E.g. `val, derivative = 4V(π) ...`

src/workarounds/forwarddiff_rules.jl
```diff
@@ -302,6 +302,13 @@ end
 end
 
 function hankel(r::AbstractVector, r2_f::AbstractVector, l::Integer, p::TT) where {TT <: ForwardDiff.Dual}
```
mfherbst, line 307: Historically we have not done this, but I think for such cases of isolated functions the primal and the dual should be close. Should we move this function to the `src/common/hankel.jl` as well ?

src/density_methods.jl
```diff
@@ -212,6 +213,25 @@ function atomic_density(element::Element, Gnorm::T,
...
 end
 
+# Generic vectoriezed version of the above
+function atomic_density(element::Element, Gnorms::AbstractVector{T},
+                        ::AtomicDensity) where {T <: Real}
+    arch = architecture(Gnorms)
+    to_device(arch,
+              map(Gnorm -> gaussian_valence_charge_density_fourier(element, Gnorm),
+                  to_cpu(Gnorms)))
+end
```
mfherbst: This looks wrong. I think it should be:
```suggestion
function atomic_density(element::Element, Gnorms::AbstractVector{T},
                        dens::AtomicDensity) where {T <: Real}
    arch = architecture(Gnorms)
    to_device(arch, map(Gnorm -> atomic_density(element, Gnorm, dens), to_cpu(Gnorms)))
end
```
Again the same question whether the single-element or the vector version should be the default arises (see comment in `src/pseudo/PspUpf.jl`.

src/pseudo/PspUpf.jl
abussy: I think this is an important discussion, and here is my take on it:

On a conceptual level, I think it makes more sense to have a "single-`p`" version as a default, and a "vector-`p`" as a special case: it follows human thinking a bit more closely. That being said, optimized "vector-`p`" functions are many times more efficient on the GPU, and also have a strong potential for CPU shared memory parallelism (OMP threading style), if this becomes necessary.

I also agree that using the "vector-`p`" as a default would push future developments in the right direction from the start. It is however not straight forward to write a GPU compatible `map` call, and it is likely never gonna happen by chance. Since most people develop and test on the CPU, there is a real risk of introducing faulty code. In that regard, using the current pattern of:
https://github.com/JuliaMolSim/DFTK.jl/blob/b3c1e9a4b621d8ef343ba50071f80baa2c8f4f45/src/pseudo/NormConservingPsp.jl#L89-L93
carries an explicit message regarding GPUs: this piece of code does not work (or has not been tested) on the GPU. Therefore, I personally slightly lean towards "single-`p`" version as a default.

> does it make sense to put rgrid and r2_ρcore onto the GPU from the start ?

I think this would introduce a lot of complication: everything in the PSP code would need to work with GPU arrays. Performance wise, the data transfer towards the GPU is cheap, because arrays are small, and the PSP functions are rarely called (mostly basis instantiation, so 2-3 times per calculation).

src/workarounds/forwarddiff_rules.jl
abussy: I didn't think about it, I see no reason not to do it.

src/workarounds/forwarddiff_rules.jl
abussy: It could make sense. There is a real risk that `src/workaround/forwarddiff_rules.jl` explodes in size in the long run. Keeping all special `Dual` rules in a single place makes it easier to navigate, though. What do you think @niklasschmitz ?

src/density_methods.jl
abussy: Indeed, it is wrong. The `:minimal` tests probably still pass because `gaussian_valence_charge_density_fourier` is the only type of `AtomicDensity` showing up there. The `test/PspLinComb.jl` fails though.

src/pseudo/PspUpf.jl
mfherbst: > there is a real risk of introducing faulty code.

Is this true ? If people develop new pseudos they are likely copying from existing implementations. If we do it right there, would it not either immediately work in a new implementation or reduce the effort in refactoring such that it works efficiently on the GPU ? In the sense that it makes it in the long run easier to fix that faulty code rather than refactoring one by one every function such that there also is a vector version ? If you think I'm wrong here and having the vector code around will not reduce efforts in refactoring, then I agree that because of the conceptional simplicity, it makes sense to default on the single-`p`-version.

> In that regard, using the current pattern of:

Ok, but then we should use this consistently everywhere, with possibly a macro to avoid writing the same code 20 times (if it's not too cumbersome ... macros have always the danger to obfuscate). It's weird that the vector version only exists for some PSP functions and not others.

> I think this would introduce a lot of complication: 

That makes sense. Let's not do it for now.

abussy: > Is this true ?

I think the risk exists, yes. For example, while working on #1163, I looked into vectorizing PspHgh:

https://github.com/JuliaMolSim/DFTK.jl/blob/b3c1e9a4b621d8ef343ba50071f80baa2c8f4f45/src/pseudo/PspHgh.jl#L110-L135

A function like the following works fine on the CPU, but fails on the GPU because instances of `Polynomial` are not isbits:
```
function eval_psp_local_fourier(psp::PspHgh, ps::AbstractVector{T}) where {T <: Real}
    map(ps) do p
        if p == 0
            zero(T)  # Compensating charge background
        else
            t = p * psp.rloc
            psp_local_polynomial(T, psp, t) * exp(-t^2 / 2) / t^2
        end
    end
end
```

I am sure there are ways to circumvent this, but I didn't spend the time because I didn't identify obvious performance bottlenecks.

It would probably be possible to rework all current PSP implementations to be vectorized by default in a GPU compatible way. If properly commented, these could also be used as a base for new Psp implementations, which would hopefully result in GPU compatible code from the start.

This is a lot of work, and a major refactoring of the code. I am willing to commit to it, but I'd rather have everybody on board with it before I do.

src/workarounds/forwarddiff_rules.jl
niklasschmitz: I would vote for keeping this workaround in `forwarddiff_rules.jl` for now. Likely there is a better way of doing it, which would either avoid the explicit Dual overload entirely or somehow separate type promotion logic from the call site. I agree `forwarddiff_rules.jl` should not explode in the long run (it would to me be a sign that we are doing something wrong with our abstractions), but importantly the fix should also NOT be to spread custom ForwardDiff.Dual logic throughout the codebase here and there. So I think as is is good enough for this PR, and we can keep an eye of boiling down `forwarddiff_rules.jl` at some point, but in an isolated way if that makes sense

src/pseudo/PspUpf.jl
mfherbst: Regarding the Polynomial: This was implemented a long, long time ago for some idea that never really matured. So if this is the obstacle, it can be completely removed. In fact this would imply we can even remove the `Polynomials` dependency from DFTK completely, which is a good thing ... given that this rather obscure use case is the only place we need the package at all.

> This is a lot of work, and a major refactoring of the code. I am willing to commit to it, but I'd rather have everybody on board with it before I do.

I agree with that and I think we should indeed only do it, if it's worth it. What is most important in my opinion is consistency across our API and something that works by default. I think for now it would be fine to just consistently introduce vector-based interfaces for all similar PSP function calls, that fall back to the single-`p` versions. It's just a little weird to single out some functions over others, in my opinion.

I think somewhat independently is the question whether the vector version should be the default and the dispatch reverted. This we can discuss elsewhere (in an issue ?)

abussy: Ok, sounds reasonable. I'll keep working on this PR and systematically introduce vectorized alternatives to the PSP functions. I'll try to reduce code duplication if I can, I'll see if some clever macro can be used. I'll also open an issue for the discussion on what should be the default.

mfherbst: Thanks, perfect. If you prefer, we can also first merge this PR and #1265 or somehow arrange things differently.

abussy: I think it's ok to do it as part of this PR, and adapt #1265 afterwards. This way, things are clean from the start. If someone has an urgent need to perform lots of fast stress calculations in the near future, then we can merge first.

mfherbst: Don't think so, let's not rush.

azadoks: Just a few thoughts here:
1. The vector-`p`-as-default makes a lot of sense if using `NumSBT` or another FFT-based algorithm for the spherical Bessel transform. There, the single-`p` implementation would be calling the vector-`p` and returning a single value (very bad, performance-wise).
2. `Polynomials` will probably be needed for ultrasoft/PAW as far as I remember. There, the augmentation charges close to the core are often represented as polynomials.

src/elements.jl
```diff
@@ -186,6 +176,24 @@ function core_charge_density_fourier(el::ElementPsp, p::T) where {T <: Real}
...
+    if has_valence_density(el.psp)
+        eval_psp_density_valence_fourier(el.psp, ps)
+    else
+        gaussian_valence_charge_density_fourier(el, ps)
+    end
+end
+function core_charge_density_fourier(el::ElementPsp, ps::AbstractVector{T}) where {T <: Real}
+    eval_psp_density_core_fourier(el.psp, ps)
+end
+
```
abussy: I find the code duplication annoying here. I tried to define the Psp dispatch in a more compact way:

```
function local_potential_fourier(el::ElementPsp, p) 
    eval_psp_local_fourier(el.psp, p)
end
```
or with
```
function local_potential_fourier(el::ElementPsp, p::Union{T, AbstractVector{T}}) where {T <: Real}
    eval_psp_local_fourier(el.psp, p)
end
```

But in either case, I run into ambiguity issues with the generic `Element` fallback with signature:
`local_potental_fourier(el::Element, ps::AbstractVector{T}) where {T}`

I am happy to take suggestions.

mfherbst: I see your problem. Would it not be a possible solution to pass on all evaluations of this function to the `el.psp`. i.e. simply have functions of the form
```julia
core_charge_density_fourier(el::ElementPsp, p) = eval_psp_density_core_fourier(el.psp, p)
```
and then it's the job of the `PspHgh` and consorts to provide both a version for `p::AbstractVector` and `p::Real` ?

src/elements.jl
```diff
@@ -288,6 +296,27 @@ function local_potential_fourier(el::ElementGaussian, p::Real)
...
+        end
+    end
+end
+
+# Generic vectorized element functions
+@vectorize_element_function DFTK.valence_charge_density_fourier
+@vectorize_element_function DFTK.gaussian_valence_charge_density_fourier
+@vectorize_element_function DFTK.core_charge_density_fourier
+@vectorize_element_function DFTK.local_potential_fourier
+@vectorize_element_function DFTK.local_potential_real
```
mfherbst: This is one way to do this. What I find often nicer for such local cases is to do it as such:
```julia
for fn in [:valence_charge_density_fourier, :gaussian_valence_charge_density_fourier, :core_charge_density_fourier, :local_potential_fourier, :local_potential_real]
    @eval begin
        function DFTK.$fn(el::Element, arg::AbstractVector{T}) where {T <: Real}
            arch = architecture(arg)
            to_device(arch, map(p -> $fn(el, p), to_cpu(arg)))
        end
    end
end
```

mfherbst: That's usually easier to process.

src/pseudo/NormConservingPsp.jl
```diff
@@ -32,6 +31,40 @@ abstract type NormConservingPsp end
...
+    end
+end
+macro vectorize_psp_projector_function(fn, PspType)
+    quote
+        function $fn(psp::$PspType, i, l, vec::AbstractVector{T}) where {T <: Real}
+            arch = architecture(vec)
+            to_device(arch, map(p -> $fn(psp, i, l, p), to_cpu(vec)))
+        end
+    end
+end
```
mfherbst: Should be doable in one macro, where a list of extra arguments (such as `i, l`) are taken as e.g. the second argument ... but this may be more annoying than I think, so if it's not obvious to you or increases macro code complexity a lot, than this is very much fine.

mfherbst: Also my intuition is that the type should be the first argument and `fn` the second. That way you can also align better the formatting to guide the eye (type always has the same length, function name lengths differ.

mfherbst (commented):
I think I agree with this mostly. We should see if we can remove some code duplication at the element level by deferring the `vector -> map over single elements` dispatch fully towards the dispatches on PSP structs.

src/elements.jl
mfherbst: See my review. Hope this works, if not, happy to think more about it.

abussy: That's one of the things I initially tried, but at the moment it does not work because of the `@vectorize_element_function` macro which defines `core_charge_density_fourier(el::Element, p::AbstractVector{T})`.

Because `ElementPsp <: Element`, but `p::AbstractVector{T}` is more specific than just `p`, there is an ambiguity.

I see a potential way out using another of your suggestion:
```
for fn in [:valence_charge_density_fourier, :gaussian_valence_charge_density_fourier, :core_charge_density_fourier, :local_potential_fourier, :local_potential_real]
    @eval begin
        function DFTK.$fn(el::Element, arg::AbstractVector{T}) where {T <: Real}
            arch = architecture(arg)
            to_device(arch, map(p -> $fn(el, p), to_cpu(arg)))
        end
    end
end
```
where I could also do a loop over all Elements except ElementPsp. I'll look into it.

abussy:
I reorganized a bit, and I was able to reduce code duplication in `elements.jl`. Unfortunately, that comes at the cost of extra complication in the generation of vectorized code (extra loop on Element type, excluding ElementPsp)

src/elements.jl
```diff
@@ -39,25 +39,23 @@ eval_psp_energy_correction(T, ::Element) = zero(T)
...
     gaussian_valence_charge_density_fourier(el, p)
 end
 
 """Gaussian valence charge density using Abinit's coefficient table, in Fourier space."""
 function gaussian_valence_charge_density_fourier(el::Element, p::T)::T where {T <: Real}
     charge_ionic(el) * exp(-(p * atom_decay_length(el))^2)
 end
+function gaussian_valence_charge_density_fourier(el::Element, ps::AbstractVector{T}) where {T <: Real}
+    arch = architecture(ps)
+    to_device(arch, map(p -> gaussian_valence_charge_density_fourier(el, p), to_cpu(ps)))
```
mfherbst: I'm surprised this needs to be executed on the CPU. The implementation above looks so simple.

src/elements.jl
```diff
@@ -288,6 +276,24 @@ function local_potential_fourier(el::ElementGaussian, p::Real)
 end
 # TODO Strictly speaking needs a eval_psp_energy_correction
 
+# Generic API expectations: all element functions taking a real space scalar |r| or a
+# reciprocal space scalar |p| should have a vectorized version accepting vectors of |r| or |p|.
+# The following loop vectorizes element functions by calling the single-value version
+# elementwise. This is GPU safe and generic. Performance critical functions should have their
+# own GPU-optimized implementation. Note: explicit loop over Element types in order to avoid
+# ambiguities with the specific ElementPsp functions.
+for fn in [:gaussian_valence_charge_density_fourier, :core_charge_density_fourier,
```
mfherbst: I trust  you checked we still really need this after the simplifications in the `ElementPsp` interface you did most recently ... so while this is still cursed, it's probably the least evil solution.

But we should make sure that once new `ElementXyz` types are added, the tests break loudly until one adds the element type to the list below. Is this the case ?

src/elements.jl
abussy: It should work on the GPU, I was on auto pilot. I'll check and change

src/elements.jl
abussy: Yes, while it is more compact than the previous solution, it is absolutely necessary that the abstract `Element` is not used here (because of ambiguity with the `ElementPsp` functions). If there was a way to just say "All elements, except ElementPsp`, we would be less error prone when new element types are added. I do not know of such a way though.

That being said, when adding a new element type, tests will not pass if the important vectorized functions are not provided.

src/elements.jl
abussy: Well, actually it is not that straight forward, because `Element` is not `isbit`. A simple solution like
```
map(p -> gaussian_valence_charge_density_fourier(el, p), ps)
```
would not work. It would be possible to port it given some refactoring, but I don't think it is worth the trouble, because it never shows up as a performance bottleneck.

src/elements.jl
abussy: I think this double loop is showing some limitations. One of the `examples` test currently fails with
```
LoadError: MethodError: no method matching local_potential_fourier(::Main.var"##226".Element2DCoulomb, ::Vector{Float64})
```
But Element2DCoulomb is defined nowhere within DFTK. It would be caught if the default vectorized functions were defined for generic `Elements`, but this leads to the usual ambiguities.

I am slowly running out of ideas.

mfherbst:
@abussy I finally got around to giving this another look. My model was that this

```julia
abstract type Element end                                                                          
function evaluate(x::Element, ps::AbstractVector)                                                  
    map(p -> evaluate(x, p), ps)                                                                   
end                                                                                                
                                                                                                   
struct ElementPsp{T} <: Element                                                                    
    inner::T                                                                                       
end                                                                                                
evaluate(x::ElementPsp, ps)                 = evaluate(x.inner, ps)                                
evaluate(x::ElementPsp, ps::AbstractVector) = evaluate(x.inner, ps::AbstractVector)       # <--- This resolves ambiguity
                                                                                                                   
struct ZeroElement <: Element end                                                                                  
evaluate(::ZeroElement, p::Number) = 0                                                                             
                                                                                                                   
abstract type NormConservingPsp end                                                                                
struct OnePsp   <: NormConservingPsp end                                                                           
struct OtherPsp <: NormConservingPsp end                                                                           
                                                                                                                   
function evaluate(x::NormConservingPsp, ps::AbstractVector)                                                        
    map(p -> evaluate(x, p), ps)                                                                                   
end                                                                                                                
                                                                                                                   
evaluate(::OnePsp,   p::Number) = 1                                                                                
evaluate(::OtherPsp, ps::AbstractVector) = copy(ps)                                                                
evaluate(::OtherPsp, p::Number) = p+1                                                                              
                                                                                                                   
                                                                                                                   
@assert evaluate(ElementPsp(OnePsp()), [2, 4]) == [1, 1]                                           
@assert evaluate(ElementPsp(OnePsp()), 2) == 1                                                     
@assert evaluate(ElementPsp(OtherPsp()), [2, 4]) == [2, 4]                                         
@assert evaluate(ElementPsp(OtherPsp()), 2) == 3  
```

works fine, but only after the highlighted line is added. A silicon calculation runs. Let's see what the tests say and then I'll also try it on a GPU.

mfherbst:
Ok I think there is another problem in the logic here. Now `NormConservingPsp` defines functions pairs like:

```julia
  eval_psp_projector_fourier(psp::NormConservingPsp, i, l, p::AbstractVector) =                                                             
       eval_psp_projector_fourier(psp, i, l, norm(p))
                                                                                                   
  # Fallback vectorized implementation for non GPU-optimized code.                                 
  function eval_psp_projector_fourier(psp::NormConservingPsp, i, l,                                
                                      ps::AbstractVector{T}) where {T <: Real}                     
      arch = architecture(ps)                                                                      
      to_device(arch, map(p -> eval_psp_projector_fourier(psp, i, l, p), to_cpu(ps)))              
  end   
```

The idea of the first function was to take the norm for you, for convenience. i.e. you could just call the function on [1, 0, 0] and it would implicitly call the function with norm `1`. But we want is to have functionality where you can evaluate the projectors on a list of norms. These two interfaces together have the potential for a lot of confusion and wrong behaviour and we should not have both of them.

Removing the versions which implicitly take the norm (which is what I propose) has the additional advantage that it should remove the method ambiguities and all the annoying macro magic could be removed also at the PSP level.

mfherbst:
@abussy  I have not checked the perf implications, but the current code runs on the GPU without any issues and looks reasonably clean.

Could you check this completes the intention of your PR without regression ? If yes, I'd say we merge this.

abussy:
I think your solution is good. Certainly more elegant than the macros I used before. I couldn't see any performance regression either. I only fixed a `map` construct in which variables of the `Element` type where called: these are not `isbit`, and the kernel won't compile on GPUs. It's good to go imo.

PR #1263, opened 2026-02-23
Add interface and tests for hybrid functionals

mfherbst:
This PR does the following:
- Gets PBE0 working
- Rename `coulomb_kernel_model` to `singularity_treatment`
- Revert default of `AceExx` to `VanillaExx` due to implications to ACE giving virtual orbitals. We should first implement a follow-up SCF step after convergence, which re-diagonalises the Hamiltonian without ACE if ACE is used or be smarter in which orbitals are used for sketching. To be investigated

CC @toschaefer especially regarding the last point.

### DIFF
--- a/src/terms/hubbard.jl
+++ b/src/terms/hubbard.jl
@@ -120,6 +120,7 @@ function Hubbard(manifolds::Vector{OrbitalManifold}, U::Vector{T}) where {T}
     U = austrip.(U)
     Hubbard{eltype(U)}(manifolds, U)
 end
+Hubbard() = Hubbard(OrbitalManifold[], Float64[])
 function Hubbard(manifold_to_U::Vararg{T}) where {T <: Pair}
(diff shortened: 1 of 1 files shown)


toschaefer:
>    Revert default of `AceExx` to `VanillaExx` due to implications to ACE giving virtual orbitals. We should first implement a follow-up SCF step after convergence, which re-diagonalises the Hamiltonian without ACE if ACE is used or be smarter in which orbitals are used for sketching. To be investigated

Both strategies are valid to get the right HF band gap and virtuals:
* include all currently considered bands in the ACE set-up step. In particular when temperature and smearing is used, this is recommendable. This makes ACE more expensive by a factor of (N_bands/N_occ)^3. Is this number with your solvers usually well below 2? Probably hard to say...
* only perform a final diagonalization with VanillaExx. The computational cost is just one additional iteration. How well does this fit into the DFTK design?

I would currently prefer the first strategy, as it is probably somewhat safer (also in terms of checkpointing and non-converged SCF).

PR #1271, opened 2026-02-27
Derivative support in mgga and mggal potential computations

Technici4n:
This is derivative hell. The tests pass but they also don't seem super robust (e.g. swapping a few cross-derivatives still makes them pass... this of course also depends on the exact used xc functional). So I triple-checked the order of derivatives, and it seems correct. We could maybe use macros/generated functions but it will reduce readability.

As a side-note, it's weird that mggal is seen as something more general than mgga; usually it's more of a bandaid when tau is not available? (E.g. `mgga_x_r2scanl`). In principle nothing stops us from doing DFPT with $\nabla^2 \rho$-dependent functionals such as r2scanl, but this interface requiring `tau` does...

It might be better to rethink this whole interface such that the evaluation of a functional is done with maximum flexibility and only considers the relevant contributions to the potential and the kernel.

### DIFF
--- a/.github/workflows/documentation.yaml
+++ b/.github/workflows/documentation.yaml
@@ -36,7 +36,7 @@ jobs:
             DocMeta.setdocmeta!(DFTK, :DocTestSetup, :(using DFTK); recursive=true)
             doctest(DFTK)'
       - name: Upload Docs
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: documentation
           path: docs/build
--- a/examples/exact_exchange.jl
+++ b/examples/exact_exchange.jl
@@ -22,5 +22,5 @@ basis  = PlaneWaveBasis(model; basis.Ecut, basis.kgrid)
 scfres = self_consistent_field(basis;
                                solver=DFTK.scf_damping_solver(damping=1.0),
                                is_converged=ScfConvergenceEnergy(1e-7), 
-                               scfres.ρ, scfres.ψ, scfres.occupation,
+                               scfres.ρ, scfres.ψ, scfres.occupation, scfres.eigenvalues,
                                diagtolalg=DFTK.AdaptiveDiagtol(; ratio_ρdiff=5e-4))
--- a/src/DFTK.jl
+++ b/src/DFTK.jl
@@ -101,7 +101,9 @@ include("supercell.jl")
 export Energies
 include("Energies.jl")
 
-export NeglectSingularity, ProbeCharge, SphericallyTruncated
+export Coulomb, SphericallyTruncatedCoulomb
+export ShortRangeCoulomb, LongRangeCoulomb
+export ProbeCharge, ReplaceSingularity
 include("coulomb.jl")
 
 export Hamiltonian
@@ -156,7 +158,8 @@ include("eigen/preconditioners.jl")
 include("eigen/diag.jl")
 
 export model_atomic, model_DFT, model_HF
-export LDA, PBE, PBE0, PBEsol, SCAN, r2SCAN
+export LDA, PBE, PBEsol, SCAN, r2SCAN
+export HybridFunctional, PBE0, HSE
 include("standard_models.jl")
 
 export KerkerMixing, KerkerDosMixing, SimpleMixing, DielectricMixing
--- a/src/coulomb.jl
+++ b/src/coulomb.jl
@@ -1,21 +1,121 @@
+#
+# Interaction models
+#
+
 @doc raw"""
-Abstract type for different methods of computing the 
-discretised Coulomb kernel ``v(G+q) = 4π/|G+q|²``.
+Abstract type for different interaction models
 
 Available models:
+- [`Coulomb`](@ref): 1/r
+- [`ShortRangeCoulomb`](@ref): erfc(μr)/r
+- [`LongRangeCoulomb`](@ref): erf(μr)/r
+- [`SphericallyTruncatedCoulomb`](@ref): θ(R-r)/r
+
+If an interaction model features a singularity, that requires some special treatment,
+the following are available:
 - [`ProbeCharge`](@ref): Gygi-Baldereschi probe charge method
-- [`NeglectSingularity`](@ref): Set G+q=0 component to zero
-- [`SphericallyTruncated`](@ref): Spherical truncation at radius Rcut
+- [`ReplaceSingularity`](@ref): Set G+q=0 component to given value (default is zero)
 
-See also: [`compute_coulomb_kernel`](@ref)
+See also: [`compute_kernel_fourier`](@ref)
 """
-abstract type CoulombSingulartyTreatment end
+abstract type InteractionKernel end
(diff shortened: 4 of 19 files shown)


src/DispatchFunctional.jl
```diff
@@ -123,12 +97,105 @@ end
...
+# and the perturbations, by summing up all the spin combinations.
+@views function libxc_assemble_δVρ(Vρρ, δρ, Vρσ=nothing, δσ=nothing,
+                                            Vρτ=nothing, δτ=nothing,
+                                            Vρl=nothing, δl=nothing)
+    if size(δρ, 1) == 1
+        δVρ = Vρρ .* δρ
+        isnothing(Vρσ) || (δVρ .+= Vρσ .* δσ)
+        isnothing(Vρτ) || (δVρ .+= Vρτ .* δτ)
+        isnothing(Vρl) || (δVρ .+= Vρl .* δl)
+        δVρ
```
mfherbst: I'd add an explicit return for such nested ifs for clarity.

src/DispatchFunctional.jl
```diff
@@ -123,12 +97,105 @@ end
...
+    else
+        δVρ1 = @. Vρρ[1,:] * δρ[1,:] + Vρρ[2,:] * δρ[2,:]
+        δVρ2 = @. Vρρ[2,:] * δρ[1,:] + Vρρ[3,:] * δρ[2,:]
+        if !isnothing(Vρσ)
+            δVρ1 += @. Vρσ[1,:] * δσ[1,:] + Vρσ[2,:] * δσ[2,:] + Vρσ[3,:] * δσ[3,:]
```
mfherbst: Can also do the addition in-place (i.e. move the `@.` to be the first thing in the line (or maybe it even works if put before the `if` ?)

src/DispatchFunctional.jl
```diff
@@ -123,12 +97,105 @@ end
...
+    if size(δρ, 1) == 1
+        δVσ = Vρσ .* δρ .+ Vσσ .* δσ
+        isnothing(Vστ) || (δVσ .+= Vστ .* δτ)
+        isnothing(Vσl) || (δVσ .+= Vσl .* δl)
+        δVσ
+    else
+        δVσ1 = @. Vρσ[1,:] * δρ[1,:] + Vρσ[4,:] * δρ[2,:]
+        δVσ2 = @. Vρσ[2,:] * δρ[1,:] + Vρσ[5,:] * δρ[2,:]
+        δVσ3 = @. Vρσ[3,:] * δρ[1,:] + Vρσ[6,:] * δρ[2,:]
+        δVσ1 .+= @. Vσσ[1,:] * δσ[1,:] + Vσσ[2,:] * δσ[2,:] + Vσσ[3,:] * δσ[3,:]
```
mfherbst, line 147: Or like this is also fine, I think.

src/DispatchFunctional.jl
```diff
@@ -123,12 +97,105 @@ end
...
+end
+@views function libxc_assemble_δVτ(Vρτ, δρ, Vστ, δσ, Vττ, δτ,
+                                   Vlτ=nothing, δl=nothing)
+    if size(δρ, 1) == 1
+        δVτ = Vρτ .* δρ .+ Vστ .* δσ .+ Vττ .* δτ
+        isnothing(Vlτ) || (δVτ .+= Vlτ .* δl)
+        δVτ
+    else
+        δVτ1 = @. Vρτ[1,:] * δρ[1,:] + Vρτ[3,:] * δρ[2,:]
+        δVτ2 = @. Vρτ[2,:] * δρ[1,:] + Vρτ[4,:] * δρ[2,:]
```
mfherbst: Adding two spaces after the `=` helps the eye

src/DispatchFunctional.jl
```diff
@@ -123,12 +97,105 @@ end
...
+end
+@views function libxc_assemble_δVσ(Vρσ, δρ, Vσσ, δσ, Vστ=nothing, δτ=nothing,
+                                                     Vσl=nothing, δl=nothing)
+    if size(δρ, 1) == 1
+        δVσ = Vρσ .* δρ .+ Vσσ .* δσ
+        isnothing(Vστ) || (δVσ .+= Vστ .* δτ)
+        isnothing(Vσl) || (δVσ .+= Vσl .* δl)
+        δVσ
+    else
+        δVσ1 = @. Vρσ[1,:] * δρ[1,:] + Vρσ[4,:] * δρ[2,:]
```
mfherbst: Add spaces to help the eye see the alignment

src/DispatchFunctional.jl
```diff
@@ -123,12 +97,105 @@ end
...
+            δVτ2 .+= @. Vlτ[2,:]*δl[1,:] + Vlτ[4,:]*δl[2,:]
+        end
+        libxc_combine_spins(δVτ1, δVτ2)
+    end
+end
+@views function libxc_assemble_δVl(Vρl, δρ, Vσl, δσ, Vlτ, δτ, Vll, δl)
+    if size(δρ, 1) == 1
+        Vρl .* δρ .+ Vσl .* δσ .+ Vlτ .* δτ .+ Vll .* δl
+    else
+        δVl1 = @. Vρl[1,:] * δρ[1,:] + Vρl[3,:] * δρ[2,:]
```
mfherbst: spaces

src/DispatchFunctional.jl
```diff
@@ -178,45 +239,134 @@ end
...
+    Vρρ = terms.v2rho2
+    Vρσ = terms.v2rhosigma
+    Vρτ = terms.v2rhotau
+    Vρl = terms.v2rholapl
+    Vσσ = terms.v2sigma2
+    Vστ = terms.v2sigmatau
+    Vσl = terms.v2sigmalapl
+    Vll = terms.v2lapl2
+    Vlτ = terms.v2lapltau
+    Vττ = terms.v2tau2
```
mfherbst, line 332: What a nightmare ...

test/kernel.jl
```diff
@@ -130,58 +130,103 @@ end
...
-            # LibxcDensities with max_derivative=1 gives both ρ and σ = |∇ρ|²
-            density = DFTK.LibxcDensities(basis, 1, ρ0, nothing)
-            ρ = reshape(density.ρ_real, size(density.ρ_real, 1), :)  # (n_spin, n_p)
-            σ = reshape(density.σ_real, size(density.σ_real, 1), :)  # (n_spin_σ, n_p)
+            ε_fd = 1e-4
+            function do_fd(f)
+                f1 = f(-2ε_fd)
+                f2 = f(-ε_fd)
+                f3 = f(ε_fd)
+                f4 = f(2ε_fd)
```
mfherbst: I would do:
```suggestion
                f_m2ε = f(-2ε_fd)
                f_m1e = f(-ε_fd)
                f_p1e = f(ε_fd)
                f_p2e = f(2ε_fd)
```
and similarly for the `y`s as this contains more meaning than just 1 to 4

PR #1282, opened 2026-03-11
[WIP] Additional exx kernels: Voxel averaged and Wigner-Seitz truncation

toschaefer:
Two additional useful strategies for the Coulomb kernel for electron repulsion integrals (e.g. for exx).

Wigner-Seitz truncation
[R. Sundararaman, T. A. Arias. Phys. Rev. B **87**, 165122 (2013)](https://doi.org/10.1103/PhysRevB.87.165122)

Voxel averaged (mean potential)
[T. Schaefer et al, J. Chem. Phys. 160, 051101 (2024)](https://doi.org/10.1063/5.0182729)
This requires Gauss-Legendre quadrature points, hence I added the package FastGaussQuadrature to the dependencies.


antoine-levitt:
Nice! Particularly interested in the Wigner-Seitz truncation - is there any reason to do anything else? It seems to me this is the superior approach? Do we really want to support the voxel averaged? Also maybe some opportunities to share code with ewald.jl, but I haven't looked too closely. 

It'd be nice to have a generalized API for all these functions that return centered functions (we have a couple of those: densities guesses, local pseudos, projectors, coulomb kernels, etc.). I think long-term we want the ability to get their real-space and partial Fourier transforms (wrt 1 or 2 dimensions), so we can do eg 2D materials. But maybe it's not worth unifying since they have quite different requirements (numerical psp vs analytic but singular coulomb really are quite different computationally)

CC @denizunel

mfherbst:
> Also maybe some opportunities to share code with ewald.jl, 

I think so, but I'd keep that for later, same as the generalized API. I'm currently looking at this on the side to streamline the code a bit, will hopefully push a review in the next week.

toschaefer:
> Nice! Particularly interested in the Wigner-Seitz truncation - is there any reason to do anything else? It seems to me this is the superior approach? 

For EXX of gapped systems Wigner-Seitz truncation is probably the best approach to reach the thermodynamic llimit. 
But for beyond-EXX stuff like many-electron correlation methods (MP2, CC, ...) it may not be quite as superior. In particular when we consider relatively small cells where a truncation is simply too rough (no matter if spehrically or Wigner-Seitz).

> Do we really want to support the voxel averaged? 

We need it for surface science (strongly anisotropic cells) with coupled cluster theory. Since our plan is to make DFTK a standard workflow, voxel averaged would be essential. 
It would also be nice to have it for tests with the new (planned) mixers for SCF calculations with EXX. In particular for metals, voxel averaged could be advantageous.

toschaefer:
rebased to make merge easier

antoine-levitt:
Is this still WIP or ready for review?

toschaefer:
from my end, ready for review

src/coulomb.jl
```diff
@@ -10,6 +12,7 @@ Available models:
 - [`ShortRangeCoulomb`](@ref): erfc(μr)/r
 - [`LongRangeCoulomb`](@ref): erf(μr)/r
 - [`SphericallyTruncatedCoulomb`](@ref): θ(R-r)/r
+- [`WignerSeitzTruncatedCoulomb`](@ref): χ(r)/r where χ(r)=1 inside Wigner-Seitz cell, otherwise 0.
 
 If an interaction model features a singularity, that requires some special treatment,
```
antoine-levitt: no comma after singularity (germanism)

antoine-levitt: features a singularity -> is long-range

src/coulomb.jl
```diff
@@ -26,7 +29,7 @@ Base.Broadcast.broadcastable(k::InteractionKernel) = Ref(k)
 #   eval_probe_charge_integral(::InteractionKernel, α)
 #      Should return ∫_{BZ}  kernel(q) * e^(-α * q^2) dq
 #      This is needed for the ProbeCharge regularisation. Note, that no factor 1/Γ
-#      where Γ is BZ volume) is used.
+#      (where Γ is BZ volume) is used.
 #   _compute_kernel_fourier(::InteractionKernel, basis, qpt, q)
 #      The single q-point version of compute_kernel_fourier
```
antoine-levitt: there's compute_ and eval_, should all these be merged?

antoine-levitt: (we really need to have a unified API for localized functions that can be evaluated in real or reciprocal space)

antoine-levitt: also should we have a eval_kernel_real? This is often useful for debug purposes (can be left as TODO if you don't feel like doing it)

src/coulomb.jl
antoine-levitt, line 44: AbstractCoulombKernel? Interaction seems a bit generic (these are all Coulomb-like)

antoine-levitt: I think we use relatively consistently (or at least we wanted to) p for the Fourier-space variable in \R^3 (so usually things like p=G+k, and q=k-k'). Why mention G at all in this file? Can't you just use p?

antoine-levitt: give the equation defining this. Gsq -> p? Or ps?)

src/coulomb.jl
```diff
@@ -26,7 +29,7 @@ Base.Broadcast.broadcastable(k::InteractionKernel) = Ref(k)
 #   eval_probe_charge_integral(::InteractionKernel, α)
 #      Should return ∫_{BZ}  kernel(q) * e^(-α * q^2) dq
 #      This is needed for the ProbeCharge regularisation. Note, that no factor 1/Γ
-#      where Γ is BZ volume) is used.
+#      (where Γ is BZ volume) is used.
 #   _compute_kernel_fourier(::InteractionKernel, basis, qpt, q)
```
antoine-levitt: not clear why this needs to exist. I don't think this file should know about the existence of basis at all?

src/coulomb.jl
antoine-levitt: be parametric on T

antoine-levitt, line 73: Here and elsewhere having a container struct be lower case single letter is weird to me. Maybe `kernel`?

src/coulomb.jl
antoine-levitt: Also eval takes |p|^2 but this one takes p, is that correct?

src/coulomb.jl
antoine-levitt, line 79: define a (e^x-1)/x function (I think in exponential integrator land they call this the phi function) and avoid this special casing?

antoine-levitt: as above, be parametric

antoine-levitt, line 92: that's a lot of boilerplate just for the user to use units. Don't we usually ask the user to take responsability for stripping their units?

src/coulomb.jl
antoine-levitt: Also it's nice to have default methods (maybe unoptimized), so if one can be written in terms of the other in general maybe do that?

src/coulomb.jl
```diff
@@ -152,6 +129,143 @@ function compute_kernel_fourier(kernel::InteractionKernel, basis::PlaneWaveBasis
...
+    Rcut::T = nothing
+end   
+function eval_kernel_fourier(k::SphericallyTruncatedCoulomb, Gsq::T) where {T}
+    4T(π) / Gsq * (1 - cos(T(k.Rcut) * sqrt(Gsq)))
+end
+function _compute_kernel_fourier(k::SphericallyTruncatedCoulomb, basis, qpt, q)
+    # TODO: This is a bit hackish as the parameter needs to be re-computed every kernel
+    #       evaluation. Cleaner would be to move this further up in the call hierarchy,
+    #       such that compute_kernel_fourier is never called without Rcut being set to
+    #       not nothing
```
antoine-levitt: yeah we have this pattern for the terms (one struct specifies and the other is instantiated with the basis), it's very annoying. Let's cross this bridge when we actually need it.

src/coulomb.jl
```diff
@@ -152,6 +129,143 @@ function compute_kernel_fourier(kernel::InteractionKernel, basis::PlaneWaveBasis
...
+    Ω = basis.model.unit_cell_volume  
+    Rcut = @something k.Rcut cbrt(3Ω/(4π))
+    kRcut = SphericallyTruncatedCoulomb(Rcut)
+
+    # Use ReplaceSingularity regularisation to explicitly set as the G==0
+    # component the exact limit of the kernel for G->0
+    _compute_kernel_fourier(kRcut, ReplaceSingularity(2π*Rcut^2), basis, qpt, q)
```
antoine-levitt, line 171: same as above, I don't like this replacesingularity business just to compensate for the fact that the implementation above is unstable (it mixes physics with a purely numerical issue). As with the exponential, define a stable (cos(x)-1)/x function

src/coulomb.jl
```diff
@@ -152,6 +129,143 @@ function compute_kernel_fourier(kernel::InteractionKernel, basis::PlaneWaveBasis
...
+end
+
+
+"""
+Truncate Coulomb interaction at the Wigner-Seitz cell boundary.
+
+Computational approach: We expand 1/r = erfc(ωr)/r + erf(ωr)/r and chose ω such that the
+short-range part erfc(ωr)/r is virtually unaffected by the truncation.
```
antoine-levitt, line 179: this is not actually true (it would be if epsilon was machine precision).

I would instead frame this as an error balance. First introduce the actual ewald-like computational scheme : we split 1/r into a short-range, singular at 0 part and a long-range non-singular at 0 part. We compute the short-range analytically, replacing the BZ by the whole space (making an error exp(-omega Rin)). We compute the long-range, non-singular at 0, in Fourier space, by an FFT on the grid we have (note we could possibly do this on another grid), incurring an error exp(-G/omega). Balancing the two to optimize the error, we get the expression you wrote. Is this correct or am I missing something?

src/coulomb.jl
```diff
@@ -152,6 +129,143 @@ function compute_kernel_fourier(kernel::InteractionKernel, basis::PlaneWaveBasis
...
+@views function _compute_kernel_fourier(k::WignerSeitzTruncatedCoulomb, 
+                                        basis::PlaneWaveBasis{T}, qpt, q) where {T}
+    model = basis.model
+    NG = length(qpt.G_vectors)
+    kernel_fourier = zeros(T, NG)
+    q = qpt.coordinate
+    
+    # === Calculate inradius R_in of Wigner-Seitz cell ===
+    
+    # R_in is largest possible R_in = (sum_i n_i * a*i) / 2 with integers n_i
```
antoine-levitt, line 212: I think we have this logic in the construction of the plane wave basis already?

antoine-levitt: otherwise it's a generic thing you can maybe move out somewhere (possibly in common/)?

src/coulomb.jl
```diff
@@ -152,6 +129,143 @@ function compute_kernel_fourier(kernel::InteractionKernel, basis::PlaneWaveBasis
...
+        r_cart = model.lattice * r_centered
+        d_min = norm(r_cart)
+        for dx in -nx:nx, dy in -ny:ny, dz in -nz:nz # Check neighbors for non-orthorhombic cells
+            dx == 0 && dy == 0 && dz == 0 && continue 
+            r_shifted = r_centered - T[dx, dy, dz]
+            d = norm(model.lattice * r_shifted)
+            d_min = min(d_min, d)
+        end
+        if d_min > sqrt(eps(T))
+            V_lr_real[idx] = erf(ω * d_min) / d_min
```
antoine-levitt, line 264: as above, define a erf(x)/x function?

antoine-levitt: Hm I think we should have a general way of doing f(x)/x stably and AD-friendly, this is coming up pretty frequently (also with @Technici4n in the spherical harmonics stuff)

antoine-levitt: Ah but of course it's my old friend the divided difference (high order in the case of f(x)/x^l). We have a general solution in https://github.com/xuequan818/MatrixFuns.jl but it might be a bit overkill...?

src/coulomb.jl
```diff
@@ -152,6 +129,143 @@ function compute_kernel_fourier(kernel::InteractionKernel, basis::PlaneWaveBasis
...
+        @warn "Coarse grid for Wigner-Seitz truncation. Effective error: $ε_actual"
+    end
+
+    # == FFT of long-range term erf(ωr)/r restricted to Wigner-Seitz cell ===
+
+    r_vectors = DFTK.r_vectors(basis)
+    V_lr_real = zeros(Complex{T}, basis.fft_size...)
+    for idx in CartesianIndices(V_lr_real)
+        r_frac = r_vectors[idx]
+        r_centered = r_frac .- round.(r_frac) # MIC
```
antoine-levitt: write it out explicitly (I had to think to get what it was referring to)

src/coulomb.jl
```diff
@@ -152,6 +129,143 @@ function compute_kernel_fourier(kernel::InteractionKernel, basis::PlaneWaveBasis
...
+        r_cart = model.lattice * r_centered
+        d_min = norm(r_cart)
+        for dx in -nx:nx, dy in -ny:ny, dz in -nz:nz # Check neighbors for non-orthorhombic cells
```
antoine-levitt: I have to say I don't understand what's going on here, why isn't this just erf(r)/r for r in r_vectors?, maybe factor out the geometry stuff to another function?

src/coulomb.jl
antoine-levitt: (the reason I want to get this right is otherwise it will come  up to bite us in the ass with AD)

src/coulomb.jl
```diff
@@ -152,6 +129,143 @@ function compute_kernel_fourier(kernel::InteractionKernel, basis::PlaneWaveBasis
...
+    kernel_fourier_lr .*= sqrt(model.unit_cell_volume)
+    
+    # === Analytic short-range term + long-range term ===
+
+    for (iG, G) in enumerate(to_cpu(qpt.G_vectors))
+        G_cart = model.recip_lattice * (G+q)
+        Gnorm2 = sum(abs2, G_cart)
+        found_singularity = (iG==1 && iszero(q))
+        Rcut = cbrt(basis.model.unit_cell_volume*3/4/π)
+        if !found_singularity
```
antoine-levitt: why don't you just write the condition here? Also why not iszero(G_cart)?

antoine-levitt: (also same comment as above, hopefully we can avoid this special casing)

src/coulomb.jl
```diff
@@ -225,3 +339,137 @@ end
     end
     kernel_fourier
 end
+
```
antoine-levitt, line 364: Running out of time today, note to self to resume the review from here.

antoine-levitt (commented):
Very nice!

src/coulomb.jl
antoine-levitt: I want to do things like 
```
function expm1_over_x(x)
    if abs(x) < 1000floatmin(Float64)
        x = 1000floatmin(Float64)
    end
    expm1(x) / x
end
```
but of course this screws up AD at 0. We can have a cutoff point and switch to a finite order taylor expansion. This is ugly and getting good precision of higher derivatives is annoying, but I'm not sure what else to do.

src/coulomb.jl
toschaefer: `compute_` sets up and returns the full vector with all the complexity of how the sampling points (including the singularity) are defined for each G+q while `eval_` is more the mathematical evaluation of the kernel at a given point G+q. 

I added a TODO for eval_kernel_real.

src/coulomb.jl
toschaefer: After some back-and-forth we called it InteractionKernel in https://github.com/JuliaMolSim/DFTK.jl/pull/1223#discussion_r2851810062

toschaefer: Principally, I am happy to follow your stylistic preferences. I don't think we've mentioned this yet during the EXX development, and if I shall change it, I would prefer to postpone it for now, since I'm already working on the next PR (which also includes some changes in src/coulomb.jl).

toschaefer: Gsq refers to G squared so psq? But I am not sure if I understand correctly...

src/coulomb.jl
toschaefer: `eval_probe_charge_integral` only takes the width `α` of the probe charge density.

Not sure what you mean with default method?

src/coulomb.jl
toschaefer: good point, thanks!

toschaefer: I see and I principally agree. Once more, this stylistic question is about parts of the code that are already in the master branch and aren't affected by this PR. I'm a bit worried that I might run into trouble since I refactor this file already for another PR that hasn't been opened yet.

I will add a TODO.

toschaefer: Very elegant idea.

I will add a TODO for that.

EDIT: Reading your comment with f(x)/x below, I will write a more general TODO, so that we can refactor this entire f(x)/x in a more elegant way.

toschaefer: done

toschaefer: Good question. If I remember correctly @mfherbst introduced that in 29c63808c .

src/coulomb.jl
toschaefer: ok

src/coulomb.jl
toschaefer: ok. add a TODO.

src/coulomb.jl
toschaefer: Ok, I will extend the TODO comments by a general statement, that we need a clever and AD-friendly solution for a possible f(x)/x machinery.

src/coulomb.jl
toschaefer: Sounds right to me. I remember that the math is done in the reference cited below.
My interpretation was that the our construction of ω actually leads to epsilon as machine precision.

src/coulomb.jl
toschaefer: You are right. Using `estimate_integer_lattice_bounds` from stc/structure.jl, I could replace
```julia
    L_min = minimum(norm, eachcol(model.lattice))
    n_bounds = zeros(Int, 3)
    for i in 1:3
        b_vec = model.recip_lattice[:, i]
        b_len = norm(b_vec)
        N = (L_min * b_len) / (2π)
        n_bounds[i] = ceil(Int, N)
    end
    nx, ny, nz = n_bounds # in case of a cubic cell nx=ny=nz=1
```
with
```julia
    L_min = minimum(norm, eachcol(model.lattice))                     
    nx, ny, nz = estimate_integer_lattice_bounds(model.lattice, L_min)
```

src/coulomb.jl
toschaefer: done

src/coulomb.jl
toschaefer: Took me a while to understand what I did here. Added better comments to make it understandable.

src/coulomb.jl
toschaefer: ok

toschaefer:
Thanks for the review!
I have another branch for the EXX k-points feature waiting on this one, so I would love to get this merged soon to keep the rebase simple. Does everything look ready to go on your end?

antoine-levitt:
I'll try to get to it soon but it's pretty hectic at the moment, sorry... Since you're very active on this I don't want to be a blocker, but at the same time I think one of the key differentiator of DFTK is that we have relatively clean code (not really in the sense that we follow best practices, but rather in the sense that the code is relatively minimal and understandable), and code review really helps with that I find (in the sense that the code we let in has been understood at least once by at least two people).

I think the best way to do that is that you fork DFTK and make your changes there. Then, when we merge a PR on the main DFTK repo (or even along the way when you make commits to the PR), you merge it on to your fork. Since the git history is preserved I think the whole process should be relatively painless (in the sense that you shouldn't have to deal with much conflicts) and you don't have to wait on us to merge PRs to progress. (also AI made everything git much smoother, I find) In exchange I promise I'll get to it eventually in a reasonable timeframe. @mfherbst might have a better idea?

toschaefer:
I totally agree with all of that. That’s why DFTK is my choice ;)

Regarding the fork strategy, normally I do it exactly like that. In this case, I shot myself in the foot because I refactored the coulomb.jl files in both branches in a somewhat non-orthogonal way. This is why I put the k-points branch on hold until this one merged.

That’s the only reason I brought it up, but please don't stress. I appreciate you taking the time to review it properly when you find time.

src/coulomb.jl
antoine-levitt: Hm OK, fine.

antoine-levitt: OK, maybe leave a TODO? I think generally this file should be about just providing interaction(p), regardless of where p is coming from (G+q, G...)

src/coulomb.jl
antoine-levitt: I don't get this. Both compute the same thing and should be consistent. If the point is that _compute_kernel_fourier(::InteractionKernel, basis, qpt, q) takes multiple kpoints then that should be a separate method of the same function. To be clear, I'd suggest
```
# computes khat(p). Not guaranteed to be fast
eval_kernel_fourier(::Coulomb, p) = 1/p^2

# eval_kernel_fourier computes khat(G+pshift) for all G in basis using a possibly faster algorithm than just looping over all G+pshift
# default implementation
function eval_kernel_fourier(k::InteractionKernel, basis, pshift)
    map(G_vectors(basis)) do G
        eval_kernel_fourier(G+pshift)
    end
end

function eval_kernel_fourier(k::WignerSeitz, p)
    r = eval_kernel_fourier(ShortRange(k.ω))
    # add to r the long-range by "slow Fourier transform", just looping over all r_vectors
end
function eval_kernel_fourier(k::WignerSeitz, pshift)
    # FFT based algorithm
end
# add a test that this overload is correct: for all kernels, the result eval_kernel_fourier(k,basis,pshift)
```

src/coulomb.jl
antoine-levitt: sorry I really should have reviewed the previous PR...

antoine-levitt: @mfherbst I think we should keep the unit stuff away from the code as much as possible

src/coulomb.jl
antoine-levitt: It's not though - unless the r grid is super super fine. In particular we should make clear that this algorithm introduces an error - it's not actually computing the Fourier transform of the Wigner-Seitz kernel up to machine precision. Indeed we could very well use a finer real-space grid for this computation. It's probably overkill but we might need a finer grid at some point anyway, so we should definitely make sure it's clear in the comment - again feel free to leave a TODO.

src/coulomb.jl
antoine-levitt: nice!

antoine-levitt (commented):
OK, I'm fine with merging the current code with lots of todo, to be adressed (by you, me or Claude :-p) at some point in the future. To me the API is the most pressing concern - it'd be nice to merge eval_kernel_fourier and _compute_kernel_fourier and clarify the relationship between the two.

src/coulomb.jl
toschaefer: done

src/coulomb.jl
toschaefer: I see. I think there was a misleading comment. I updated the doc string for InteractionKernel to explain why we have `eval_kernel_fourier` and `compute_kernel_fourier` for the splitting: (i) formula and (ii) discretization/singularity handling.

```julia
@doc raw"""                                                                                                        
Abstract type for different interaction models.                                                                    
                                                                                                                   
### Architecture                                                                                                   
                                                                                                                   
Computing interaction kernels is split into two parts: the mathematical formula (e.g. 4\pi/G^2) and the grid discretization. This split is primarily driven by the need to handle singularities in long-range kernels.                
                                                                                                                   
1. **InteractionKernel:** Defines the pure mathematical formula (via `eval_kernel_fourier`).                       
2. **regularization:** Necessary for long-range kernels (like `Coulomb` and `LongRangeCoulomb`) diverge as ``G+q \to 0``. Evaluating them on a periodic grid requires a specific strategy to handle this divergence.                  
                                                                                                                   
Because of this divergence, long-range `InteractionKernel`s contain a `regularization` field to dictate how the ``G
+q=0`` component is built via `_compute_kernel_fourier`. Short-range kernels have a finite limit at `G+q \to 0``` a
nd don't need a regularizatin.                                                                                     
                                                                                                                   
### Available models:                                                                                              
- [`Coulomb`](@ref): 1/r                                                                                           
- [`ShortRangeCoulomb`](@ref): erfc(μr)/r                                                                          
- [`LongRangeCoulomb`](@ref): erf(μr)/r                                                                            
- [`SphericallyTruncatedCoulomb`](@ref): θ(R-r)/r                                                                  
- [`WignerSeitzTruncatedCoulomb`](@ref): χ(r)/r (1 inside Wigner-Seitz cell, 0 otherwise)                          
                                                                                                                   
### Available singularity corrections (regularizations):                                                           
- [`ProbeCharge`](@ref): Gygi-Baldereschi probe charge method                                                      
- [`ReplaceSingularity`](@ref): Set the G+q=0 component to a specific value                                        
- [`VoxelAveraged`](@ref): Average the continuous kernel over the Brillouin zone voxel                             
                                                                                                                   
See also: [`compute_kernel_fourier`](@ref)                                                                         
"""                                                                                                                
abstract type InteractionKernel end                                                                                
Base.Broadcast.broadcastable(k::InteractionKernel) = Ref(k)                                                        
                                                                                                                   
# TODO: should we have a eval_kernel_real?                                                                         
# TODO: rename "k" in _compute_kernel_fourier(k...                                                                 
# TODO: change notation: p instead of G, G+q, ...                                                                  

```

src/coulomb.jl
toschaefer: done

Project.toml
```diff
@@ -14,6 +14,7 @@ DftFunctionals = "6bd331d2-b28d-4fd3-880e-1a1c7f37947f"
 DifferentiationInterface = "a0c0ee7d-e4b9-4e03-894e-1c5f64a51d63"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
+FastGaussQuadrature = "442a2c76-b920-505d-bb47-c5924d526838"
```
mfherbst: I'm not too happy with this extra dependency at this stage. I'll move this out into an extmodule.

src/coulomb.jl
```diff
@@ -1,34 +1,38 @@
...
-#
+using FastGaussQuadrature
 
 @doc raw"""
-Abstract type for different interaction models
+Abstract type for different interaction models.
+
+### Architecture
+
+Computing interaction kernels is split into two parts: the mathematical formula (e.g. 4\pi/G^2) and the grid discretization. This split is primarily driven by the need to handle singularities in long-range kernels.
```
mfherbst: Refer to the new issue here to make it clear this is not yet finalised.

src/coulomb.jl
```diff
@@ -1,34 +1,38 @@
...
 Base.Broadcast.broadcastable(k::InteractionKernel) = Ref(k)
 
-# Each InteractionKernel should support the following functions:
-#   eval_kernel_fourier(::InteractionKernel, Gsq)
-#   eval_probe_charge_integral(::InteractionKernel, α)
-#      Should return ∫_{BZ}  kernel(q) * e^(-α * q^2) dq
-#      This is needed for the ProbeCharge regularisation. Note, that no factor 1/Γ
-#      where Γ is BZ volume) is used.
-#   _compute_kernel_fourier(::InteractionKernel, basis, qpt, q)
-#      The single q-point version of compute_kernel_fourier
```
mfherbst, line 31: I would not remove this, it's a pretty concise and compact overview of what needs to be done.

src/coulomb.jl
```diff
@@ -58,13 +62,15 @@ function _compute_kernel_fourier(k::ShortRangeCoulomb, basis, qpt, q)
     # component the exact limit of the kernel for G->0, namely π/μ^2
     _compute_kernel_fourier(k, ReplaceSingularity(π/k.μ^2), basis, qpt, q)
 end
+# TODO: introduce a clever and AD-friendly way to deal with f(x)/x for x->0. E.g. intoduce phi(x) = iszero(x) ? one(x) : expm1(x) / x 
+# Also very useful for other InteractionKernels.
```
mfherbst: This feels a bit out of place here.

src/coulomb.jl
mfherbst: I agree with the first sentence, but I don't agree with the second (in the sense that this would be the solution. Maybe just have the first sentence and let us consider how to solve it when it actually becomes an issue. It does not cost that much compute after all ...

src/coulomb.jl
mfherbst: Has this been already done ?

src/coulomb.jl
mfherbst: Has this been already done ?

test/coulomb.jl
```diff
@@ -70,6 +70,20 @@
 
         # TODO: Test this gives a spherically truncated function.
     end
+
+    @testset "WignerSeitzTruncatedCoulomb" begin
+        k_wstrunc = compute_kernel_fourier(WignerSeitzTruncatedCoulomb(), basis)
+        E_wstrunc = exx_energy_only(basis, kpt, k_wstrunc, ψk_real, occk)
+        E_ref = -2.3456813523805415
+        @test abs(E_ref - E_wstrunc) < 1e-6
```
mfherbst, line 79: Is it possible to add a test that this is somewhat related to SphericallyTruncatedCoulomb, e.g. test that we get similar entries in k_wstrunc versus k_strunc or that some behaviour between the two agrees ? One may need to set `Rcut` for the SphericallyTruncatedCoulomb to a special value for this (e.g. the actual size of the cell etc.), but that sounds fair to me.

A good way to test this could be to consider evaluating the kernel on cells of the form a * Diagonal([1, 1, 1]) with larger and larger values of a, I would expect the two implementations to agree more and more as we make a larger.

mfherbst (commented):
Some minor comments. Main point is still the missing test. I've added some ideas what one could test. Feel free to object and defer if you don't have the time now, but I think it will help in the long run to have such consistency tests.

test/coulomb.jl
mfherbst: I'll also add a few tests on non-cubic cells, just that we have something where Wigner-Seitz should make a real difference.

src/coulomb.jl
toschaefer: yes, below `abstract type InteractionKernel end`.

src/coulomb.jl
toschaefer: moved it to the other TODOs below `InteractionKernel`

src/coulomb.jl
toschaefer: done

src/coulomb.jl
toschaefer: yes, below abstract type InteractionKernel end.

test/coulomb.jl
toschaefer: Thanks for the non-cubic test.

I am not sure if the sampling points of the kernel of SphericallyTruncatedCoulomb and WignerSeitzTruncatedCoulomb will get more and more similar when you increase the volume of a cubic cell. When I look at the spherically truncated case, the cos function will oscillate wildly with increasing R: v(G) = 4π/G² * (1 - cos(GR)). 
Wigner-Seitz will also oscillate like crazy but probably not in the same way.

In the limit, the physics will be the same, but the individual sampling points probably not.

mfherbst: In the limit, the physics will be the same, but the individual sampling points probably not. But does this not imply that at least the energy should get similar ?

Well if it does not really work, than let's not get too hung up about it.

toschaefer: Yes, the energy should get similar, but in the limit this should hold for all kernels. Of course, converging at different rates...
So what you mean is to define a large `a` and then calculate exx_energy_only. The energies will agree to some extend and we simply define a test like `@test E_wstrunc ≈ E_strunc atol=1e-3` with some apropriate `atol`?

mfherbst: Yes, essentially this captures we get the right limit and that things converge at the right order (if we take two different as ... at least in theory

toschaefer: done. Should be ready to merge.

mfherbst:
I refactored the test a bit @toschaefer. If tests pass and you give your ok to what I wrote (please have a look I did not mess up), I will merge.

toschaefer:
Thank you @mfherbst 
Looks fine, ready to merge.

PR #1287, opened 2026-03-12
Support Laplacian-mGGA response and second derivatives of XC via Libxc

mfherbst:
Too much code duplication for my taste and no real test yet, but it seems to run, at least.

Needs a test for derivatives of laplace-based functionals.


src/DispatchFunctional.jl
```diff
@@ -182,14 +318,18 @@ end
 end
 @views function libxc_assemble_δVl(Vρl, δρ, Vσl, δσ, Vlτ, δτ, Vll, δl)
     if size(δρ, 1) == 1
-        return Vρl .* δρ .+ Vσl .* δσ .+ Vlτ .* δτ .+ Vll .* δl
+        δVl = Vρl .* δρ .+ Vσl .* δσ .+ Vll .* δl
+        !isnothing(Vlτ) && (δVl += Vlτ .* δτ)
```
Technici4n:
```suggestion
        !isnothing(Vlτ) && (δVl .+= Vlτ .* δτ)
```

Technici4n: This got undone apparently.

src/terms/xc.jl
```diff
@@ -406,11 +425,12 @@ end
 function apply_kernel(term::TermXc, basis::PlaneWaveBasis{T}, δρ::AbstractArray{Tδρ};
                       ρ, q=zero(Vec3{T}), kwargs...) where {T, Tδρ<:Union{T,Complex{T}}}
     isempty(term.functionals) && return nothing
-    @assert all(family(xc) in (:lda, :gga) for xc in term.functionals)
+    @assert (all(family(xc) in (:lda, :gga) for xc in term.functionals) || 
+             all(family(xc) == :mggal && !needs_τ(xc) for xc in term.functionals))
```
Technici4n:
```suggestion
    @assert (all(family(xc) in (:lda, :gga, :mggal) && !needs_τ(xc) for xc in term.functionals))
```

src/DispatchFunctional.jl
```diff
@@ -24,24 +34,135 @@ function LibxcFunctional(identifier::Symbol)
...
+                                       σ::AbstractMatrix{Float64}, τ::AbstractMatrix{Float64}, args...)
+    libxc_energy_density(func; rho=ρ, sigma=σ, tau=τ)
+end
+function DftFunctionals.energy_density(func::LibxcFunctional{:mggal},
+                                       ρ::AbstractMatrix{Float64},
+                                       σ::AbstractMatrix{Float64},
+                                       τ::Union{Nothing,AbstractMatrix{Float64}},
+                                       Δρ::AbstractMatrix{Float64})
+    libxc_energy_density(func; rho=ρ, sigma=σ, tau=τ, lapl=Δρ)
 end
```
Technici4n: Would it be worth collapsing all of these by defaulting the arguments with `=nothing`?

mfherbst: Ah yes, now this works ... did not before I made the change in Libxc.jl

src/DispatchFunctional.jl
```diff
@@ -380,19 +470,32 @@ end
...
-hybrid_parameters(::Functional{:gga})  = nothing
-hybrid_parameters(::Functional{:mgga}) = nothing
+# Matrix element types that can dispatch to Libxc for energy computations
+const LibxcDispatchFloatEnergy = Union{LibxcDispatchFloat,Dual{<:Any,<:Dual{<:Any,Float64}}}
+function DftFunctionals.energy_density(fun::DispatchFunctional, ρ::Matrix{<:LibxcDispatchFloatEnergy}, args...)
+    energy_density(fun.inner, ρ, args...)
+end
+function DftFunctionals.energy_density(fun::DispatchFunctional, ρ::AbstractMatrix, args...)
+    energy_density(DftFunctional(identifier(fun)), ρ, args...)
+end
```
abussy, line 493: With AMDGPU, I get `@allowscalar` error from this.

When `ρ` is a ROCArray, the function still dispatches to DftFunctional, despite it not being GPU compatible. There should be some guards against it, to make sure the density is transferred to the CPU first. I'll look into it

mfherbst: Yes you need to add an appropriate function in `gpu/gpu_arrays.jl` for the new `energy_density` dispatch. I forgot to do this.

abussy: Indeed, all the GPU tests pass once this is added.

PR #1297, opened 2026-04-07
Reduce allocations in LOBPCG

abussy:
Reducing allocations in LOBPCG by pre-allocating arrays and taking views, rather than reallocating at each iteration.

Performance wise, I observe a speedup of LOBPCG from 0% to 10% depending on system size and architecture. It is never slower.

Less allocations also means less pressure on the GC in general, and a lower risk of memory fragmentation on the GPU (as far as I understand). This is hard to measure though.

### DIFF
--- a/src/eigen/lobpcg_hyper_impl.jl
+++ b/src/eigen/lobpcg_hyper_impl.jl
@@ -105,27 +105,37 @@ Base.adjoint(A::LazyHcat) = Adjoint(A)
     end
 end
 
+mul_hermi(A, B) = Hermitian(A * B)
+function mul_hermi(A::Adjoint{T,<:LazyHcat}, B::LazyHcat) where {T}
+    _mul(A, B; hermitian=Val(true))
+end
+
 Base.:*(A::Adjoint{T,<:LazyHcat}, B::LazyHcat) where {T}       = _mul(A, B)
 Base.:*(A::Adjoint{T,<:LazyHcat}, B::AbstractMatrix) where {T} = A * LazyHcat(B)
 
-@views function *(Ablock::LazyHcat, B::AbstractMatrix)
-    res = Ablock.blocks[1] * B[1:size(Ablock.blocks[1], 2), :]  # First multiplication
+function _mul_blocks!(res::AbstractMatrix, Ablock::LazyHcat, B::AbstractMatrix, α::Number)
     offset = size(Ablock.blocks[1], 2)
     for block in Ablock.blocks[2:end]
-        mul!(res, block, B[offset .+ (1:size(block, 2)), :], 1, 1)
+        mul!(res, block, B[offset .+ (1:size(block, 2)), :], α, 1)
         offset += size(block, 2)
     end
     res
 end
 
-function LinearAlgebra.mul!(res::AbstractMatrix, Ablock::LazyHcat,
-                            B::AbstractVecOrMat, α::Number, β::Number)
-    mul!(res, Ablock*B, I, α, β)
+@views function *(Ablock::LazyHcat, B::AbstractMatrix)
+    res = Ablock.blocks[1] * B[1:size(Ablock.blocks[1], 2), :]  # First multiplication
+    _mul_blocks!(res, Ablock, B, 1)  # Rest of A blocks
 end
 
-mul_hermi(A, B) = Hermitian(A * B)
-function mul_hermi(A::Adjoint{T,<:LazyHcat}, B::LazyHcat) where {T}
-    _mul(A, B; hermitian=Val(true))
+@views function mul!(res::AbstractMatrix, Ablock::LazyHcat, B::AbstractMatrix)
+    mul!(res, Ablock.blocks[1], B[1:size(Ablock.blocks[1], 2), :])  # First multiplication
+    _mul_blocks!(res, Ablock, B, 1)  # Rest of A blocks
+end
+
+@views function LinearAlgebra.mul!(res::AbstractMatrix, Ablock::LazyHcat,
+                                   B::AbstractMatrix, α::Number, β::Number)
+    mul!(res, Ablock.blocks[1], B[1:size(Ablock.blocks[1], 2), :], α, β)
+    _mul_blocks!(res, Ablock, B, α)  # Rest of A blocks
 end
 
 # Perform a Rayleigh-Ritz for the N first eigenvectors.
@@ -375,6 +385,12 @@ end
     AP = zero(X)
     R = zero(X)
     AR = zero(X)
+    # Pre-allocating work arrays
+    new_R = zero(X)
+    new_X = copy(X)
+    new_AX = copy(AX)
+    new_P = zero(X)
+    new_AP = zero(X)
     if B != I
         BR = zero(X)
         # BX was already computed
@@ -388,8 +404,6 @@ end
     nlocked = 0
     niter = 0  # the first iteration is fake
     λs = compute_λ(X, AX, BX)
-    new_X  = X
-    new_AX = AX
     new_BX = BX
     # The full_ arrays contain all the vectors, the others only get the active ones
     full_X  = X
@@ -420,14 +434,14 @@ end
             # wait on updating P because we have to know which vectors
             # to lock (and therefore the residuals) before computing P
             # only for the unlocked vectors. This results in better convergence.
-            new_X  = Y * cX
-            new_AX = AY * cX  # no accuracy loss, since cX orthogonal
+            mul!(new_X, Y, cX)
+            mul!(new_AX, AY, cX)  # no accuracy loss, since cX orthogonal
             new_BX = (B == I) ? new_X : BY * cX
         end
 
         ### Compute new residuals
         @timing "Update residuals" begin
-            new_R = new_AX .- new_BX .* λs'
+            new_R .= new_AX .- new_BX .* λs'
             norms = to_cpu(columnwise_norms(new_R))
             @views resid_history[1 + nlocked: size(new_R, 2) + nlocked, niter+1] .= norms[:]
         end
@@ -491,9 +505,15 @@ end
             # orthogonalize against all Xn (including newly locked)
             ortho!(cP, cX, cX, tol=ortho_tol)
 
+            @views begin
+                new_P = new_P[:, Xn_indices]
+                new_AP = new_AP[:, Xn_indices]
+                B != I && (new_BP = new_BP[:, Xn_indices])
+            end
+
             # Get new P
-            new_P  = Y  * cP
-            new_AP = AY * cP
+            mul!(new_P, Y, cP)
+            mul!(new_AP, AY, cP)
             if B != I
                 new_BP = BY * cP
             else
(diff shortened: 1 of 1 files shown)


src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -105,27 +105,37 @@ Base.adjoint(A::LazyHcat) = Adjoint(A)
...
+function mul_hermi(A::Adjoint{T,<:LazyHcat}, B::LazyHcat) where {T}
+    _mul(A, B; hermitian=Val(true))
+end
+
 Base.:*(A::Adjoint{T,<:LazyHcat}, B::LazyHcat) where {T}       = _mul(A, B)
 Base.:*(A::Adjoint{T,<:LazyHcat}, B::AbstractMatrix) where {T} = A * LazyHcat(B)
 
-@views function *(Ablock::LazyHcat, B::AbstractMatrix)
-    res = Ablock.blocks[1] * B[1:size(Ablock.blocks[1], 2), :]  # First multiplication
+function _mul_blocks!(res::AbstractMatrix, Ablock::LazyHcat, B::AbstractMatrix, α::Number)
```
antoine-levitt: Why not just overload the 5args mul! here?

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -105,27 +105,37 @@ Base.adjoint(A::LazyHcat) = Adjoint(A)
...
     offset = size(Ablock.blocks[1], 2)
     for block in Ablock.blocks[2:end]
-        mul!(res, block, B[offset .+ (1:size(block, 2)), :], 1, 1)
+        mul!(res, block, B[offset .+ (1:size(block, 2)), :], α, 1)
```
antoine-levitt: use true and false for strong 1 and 0 (which the underlying optimization can potentially optimize away)

antoine-levitt: no `@views`?

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -105,27 +105,37 @@ Base.adjoint(A::LazyHcat) = Adjoint(A)
...
     end
     res
 end
 
-function LinearAlgebra.mul!(res::AbstractMatrix, Ablock::LazyHcat,
-                            B::AbstractVecOrMat, α::Number, β::Number)
-    mul!(res, Ablock*B, I, α, β)
+@views function *(Ablock::LazyHcat, B::AbstractMatrix)
+    res = Ablock.blocks[1] * B[1:size(Ablock.blocks[1], 2), :]  # First multiplication
+    _mul_blocks!(res, Ablock, B, 1)  # Rest of A blocks
```
antoine-levitt: Is this just to get the correct size array? Is this really necessary?

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -105,27 +105,37 @@ Base.adjoint(A::LazyHcat) = Adjoint(A)
...
 end
 
-mul_hermi(A, B) = Hermitian(A * B)
-function mul_hermi(A::Adjoint{T,<:LazyHcat}, B::LazyHcat) where {T}
-    _mul(A, B; hermitian=Val(true))
+@views function LinearAlgebra.mul!(res::AbstractMatrix, Ablock::LazyHcat,
+                                   B::AbstractMatrix, α::Number, β::Number)
+    mul!(res, Ablock.blocks[1], B[1:size(Ablock.blocks[1], 2), :], α, β)
+    _mul_blocks!(res, Ablock, B, α)  # Rest of A blocks
 end
```
antoine-levitt, line 137: there's just too many functions here, I'm lost. Need to rationalize all this.

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -375,6 +385,12 @@ end
     AP = zero(X)
     R = zero(X)
     AR = zero(X)
+    # Pre-allocating work arrays
+    new_R = zero(X)
+    new_X = copy(X)
```
antoine-levitt, line 389: why are these copy and the others zero?

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -420,14 +434,14 @@ end
...
             new_BX = (B == I) ? new_X : BY * cX
         end
 
         ### Compute new residuals
         @timing "Update residuals" begin
-            new_R = new_AX .- new_BX .* λs'
+            new_R .= new_AX .- new_BX .* λs'
             norms = to_cpu(columnwise_norms(new_R))
             @views resid_history[1 + nlocked: size(new_R, 2) + nlocked, niter+1] .= norms[:]
         end
```
antoine-levitt, line 446: while I'm here, there's a precondprep!(precon, X) here that should be precondprep!(precon, new_X) (shouldn't change much)

antoine-levitt (commented):
Looks globally fine but there needs to be some cleanup of the mul functions

src/eigen/lobpcg_hyper_impl.jl
abussy: I think it would be misleading, because this is not the standard `C = beta*C + alpha*A*B`. This only goes over the blocks of A, from block # 2. The only reason I introduced this `_mul_blocks!` function is because the pattern is shared by 3 functions, and I try to reduce code duplication.

src/eigen/lobpcg_hyper_impl.jl
abussy: > no @views?

Good catch, I'll fix it.

> use true and false for strong 1 and 0 (which the underlying optimization can potentially optimize away)

I don't really understand. The 5 argument `mul!` is called anyways, shouldn't these kind of optimizations be taken care internally anyways?

src/eigen/lobpcg_hyper_impl.jl
abussy: This is just a refactoring of the existing `*(Ablock::LazyHcat, B::AbstractMatrix)` overload. I've decided to abstract away the loop over blocks with the `_mul_blocks!` function, since the pattern is shared by other functions.

So, it's not really necessary to defined and call `_mul_blocks!`, it merely avoid code duplication.

src/eigen/lobpcg_hyper_impl.jl
abussy: I am not sure what you mean here, sorry.

src/eigen/lobpcg_hyper_impl.jl
abussy: No good reason, I think these can safely all be zero. I'll make the change

src/eigen/lobpcg_hyper_impl.jl
abussy: Ok, I'll change that, and check if I see any difference in the convergence rate of the solver

src/eigen/lobpcg_hyper_impl.jl
antoine-levitt: There's a subtlety: 
```
julia> false*NaN
0.0

julia> 0*NaN
NaN
```
so `false` is a "strong zero" for which more optimizations are valid (false * x = zero(x) for all floats). I'm not sure if they're actually done, but it's good practice to expose as much semantic information as possible to downstream implementations.

src/eigen/lobpcg_hyper_impl.jl
antoine-levitt: Sorry I misread. Then give them a better name like _mul_from_second! or something (it would be even better if they were not needed)

src/eigen/lobpcg_hyper_impl.jl
antoine-levitt: Why can't we just make a zero and just accumulate all blocks?

src/eigen/lobpcg_hyper_impl.jl
antoine-levitt: I just checked there's no such thing as a "strong one" but might as well just use true and false anyway

src/eigen/lobpcg_hyper_impl.jl
abussy: That should work, I'll give it a try. Maybe we can get rid of some code.

src/eigen/lobpcg_hyper_impl.jl
abussy: Actually, there is a good reason: at the first iteration, `new_X` and `new_AX` are not calculated, but used nonetheless. Therefore, they need to be initialized at the values of `X` and `AX`.

src/eigen/lobpcg_hyper_impl.jl
antoine-levitt: Sorry that's more than your bargained for, but this piece of code is pretty delicate (it's definitely the most tricky numerical code I've ever written), there's a lot going on, so it's great if the logic can be as transparent as possible. I can help if needed.

src/eigen/lobpcg_hyper_impl.jl
antoine-levitt: You can also just move the         # Update all X, even newly locked
block into the if niter > 0 block? Or even more?

antoine-levitt: mh no, that would be tricky, just leave it I think

abussy:
With that latest commit, I define the 3-argument `mul!` and `Base.:*` as a function of the 5-argument mul!. This should clarify things, and reduce code duplication.

I also use `true` and `false` instead of 1 and 0, although I don't think it makes the code any clearer.

I'd like to run a few tests to check if there is any performance impact still.

antoine-levitt:
Nice!

antoine-levitt:
I took a look at the docs and tried to clarify it, I pushed to your branch directly (hope that's OK)

abussy:
Performance looks good.

src/eigen/lobpcg_hyper_impl.jl
mfherbst: I think Antoine wants some comments why e.g. here the `mul_hermi` and `_mul` functions exist at all.

src/eigen/lobpcg_hyper_impl.jl
```diff
@@ -69,24 +76,23 @@ function LazyHcat(arrays::AbstractArray...)
...
     m = sum(size(block, 2) for block in A.blocks)
     (n, m)
 end
-
 Base.Array(A::LazyHcat)   = stack(A.blocks)
 Base.adjoint(A::LazyHcat) = Adjoint(A)
 
-# Computes A*B matrix product for LazyHcat type. Special case if product is assumed to be Hermitian
+# Computes A*B matrix product when B is a LazyHcat and A is a LazyVcat (adjoint of LazyHcat).
+# Special case if product is known to be Hermitian
```
mfherbst: E.g. here one could add a one-line rationalisation why this exists.

mfherbst (commented):
Very nice. Some more comments and we are good to go.

abussy:
I added an additional comment on `_mul` and `mul_hermi`, but I feel like it's redundant given Antoine's last commit and the function signatures themselves.

Note: up to date with current master

PR #1300, opened 2026-04-11
Add a CLAUDE.md

antoine-levitt:
I just asked Claude to generate something and it looks reasonable. I told it to add specific instructions about testing and being patient.


Technici4n:
Reminder to consider merging with squash for a cleaner history 😄

antoine-levitt:
Woops, for some reason it didn't ask me to do that, sorry.

PR #1301, opened 2026-04-11
Add non-periodic system support with truncated Coulomb

antoine-levitt:
…006)

Implement isolated/semi-periodic electrostatics via the truncated Coulomb method of Rozzi et al. (Phys. Rev. B 73, 205119, 2006) for 0D and 2D slab geometries (1D wire left as error stub for a future PR).

Key changes:
- `Model` gains a `periodicity` field: a 3-tuple where each entry is `true` (fully periodic), `:wavefunctions_only` (plane-wave wavefunctions with isolated electrostatics), or `false` (reserved for future fully- isolated wavefunctions; currently identical to `:wavefunctions_only`). The constructor validates that isolated directions are orthogonal to periodic ones.
- New `src/terms/truncated_coulomb.jl`: `truncated_coulomb_radius` and `truncated_coulomb_fourier` implement the Rozzi kernels for 0D (spherical cutoff) and 2D slab (Bessel-based) cases.
- `Hartree`: `compute_poisson_green_coeffs` uses the truncated kernel instead of 4π/|G|² when `!is_fully_periodic_electrostatics(model)`.
- `AtomicLocal`: `compute_local_potential` applies a per-atom correction via `add_truncated_coulomb_ionic_correction!` that replaces the standard -Z/|G|² long-range tail with the truncated kernel (including the finite G=0 contribution V_short(0) - Z·v_c(0)).
- `Ewald`: non-periodic models use a bare direct pair sum (`energy_forces_ewald_direct`) instead of the Ewald summation — no periodic images, no compensating background.
- `PspCorrection`: returns zero energy for non-periodic models to avoid double-counting the V_short(G=0) term that is now embedded in V_ion(G=0).
- New helper predicates exported: `is_wavefunctions_periodic`, `is_electrostatics_periodic`, `n_periodic_electrostatics`, `is_fully_periodic_electrostatics`.
- `todict!` serialises the `periodicity` field.
- New test file `test/truncated_coulomb.jl` (5 test items): model periodicity field construction/validation, kernel values, direct Ewald pair sum, ionic potential G=0 value, and isolated-dipole SCF convergence.

Known limitations (TODOs in code):
- 1D wire Coulomb truncation not implemented (raises error).
- Forces for the truncated-Coulomb ionic correction not implemented.
- Phonon/dynmat and q≠0 ionic corrections not implemented.

https://claude.ai/code/session_018Y7AgeMV3ZTCA1G4CLZz4F

### DIFF
--- a/src/DFTK.jl
+++ b/src/DFTK.jl
@@ -69,6 +69,8 @@ include("SymOp.jl")
 
 export Smearing
 export Model
+export is_wavefunctions_periodic, is_electrostatics_periodic
+export n_periodic_electrostatics, is_fully_periodic_electrostatics
 export FFTGrid
 export MonkhorstPack, ExplicitKpoints
 export PlaneWaveBasis
--- a/src/Model.jl
+++ b/src/Model.jl
@@ -21,6 +21,20 @@ struct Model{T <: Real, VT <: Real}
     unit_cell_volume::T
     recip_cell_volume::T
 
+    # Boundary conditions along each of the three lattice directions.
+    # Each entry is one of:
+    #     true                  -- fully periodic
+    #     :wavefunctions_only   -- wavefunctions are periodic (standard plane-wave
+    #                              Bloch basis), but the electrostatics is treated
+    #                              as isolated via the truncated Coulomb method of
+    #                              Rozzi et al. (Phys. Rev. B 73, 205119 (2006))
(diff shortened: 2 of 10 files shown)


antoine-levitt:
That's a mistake I meant to develop it locally

PR #1306, opened 2026-04-17
📚 `tutorial.jl`: remove duplication

mbercx:
"There is also a Mathematical Tutorial." is said in both the text and the tip.


mfherbst:
Thanks !

Issue #1310, opened 2026-04-24
Phonon +q instead of -q?


Technici4n:
The dynamical matrix is supposed to be a mixed derivative wrt one monochromatic displacement at $+q$ and one monochromatic displacement at $-q$. From the code it seems to me that we compute the $\delta \rho$ at momentum $+q$. So for example I would expect something like $\delta V_{-q} \delta \rho_{+q}$ for the local term, but instead I see that the potential change is also computed at +q?

https://github.com/JuliaMolSim/DFTK.jl/blob/b6b1f7d2a9c8e48223b9737109c14ffb0a63b297/src/terms/local.jl#L193-L197

Maybe we are lucky by chance and we are actually computing $D(-q)$ consistently by accident (I think this could be the case but I am not 100% sure).

antoine-levitt:
Yeah I remember this was very confusing. I think we didn't want to take into account time reversal symmetry explicitly so we wanted to define the dynamical matrix as mapping dR to dF under a complex displacement dR. But then that mixes q and -q anyway because of the psibar. In the end there's a trick that they use in Baroni and I don't remember if we use the same one. Maybe @epolack remembers? We definitely should have written this down at the time...

antoine-levitt:
Right the trick was that the delta rho is psibar dpsi + dpsibar psi. We don't like the second one because it couples the +q and -q phonon. So what we do is to compute it as 2 psibar dpsi. I think this is in fact fine because of a funny complex analysis reason. Surely we must have written this down somewhere...

antoine-levitt:
So the ground truth is the supercell. There, the map χ0 = δV -> δψ -> δρ is a priori a translation invariant map from real to real. The map δV -> δψ is complex-linear, but *not* δψ->δρ (because δρ = ψbar δψ + δψbar ψ). So strictly speaking the Bloch theorem does *not* apply (because it requires complex vector spaces). We can't a priori just take δV = δVq e^iqx to compute δρ = δρq e^iqx: if we do that, we will have to compute δψ at +q and -q. However (plus or minus signs, factors and multiple bands),

χ0 = ∑_k,q' |ψkbar ψk+q'><ψkbar ψk+q'| / (εk-εk+q').

So this actually extends χ0 to a C-linear map (there's nothing preventing me to take a complex δV in this formula). If we hit it with δV = δVq e^iqx, this excites only the q'=q term, and 

[...]

PR #1311, opened 2026-04-26
Add NLCC support for phonon computations

Technici4n:
Cleaner than I expected. Nice framework that you built @epolack and @antoine-levitt, especially the automated testing!


Technici4n:
I had to remove `@inferred` from a test unfortunately, but I just ran the full GPU (CUDA) test suite and it still works, so that shouldn't be a problem.

examples/phonons.jl
```diff
@@ -8,7 +8,7 @@
 #     calculations. Some of the limitations are:
 #     - symmetries must be disabled (pass `symmetries=false` to the model)
 #     - only LDA functionals are supported
-#     - non-linear core corrections from the pseudopotentials are not supported
+#     - q-points must be commensurate with the k-point grid of the SCF calculation
```
antoine-levitt: That's not really a limitation, it's kind of intrinsic to phonons

src/postprocess/phonon.jl
```diff
@@ -66,7 +66,10 @@ end
 
 @doc raw"""
 Compute the dynamical matrix in the form of a ``3×n_{\rm atoms}×3×n_{\rm atoms}`` tensor
-in reduced coordinates.
+in reduced coordinates:
+```
+dynmat[β, t, α, s] = ∂²E/∂u_sα(-q)∂u_tβ(q).
+```
```
antoine-levitt, line 72: really not a fan of this interpretation, defining it as the derivative of the R -> F map seems much simpler to me...

src/postprocess/phonon.jl
```diff
@@ -78,7 +81,7 @@ in reduced coordinates.
     δψs = [zero.(ψ) for _ = 1:3, _ = 1:n_atoms]
     for s = 1:n_atoms, α = 1:basis.model.n_dim
         # Get δH ψ
-        δHψs_αs = compute_δHψ_αs(basis, ψ, α, s, q)
+        δHψs_αs = compute_δHψ_αs(basis, ψ, α, s, q; ρ)
```
antoine-levitt, line 84: We really need this state refactor business...

src/terms/xc.jl
```diff
@@ -482,6 +481,71 @@ function apply_kernel(term::TermXc, basis::PlaneWaveBasis{T}, δρ::AbstractArra
...
+    end
+
+    dynmat = zeros(complex(T), 3, n_atoms, 3, n_atoms)
+    # Assemble first term directly
+    for s = 1:n_atoms, α = 1:n_dim
+        # We actually receive ∂ρ/∂u_sα(q) (note: +q and not -q),
+        # however since the potential is real, we conjugate to obtain the -q term:
+        # δV =       Kxc (∂ρ/∂u_sα(-q) + ∂ρcore/∂u_sα(-q))
+        #    = conj( Kxc (∂ρ/∂u_sα( q) + ∂ρcore/∂u_sα( q)) )
+        δV_αs = conj.(apply_kernel(term, basis, δρs[α, s] + δρcores[α, s]; ρ, q))
```
antoine-levitt, line 511: Haven't followed the complete derivation but I'm scared by this line; the whole point of the explanation I gave in https://github.com/JuliaMolSim/DFTK.jl/issues/1310 is that we never conjugate q stuff. Are you sure about this conj?

antoine-levitt (commented):
Haven't checked everything but looks reasonable!

examples/phonons.jl
Technici4n: I'm not so sure. What you need is a way to compute $H_{k+q}$ for an arbitrary q, as well as the corresponding $\psi_{n,k+q}$. I assume it can be done by interpolating the bands (i.e. `compute_bands`). Maybe this "just works". QE should be able to do it. I'm not so sure Hellmann-Feynman is still valid though...

src/postprocess/phonon.jl
Technici4n: That's basically what it is since dE/du is the force. But with an extra q dependency here. The main reason I put the formula there is to know in which order to write the dynmat entries, otherwise you can get the transpose by mistake.

src/terms/xc.jl
Technici4n: Yes 100% sure. It's also there in the local term, just hidden by Parseval's theorem. For a real function (the density is real) the conjugate should always be fine. You only get problems if you directly conjugate the psis since those are complex. I am not sure how DFPT works for spin orbit coupling or magnetic fields. Surely Dal Corso has some paper about it, but I haven't read it.

examples/phonons.jl
Technici4n: Okay, I removed this remark. :D

Technici4n:
@antoine-levitt do you want to review this or should I ask Michael? :D

src/postprocess/phonon.jl
antoine-levitt: right but we don't use this notation elsewhere so maybe deltaF[alpha, s] = dynmat[] deltaR[]...? Any particular reason why it's beta alpha rather than alpha beta?

src/terms/xc.jl
```diff
@@ -482,6 +481,71 @@ function apply_kernel(term::TermXc, basis::PlaneWaveBasis{T}, δρ::AbstractArra
...
 
+@views function compute_dynmat(term::TermXc, basis::PlaneWaveBasis{T}, ψ, occupation;
+                               ρ, δρs, q=zero(Vec3{T}), kwargs...) where {T}
+    isnothing(term.ρcore) && return nothing
+
+    # Hellmann-Feynman:
+    # ∂E/∂u_tβ(q) = ∫ Vxc ∂ρcore/∂u_tβ(q)
+    # Differentiate again wrt. u_sα(-q) to get the dynamical matrix element:
+    # ∂²E/∂u_sα(-q)∂u_tβ(q) = ∫ Kxc (∂ρ/∂u_sα(-q) + ∂ρcore/∂u_sα(-q)) (∂ρcore/∂u_tβ(q))
+    #                       + ∫ Vxc ∂²ρcore/∂u_sα(-q)∂u_tβ(q)
```
antoine-levitt, line 492: I'm still not very comfortable with this d2E business, and it contrasts with the way everything else is written (as dR -> dF). Can't you write it like that?

src/terms/xc.jl
```diff
@@ -482,6 +481,71 @@ function apply_kernel(term::TermXc, basis::PlaneWaveBasis{T}, δρ::AbstractArra
...
+        dynmat[β, s, α, s] += sum(Vxc .* δ²ρcore) * basis.dvol
+    end
+
+    dynmat
+end
+
+function compute_δHψ_αs(term::TermXc, basis::PlaneWaveBasis, ψ, α, s, q; ρ)
+    isnothing(term.ρcore) && return nothing
+
+    # With an NLCC, an atom displacement triggers a change in the XC potential.
```
antoine-levitt: that comment is not super helpful, just remove?

src/density_methods.jl
```diff
@@ -143,41 +145,49 @@ end
...
-    form_factors, iG2ifnorm = atomic_density_form_factors(basis, method)
+                                      coefficients=ones(T, length(basis.model.atoms)),
+                                      positions=basis.model.positions,
+                                      q=zero(Vec3{T})) where {T}
+    form_factors, iG2ifnorm = atomic_density_form_factors(basis, method; q)
 
     # Pre-allocation of large arrays for GPU efficiency
-    Gs = G_vectors(basis)
-    ρ = to_device(basis.architecture, zeros(Complex{T}, length(Gs)))
+    Gqs = map(G -> G+q, G_vectors(basis))
```
antoine-levitt, line 154: I think we call those Gpk or something somewhere else?

antoine-levitt (commented):
LGTM modulo the comments

src/terms/xc.jl
Technici4n: This is how Baroni writes it. I don't like to just use dF and dR as it doesn't make it clear which has q and which has -q. This derivation also directly matches the implementation

src/postprocess/phonon.jl
Technici4n: I put it in that order such that the RHS matches Baroni, and it also makes sense to me with Julia's column-major indexing to write it like that, no?

antoine-levitt: I mean it's usually xi = Aij xj, not xj = Aji xi

src/terms/xc.jl
antoine-levitt: What I don't understand is what's up with the -q? The supercell mapping from dR to dF is cell-periodic, and therefore bloch theorem applies so a displacement dR_q e^iqx produces a force dF_q e^iqx, and we compute the mapping dR_q -> dF_q. No need to involve -q. Should we zoom to discuss?

test/guess_density.jl
```diff
@@ -29,14 +29,14 @@
 
     @testset "Atomic $(string(typeof(method)))" for (method, elements) in zip(methods, elements)
         basis = build_basis(elements, :none)
-        ρ = @inferred guess_density(basis, method)
+        ρ = guess_density(basis, method)
```
mfherbst, line 32: Just a remark that we had real trouble with type instabilities here, so this is why the `@inferred` tests exist. So this was not a GPU thing but a general performance thing. I think it's a bad sign you need to remove it again and we should think very carefully if this is the right thing to do (instead of fixing the issue)

Technici4n: It was added in this PR, so it's presumably safe to remove as long as DFTK still runs on the GPU: https://github.com/JuliaMolSim/DFTK.jl/pull/850.

A proper fix is to make the q==0 case encoded in the type system (i.e. with a different type), which I would consider out of scope for this PR.

src/density_methods.jl
Technici4n: The naming is not super consistent. I used `Gqs` because that is what the local term (my inspiration) uses.

src/terms/xc.jl
Technici4n: Sure, since a displacement dR_q e^iqR produces a force dF_q e^iqR, you apply a displacement dR_q e^iqR, compute the corresponding force, and then multiply by e^-iqR to recover dF_q. So what you suggest is computing the force "normally" with d/du and then multiplying by e^-iqR. What I suggest is directly computing the force with the phase removed with d/du(-q). Does this make sense? The two should be equivalent, but I prefer the latter approach since I find the required code changes more obvious that way.

If you want we can have a call, maybe tomorrow somewhere between 13:30 and 16:00, or on Friday somewhere between 13:30 and 16:00?

src/postprocess/phonon.jl
Technici4n: I tried to write it like deltaF[alpha, s] = dynmat[] deltaR[] as well, but I had to give up because it's really confusing and unusual (it's difficult to keep track of what is R-periodic and what is not).

Technici4n:
FWIW it would be good if this PR would be merged rather soon as I have follow-up PRs that depend on this one.

test/guess_density.jl
mfherbst: Well, probably (did you test ?).

In any case at least make sure we document this clearly, because with this sort of stuff you never know if does not end up biting us back some day. The density and guess densities go quite deep into the code and not having the type inferrable could easily have nasty performance regressions. So please add a comment about this here (i.e. that we had the `@inferred` here but had to remove it because the `q` handling breaks type inferability) and in the atomic superposition functions themselves with the proposed solution to have a `ZeroVec3{T}` type.

Technici4n: Hmm yeah making all uses of ρ non-inferrable is quite bad. Maybe we have to bite the bullet and actually encode `q==0` in the type system

Technici4n:
Needs to account for this bugfix: https://github.com/JuliaMolSim/DFTK.jl/pull/1370

PR #1312, opened 2026-04-27
Add some documentation to phonons

antoine-levitt:
Just linked to https://github.com/JuliaMolSim/DFTK.jl/issues/1310


antoine-levitt:
Fixes #1310

src/densities.jl
```diff
@@ -77,8 +77,11 @@ end
 
     # The variation of the orbital ψ_k defined in the basis ℬ_k is δψ_{[k+q]} in ℬ_{[k+q]},
     # where [k+q] is equivalent to the basis k+q (see find_equivalent_kpt).
-    # The perturbation of the density
-    #   |ψ_{n,k}|² is 2 ψ_{n,k} * δψ_{n,k+q}.
+    # The perturbation of the density |ψ_{n,k}|² is
+    # 2 ψ_{n,k} * δψ_{n,k+q}
+    # and not the real part of that; for a discussion see
+    # See https://github.com/JuliaMolSim/DFTK.jl/issues/1310#issuecomment-4318880401
```
Technici4n, line 83: I'm not sure this is so helpful like that. I'll have to read your derivation again and see if it makes sense to me, maybe we could inline it in DFTK if so. The classic Baroni derivation is a bit magical and of course we'd rather not assume TRS, agreed there.

antoine-levitt:
Yeah this explanation is bad anyway, closing

PR #1313, opened 2026-04-30
Safe GPU-aware MPI communication

abussy:
Before MPI communication of GPU buffers, the device must be synchronized. If not, there is a real risk of race condition because CPU and GPU run asynchronously. This has no effect on CPU performance. On the GPU, this cannot be avoided.

For reference, here is a [PR](https://github.com/JuliaParallel/MPI.jl/pull/949) updating the relevant documentation in MPI.jl. The problem has been known for a while, but never properly documented. The PR also links to a slack thread where the problem is discussed.

### DIFF
--- a/src/common/mpi.jl
+++ b/src/common/mpi.jl
@@ -11,20 +11,24 @@ mpi_master(comm::MPI.Comm) = (MPI.Init(); MPI.Comm_rank(comm) == 0)
 @deprecate mpi_nprocs() mpi_nprocs(MPI.COMM_WORLD)
 @deprecate mpi_master() mpi_master(MPI.COMM_WORLD)
 
-# Wrappers around standarf MPI operations. Avoid using MPI.COMM_WORLD unless strictly necessary.
-mpi_sum(  arr, comm::MPI.Comm) = MPI.Allreduce( arr,   +, comm)
-mpi_sum!( arr, comm::MPI.Comm) = MPI.Allreduce!(arr,   +, comm)
-mpi_min(  arr, comm::MPI.Comm) = MPI.Allreduce( arr, min, comm)
-mpi_min!( arr, comm::MPI.Comm) = MPI.Allreduce!(arr, min, comm)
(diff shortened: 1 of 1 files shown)


Technici4n (approved):
Ah yes I remember this

src/common/mpi.jl
```diff
@@ -11,20 +11,24 @@ mpi_master(comm::MPI.Comm) = (MPI.Init(); MPI.Comm_rank(comm) == 0)
...
-# Wrappers around standarf MPI operations. Avoid using MPI.COMM_WORLD unless strictly necessary.
-mpi_sum(  arr, comm::MPI.Comm) = MPI.Allreduce( arr,   +, comm)
-mpi_sum!( arr, comm::MPI.Comm) = MPI.Allreduce!(arr,   +, comm)
-mpi_min(  arr, comm::MPI.Comm) = MPI.Allreduce( arr, min, comm)
-mpi_min!( arr, comm::MPI.Comm) = MPI.Allreduce!(arr, min, comm)
-mpi_max(  arr, comm::MPI.Comm) = MPI.Allreduce( arr, max, comm)
-mpi_max!( arr, comm::MPI.Comm) = MPI.Allreduce!(arr, max, comm)
+# GPU-aware MPI requires device synchronization to avoid race conditions. No overhead on CPU.
+sync(arr::AbstractGPUArray) = synchronize_device(architecture(arr))
+sync(arr) = nothing
```
mfherbst, line 16: I'd move this to architecture.jl and avoid introducing a second function (just add another set of methods to `synchronize_device`). Comment makes sense to have here, though.

abussy: It was intended as a local utility function for compactness, because writing `synchronize_device(architecture(arr))` everywhere feels heavy. I guess I could move those to `architecture.jl` and rename them `synchronize_device` for multiple dispatch, but these aren't really generic functions.

I am in favor of leaving the code as is, but I can make the change is that's the consensus.

PR #1314, opened 2026-05-01
Warn on TRS violation in compute_δρ at q ≠ 0

hsugawa8651:
Per @antoine-levitt's [comment in #1310](https://github.com/JuliaMolSim/DFTK.jl/issues/1310#issuecomment-4333054728) ("We need to add an assert and explain this better"), this PR adds a soft `@warn` at the entry of `compute_δρ` when `q ≠ 0` and `basis.model.spin_polarization != :none`.                                                                  
                                              
I went with `@warn maxlog=1` rather than `@assert` because the existing                                   phonon test (`test/phonon/lowlevel.jl`) uses `spin_polarization=:collinear`                                and a hard assert would break runtests. Happy to change it to `@assert` if you prefer. 

Refs #1310

### DIFF
--- a/src/densities.jl
+++ b/src/densities.jl
@@ -58,6 +58,18 @@ end
 @views @timing function compute_δρ(basis::PlaneWaveBasis{T}, ψ, δψ, occupation,
                                    δoccupation=zero.(occupation);
                                    occupation_threshold=zero(T), q=zero(Vec3{T})) where {T}
+    # The phonon δρ formula below uses `2 ψ̄ δψ` instead of `ψ̄ δψ + δψ̄ ψ`,
+    # implicitly relying on time-reversal symmetry to fold in the -q contribution.
+    # For systems with broken TRS (spin polarization) this is silently incorrect.
+    # See #1310 for the full discussion.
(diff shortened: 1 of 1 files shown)


Technici4n:
There is no problem with collinear spin per se. This usually only concerns magnetism or spin-orbit coupling. (Right?)

hsugawa8651:
Thanks @Technici4n: 
> This usually only concerns magnetism or spin-orbit coupling. (Right?)

Yes, that matches my understanding. 
I can narrow the condition to `:full || (:collinear && any(!iszero, magnetic_moments))`. 
To address @antoine-levitt's `@assert` request in #1310 without breaking users with existing magnetic-phonon workflows, what about exposing a `DFTK.TRS_PHONON_CHECK::Ref{Symbol}` (`:warn` default, `:error`, `:silent`)? 
Default keeps current behavior, and DFTK's CI / strict callers can opt into `:error`. 
Happy to push that combined patch if you both agree.

antoine-levitt:
Hi, thanks for the PR! However this needs a more comprehensive approach to get done properly, I'll get around to it.

PR #1315, opened 2026-05-03
Add a CONTRIBUTING.md guide (and an example on the README)

antoine-levitt:


Technici4n (approved):
Indeed, the AI clause is now unfortunately necessary.

README.md
```diff
@@ -42,6 +42,34 @@ Our performance is of the same order of magnitude
 as other plane-wave DFT packages and  systems of 1000 electrons
 can be routinely treated.
 
+Basic usage is demonstrated below:
+```julia
```
mfherbst, line 46: I'm hesitant with this, because it's better to have no example in the readme than a broken example ... and unlike the tutorial (to which we can link !) this code will not be tested in CI.

CONTRIBUTING.md
```diff
@@ -0,0 +1,19 @@
+# Contributions
+Contributions to the code in any form is very welcome,
+just [submit a pull request](https://github.com/JuliaMolSim/DFTK.jl/pulls)
+on github. If you want to contribute but are unsure where to start, take a look
+at the list of issues tagged [good first issue](https://github.com/JuliaMolSim/DFTK.jl/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22)
+(relatively easy tasks suitable for newcomers) or [help wanted](https://github.com/JuliaMolSim/DFTK.jl/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22)
+(more sizeable but well-defined and isolated).
+Don't hesitate to ask for help, through github,
+email, [JuliaMolSim zulip][chat-url] or [Matrix chat][matrix-url].
```
mfherbst, line 12: This is likely broken

mfherbst (commented):
I agree with starting the contributing guide. See my comment on the example, though.

README.md
antoine-levitt: I think if we break this we have bigger problems... It's very nice to have an example of basic usage on the readme of packages, and it's a copy paste of the silicon example which is tested

CONTRIBUTING.md
antoine-levitt: Do we remove both?

mfherbst: Just paste the actual urls from the README ?

README.md
mfherbst: Ok, we just have to remember this

PR #1316, opened 2026-05-07
Add breaks_TRS term predicate and guard compute_dynmat

antoine-levitt:
## Summary

- Adds a `breaks_TRS` dispatch on term builders, mirroring the existing `breaks_symmetries`. Default is `false`; `Magnetic` and `Anyonic` are marked as breaking TRS.
- `compute_dynmat` now asserts that no term in the model breaks TRS. The phonon response solver folds the +q and −q Sternheimer equations into a single one at +q (with a factor of 2), which is only valid under TRS — see #1310 and Dal Corso, https://arxiv.org/abs/1906.11673.
- Documents the derivation of the dropped \`δψ̄·ψ\` term in \`compute_δρ\` from scratch (without going through χ₀): tracks the four Fourier components of \`δρ\`, picks out the +q part, and uses TRS (\`ψ_{−k} = ψ_k*\`) to identify the second +q contribution with the first, justifying the factor 2.

## Test plan

- [ ] `Pkg.test("DFTK"; test_args=["minimal"])` passes
- [ ] Existing phonon tests still pass (no Magnetic / Anyonic terms in those models)
- [ ] Confirm the assertion fires on a phonon-style call with a `Magnetic` term in the model

### DIFF
--- a/src/densities.jl
+++ b/src/densities.jl
@@ -55,37 +55,6 @@ using an optional `occupation_threshold`. By default all occupation numbers are 
 end
 
 # Variation in density corresponding to a variation in the orbitals and occupations.
-#
-# Density response and the "factor of 2" / dropped δψ̄·ψ term
-# ---------------------------------------------------------
-# A real perturbation at wavevector q has two complex Fourier components,
-# δV(r) = δV_q e^{iq·r} + δV_q* e^{−iq·r}. Each Bloch state ψ_k acquires a response
-# with two pieces: δψ_k^{(+)} at momentum k+q (sourced by δV_q) and δψ_k^{(−)} at
-# k−q (sourced by δV_q*). Differentiating ρ = ∑_k f_k |ψ_k|² gives
-#   δρ = ∑_k f_k (ψ̄_k δψ_k + δψ̄_k ψ_k).
-# Picking out the +q Fourier component of δρ, by counting momenta:
-#   • ψ̄_k δψ_k^{(+)}   contributes at  −k + (k+q) = +q   ← what we compute below
-#   • ψ̄_k δψ_k^{(−)}   contributes at  −q                (drops out of δρ_{+q})
-#   • δψ̄_k^{(+)} ψ_k   contributes at  −q                (drops out)
-#   • δψ̄_k^{(−)} ψ_k   contributes at  +q                ← *not* computed directly
-# So naively δρ_{+q} would need both δψ_k^{(+)} (Sternheimer at +q) and δψ_k^{(−)}
-# (Sternheimer at −q).
-#
-# Time-reversal symmetry rescues us. Under TRS the Bloch state ψ_{−k} = ψ_k* is
-# also occupied with the same energy, and conjugating its Sternheimer equation
-# at +q,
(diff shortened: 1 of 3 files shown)


src/densities.jl
```diff
@@ -55,6 +55,37 @@ using an optional `occupation_threshold`. By default all occupation numbers are
 end
 
 # Variation in density corresponding to a variation in the orbitals and occupations.
+#
+# Density response and the "factor of 2" / dropped δψ̄·ψ term
```
Technici4n: This is AI generated, no? While technically true it seems to have lower information density than something a human would write. :/

src/terms/terms.jl
```diff
@@ -50,6 +50,12 @@ include("Hamiltonian.jl")
 # is invalid)
 breaks_symmetries(::Any) = false
 
+# breaks_TRS on a term builder answers true if this term breaks
+# time-reversal symmetry. Some algorithms (notably the phonon response
+# solver, see compute_dynmat) rely on TRS to avoid solving Sternheimer
+# equations at both +q and -q (cf. discussion in JuliaMolSim/DFTK.jl#1310).
+breaks_TRS(::Any) = false
```
Technici4n: Not sure about the acronym vs the full version.

src/postprocess/phonon.jl
```diff
@@ -71,6 +71,13 @@ in reduced coordinates.
 @timing function compute_dynmat(basis::PlaneWaveBasis{T}, ψ, occupation; q=zero(Vec3{T}),
                                 ρ=nothing, ham=nothing, εF=nothing, eigenvalues=nothing,
                                 kwargs...) where {T}
+    # The phonon response solver assumes time-reversal symmetry: the trick used
+    # to compute δρ from a single Sternheimer equation at +q (instead of one at
+    # +q and one at -q) is only valid under TRS. See the discussion in
+    # JuliaMolSim/DFTK.jl#1310 and Dal Corso, https://arxiv.org/abs/1906.11673.
+    @assert !any(breaks_TRS, basis.model.term_types) (
```
Technici4n: It seems reasonable to have it here to fail-fast. But it should also be in compute_δρ where the assumption is made. :)

antoine-levitt: I wondered about that but the assumption is actually pretty pervasive. But it's better to have all the info in the same place, so not super sure where this all should go. Maybe a big comment on top of the phonon.jl file?

src/densities.jl
antoine-levitt: Yeah this is draft, I'm planning on a rewrite

antoine-levitt:
Ended up rewriting the whole comment, AI is not great for this type of non-trivial explanation. Ready to review now.

Technici4n (approved):
I like this, thanks

src/terms/terms.jl
```diff
@@ -50,6 +50,12 @@ include("Hamiltonian.jl")
 # is invalid)
 breaks_symmetries(::Any) = false
 
+# breaks_TRS on a term builder answers true if this term breaks
```
Technici4n: Name should be updated here

PR #1317, opened 2026-05-08
Use time reversal symmetry in kpoint reduction 

antoine-levitt:
This is fully written by Claude, prodded by me. Looks reasonable and tests pass, but still needs careful review.


antoine-levitt:
This is my first substantial PR for AI code intended to get into dftk so if it's of interest to anybody, here's my workflow. I started by brainstorming with opus, telling it to write a plan.md file. Then I switched to a fresh sonnet session and told it to keep notes in a progress.md file in case the session got aborted. Running tests went fine once I made sure julia setup was OK (julia --project). I used remote control to control it from my phone. Whole thing took less than a day split over a few sessions. The most annoying bit was the stop and go part: approving commands, waiting for it to think or tests to run, etc. I should use auto approve, julia mcp with revise and notifications. 

Things it's great all : coding style, git stuff, tests (designing, running and debugging), getting classical formulas by itself, knowing about physics . Nonzero chance it's so good because it knows quantum espresso source code by heart...

Things it's bad at : getting the right fix as opposed to the quickest fix, inserting itself in existing code rather than adding new hacks, limiting code size (it loves copy paste), sometimes getting stuck (eg its commits sometimes hung, I don't know why). Also an annoying tendency to prioritize what it's doing atm as opposed to the broader context (eg adding a 10 lines comment related to trs inside a nontrivial function doing something else). All this possibly fixable with more explicit instructions.

antoine-levitt:
Oh and all in all it used up about a third of my weekly quota (probably mainly because I switched back to opus at the end)

antoine-levitt:
There's a subtlety with spin which this PR somewhat conflates. The actual physical TRS (the one that reverses the sign of p times sigma) is (0 1;-1 0) times the conjugation. This symmetry is broken in ferromagnetic states, but is preserved in one type of systems: anti-ferromagnets (TRS plus translation). In ferromagnetic states however, the symmetry of just conjugation (not acting on spins) is preserved. So the most general SymOp would allow for both full TRS and also just conjugation; in fact, with noncollinear spin it would allow general SU(2) matrices. I think we should go with just conjugation (so not really TRS) for now and leave spin alone, which would simplify this PR a bit. Longer term when we have noncollinear spins we should allow full SU(2).

src/symmetry.jl
```diff
@@ -86,20 +87,17 @@ on each of the atoms. The symmetries are determined using spglib.
...
-            # Keep only the symmetries that don't flip the spins
-            rotations[spin_flips.==1], translations[spin_flips.==1]
-            # Taking into account the symmetries that flip spins here would mean detecting 
-            # antiferromagnetic order (where the down orbitals are the up orbitals 
-            # translated by half a cell) and optimizing by computing only the up channel of 
-            # the orbitals and building the density by symmetry. So, the density has two 
-            # components, but the orbitals have only one spin channel.
-            # This would cut runtime by a factor 2.
+            # Keep only spin-preserving (spin_flips==1) symmetries. Spin-flipping symops
+            # (AFM order) are not yet exploited.
```
antoine-levitt, line 98: add a TODO

src/SymOp.jl
```diff
@@ -30,41 +30,54 @@ function is_approx_integer(r; atol=100eps(eltype(r)))
 end
 
 struct SymOp{T <: Real}
-    # (Uu)(x) = u(W x + w) in real space
+    # (Uu)(x) = u(W x + w) in real space           (θ = +1, unitary)
+    # (Uu)(x) = conj(u(W x + w)) in real space     (θ = -1, antiunitary / conjugation)
     W::Mat3{Int}
     w::Vec3{T}
```
antoine-levitt, line 38: add a comment that eventually this should include a SU(2) matrix acting on spin

docs/src/developer/symmetries.md
```diff
@@ -10,13 +10,23 @@ W \mathcal{C} + w = \mathcal{C}.
...
+   \overline{u\left( W x + w \right)} & \text{if } θ = -1.
+ \end{cases}
 ```
 We assume that the atomic potentials are radial and that any self-consistent potential
-also respects this symmetry, so that ``U`` commutes with the Hamiltonian.
+also respects these symmetries, so that ``U`` commutes with the Hamiltonian.
+Conjugation (``θ = -1``) is a valid symmetry whenever no term in the
+Hamiltonian breaks it (no external magnetic field, no anyonic statistics, etc.).
+It acts diagonally on spin channels: each spin channel maps independently under
+``k \to -k`` with complex conjugation, without mixing the two channels.
```
antoine-levitt, line 29: add a sentence that eventually this should include spin symmetries but doesn't yet

antoine-levitt:
Ready for review

Technici4n:
After further thinking, I am not sure that we want the TRS symmetry to be included in the symmetry operations since it usually comes for free, i.e. there is no need to symmetrize densities, forces, etc...

antoine-levitt:
Symmetries is there to list the symmetries of the system: whether we actually use them for computation is kind of besides the point. But also yes we do want to symmetrize explicitly wrt TRS in case the BZ is reduced, it's just that densities and forces don't need it. But eg if you compute the current with TRS and BZ reduction you have to symmetrize or you'll get something nonzero, I think.

antoine-levitt:
(which this PR doesn't do I see... needs to be fixed but I'd like a review first)

Technici4n:
> Symmetries is there to list the symmetries of the system: whether we actually use them for computation is kind of besides the point.

A boolean flag for the TRS symmetry makes more sense though, no? Sure we can fold it with the other symmetries but since it operates in a fundamentally different way, I don't think we get anything from that except more complicated code :P

> But also yes we do want to symmetrize explicitly wrt TRS in case the BZ is reduced, it's just that densities and forces don't need it. But eg if you compute the current with TRS and BZ reduction you have to symmetrize or you'll get something nonzero, I think.

Yeah I think you're right here.

antoine-levitt:
> since it operates in a fundamentally different way

Why do you say that? The symmetries of a Hilbert space are exactly the unitary and anti-unitary operators (Wigner's theorem). TRS just happens to be anti-unitary. It's a clean abstraction I would say, it just makes life a bit more complicated because sometimes things have to be conjugated, but that's a fundamental fact rather than a technical annoyance.

Technici4n:
Ah interesting, I did not know this concept of anti-unitary operators. Are there some "irreducible" anti-unitary symmetries that we might care about in practice? Where we have to combine a non-trivial unitary matrix with a complex conjugation?

antoine-levitt:
That comes from the group structure, where we assume that we list *all* symmetries (except translational, which is always assumed). So TRS + spatial rotation or reflection is a valid symmetry. Or do you mean something else?

Technici4n:
Concretely, I am asking if there would be some systems where having only the unitary symmetries in `symmetries` + a boolean flag for TRS (as I suggested above) would prevent us from representing some symmetries of the system. If yes then that would be a strong reason to go for this additional complexity. For example this could happen if the system does not have TRS symmetry, but has some anti-unitary symmetry.

jaemolihm:
Sorry to jump in, but one example is antiferromagnets, which doesnt have TRS but has TRS + R/2 translation (or TRS + other spatial symmetry). I've been doing something similar in my code (based on DFTK as of 2022 or so). Recent versions of spglib also gives such composite TRS&spatial symmetries. (Magnetic point group)

https://github.com/jaemolihm/ElectronPhonon.jl/blob/0e563c540bcd29e3697fd9992dd2f2dbc86e82dc/src/common/symmetry.jl#L210

Technici4n:
Don't be sorry @jaemolihm, this is a very useful comment! So indeed this approach of representing _all_ symmetries is the right way to go.

antoine-levitt:
@jaemolihm you chiming in is always appreciated! I'm going blind on most of this, I have no physical intuition nor experience with actual materials, so I'm just going with the mathematical structure of the thing and what I can understand from what Claude tells me.

So yes we need the full structure I think. See also https://github.com/JuliaMolSim/DFTK.jl/pull/1317#issuecomment-4412442353. Right now the symop supports only conjugation (note how it refers only to conjugation, not to TRS: with magnetism those two things are different)

Technici4n:
Relevant document: https://docs.abinit.org/theory/wavefunctions/#magnetic-space-groups

In particular the "Shubnikov type III" groups cannot be separated into a direct product of the real-space symop with TRS.

PR #1318, opened 2026-05-13
Fix differentiation of nonlocal projectors at p=0 by switching to solid harmonics

Technici4n:
Spherical coordinates at p=0 are awkward, and this causes problems for AD wrt. k at G+k=0 for projectors that have angular momentum l=1: the derivative should not be zero, but our guards against small p make it 0. Additionally, only the product of the Hankel transform and the spherical harmonics is differentiable, but not each of them on their own.

The best fix I could find find (thanks Claude, although all the code is mine) was to move the `1/p^l` from the spherical harmonic to the Hankel transform. This gives a "regularized" Hankel transform (whose derivative is always 0 at p=0) times a solid harmonic. With this trick the derivative is $H(0) \cdot \partial_\alpha R_{lm}(0)$ which is always well-defined! And since $R_{lm}$ is a homogeneous polynomial of degree l this cleanly explains why only projectors at l=1 contribute to the derivative.

### DIFF
--- a/docs/src/developer/useful_formulas.md
+++ b/docs/src/developer/useful_formulas.md
@@ -28,6 +28,14 @@ of the conventions used in the equations.
   ```
   This also holds true for real spherical harmonics.
 
+  Introducing the solid harmonics ``R_l^m(x) = x^l \, Y_l^m(x/|x|)``
+  this can equivalently be written as
+  ```math
+  \hat f( q) = 4 \pi R_{l}^{m}(q) (-i)^{l}
+  \int_{{\mathbb R}^+} r^2 R(r) \ j_{l}(|q| r) / |q|^l dr,
+  ```
+  which is better behaved for algorithmic differentiation at ``q=0``.
+
 ## Spherical harmonics
 - Plane wave expansion formula
   ```math
--- a/docs/src/developer/useful_formulas.md
+++ b/docs/src/developer/useful_formulas.md
@@ -32,9 +32,11 @@ of the conventions used in the equations.
   this can equivalently be written as
   ```math
   \hat f( q) = 4 \pi R_{l}^{m}(q) (-i)^{l}
-  \int_{{\mathbb R}^+} r^2 R(r) \ j_{l}(|q| r) / |q|^l dr,
+  \int_{{\mathbb R}^+} r^2 R(r) j_{l}(|q| r) / |q|^l dr.
   ```
-  which is better behaved for algorithmic differentiation at ``q=0``.
+  The main advantage of this formulation is that the solid harmonics are
+  simple homogeneous polynomials
+  (which are e.g. well behaved for algorithmic differentiation close to ``q=0``).
 
 ## Spherical harmonics
 - Plane wave expansion formula
--- a/src/common/hankel.jl
+++ b/src/common/hankel.jl
@@ -1,13 +1,17 @@
 @doc raw"""
     hankel(r, r2_f, l, p)
 
-Compute the Hankel transform
+Compute the **regularized** Hankel transform
 ```math
-    H[f] = 4\pi \int_0^\infty r f(r) j_l(p·r) r dr.
+    H[f] = \frac{1}{p^l} ⋅ 4\pi \int_0^\infty r f(r) j_l(p·r) r dr.
 ```
 The integration is performed by simpson quadrature, and the function takes
 as input the radial grid `r`, the precomputed quantity r²f(r) `r2_f`, angular
 momentum / spherical bessel order `l`, and the Hankel coordinate `p`.
+
+The regularization by 1/p^l avoids numerical issues for small p:
+instead of computing the unregularized Hankel transform times a spherical harmonic,
+we can compute the regularized Hankel transform times a solid harmonic.
 """
 function hankel(r::AbstractVector, r2_f::AbstractVector, l::Integer, p::T)::T where {T<:Real}
     quadrature = default_psp_quadrature(r)
@@ -16,7 +20,17 @@ end
 
 function hankel(quadrature, r::AbstractVector, r2_f::AbstractVector, l::Integer, p::T)::T where {T<:Real}
     @assert length(r) == length(r2_f)
-    4T(π) * quadrature(r) do i, ri
+    if abs(p) <= 10 * eps(T)
+        # Use j_l(x) ≈ x^l / (2l+1)!! for small p
+        l == 0 && return 4T(π) * quadrature((i, ri) -> r2_f[i], r)
+        l == 1 && return 4T(π) * quadrature((i, ri) -> r2_f[i] * ri, r) / 3
+        l == 2 && return 4T(π) * quadrature((i, ri) -> r2_f[i] * ri^2, r) / 15
+        l == 3 && return 4T(π) * quadrature((i, ri) -> r2_f[i] * ri^3, r) / 105
+        l == 4 && return 4T(π) * quadrature((i, ri) -> r2_f[i] * ri^4, r) / 945
+        l == 5 && return 4T(π) * quadrature((i, ri) -> r2_f[i] * ri^5, r) / 10395
+        throw(BoundsError()) # specific l not implemented
+    end
+    1/p^l * 4T(π) * quadrature(r) do i, ri
         r2_f[i] * sphericalbesselj_fast(l, p * ri)
     end
 end
--- a/src/common/hankel.jl
+++ b/src/common/hankel.jl
@@ -1,7 +1,7 @@
 @doc raw"""
     hankel(r, r2_f, l, p)
 
-Compute the **regularized** Hankel transform
+Compute the **modified** Hankel transform
 ```math
     H[f] = \frac{1}{p^l} ⋅ 4\pi \int_0^\infty r f(r) j_l(p·r) r dr.
 ```
@@ -9,9 +9,9 @@ The integration is performed by simpson quadrature, and the function takes
 as input the radial grid `r`, the precomputed quantity r²f(r) `r2_f`, angular
 momentum / spherical bessel order `l`, and the Hankel coordinate `p`.
 
-The regularization by 1/p^l avoids numerical issues for small p:
-instead of computing the unregularized Hankel transform times a spherical harmonic,
-we can compute the regularized Hankel transform times a solid harmonic.
+The multiplication by 1/p^l avoids numerical issues for small p:
+instead of computing the usual Hankel transform times a spherical harmonic,
+we can compute the modified Hankel transform times a solid harmonic.
 """
 function hankel(r::AbstractVector, r2_f::AbstractVector, l::Integer, p::T)::T where {T<:Real}
     quadrature = default_psp_quadrature(r)
--- a/src/common/hankel.jl
+++ b/src/common/hankel.jl
@@ -20,8 +20,10 @@ end
 
 function hankel(quadrature, r::AbstractVector, r2_f::AbstractVector, l::Integer, p::T)::T where {T<:Real}
     @assert length(r) == length(r2_f)
+    # TODO: 10*eps is somewhat arbitrary, should find a proper bound
+    #       based on the stability of sphericalbesselj_fast
     if abs(p) <= 10 * eps(T)
-        # Use j_l(x) ≈ x^l / (2l+1)!! for small p
+        # Use j_l(x) ≈ x^l / (2l+1)!! for small x
         l == 0 && return 4T(π) * quadrature((i, ri) -> r2_f[i], r)
         l == 1 && return 4T(π) * quadrature((i, ri) -> r2_f[i] * ri, r) / 3
         l == 2 && return 4T(π) * quadrature((i, ri) -> r2_f[i] * ri^2, r) / 15
--- a/src/common/hankel.jl
+++ b/src/common/hankel.jl
@@ -21,7 +21,9 @@ end
 function hankel(quadrature, r::AbstractVector, r2_f::AbstractVector, l::Integer, p::T)::T where {T<:Real}
     @assert length(r) == length(r2_f)
     # TODO: 10*eps is somewhat arbitrary, should find a proper bound
-    #       based on the stability of sphericalbesselj_fast
+    #       based on the stability of sphericalbesselj_fast. The current
+    #       value is likely too lose and has the wrong scaling in eps().
+    #       In particular the scaling in eps() should depend on `l`.
     if abs(p) <= 10 * eps(T)
         # Use j_l(x) ≈ x^l / (2l+1)!! for small x
         l == 0 && return 4T(π) * quadrature((i, ri) -> r2_f[i], r)
--- a/src/common/spherical_harmonics.jl
+++ b/src/common/spherical_harmonics.jl
@@ -6,43 +6,57 @@ Returns the ``(l,m)`` real spherical harmonic ``Y_l^m(r)``. Consistent with
 [Wikipedia](https://en.wikipedia.org/wiki/Table_of_spherical_harmonics#Real_spherical_harmonics).
 """
 function ylm_real(l::Integer, m::Integer, rvec::AbstractVector{T}) where {T}
+    if l == 0 && m == 0
+        return sqrt(1 / 4T(π))
+    end
+
+    r = norm(rvec)
+    # Catch cases of numerically very small r
+    if r <= 10 * eps(T)
+        return zero(T)
+    end
+
+    solid_harmonic_real(l, m, rvec / r)
+end
+
+"""
+Returns the ``(l,m)`` real solid harmonic, defined as ``R_l^m(r) = r^l Y_l^m(r)``.
+The solid harmonics are homogeneous polynomials of degree l,
+which makes them cleanly defined and differentiable at the origin,
+unlike the spherical harmonics.
+"""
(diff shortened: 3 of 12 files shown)


antoine-levitt:
Hm, feels wrong to change the interface just to catch a numerical stability bug. Looking at the original function I think just writing iszero(r) && return 0; r = normalize(r)) at the beginning of the function should fix any issue?

antoine-levitt:
To be clear, the code on master is bad: the potential numerical issue has nothing to do with eps(), it's really about under and overflow, and normalize() fixes this. It's potentially slower though (needs a sqrt), so maybe we want to do the normalization only in case of potential under/overflow (something like max(abs(r))^l > typemax(T) or something)? Or we find a library that actually has thought about such things and steal it.

antoine-levitt:
Why don't we just depend on https://github.com/jishnub/SphericalHarmonics.jl?

Technici4n:
Well the spherical harmonic is fundamentally not differentiable at 0. Only the product of the spherical harmonic and the hankel transform is. The problem is precisely that for l=1 the derivative should not be zero!

src/common/hankel.jl
```diff
@@ -1,13 +1,17 @@
...
-    H[f] = 4\pi \int_0^\infty r f(r) j_l(p·r) r dr.
+    H[f] = \frac{1}{p^l} ⋅ 4\pi \int_0^\infty r f(r) j_l(p·r) r dr.
 ```
 The integration is performed by simpson quadrature, and the function takes
 as input the radial grid `r`, the precomputed quantity r²f(r) `r2_f`, angular
 momentum / spherical bessel order `l`, and the Hankel coordinate `p`.
+
+The regularization by 1/p^l avoids numerical issues for small p:
+instead of computing the unregularized Hankel transform times a spherical harmonic,
+we can compute the regularized Hankel transform times a solid harmonic.
```
antoine-levitt: I don't like the word regularized here, as regularization implies the removal of a singularity, but the singularity here is one level up; this version of the hankel transform is, if anything, "more singular". "Modified"?

src/common/hankel.jl
```diff
@@ -16,7 +20,17 @@ end
...
     @assert length(r) == length(r2_f)
-    4T(π) * quadrature(r) do i, ri
+    if abs(p) <= 10 * eps(T)
+        # Use j_l(x) ≈ x^l / (2l+1)!! for small p
+        l == 0 && return 4T(π) * quadrature((i, ri) -> r2_f[i], r)
+        l == 1 && return 4T(π) * quadrature((i, ri) -> r2_f[i] * ri, r) / 3
+        l == 2 && return 4T(π) * quadrature((i, ri) -> r2_f[i] * ri^2, r) / 15
+        l == 3 && return 4T(π) * quadrature((i, ri) -> r2_f[i] * ri^3, r) / 105
+        l == 4 && return 4T(π) * quadrature((i, ri) -> r2_f[i] * ri^4, r) / 945
+        l == 5 && return 4T(π) * quadrature((i, ri) -> r2_f[i] * ri^5, r) / 10395
```
antoine-levitt, line 34: just have a normalization tuple and return quadrature() / normalization[l+1]

src/common/hankel.jl
antoine-levitt: Also change the name to modified_hankel()

src/common/hankel.jl
```diff
@@ -16,7 +20,17 @@ end
 
 function hankel(quadrature, r::AbstractVector, r2_f::AbstractVector, l::Integer, p::T)::T where {T<:Real}
     @assert length(r) == length(r2_f)
-    4T(π) * quadrature(r) do i, ri
+    if abs(p) <= 10 * eps(T)
```
antoine-levitt, line 27: The eps here is wrong; this has nothing to do with eps, as fp are always relative and manipulating very small numbers and dividing with other very small numbers is completely fine. The issue here is underflow and p==0, so something with floatmin (with very generous margins, probably) is more appropriate.

antoine-levitt: Actually we can bypass this altogether with something like if abs(p) < minfloat p = minfloat and it should be fine?

docs/src/developer/useful_formulas.md
```diff
@@ -28,6 +28,14 @@ of the conventions used in the equations.
   ```
   This also holds true for real spherical harmonics.
 
+  Introducing the solid harmonics ``R_l^m(x) = x^l \, Y_l^m(x/|x|)``
+  this can equivalently be written as
+  ```math
+  \hat f( q) = 4 \pi R_{l}^{m}(q) (-i)^{l}
+  \int_{{\mathbb R}^+} r^2 R(r) \ j_{l}(|q| r) / |q|^l dr,
```
antoine-levitt: what's this backslash before the j_l ?

docs/src/developer/useful_formulas.md
```diff
@@ -28,6 +28,14 @@ of the conventions used in the equations.
...
+  ```
+  which is better behaved for algorithmic differentiation at ``q=0``.
```
antoine-levitt: don't necessarily mention AD explicitly here? Say we choose in DFTK to work with solid harmonics because they're nicer (polynomials) and this helps in particular with AD. Also introduce the modified hankel transform here?

src/common/spherical_harmonics.jl
```diff
@@ -6,43 +6,57 @@ Returns the ``(l,m)`` real spherical harmonic ``Y_l^m(r)``. Consistent with
 [Wikipedia](https://en.wikipedia.org/wiki/Table_of_spherical_harmonics#Real_spherical_harmonics).
 """
 function ylm_real(l::Integer, m::Integer, rvec::AbstractVector{T}) where {T}
+    if l == 0 && m == 0
+        return sqrt(1 / 4T(π))
+    end
+
+    r = norm(rvec)
+    # Catch cases of numerically very small r
```
antoine-levitt, line 14: same remark as above, nothing to do with eps

src/common/spherical_harmonics.jl
```diff
@@ -6,43 +6,57 @@ Returns the ``(l,m)`` real spherical harmonic ``Y_l^m(r)``. Consistent with
...
+        return zero(T)
+    end
+
+    solid_harmonic_real(l, m, rvec / r)
+end
+
+"""
+Returns the ``(l,m)`` real solid harmonic, defined as ``R_l^m(r) = r^l Y_l^m(r)``.
+The solid harmonics are homogeneous polynomials of degree l,
+which makes them cleanly defined and differentiable at the origin,
```
antoine-levitt: smooth

src/pseudo/NormConservingPsp.jl
```diff
@@ -49,11 +49,11 @@ eval_psp_projector_real(psp::NormConservingPsp, i, l, r::AbstractVector) =
     eval_psp_projector_fourier(psp, i, l, p)
 
 Evaluate the radial part of the `i`-th projector for angular momentum `l`
-at the reciprocal vector with modulus `p`:
+at the reciprocal vector with modulus `p`, regularized by p^l:
```
antoine-levitt: not regularized

src/pseudo/PspHgh.jl
```diff
@@ -135,7 +135,8 @@ function eval_psp_local_real(psp::PspHgh, r::T) where {T <: Real}
 end
 
 
-# [HGH98] (7-15) except they do it with plane waves normalized by 1/sqrt(Ω).
+# [HGH98] (7-15) except they do it with plane waves normalized by 1/sqrt(Ω)
+# and we regularize by 1/p^l.
```
antoine-levitt: as above

antoine-levitt: This is very annoying: you're changing the contract of eval_psp_projector_fourier (which doesn't return a fourier transform anymore), but you don't do this for the local part. Is this really really necessary? If so we need to rename things

src/workarounds/forwarddiff_rules.jl
```diff
@@ -303,29 +303,23 @@ end
        scfres.seed, scfres.algorithm, scfres.runtime_ns)
 end
 
-function hankel(r::AbstractVector, r2_f::AbstractVector, l::Integer, p::TT) where {TT <: ForwardDiff.Dual}
+function hankel(quadrature, r::AbstractVector, r2_f::AbstractVector, l::Integer, p::TT) where {TT <: ForwardDiff.Dual}
```
antoine-levitt: maybe this is not necessary anymore?

antoine-levitt (commented):
Hm, I think I buy your argument but I'm not happy about it... Apart from minor stuff, the important bit is that we can't have psp_fourier be one thing and local_fourier be another.

src/common/hankel.jl
Technici4n: I did it like this to keep a literal pow for speed:tm: :D

src/common/hankel.jl
Technici4n: That's just taken from `ylm_real`

src/pseudo/PspHgh.jl
Technici4n: The local part has l=0 so the regularized vs unregularized hankel transforms are equivalent there. Same for the NLCC as well

src/common/hankel.jl
antoine-levitt: if it's a tuple both are equivalent for the compiler so it should be the same speed?

src/common/hankel.jl
antoine-levitt: I know, it was wrong there as well :-p

src/pseudo/PspHgh.jl
antoine-levitt: OK, that's fair, it never actually returned a fourier transform. I think we need to clarify what "radial part" means for such a function, since the comment 
```
Evaluate the radial part of the `i`-th projector for angular momentum `l`
in real-space at the vector with modulus `r`.
```
is imprecise.

docs/src/developer/useful_formulas.md
mfherbst: I agree. AD is just one thing where this helps.

src/common/hankel.jl
mfherbst: That quick fix was my bad. Avoiding underflow is better of course.

src/common/hankel.jl
mfherbst: Think so too. Only thing is to take care of proper type propagation to make sure that things are inlined (e.g. one may need an `@inbounds` to suppress the bounds checking code)

src/common/hankel.jl
mfherbst: Dito, please avoid "regularised"

src/common/spherical_harmonics.jl
```diff
@@ -6,43 +6,57 @@ Returns the ``(l,m)`` real spherical harmonic ``Y_l^m(r)``. Consistent with
...
+unlike the spherical harmonics.
```
mfherbst, line 26: Also refer to wikipedia in case we need to extend this later ?

src/pseudo/PspHgh.jl
mfherbst: As a suggestion, to make this clearer we could rename the `_fourier` functions to `_reciprocal` where they are not actually fourier transforms. This may help, but it may also add confusion; I'm divided.

src/workarounds/forwarddiff_rules.jl
mfherbst: Why should that be ?

test/PspHgh.jl
```diff
@@ -109,7 +109,7 @@ end
         psp = load_psp(family, element)
         for l = 0:psp.lmax, i = 1:count_n_proj_radial(psp, l)
             for p in [0.01, 0.1, 0.2, 0.5, 1, 2, 5, 10]
-                reference = quadgk(r -> integrand(psp, i, l, p, r), 0, Inf)[1]
+                reference = 1/p^l * quadgk(r -> integrand(psp, i, l, p, r), 0, Inf)[1]
```
mfherbst, line 113: This is not obvious to someone just reading / refactoring the tests. Needs a comment why this is needed to make the two things agree.

test/PspUpf.jl
```diff
@@ -163,8 +163,8 @@ end
         for l = 0:psp.lmax, i in count_n_proj_radial(psp, l)
             ir_cut = min(psp.ircut, length(psp.r2_projs[l+1][i]))
             for p in (0.01, 0.1, 0.2, 0.5, 1., 2., 5., 10.)
-                reference = quadgk(r -> integrand(psp, i, l, p, r),
-                                   psp.rgrid[ir_start], psp.rgrid[ir_cut])[1]
+                reference = 1/p^l * quadgk(r -> integrand(psp, i, l, p, r),
```
mfherbst, line 167: Same as  above:

This is not obvious to someone just reading / refactoring the tests. Needs a comment why this is needed to make the two things agree.

src/common/hankel.jl
Technici4n: I would like to keep the explicit `^ <literal>` here.

src/common/hankel.jl
Technici4n: `sphericalbesselj_fast` is quite numerically unstable for small $p$ anyway so better keep it like this than use minfloat.

src/common/spherical_harmonics.jl
Technici4n: I think that epsilon is reasonable since the norm computation is itself approximate. Perhaps sqrt(epsilon) would be more correct.

src/common/spherical_harmonics.jl
Technici4n: I don't find a clean table on Wikipedia. Better use the real spherical harmonic table.

src/common/hankel.jl
antoine-levitt: Oh yeah sorry, I was looking at ylm_real which is stable, but this sphericalbesselj_fast is very unstable, eg (sin(x) - cos(x) * x) is bad

src/common/spherical_harmonics.jl
antoine-levitt: Nah that's not an issue, norm(lambda x) = lambda norm(x) holds true (in a relative error sense) for a very wide range of lambdas, up to under/overflow, the magnitude just goes into the exponent

src/common/hankel.jl
Technici4n: Ideally the bound should be something like `\epsilon^(1/l)` or maybe `\epsilon^(1/l+1)`? But I am hesitant to write it like this because it might not be very GPU friendly?

src/common/spherical_harmonics.jl
Technici4n: Indeed you are right, for most $\mathbf r$ values the spherical harmonics are very stable (until you get to subnormal numbers, more or less). But `issubnormal` is exclusive to IEEE float types, and there are a few subtractions in the solid harmonics, so maybe an eps-based check is still reasonable as a conservative bound?

src/common/spherical_harmonics.jl
mfherbst: sure, but what I meant is refer to the spherical harmonics table and indicate "what one has to do" to get to these expressions.

src/common/hankel.jl
```diff
@@ -20,8 +20,10 @@ end
 
 function hankel(quadrature, r::AbstractVector, r2_f::AbstractVector, l::Integer, p::T)::T where {T<:Real}
     @assert length(r) == length(r2_f)
+    # TODO: 10*eps is somewhat arbitrary, should find a proper bound
+    #       based on the stability of sphericalbesselj_fast
```
mfherbst:
```suggestion
    # TODO: 10*eps is somewhat arbitrary, should find a proper bound
    #       based on the stability of sphericalbesselj_fast. The current
    #       value is likely too lose and has the wrong scaling in eps().
    #       In particular the scaling in eps() should depend on `l`.
```

src/pseudo/PspUpf.jl
```diff
@@ -234,8 +234,8 @@ function _eval_psp_local_fourier(quadrature, rgrid, vloc, Zion, p::T)::T where {
     # ABINIT uses a more 'pure' Coulomb term with the same asymptotic behavior
     # C(r) = -Z/r; H[-Z/r] = -Z/p^2
     p == 0 && return zero(T)  # Compensating charge background
-    I = quadrature(rgrid) do i, r
-         r * (r * vloc[i] - -Zion * erf(r)) * sphericalbesselj_fast(0, p * r)
+    I = 1/p * quadrature(rgrid) do i, r
```
mfherbst:
```suggestion
    # Equal to \int r (r * vloc[i] - (-Zion) erf(r)) * sin(p*r)/(p*r)
    I = 1/p * quadrature(rgrid) do i, r
```

src/pseudo/PspUpf.jl
```diff
@@ -234,8 +234,8 @@ function _eval_psp_local_fourier(quadrature, rgrid, vloc, Zion, p::T)::T where {
...
+         (r * vloc[i] - -Zion * erf(r)) * sin(p * r)
```
mfherbst:
```suggestion
         (r * vloc[i] - (-Zion) * erf(r)) * sin(p * r)
```

mfherbst (commented):
Ok for me like this.

Technici4n:
@mfherbst ping for review

PR #1325, opened 2026-06-17
Remove negative rhocore/taucore warnings

### LINKED ISSUE #1324: Remove negative ρcore warnings
I have never found these warnings useful. They clutter the logs and it gets even worse for meta-GGAs.

Technici4n:
Fixes #1324


Technici4n:
After discussion: keep the warnings but make the bounds less tight, e.g. by looking at silicon. Could also make it relative.

mfherbst:
I experimented with this a bit and it seems it's actually pretty tricky to do this reliably. I will take care of this in a follow-up PR implementing more changes for meta-GGA SCFs. My current suggestion would be to make these as `@debug` instead of `@warn`, such that one can still see them if one wants to (by setting debug-level logging for `xc.jl`) but by default these messages would no longer be printed.

Technici4n:
OK thanks! Maybe only checking for negative alpha by default is enough (in every xc evaluation, not necessarily only once?).

PR #1329, opened 2026-06-20
fix code-block backgrounds in dark themes

aouinaayoub:
custom.css hardcoded a light background for code blocks, making the code illegible in dark themes. Removed it and made output blocks transparent so they follow the active theme.


antoine-levitt:
Nice I was wondering why that was. But is that not autogenerated by something? @mfherbst will know

mfherbst:
I have no idea where that is from. I likely stole it from somewhere. If this improves things, I don't care.

Thanks @aouinaayoub !

PR #1335, opened 2026-06-30
Use multithreading for PSP evaluations

### LINKED ISSUE #1326: Better multithreading in basis construction
SCF multithreading support seems reasonable, however a lot of time can be spent instantiating the basis both before the SCF and especially after (e.g. stresses or higher-order energy derivatives), so it would be good to exploit multithreading there as well.

Technici4n:
Same pattern as the GPU support, but on the CPU: use a parallel map! for the Hankel transform evaluations. Should mostly fix #1326.

Needs benchmarking still. Thoughts on the approach?

### DIFF
--- a/src/common/threading.jl
+++ b/src/common/threading.jl
@@ -86,3 +86,25 @@ function parallel_loop_over_range(fun, range, storages)
 
     return storages
 end
+
+function parallel_map(f, T, src::AbstractVector)
+    dest = similar(src, T)
+    parallel_map!(f, dest, src)
+end
+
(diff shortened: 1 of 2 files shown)


antoine-levitt:
Looks reasonable but `@threads` has become much more featureful in recent releases so maybe we can use it directly? Also don't we have this construction somewhere else in the code? Also also parallel_map -> threaded_map?

mfherbst:
I also agree in principle, given there is no natural julia way to do this and module bikesheding names.

PR #1336, opened 2026-07-02
Implementation of the hybrid preconditioner for ferromagnetic systems

ClementineBarat:
This pull requests implements a new preconditioner meant for magnetic systems, that is based on a diagonal model for the non-interacting susceptibility (https://hal.science/hal-05658631).

### DIFF
--- a/src/scf/chi0models.jl
+++ b/src/scf/chi0models.jl
@@ -15,7 +15,7 @@ where ``D_\text{loc}`` is the local density of states and ``D`` the density of s
 For details see [Herbst, Levitt 2020](https://arxiv.org/abs/2009.01665).
 
 By default the LdosModel is constructed using the smearing of
-`basis.model.smearing` and a temperature of `min(50basis.model.temperature, 0.1)`,
+`Smearing.Gaussian` and a temperature of `min(100basis.model.temperature, 0.1)`,
 but this may be changed using the `smearing` and `temperature` arguments.
 """
 @kwdef struct LdosModel <: χ0Model
@@ -86,12 +86,30 @@ end
 Applyχ0Model(; kwargs_apply_χ0...) = Applyχ0Model(kwargs_apply_χ0)
 function (χ0::Applyχ0Model)(basis; ham, eigenvalues, ψ, occupation, εF,
                             kwargs...)
+    function apply!(δρ, δV, α=1)
+        χ0δV = apply_χ0(ham, ψ, occupation, εF, eigenvalues, δV;
+                        χ0.kwargs_apply_χ0...).δρ
+        δρ .+= α .* χ0δV
+    end
+end
+
+"""
+Diagonal approximation of ``χ_0``, that disables the Sternheimer solver and uses a modified 
+smearing temperature and smearing function.
+
+By default the modified temperature is `min(100basis.model.temperature, 0.1)`, and the 
+modified smearing is `Smearing.Gaussian` but this may be changed using the `smearing` and 
+`temperature` arguments.
+"""
+@kwdef struct DiagonalModel <: χ0Model
+    smearing::Union{Nothing,Smearing.SmearingFunction} = nothing
+    temperature::Union{Nothing,Float64} = nothing
+end
+function (χ0::DiagonalModel)(basis; ham, eigenvalues, ψ, occupation, εF,
+                            kwargs...)
     defaults = default_smearing_temperature(basis.model)
     temperature = @something(χ0.temperature, defaults.temperature)
     smearing    = @something(χ0.smearing,    defaults.smearing)
-    function apply!(δρ, δV, α=1)
-        χ0δV = apply_χ0(ham, ψ, occupation, εF, eigenvalues, δV;
-                        smearing, temperature, χ0.kwargs_apply_χ0...).δρ
-        δρ .+= α .* χ0δV
-    end
+    kwargs_apply_χ0 = (; smearing, temperature, disable_sternheimer=true)
+    Applyχ0Model(kwargs_apply_χ0)(basis; ham, eigenvalues, ψ, occupation, εF, kwargs...)
 end
--- a/src/scf/chi0models.jl
+++ b/src/scf/chi0models.jl
@@ -77,57 +77,6 @@ function (χ0::DielectricModel)(basis; kwargs...)
 end
 
 """
-χ0 reduced to eigenvalue variations and Fermi-level variations
-    χ0_Δε = ∑_i f'(εᵢ-εF) |ρᵢᵢ⟩⟨ρᵢᵢ| + 1/DOS |LDOS⟩⟨LDOS|.
-"""
-@kwdef struct DiagonalModel <: χ0Model
-    temperature = 0.01
-end
-function (χ0::DiagonalModel)(basis::PlaneWaveBasis{T}; eigenvalues, ψ, εF, kwargs...) where {T}
-
-    mixing_temperature = χ0.temperature
-    ldos = compute_ldos(εF, basis, eigenvalues, ψ)
-    tdos = sum(sum, ldos) * basis.dvol  # Integrate LDOS to form total DOS
-
-    function apply!(δρ, δV, α=1)
-
-        filled_occ = filled_occupation(basis.model)
-        weights = deepcopy(eigenvalues)
-
-        # Compute the weights that will be use in compute_density 
-        # in place of the occupations : w_i = f'(εᵢ-εF)⟨ρᵢᵢ|δV⟩
-        for ik=1:length(basis.kpoints)
-            for (iband, ε) in enumerate(eigenvalues[ik])
-                enred = (ε - εF) / mixing_temperature
-                fp = (Smearing.occupation_derivative(basis.model.smearing, enred) 
-                        / mixing_temperature)
-                if abs(fp) > eps()
-                    ρii = abs.(ifft(basis.fft_grid, basis.kpoints[ik], ψ[ik][:, iband])).^2
-                    if size(δV, 4) == 1
-                        weights[ik][iband] = filled_occ * fp * dot(δV, ρii) * basis.dvol
-                    elseif size(δV, 4) == 2
-                        weights[ik][iband] = (filled_occ * fp 
-                            * dot(δV[:, :, :, basis.kpoints[ik].spin], ρii) * basis.dvol
-                        )
-                    end
-                else
-                    weights[ik][iband] = 0.
-                end
-            end
-        end
-
-        δρ .+= α * compute_density(basis, ψ, weights; warn=false)
-
-        # Fermi-level variation
-        δεF = dot(ldos, δV) .* basis.dvol
-        δρ .+= α .* (ldos .* δεF ./ tdos)
-
-    return δρ
-
-    end
-end
-
-"""
 Full χ0 application, optionally dropping terms or disabling Sternheimer.
 All keyword arguments passed to [`apply_χ0`](@ref).
 """
--- a/src/scf/chi0models.jl
+++ b/src/scf/chi0models.jl
@@ -77,15 +77,17 @@ function (χ0::DielectricModel)(basis; kwargs...)
 end
 
 """
-χ0 reduced to eigenvalue variations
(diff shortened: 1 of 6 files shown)


src/scf/chi0models.jl
```diff
@@ -76,6 +76,68 @@ function (χ0::DielectricModel)(basis; kwargs...)
...
 end
 
+"""
+χ0 reduced to eigenvalue variations
+    χ0_Δε = ∑_i f'(εᵢ-εF) |ρᵢᵢ⟩⟨ρᵢᵢ|.
+"""
+@kwdef struct ΔεModel <: χ0Model
+    temperature = 0.01
+end
+function (χ0::ΔεModel)(basis::PlaneWaveBasis{T}; eigenvalues, ψ, εF, kwargs...) where {T}
```
antoine-levitt: can this not be applied as an option to apply_chi0?  I think we had this before

src/scf/chi0models.jl
```diff
@@ -76,6 +76,68 @@ function (χ0::DielectricModel)(basis; kwargs...)
...
+
+        δρ .+= α * compute_density(basis, ψ, weights; warn=false)
+
+    return δρ
+
+    end
+end
+
+"""
+χ0 reduced to the fermi-level variation
```
antoine-levitt: I don't know if this makes sense as a standalone chi0 model, it's pretty useless by itself? Maybe do this inside others?

src/scf/mixing.jl
```diff
@@ -277,18 +296,110 @@ end
...
+    maxiter::Int = 20       # Maximum number of iterations for the GMRES
+end
+function Base.show(io::IO, mixing::mixedχ0Mixing)
+    χ0terms_vc   = mixing.χ0terms_vc
+    χ0terms_Kxc  = mixing.χ0terms_Kxc
+    χ0terms_Khxc = mixing.χ0terms_Khxc
+
+    if isempty(χ0terms_Khxc) && length(χ0terms_vc) == 1 && χ0terms_vc[1] isa LdosModel &&
+       length(χ0terms_Kxc) == 2 && χ0terms_Kxc[1] isa ΔεModel && χ0terms_Kxc[2] isa ΔεFModel
+        print(io, "HybridDiagonalMixing(temperature=$(χ0terms_Kxc[1].temperature), ")
```
antoine-levitt: meh, that's quite brittle just for printing (it's "python"-style in the sense that functions try to do something based on the types of its arguments, whereas in Julia the goal would be to push this logic into dispatch)

antoine-levitt: My preference is to not special-case it and just print the generic thing

antoine-levitt (commented):
In general LGTM, but one thing I don't like is that it looks like this lives separately from the existing mechanism rather than extend it. Is there no way to fold everything into one framework? I also don't like too much the fact that the hartree and  xc terms are singled out (DFTK tries to be somewhat agnostic to terms). Would it make sense to have something like a term => chi0 approximation dict?

src/scf/chi0models.jl
ClementineBarat: Yes i changed that

src/scf/mixing.jl
ClementineBarat: I removed it, along with pre-existing printing functions for the LDOS, hybrid preconditioner.

src/scf/mixing.jl
```diff
@@ -227,70 +227,169 @@ Important `kwargs` passed on to [`χ0Mixing`](@ref)
...
+convergence is controlled by `reltol` and `maxiter`.
+`verbose=true` lets the GMRES run in verbose mode (useful for debugging).
 """
 @kwdef struct χ0Mixing <: Mixing
-    χ0terms   = χ0Model[Applyχ0Model()]  # The terms to use as the model for χ0
-    RPA::Bool = true        # Use RPA, i.e. only apply the Hartree and not the XC Kernel
+    χ0terms_mapping::Dict = Dict(Union{TermHartree, TermXc} => [Applyχ0Model()])
     verbose::Bool = false   # Run the GMRES verbosely
-    reltol::Float64 = 0.01  # Relative tolerance for GMRES
+    reltol::Float64 = 1e-6  # Relative tolerance for the GMRES.
```
mfherbst: This is a lot tighter. Do we really need this in production ? On larger systems the gmres can cost so we should not oversolve.

src/scf/mixing.jl
```diff
@@ -227,70 +227,169 @@ Important `kwargs` passed on to [`χ0Mixing`](@ref)
...
+    mixed_ΔV = similar(ΔV)
+    mixed_ΔV, info = linsolve(ε_op, ΔV;
+        verbosity=(mixing.verbose ? 3 : 0),
+        rtol=T(mixing.reltol),
+        krylovdim=mixing.maxiter,
+        maxiter=1,
+        ishermitian=false,
+        isposdef=false,
+    )
+    if MPI.Comm_rank(MPI.COMM_WORLD) == 0
```
mfherbst: use mpi_master function

src/scf/mixing.jl
```diff
@@ -227,70 +227,169 @@ Important `kwargs` passed on to [`χ0Mixing`](@ref)
...
+        kwargs...) where {T}
+
+    ε_op = get_ε_op(mixing, basis; ρin, kwargs...)
+    ε_op == identity && return mix_potential(SimpleMixing(), basis, ΔV)
+    
+    mixed_ΔV = similar(ΔV)
+    mixed_ΔV, info = linsolve(ε_op, ΔV;
+        verbosity=(mixing.verbose ? 3 : 0),
+        rtol=T(mixing.reltol),
+        krylovdim=mixing.maxiter,
```
mfherbst, line 398: 100 densities is potentially a memory bottle neck. Do we need this ?

src/scf/mixing.jl
```diff
@@ -227,70 +227,169 @@ Important `kwargs` passed on to [`χ0Mixing`](@ref)
...
+    MPI.Bcast!(mixed_Δρ, 0, MPI.COMM_WORLD) 
+
+    # Ensuring that the mean value of Δρ is unchanged 
+    # (conservation of electron number).
+    return mixed_Δρ .+ mean(Δρ) .- mean(mixed_Δρ)
 end
 
-@timing "χ0Mixing" function mix_potential(mixing::Mixing, basis::χ0Mixing, δF::AbstractArray; kwargs...)
-    error("Not yet implemented.")
+@timing "χ0Mixing" function mix_potential(mixing::χ0Mixing, basis, ΔV::AbstractArray{T};
```
mfherbst, line 388: Does this finally work on the potential ?

src/scf/mixing.jl
```diff
@@ -227,70 +227,169 @@ Important `kwargs` passed on to [`χ0Mixing`](@ref)
...
-        # Apply Kernel (just vc for RPA and (vc + K_{xc}) if not RPA)
-        δV = apply_kernel(basis, δF; ρ=ρin, mixing.RPA)
-        δV .-= mean(δV)
-        εδF = copy(δF)
-        for apply_term! in χ0applies
-            apply_term!(εδF, δV, -1)  # εδF .-= χ₀ * δV
+"""
+Get the model dielectric operator used for this mixing.
+"""
+function get_ε_op(mixing::χ0Mixing, basis::PlaneWaveBasis; ρin, kwargs...)
```
mfherbst, line 328: Would it make sense to combine this with the DielectricOperator struct ? If you don't want to do it now, make a todo please

src/scf/mixing.jl
```diff
@@ -227,70 +227,169 @@ Important `kwargs` passed on to [`χ0Mixing`](@ref)
...
 end
-function Base.show(io::IO, mixing::χ0Mixing)
-    χ0terms = mixing.χ0terms
-    if length(χ0terms) == 1 && χ0terms[1] isa Applyχ0Model
-        print(io, "χ0Mixing([Applyχ0Model()], ")
-    elseif length(χ0terms) == 1 && χ0terms[1] isa LdosModel
-        print(io, "LdosMixing(")
-    elseif length(χ0terms) == 2 && χ0terms[2] isa LdosModel && χ0terms[1] isa DielectricModel
-        print(io, "HybridMixing(")
+function χ0Mixing(χ0terms::Vector; RPA::Bool=true, kwargs...)
```
mfherbst: This is anyway only internal. You can remove it for the more general dict interface.

src/scf/mixing.jl
```diff
@@ -227,70 +227,169 @@ Important `kwargs` passed on to [`χ0Mixing`](@ref)
...
+Hybrid mixing for ferromagnetic systems, that uses the LDOS χ0-model for the Hartree 
+kernel and a diagonal χ0-model for the exchange-correlation kernel: 
+```math
+\begin{aligned}
+    & χ_0^\text{diag} = \sum_{i=1}^\infty f'(\varepsilon_i-\varepsilon_F) |\psi_i|^2(r)+|\psi_i|^2(r') + \frac1D D_\text{loc}(r) D_\text{loc}(r') \\
+    & \varepsilon^\dagger = I - \chi_0^text{LDOS} K_H - \chi_0^\text{diag} K_\text{XC}
+\end{aligned}
+```
+"""
+function HybridDiagonalMixing(; verbose=false, maxiter=20, reltol=1e-6, kwargs...)
```
mfherbst: Can we bikeshed the name ? I don't like we have HybridMixing and HybridDiagonalMixing and the latter is not an approx of the former even though it sounds like it.

HybridMixing I never liked either as a name because it's so meaningless and does not really say what it is.

src/scf/mixing.jl
```diff
@@ -227,70 +227,169 @@ Important `kwargs` passed on to [`χ0Mixing`](@ref)
...
 
+@doc raw"""
+Hybrid mixing for ferromagnetic systems, that uses the LDOS χ0-model for the Hartree 
+kernel and a diagonal χ0-model for the exchange-correlation kernel: 
+```math
+\begin{aligned}
+    & χ_0^\text{diag} = \sum_{i=1}^\infty f'(\varepsilon_i-\varepsilon_F) |\psi_i|^2(r)+|\psi_i|^2(r') + \frac1D D_\text{loc}(r) D_\text{loc}(r') \\
+    & \varepsilon^\dagger = I - \chi_0^text{LDOS} K_H - \chi_0^\text{diag} K_\text{XC}
+\end{aligned}
+```
```
mfherbst, line 255: Wanna include a citation ?

src/scf/mixing.jl
ClementineBarat: Maybe just DiagonalMixing or XCDiagonalMixing ? And for Hybrid mixing, it could be changed to LocalDielectricMixing. But these name don't mention that the preconditioner includes the LDOS model ...

mfherbst: Hmm, I'm not too sure about the "Local" particle because all common mixing models use some degree of locality, even Kerker ... but maybe you can explain your thinking behind using that term ?

How about:
- `LdosMixing`  (as it is right now, just Ldos + coulomb kernel)
- `LdosDielectricMixing` (what is "HybridMixing" right now)
- `LdosXcDiagonalMixing` (what you propose ?)

Or are we stepping too much into the world of never-ending CamelCaseWordsThatBecomeUnreadable.

ClementineBarat: My thinking was that the dielectric model in `HybridMixing` includes a localization function and is therefore different from the model in `DielectricMixing`. But I am good with what you proposed.

mfherbst: Ah yeah that's true, I see your point. But honestly that localisation function never really worked so well and could in fact be changed or dropped. I would not be too hung up on that part.

mfherbst: > But I am good with what you proposed.

Ok, but feel free to propose better things. `LdosXcDiagonalMixing` is not exactly great either, honestly. If not, than we just use that for now and then we'll see later.

src/scf/mixing.jl
ClementineBarat: On Aluminium yes, although not as well as density mixing.
potential_mixing_DFTK

src/scf/mixing.jl
ClementineBarat: Yes, I changed it

src/scf/mixing.jl
ClementineBarat: No, I changed it to 20

PR #1338, opened 2026-07-10
Refactor coulomb

antoine-levitt:
As discussed previously. Along the way, I found a bug : the coulomb kernel was evaluated on the kpoint dependent basis instead of the full one, so the previous version was less accurate. Several independent commits there, so review by commit rather than by files. 
@toschaefer can you review?

Code written by Opus 4.8

### DIFF
--- a/ext/DFTKFastGaussQuadratureExt.jl
+++ b/ext/DFTKFastGaussQuadratureExt.jl
@@ -1,15 +1,16 @@
 module DFTKFastGaussQuadratureExt
 using FastGaussQuadrature
 using DFTK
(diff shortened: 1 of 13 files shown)


toschaefer:
Thank you! I will soon finish the k-point support for exact exchange (where I also noticed and fixed the bug with the kernel being evaluated on the k-point dependent grid). Once that's done, I'll merge the refactor coulomb.

mfherbst:
Ok, so the plan is to first merge the k-point extension from @toschaefer into master and than pick this one up.

PR #1343, opened 2026-07-12
Add electric-field dielectric response (ε∞ / polarizability)

antoine-levitt:
Vibe coded but looks like it works. Pushing this as a WIP for discussion, I'll polish it later.


antoine-levitt:
@Technici4n apparently you're also working on this so feel free to steal what you want from this PR if there's anything useful or close it if not

Technici4n:
I do indeed have a local version of this which I should be PRing at some point over the summer.

PR #1345, opened 2026-07-13
Fix flaky iron_LDA test by tightening SCF tolerance in density

niklasschmitz:
I think I found the issue why iron_LDA is flaky: We used the total energy convergence of 1e-11 but that does not guarantee the eigenvalues to be converged to sqrt(1e-11). The order is right (since SCF eigenvalues error is linear in density error) but still uncontrolled (hence the stochastic CI failures). I switched now to explicitly enforce SCF density convergence.


antoine-levitt:
nice thanks!

PR #1346, opened 2026-07-14
Numerical pseudopotentials overhauled

### LINKED ISSUE #1258: Consider using SphericalBesselTransforms.jl for log-grid pseudos
I just submitted [SphericalBesselTransforms.jl](https://github.com/azadoks/SphericalBesselTransforms.jl) the registrator.

It's a fairly direct translation of [pySBT](https://github.com/QijingZheng/pySBT) which implements the algorithm described in [Talman, Comp. Phys. Comms. 180, 332 (2009)](https://www.sciencedirect.com/science/article/pii/S0010465508003329) for the SBT/Hankel transform _*of functions on a logarithmic radial mesh*_ in O(Nlog(N)) time using FFTs.

The implementation is fairly minimal, and the interface is not super ergonomic at the moment. I spent some time with the profiler to make sure it's pretty fast, so the typing might be a bit too strict.

An implementation in DFTK would probably also require some interpolation because the transform returns `g(k)` on a logarithmic grid. For AD support, we'd also need to write a rule.

I wrote up a [quick example here](https://github.com/azadoks/SphericalBesselTransforms.jl/blob/main/examples/psp.jl) which does the transforms for HGH pseudos and compares with the analytic reference. It seems to do quite well, but I should probably also compare to simpson integration / QuadGK.jl.

antoine-levitt:
Fix https://github.com/JuliaMolSim/DFTK.jl/issues/1258

This removes the previous hankel machinery and goes uniformly through input r grid -> log r grid -> log p grid -> arbitrary p points. log r -> log p is done by SphericalBesselTransforms.jl, interpolations done by splines. Splines are constructed by BSplineKit (order 6), with the r integral end-point corrected in r space by the Gregory rule, and evaluated by a custom implementation (because BSplineKit doesn't support GPU, and because the spline is uniform so that evaluation can be done branchless)

Main win from my point of view: we can remove the caching! Code deletion <3

Performance:

| | master | this branch | |
|---|---|---|---|
| `eval_psp_local_fourier`, one p | 21 000 ns | **59 ns** | **355×** |
| `eval_psp_projector_fourier`, one p | 2 010 ns | **59 ns** | **34×** |
| local form factors | 48.9 ms | **4.2 ms** | **12×** |
| density form factors | 42.7 ms | **3.8 ms** | **11×** |
| `PlaneWaveBasis` construction | 117 ms | **28.7 ms** | **4×** |
| `compute_forces_cart` | 133 ms | **42.8 ms** | **3×** |
| **`compute_stresses_cart`** | **9 686 ms** | **1 198 ms** | **8×** |

Downsides:
| | master | this branch |
|---|---|---|
| `load_psp` (per pseudopotential) | 4.5 ms | 24.6 ms |
| `using DFTK` | 2.4 s | 3.1 s (**+30%**) |

The load-time regression is `BandedMatrices` + `ArrayLayouts` stack (677 ms). If that's too much, we can use `Dierckx` (2.8 ms of load, FITPACK). This gives numerically identical results but it would not supply the log-p prefilter, which would
then have to be hand-rolled.

Accuracy

There are two things here: the accuracy of the r integral (for a p in the log grid), and the accuracy of the p interpolation. p error is ~1e-15 for values, ~1e-10 for up to second derivatives, but degrades for higher derivatives. The previous code was better in that respect because p derivatives were analytic. This can be fixed by increasing the order (or the number of log grid points, but only to a point since noise amplification kicks in). For the accuracy of the r integration, quoting Claude for methodology:

A pseudopotential's radial data has no defined value between its mesh points, so on the real
quantities there is no ground truth — only refinement. Feed both methods a mesh downsampled by 2 and
compare against the transform of the full mesh, at an on-grid p. Error relative to max|H̃|, on Li and Ge:

| quantity | master (Simpson) | this branch | |
|---|---|---|---|
| projector l = 0 | 3e-7 … 6e-7 | **4e-10 … 4e-8** | **18–800× better** |
| vloc (erf-corrected) | 2e-11 … 2e-9 | 8e-11 … 8e-10 | ~3× better, or a wash |
| pswfc (Ge) | 1e-9 … 7e-9 | **1e-12** | **1300–6900× better** |
| pswfc (Li) | 1e-11 … 4e-11 | 6e-11 | **~2× worse** |
| ρion | 2e-12 … 6e-11 | 3e-10 … 9e-10 | **10–200× worse** |
| ρcore | 2e-12 … 5e-12 | 5e-10 | **100–300× worse** |

So we lose only on the very smooth quantities (because hankel was spectrally accurate and we lose that by resampling on the log grid). This is on the downsampled data: the new method's actual errors are likely to be 2^6=64 times lower - confirmed numerically by measuring the order by further downsampling). Acceptable I would say.

TODO

- [ ] Convince @azadoks to downgrade SphericalBesselTransforms to julia 1.10 (I think you just use it for logrange? You can either use Compat or just write it as a range)
- [ ] Tidy up the docs
- [ ] Check tests
- [ ] GPU. If Indeed BSplineKit doesn't handle it (as Claude says it doesn't), we need to make sure the hand-rolled version is GPU-friendly. @abussy can you advise?

(coded by Claude)


Technici4n:
The basis construction speedup is huge especially for stresses (and other derivatives down the line)! Very exciting. To be confident enough to make this the default strategy for numerical psps, I offered to run the acwf verification study when this PR gets to a merge-ready stage, at the very least on one configuration per element.

antoine-levitt:
CI is green (on julia latest, stable errors are because sphericalbesseltransforms needs 1.10) and I think the functional side of this is fine (apart from GPU), so I think you can go ahead with the validation. I don't expect results to change (and, if they do, I think I would expect the new code to be more accurate)

antoine-levitt:
(actually looks like there's an accuracy issue so don't do that yet)

antoine-levitt:
I remember my desktop has a GPU, so was able to test this PR:

4-atom Al, UPF with NLCC, PBE, Ecut=32, 1 k-point, tol=1e-10:
| Stage | master CPU (12t) | loggrid CPU (12t) | master GPU | loggrid GPU |
|---|---|---|---|---|
| Basis + term setup | 0.254 s | 0.121 s (2.1×) | 0.111 s | 0.030 s (3.7×) |
| SCF (18 iters) | 11.1 s | 11.4 s (0.97×) | 2.87 s | 2.68 s (1.07×) |
| Forces | 0.364 s | 0.196 s (1.9×) | 0.101 s | 0.027 s (3.7×) |
| Stresses | 19.3 s | 3.09 s (6.2×) | 1.86 s | 0.527 s (3.5×) |

So apparently this PR works on GPU, nice! It's the first time I run DFTK on a GPU, I'm surprised it's this good. Why do we bother with CPUs anymore? Also even with this PR stresses are super slow, I might tackle that later.

azadoks:
Very cool! I just bumped SphericalBesselTransforms.jl to 0.1.2 and lowered the compat; so you should at least be able to check that off the list.

azadoks:
> (actually looks like there's an accuracy issue so don't do that yet)

What seems to be the problem here?

antoine-levitt:
This PR has a 2e-6 difference vs master (and so probably QE) in total energy on Al, and this can get even worse on other materials. This is only due to the projectors, everything else is fine. But I think this PR is doing the semi-correct thing.

First, the facts. ONCVPSP builds the projectors on its internal loggrid going to rc (which can be found in the PP_INFO header of the UPF file). It then interpolates it (by a order-7 Lagrange polynomial) to a linear grid that extends far beyond rc. This creates ringing which can be at the 1e-3 level (in relative L^∞ norm), extending over a few grid points (~.05 bohr), then going to exactly zero after a few grid points (splines are compactly supported). It looks like this (cube root scale for visualization):

To complicate matters further, there is a cutoff_radius_index field that appears to be somewhat arbitrary (Claude found it in ONCVPSP: it's  (max over channels of rc), snapped up to the linear grid, plus 4 mesh points, then applied to every PP_BETA in the file; upfout.f90: nrlproj = max_l(4 + int(rc(l)/drl)).). In particular see for the l=1 it's just far above rc. Master (and QE) integrate the psp to this cutoff_radius_index, with a relatively low-order simpson rule for endpoint correction. This PR sets β[cutoff_radius_index+1:end] to zeros then spline-interpolates on a log grid which extends far above rc.

So the bad news is that everybody is wrong. ONCVPSP uses a high-order spline to interpolate a function with a kink. Master integrates to an arbitrary index (which is *not* where the pseudo ends in many cases) with a low-order simpson. This PR is wrong because it assumes the function is smooth up to the max of the log grid and resamples using splines.

Everything looks like this:

proj_tail_PR_spline_samples

There are two ringings here: the first because of the log -> lin interpolation in ONCVPSP, the second because of the lin -> log interpolation in DFTK.

(the actual ringing is not as bad as it looks: this is cube root scale to increase contrast near zero; also it's oscillatory so it doesn't contribute that much to the integral)

Now contrast this with a PAW data file, which has the data on a native log scale (Fe LDA, generated by Dal Corso's atomic):

proj_tail_PAW_Fe_cuts_pad

There cutoff_radius_index is even more arbitrary.

So on the first picture the actual thing that one should do is to trust data only up to rc (which can be obtained from PP_INFO), and integrate up to there. On the second picture I don't even know what to do, it's so weird.

The good news is that it doesn't matter too much, these are small effects. The data is, to put it mildly, imperfect, and there's no actual good reference data to compare to, so essentially anything reasonable is fine here. This PR behavior is reasonable and simple, so I'd just go with it. One question to maybe settle is how to tune the spline order in r: 6 seems too high here, but it's reasonable for smoother quantities, and it gets the bulk right, which is presumably more important than the pathological edge. But without a good notion of reference, there's not that much we can do...

Technici4n:
Some comments:
- 2e-6 energy difference is fine IMO.
- The projector is indeed supposed to be exactly 0 beyond rc.
- I'm very surprised that the projectors are not smooth at rc... At least the pseudo wavefunctions should be up to 3rd to 5th order.
- I'm surprised that the cutoff_radius_index is chosen kinda randomly. It should just be a bit after the last non zero entry. This saves a lot of time when building the projectors via Simpson.

antoine-levitt:
Here's some more data from the JTH pseudos. There's two versions: XML and UPF. XML is on log grid, UPF is on lin grid, and has the interpolation artefacts. Eg https://www.abinit.org/atomic_data/paw/ATOMICDATA/026-fe, first XML then UPF
proj_tail_PAW_JTH_Fe
proj_tail_PAW_JTH_Fe_UPF
On the UPF files, cutoff_radius_index is uniformly non-informative. The right thing to do here is to get the information on the native (non resampled) log grid, and integrate up to the (actual) cutoff. But there's not enough information in the UPF files to reconstruct this. On the l=1 example here, master (and QE as well) will bite the inaccurate integration of  the 3e-3 jump (as well as the ringing artefacts), because it will integrate up to cutoff_radius_index and therefore will not endpoint-correct.

I'd say this PR is not particularly worse than master, so we merge this and later on we try to get data on native log grids  (and adopt this grid as our grid rather than making up a new one as we do here)

antoine-levitt:
Alternative: we keep the current slow bessel transform but sample p space, and then spline it when we need it. I think that's what abinit does (not sure about QE). We lose out on the nice properties of SBT (N log N, high-p accuracy possibly), but probably we keep the speedup, and we avoid tying the r and p space grids. I'd rather go log grid all the way because it seems cleaner to me, but it's a possibility. @azadoks any thoughts?

azadoks:
Hmm - maybe I'd suggest keeping both Hankel transform backends (quadrature, SBT) followed by splining.
That way, once sane log-grid pseudos are on the market, it's easy to use SBT for them.
Linear-mesh pseudos will always be around, and it seems like quadrature is the best way to treat them at the moment.

Splining afterwards gives a consistent interface for the rest of DFTK to the pseudo quantities, brings us in line with ABINIT and recent versions of QE, and removes all the complicated caching stuff.
The only question remaining on my side is what the AD/GPU support looks like for the interpolation packages in the ecosystem.
It would be a bit of a shame to roll our own, but I guess we've already essentially done that for quadrature.

antoine-levitt:
OK I'm going to keep the splining and throw away the sbt stuff. That should get us to acceptable performance, the sbt stuff can always be brought back if necessary, and I'd rather not have two different code paths. Iiuc stresses will compute in the dual basis the derivatives of the hankel transforms and then spline them as needed, which is maybe not how it's done in other codes I think but is better for accuracy anyway. Perf should be good enough.

PR #1348, opened 2026-07-15
A couple of drive-by (mostly phonon-related) GPU fixes

antoine-levitt:


src/postprocess/phonon.jl
```diff
@@ -102,8 +102,7 @@ in reduced coordinates.
     @assert !any(breaks_time_reversal_symmetry, basis.model.term_types) (
         "Phonons are currently only implemented in the presence of time-reversal-symmetry.")
     n_atoms = length(basis.model.positions)
-    δρs = [zeros(complex(T), basis.fft_size..., basis.model.n_spin_components)
-           for _ = 1:3, _ = 1:n_atoms]
+    δρs = [zero(ρ) for _ = 1:3, _ = 1:n_atoms]
```
Technici4n: rho is real but delta rho is complex at non-zero q.

Maybe `zeros_like(\rho, complex(T))`?

src/transfer.jl
```diff
@@ -75,7 +76,8 @@ function transfer_mapping(basis_in::PlaneWaveBasis,  kpt_in::Kpoint,
         idcs_out = idcs_out[idcs_out .!= nothing]
     end
 
-    idcs_in, idcs_out
+    # the nothings have been filtered out
+    to_device(basis_in.architecture, Vector{Int}(idcs_in)), to_device(basis_out.architecture, Vector{Int}(idcs_out))
```
Technici4n: Should be split in two. And why the `Vector{Int}`?

antoine-levitt: why do you mean split in two? Vector{Int} is to get rid of the nothings at the type level

src/postprocess/phonon.jl
antoine-levitt: Huh, that was the only functional change I did myself and not with Claude, that'll teach me :-)

src/transfer.jl
Technici4n: Ah so this was a `Vector{Union{Nothing,Int}}` hence the type cast?

For the split I just meant two lines instead of a big line with a small `,` in the middle :)

antoine-levitt: > Ah so this was a Vector{Union{Nothing,Int}} hence the type cast?

Yup, that was creating the GPU issues

> For the split I just meant two lines instead of a big line with a small , in the middle :)

That was also something Claude did and I changed it to one line :-D Fine I'll change it back

src/postprocess/phonon.jl
antoine-levitt: Cleaned up by deleting it (saves allocs)

src/postprocess/phonon.jl
```diff
@@ -102,21 +102,14 @@ in reduced coordinates.
     @assert !any(breaks_time_reversal_symmetry, basis.model.term_types) (
         "Phonons are currently only implemented in the presence of time-reversal-symmetry.")
     n_atoms = length(basis.model.positions)
-    δρs = [zeros(complex(T), basis.fft_size..., basis.model.n_spin_components)
-           for _ = 1:3, _ = 1:n_atoms]
-    δoccupations = [zero.(occupation) for _ = 1:3, _ = 1:n_atoms]
-    δψs = [zero.(ψ) for _ = 1:3, _ = 1:n_atoms]
-    for s = 1:n_atoms, α = 1:basis.model.n_dim
-        # Get δH ψ
+    responses = map(CartesianIndices((3, n_atoms))) do αs
```
Technici4n:
```suggestion
    responses = map(CartesianIndices((3, n_atoms))) do (α, s)
```
Does that work?

antoine-levitt: Nope, but something else like that does

Technici4n (approved):

PR #1352, opened 2026-07-17
Fixes for empty arrays

### LINKED ISSUE #1350: Phonon tests broken
@antoine-levitt PR https://github.com/JuliaMolSim/DFTK.jl/pull/1348 broke the tests, see master commit https://github.com/JuliaMolSim/DFTK.jl/commit/0fe16a0ad9142b94cd148bb940c21f54829e3da9. Could you take a look ?

antoine-levitt:
Fix https://github.com/JuliaMolSim/DFTK.jl/issues/1350. Also a drive-by fix for the same pattern in another file (robustness to zero-size arrays in julia is generally tricky, but I prefer this to special-casing nothings)

### DIFF
--- a/src/terms/Hamiltonian.jl
+++ b/src/terms/Hamiltonian.jl
@@ -102,7 +102,7 @@ end
                                                                         ψ::AbstractArray)
     function allocate_local_storage()
         T = eltype(H.basis)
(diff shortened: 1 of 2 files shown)


antoine-levitt:
Tested locally and should be innocuous so merging now to unbreak master

PR #1356, opened 2026-07-24
Threading defaults

antoine-levitt:
There was a gotcha I wasn't aware of, which meant that threading was turned off by default. Also set nblas to 1 to avoid overthreading. Also a drive by optimization

### DIFF
--- a/src/DFTK.jl
+++ b/src/DFTK.jl
@@ -299,6 +299,9 @@ end
 function __init__()
     # Reset timer; otherwise the starting time is the time of precompilation
     reset_timer!(timer)
+
+    # nthreads() is 1 at precompile time, so apply the thread defaults at load
+    setup_threading(; verbose=false)
 end
 
 end # module DFTK
--- a/src/common/threading.jl
+++ b/src/common/threading.jl
@@ -3,17 +3,17 @@ using LinearAlgebra
 using MPI
 
 """
-Setup the number of threads used by DFTK's own threading (`n_DFTK`), by BLAS (`n_blas`) 
+Setup the number of threads used by DFTK's own threading (`n_DFTK`), by BLAS (`n_blas`)
 and by FFTW (`n_fft`). This is independent from the number of Julia threads (`Threads.nthreads()`).
 DFTK and FFTW threading are upper bounded by `Threads.nthreads()`, but not BLAS,
 which uses its own threading system.
-By default, use 1 FFT thread, and `Threads.nthreads()` BLAS and DFTK threads.
+By default, use `Threads.nthreads()` DFTK threads and a single FFT and BLAS thread.
 """
-function setup_threading(; n_fft=1, n_blas=Threads.nthreads(), n_DFTK=Threads.nthreads())
+function setup_threading(; n_fft=1, n_blas=1, n_DFTK=Threads.nthreads(), verbose=true)
     set_DFTK_threads!(n_DFTK)
     FFTW.set_num_threads(n_fft)
     BLAS.set_num_threads(n_blas)
-    mpi_master(MPI.COMM_WORLD) && @info "Threading setup: " Threads.nthreads() n_DFTK n_fft n_blas
+    verbose && mpi_master(MPI.COMM_WORLD) && @info "Threading setup: " Threads.nthreads() n_DFTK n_fft n_blas
 end
 
 """
--- a/src/terms/xc.jl
+++ b/src/terms/xc.jl
@@ -370,8 +370,9 @@ function LibxcDensities(basis::PlaneWaveBasis{T}, max_derivative::Integer, ρ, �
         ∇ρ_real = similar(ρ_real,   n_spin, basis.fft_size..., 3)
         σ_real  = similar(ρ_real, n_spin_σ, basis.fft_size...)
 
+        Gs_cart = G_vectors_cart(basis)
         for α = 1:3
-            iGα = map(G -> im * G[α], G_vectors_cart(basis))
+            iGα = map(G -> im * G[α], Gs_cart)
             for σ = 1:n_spin
                 ∇ρ_real[σ, :, :, :, α] .= irfft(basis, iGα .* @view ρ_fourier[σ, :, :, :])
             end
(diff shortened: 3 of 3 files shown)


src/common/threading.jl
```diff
@@ -3,17 +3,17 @@ using LinearAlgebra
...
 """
-Setup the number of threads used by DFTK's own threading (`n_DFTK`), by BLAS (`n_blas`) 
+Setup the number of threads used by DFTK's own threading (`n_DFTK`), by BLAS (`n_blas`)
 and by FFTW (`n_fft`). This is independent from the number of Julia threads (`Threads.nthreads()`).
 DFTK and FFTW threading are upper bounded by `Threads.nthreads()`, but not BLAS,
 which uses its own threading system.
-By default, use 1 FFT thread, and `Threads.nthreads()` BLAS and DFTK threads.
+By default, use `Threads.nthreads()` DFTK threads and a single FFT and BLAS thread.
 """
-function setup_threading(; n_fft=1, n_blas=Threads.nthreads(), n_DFTK=Threads.nthreads())
```
Technici4n, line 12: Why default n_blas to 1? We need blas threading for efficient nonlocal operator.

src/DFTK.jl
```diff
@@ -299,6 +299,9 @@ end
 function __init__()
     # Reset timer; otherwise the starting time is the time of precompilation
     reset_timer!(timer)
+
+    # nthreads() is 1 at precompile time, so apply the thread defaults at load
+    setup_threading(; verbose=false)
```
Technici4n, line 304: I don't necessarily disagree with this, and I find it very annoying to have to call `setup_threading()` in every script. At the same time, this makes a simple `using DFTK` change global settings, which needs further discussion. Thoughts @mfherbst @niklasschmitz?

src/common/threading.jl
antoine-levitt: To avoid oversubscription (nonlocal runs inside kpoint parallelism). I thought it made sense to optimize by default for small systems (nonlocal negligible, several kpoints), no?

src/DFTK.jl
antoine-levitt: I thought this was a pure bug: const DFTK_threads = Ref(Threads.nthreads()) evaluates at *precompile time*, which I definitely did not have in mind. Makes sense that a setup routine would be in init, though.

src/common/threading.jl
Technici4n: > nonlocal runs inside kpoint parallelism

julia threads are also inside kpoint parallelism. For example, `mpirun -np 2 julia -t 4` will use 2 MPI processes, each with 4 threads, and I would thus expect us to use 4 BLAS threads as well.

src/DFTK.jl
Technici4n: For sure we should set `DFTK_threads` to `Threads.nthreads()` in init. But `setup_threading` also changes global FFT and BLAS settings hence it needs more discussion IMO.

(Still, I support this change, because by default BLAS will use _all available cores_ on the machine which is incredibly stupid, especially when running with MPI where it will vastly oversubscribe the poor CPU.)

src/common/threading.jl
antoine-levitt: It will still thread over the remaining kpoints, which is a reasonable default IMO

Technici4n: Do you not see what I mean? Check "DftHamiltonian multiplication". We use thread parallelization for the ffts but not for the nonlocal operator and that's why BLAS parallelization is important.

antoine-levitt: Oh wow I was convinced that the kpoint diagonalization loop was threaded, that was completely wrong, sorry! Hm. Why don't we do that again? But OK in that case  two options for setup_threading: either we default to setting BLAS threads to nthreads, or we just leave it alone. I think leaving it alone and letting julia decide on the correct thing makes more sense, no?

src/DFTK.jl
antoine-levitt: > by default BLAS will use all available cores

On my laptop (4 cores but hyperthreading) it defaults to 4 which seems sensible to me.

src/common/threading.jl
mfherbst: > I would thus expect us to use 4 BLAS threads as well.

There is also a subtlety that n_blas=1 on openblas is not the same as n_blas=1 when using MKL. Openblas shares the threadpool with julia on n_blas=1 (but not on n_blas>1) while MKL is strictly this many blas threads per julia thread. This is why having sane defaults here is basically impossible ... and why n_blas=1 on openblas is the right thing to do. The total number of threads is determined by the `-t` flag passed to julia.

src/DFTK.jl
mfherbst: >  At the same time, this makes a simple using DFTK change global settings, which needs further discussion. 

Given the unfortunate side-effects that even differ between openblas and MKL I don't think doing this by default on every `using DFTK` is a good idea.

mfherbst: But on the other hand n_blas=1 for openblas is almost always what one wants ...

mfherbst: Also if we set this globally, we should inform the user about it (i.e. keep the verbose=true)

Technici4n:
I ran some experiments with OpenBLAS on some ZnO system.

| Julia threads | SCF time with n_blas=1 | SCF time with n_blas=n_threads |
| --- | --- | --- |
| 1 | 38.6s | same |
| 2 | 28.2s | 26.5s |
| 4 | 23.4s | 16.9s |
| 8 | 19.4s | 14.1s |

So I would strongly suggest defaulting to n_blas == n_threads as was done before, especially since OpenBLAS is the default anyway.

mfherbst:
@Technici4n Thanks.

 On what type of hardware and julia version was this ? This does not match my experience (as I said, n_blas=1 for openblas at least used to be pretty generically a good choice for smallish problems)

Technici4n:
This is Nymphalis with Julia 1.12.6. If you think about it, the operations for which we need BLAS the most (nonlocal & eigensolver internals) are in fact not multithreaded by DFTK. So I think that even with MKL it might be better to use BLAS multithreading.

mfherbst:
Thanks. It's been I while I did this, maybe they also changed the way the thread pool is shared. (Point is openblas 1 thread did not mean no parallelism in blas, but a shared thread pool with julia threads)

PR #1357, opened 2026-07-24
Symmetrize densities via precomputed symmetry stars

### LINKED ISSUE #1355: Symmetrize over stars
Claude flagged that abinit and quantum espresso's version of accumulate_over_density! uses a different algorithm that is much more efficient - it implemented it and found 65x. I rejected it, it's about 100 lines of non obvious code plus a precomputed field plus a custom gpu kernel possibly and that function is heavy only for relatively small systems, but if we ever want to kill off that function it's easy enough to.

antoine-levitt:
Fix https://github.com/JuliaMolSim/DFTK.jl/issues/1355. Turns out it's not that hard and a 60x speedup is hard to ignore

### DIFF
--- a/src/PlaneWaveBasis.jl
+++ b/src/PlaneWaveBasis.jl
@@ -83,6 +83,8 @@ struct PlaneWaveBasis{T,
     ## Symmetry operations that leave the discretized model (k and r grids) invariant.
     # Subset of model.symmetries.
     symmetries::Vector{SymOp{VT}}
+    # Symmetry-equivalent G-vector stars, used to speed up symmetrization of ρ
+    symmetry_stars::Vector{SymmetryStar{VT}}
     # Whether the symmetry operations leave the rgrid invariant
     # If this is true, the symmetries are a property of the complete discretized model.
     # Therefore, all quantities should be symmetric to machine precision
@@ -240,6 +242,10 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Real, fft_size::Tuple{Int, Int, I
 
     dvol  = model.unit_cell_volume ./ prod(fft_size)
     terms = Vector{Any}(undef, length(model.term_types))  # Dummy terms array, filled below
+    
+    # precompute stars (only used by the CPU symmetrization of grid-preserving symmetries)
+    symmetry_stars = (architecture isa CPU && symmetries_respect_rgrid) ?
+                     compute_symmetry_stars(fft_size, symmetries) : SymmetryStar{value_type(T)}[]
 
     basis = PlaneWaveBasis{T, value_type(T), Arch, typeof(fft_grid),
                            typeof(kpoints[1].G_vectors)}(
@@ -249,7 +255,7 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Real, fft_size::Tuple{Int, Int, I
         kpoints, kweights, kgrid,
         kcoords_global, kweights_global, n_irreducible_kpoints,
         comm_kpts, krange_thisproc, krange_allprocs, krange_thisproc_allspin,
-        architecture, symmetries, symmetries_respect_rgrid,
+        architecture, symmetries, symmetry_stars, symmetries_respect_rgrid,
         use_symmetries_for_kpoint_reduction, terms)
 
     # Instantiate the terms with the basis
--- a/src/SymOp.jl
+++ b/src/SymOp.jl
@@ -116,14 +116,14 @@ end
 
 # A star is the orbit {G = S·G₀ : S in the group} of a representative G-vector G₀ on the FFT
 # grid. The symmetrized density is constant over a star up to a phase, ρ_sym(S·G₀) =
-# e^{-i (S·G₀)·τ_S} ρ_sym(G₀), so `accumulate_over_symmetries!` gathers ρ_sym(G₀) once and
-# scatters it across the orbit. (Defined here, before PlaneWaveBasis, as it is a basis field.)
+# e^{-i (S·G₀)·τ_S} ρ_sym(G₀), so `accumulate_over_symmetries!` gathers ρstar = ρ_sym(G₀) once and
+# scatters it across the orbit.
 struct SymmetryStar{T}
     # gather: ρstar = Σ_isym θ·ρ[iG] with (iG, θ) = sources[isym]
     sources::Vector{Pair{Int, Complex{T}}}
     # scatter: ρsym[iG] = θ·ρstar for (iG, θ) in members
     members::Vector{Pair{Int, Complex{T}}}
-    # G₀ is not stored: it is the sources/members entry of the trivial symmetry.
+    # G₀ is not stored: it is the sources/members entry with trivial symmetry.
 end
 
 # Discover the stars by scanning the grid: each grid point not already in a star seeds a new one
--- a/src/SymOp.jl
+++ b/src/SymOp.jl
@@ -37,16 +37,17 @@ struct SymOp{T <: Real}
     # (Uu)(G) = e^{-i G τ} u(S^-1 G) in reciprocal space
     S::Mat3{Int}
     τ::Vec3{T}
+    invS::Mat3{Int}  # = inv(S), integer since det S = 1
 end
 function SymOp(W, w::AbstractVector{T}) where {T}
     w = mod.(w, 1)
     S = W'
     τ = -W \ w
-    SymOp{T}(W, w, S, τ)
+    SymOp{T}(W, w, S, τ, Mat3{Int}(inv(S)))
 end
 
 function Base.convert(::Type{SymOp{T}}, S::SymOp{U}) where {U <: Real, T <: Real}
-    SymOp{T}(S.W, T.(S.w), S.S, T.(S.τ))
+    SymOp{T}(S.W, T.(S.w), S.S, T.(S.τ), S.invS)
 end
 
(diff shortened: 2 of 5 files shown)


antoine-levitt:
Good to merge I think. @abussy this does not touch the GPU, which stays as before. Is accumulate_over_symmetries! a bottleneck for the GPU? If not might as well leave it alone. If yes I can try to make a GPU version of this.

src/SymOp.jl
```diff
@@ -112,3 +113,49 @@ function complete_symop_group(symops; maxiter=10, kwargs...)
...
 end
+
+# A star is the orbit {G = S·G₀ : S in the group} of a representative G-vector G₀ on the FFT
+# grid. This is used to speed up the symmetrization
+# ρ_sym(G) = Σ_S e^{-i G·τ} ρ(S⁻¹G)
+# The symmetrized density is constant over a star up to a phase, ρ_sym(S·G₀) =
+# e^{-i (S·G₀)·τ_S} ρ_sym(G₀), so, to symmetrize a density, gather ρstar = ρ_sym(G₀) once and
+# scatter it across the orbit.
+# This is faster than the naive sum because, for a generic star, the
+# naive sum is O(members^2) while the optimized algorithm is
```
Technici4n, line 124: Hmmm? You mean naive is O(grid points x symmetries) whereas this is only O(grid points)?

src/SymOp.jl
```diff
@@ -112,3 +113,49 @@ function complete_symop_group(symops; maxiter=10, kwargs...)
...
+# O(members)
+struct SymmetryStar{T}
```
Technici4n, line 126: I think this name needs more bike shedding. Star is quite overloaded (e.g. q point stars). Maybe `GVectorOrbits`?

src/SymOp.jl
```diff
@@ -112,3 +113,49 @@ function complete_symop_group(symops; maxiter=10, kwargs...)
...
+    # gather: ρstar = Σ_isym θ·ρ[iG] with (iG, θ) = sources[isym]
```
Technici4n, line 127: This just describes the algorithm that is already elsewhere, no?

src/SymOp.jl
```diff
@@ -112,3 +113,49 @@ function complete_symop_group(symops; maxiter=10, kwargs...)
...
+    linear_ind = LinearIndices(fft_size)
+    visited    = falses(length(Gs))
+    stars      = SymmetryStar{T}[]
+    for lin in eachindex(Gs)
+        visited[lin] && continue
+        G = Gs[lin]
+        sources = Pair{Int, Complex{T}}[]
+        for symop in symmetries
+            idx = index_G_vectors(fft_size, symop.invS * G)
+            isnothing(idx) || push!(sources, linear_ind[idx] => cis2pi(-T(dot(G, symop.τ))))
```
Technici4n: Is the explicit T needed here?

Technici4n: Can it ever happen that it's nothing?

src/PlaneWaveBasis.jl
```diff
@@ -240,6 +242,10 @@ function PlaneWaveBasis(model::Model{T}, Ecut::Real, fft_size::Tuple{Int, Int, I
 
     dvol  = model.unit_cell_volume ./ prod(fft_size)
     terms = Vector{Any}(undef, length(model.term_types))  # Dummy terms array, filled below
+    
+    # precompute stars (only used by the CPU symmetrization of grid-preserving symmetries)
+    symmetry_stars = (architecture isa CPU && symmetries_respect_rgrid) ?
```
Technici4n, line 247: That's not very idiomatic.

src/symmetry.jl
```diff
@@ -269,47 +269,31 @@ function apply_symop(symop::SymOp, basis, kpoint, ψk::AbstractVecOrMat)
...
-    #     ρ ̂_{Sk}(G) = e^{-i G \cdot τ} ̂ρ_k(S^{-1} G) 
-    if all(isone, symmetries)
-        ρaccu .= ρin
-        return ρaccu
+# Accumulate the symmetrized density Σ_S e^{-i G·τ} ρ(S⁻¹G) into ρaccu
+function accumulate_over_symmetries!(ρaccu, ρin, basis, symmetries)
+    # The precomputed stars speed up the serial CPU traversal; the GPU (already parallel) and
+    # non-basis symmetries use the direct `map!`. Stars are exact only when the symmetries permute
+    # the grid, hence the `symmetries_respect_rgrid` guard.
+    if basis.architecture isa CPU && symmetries === basis.symmetries && basis.symmetries_respect_rgrid
```
Technici4n: Just check if the stars are non-empty, no?

Technici4n (commented):
Can't say I fully understand this.

Would it be possible to make this new algorithm also work on the GPU and in the case where `symmetries_respect_rgrid` is false? Such that the previous one can be eliminated.

antoine-levitt:
> Can't say I fully understand this.

Yeah it took me a while also but I finally got it; see the comments around the data structure.

> Would it be possible to make this new algorithm also work on the GPU and in the case where symmetries_respect_rgrid is false? Such that the previous one can be eliminated.

Yes, but both would probably add quite a bit of complexity. Symmetries_respect_rgrid false I don't care particularly as it's a niche case. GPU that's why I was asking if that's an important cost: if not I'd rather keep it as it is. Also I like the fact that we have two different algorithms for computing the same thing: it makes clear that the star thing is just an optimization over a more elementary operation.

abussy:
> Is accumulate_over_symmetries! a bottleneck for the GPU? If not might as well leave it alone. If yes I can try to make a GPU version of this.
 
 `accumulate_over_symmetries!` used to be a bottleneck for GPU runs. I did some work to port it from CPU to GPU (#1097), and it is now mostly negligible. That begin said, a 60x speedup is better than what I achieved with that porting.

The current state of this PR breaks the current GPU approach we have though, where we avoid explicit GPU/CPU branching and rely on multiple dispatch and abstract `map` calls to launch kernels (so as to have a simple, unified and portable code base).

I see two options:
1) since a 60x speedup is better than the generic GPU speedup, we could do all grid-preserving operations on the CPU anyways
2) we port the `_accumulate_over_symmetries_stars!` function to run on GPUs too

Both options avoid the architecture dependent branching. If `_accumulate_over_symmetries_stars!` is easily portable, which looks like it is, options 2 is probably best. I will investigate.

antoine-levitt:
> The current state of this PR breaks the current GPU approach we have though, where we avoid explicit GPU/CPU branching and rely on multiple dispatch and abstract map calls to launch kernels (so as to have a simple, unified and portable code base).

Yeah but that's just for "technical" implementations of the same algorithm. Here two actual different algorithms are used, so I'd rather this be explicit. Again I rather like having the two algorithms for code readability purposes.

> since a 60x speedup is better than the generic GPU speedup, we could do all grid-preserving operations on the CPU anyways

Hadn't thought of this, but maybe the simpler option, also in terms of code complexity? Unless data movement is too expensive.

src/SymOp.jl
antoine-levitt: I didn't phrase it like that because you can't really vary the parameters independently and O(symmetries) doesn't really make a lot of sense, but yes it's something like that. The naive algorithm will (essentially) loop over the stars, and compute the value of rho_out for each member of the star (from rho_in of all the others). The optimized algorithm computes a rhostar for all the star and dispatches it to the members.

Feel free to suggest modifications to the comment!

src/SymOp.jl
antoine-levitt: Don't know what q point stars are, is it not the same concept? Stars appears to be a quite common name, but we can qualify it (eg GVectorStar) if it's used elsewhere?

src/SymOp.jl
antoine-levitt: In my mind the main explanation is around the data structure, and the function has a pointer to it

src/SymOp.jl
antoine-levitt: Hm no, good catch. Bad bot.

antoine-levitt: Hopefully not if the grid preserves the symmetries, I'll assert instead

src/PlaneWaveBasis.jl
antoine-levitt: Which part? The ? : or the GPU?

src/symmetry.jl
antoine-levitt: Yeah I wanted to duplicate the logic so it's clearer at the point of use but OK fine

abussy:
> Hadn't thought of this, but maybe the simpler option, also in terms of code complexity? Unless data movement is too expensive.

My tests show that moving the data is too expensive compared to the brute force approach. 

I also thought about how to port the stars approach to the GPU, and it is more complicated than I initially anticipated. Given that density symmetrization is not a bottleneck on the GPU, the status quo is probably ok.

I haven't really run many CPU calculations, is symmetrization a real bottleneck? Because I would argue that this is a case of code uglification.

antoine-levitt:
> I haven't really run many CPU calculations, is symmetrization a real bottleneck? Because I would argue that this is a case of code uglification.

30% on small systems, so not insignificant for something that really should be trivial. It's an optimization for sure, but it's not that bad: that is a strict algorithmtic improvement (same in/out, much better complexity)

antoine-levitt:
> Can it ever happen that it's nothing?

Yes actually it can, my "fix" was stupid. Also the lowpass is necessary for the delta_V (not the rhos or deltarhos) so reverting as in master

PR #1360, opened 2026-07-29
Fix GPU in nonlocal term

abussy:
#1348 broke the GPU implementation because `D` matrix of the nonlocal term was not copied to the device anymore.

This highlights again the need for a GPU CI, I'll open an issue to discuss this.

### DIFF
--- a/src/terms/nonlocal.jl
+++ b/src/terms/nonlocal.jl
@@ -18,7 +18,7 @@ function (::AtomicNonlocal)(basis::PlaneWaveBasis{T}) where {T}
     isempty(psp_groups) && return TermNoop()
     ops = map(basis.kpoints) do kpt
         P = build_projection_vectors(basis, kpt, psps, psp_positions)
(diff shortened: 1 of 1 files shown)


src/terms/nonlocal.jl
```diff
@@ -18,7 +18,7 @@ function (::AtomicNonlocal)(basis::PlaneWaveBasis{T}) where {T}
     isempty(psp_groups) && return TermNoop()
     ops = map(basis.kpoints) do kpt
         P = build_projection_vectors(basis, kpt, psps, psp_positions)
-        D = build_projection_coefficients(T, psps, psp_positions)
+        D = build_projection_coefficients(basis, psps, psp_positions)
```
antoine-levitt, line 21: factor out of the loop (k-independent)

antoine-levitt (commented):
OK, I'm a bit lost, this file is a mess there are too many methods

antoine-levitt:
Sorry for the break, let's merge this to unbreak it and I'll try to refactor later

PR #1362, opened 2026-07-29
Add hubbard_n to the 'gdensity'

Technici4n:
Incremental progress towards the state refactor... :P

The gdensity is now a tuple of the density, KED, and `hubbard_n`. A `ComponentArray` is not pratical because 1) `hubbard_n` is quite deeply nested and 2) `hubbard_n` is Complex. Regarding 2), apparently it should be real as long as time-reversal symmetry is satisfied, but we probably want to keep it complex for max generality.

Maybe we'll want to move towards a custom struct eventually with a vector-like interface. Currently I implemented addition, subtraction and damping by explicitly breaking the gdensity into its 3 components.

The `hubbard_n` is still not Anderson-accelerated, but this would be possible to implement at the level of the scf solver (quite localized which is nice).

(Started as a vibe coding experiment but then I tweaked/rewrote most of it.)

### DIFF
--- a/src/scf/mixing.jl
+++ b/src/scf/mixing.jl
@@ -23,12 +23,12 @@ function mix_potential(args...; kwargs...)
 end
 
 
-# Mixing in the generalised density (essentially an adapted tuple of ρ and τ;
-# see pack_gdensity in densities.jl): For now just fall back to ρ-only mixing
+# Mixing in the generalised density (see pack_gdensity in densities.jl):
+# For now just fall back to ρ-only mixing
 function mix_gdensity(mixing, basis, ΔD; kwargs...)
-    Δρ, Δτ  = split_gdensity(basis, ΔD)
+    Δρ, Δτ, Δhubbard_n = split_gdensity(basis, ΔD)
     Pinv_Δρ = mix_density(mixing, basis, Δρ; kwargs...)
-    pack_gdensity(basis, Pinv_Δρ, Δτ)
+    pack_gdensity(basis, Pinv_Δρ, Δτ, Δhubbard_n)
 end
 
 
--- a/src/scf/self_consistent_field.jl
+++ b/src/scf/self_consistent_field.jl
@@ -123,9 +123,16 @@ function next_density(ham::Hamiltonian,
         τ = nothing
     end
 
-    (; ψ=eigres.X, eigenvalues=eigres.λ, occupation, εF, ρ, τ, diagonalization=eigres,
-     n_bands_converge, nbandsalg.occupation_threshold,
-     n_matvec=mpi_sum(eigres.n_matvec, ham.basis.comm_kpts))
+    ihubbard = findfirst(t -> t isa TermHubbard, ham.basis.terms)
+    if !isnothing(ihubbard)
+        hubbard_n = compute_hubbard_n(ham.basis.terms[ihubbard], ham.basis, eigres.X, occupation)
+    else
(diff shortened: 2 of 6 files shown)


src/scf/self_consistent_field.jl
```diff
@@ -236,17 +234,30 @@ Overview of parameters:
...
-        ΔD = D - Din
-        Δρ, Δτ = split_gdensity(basis, ΔD)
+        # Residual of the generalised density. Only ρ and τ are mixed/accelerated, so the
+        # residual only needs their differences; hubbard_n is patched through, so it is passed
+        # along as its new value rather than as an increment (see below).
+        Δρ = ρ - ρin
+        Δτ = isnothing(τ) ? nothing : τ - τin
+        if isnothing(hubbard_nin) && !isnothing(hubbard_n)
+            hubbard_nin = zero(hubbard_n)
+        end
```
Technici4n: This is hacky and could be replaced by a proper `guess_hubbard_n` function.

src/scf/mixing.jl
```diff
@@ -23,12 +23,12 @@ function mix_potential(args...; kwargs...)
 end
 
 
-# Mixing in the generalised density (essentially an adapted tuple of ρ and τ;
-# see pack_gdensity in densities.jl): For now just fall back to ρ-only mixing
+# Mixing in the generalised density (see pack_gdensity in densities.jl):
+# For now just fall back to ρ-only mixing
 function mix_gdensity(mixing, basis, ΔD; kwargs...)
-    Δρ, Δτ  = split_gdensity(basis, ΔD)
+    Δρ, Δτ, Δhubbard_n = split_gdensity(basis, ΔD)
```
antoine-levitt, line 29: should be unpack for symmetry?

antoine-levitt (commented):
The way we do it in direct minimization is that pack/unpack take structured data and give a plain vector suitable for linear algebra (in L^2, so the packing/unpacking includes representation data). That way the solvers don't have to know about structured data. Maybe we can do that here?

Technici4n:
> The way we do it in direct minimization is that pack/unpack take structured data and give a plain vector suitable for linear algebra (in L^2, so the packing/unpacking includes representation data). That way the solvers don't have to know about structured data. Maybe we can do that here?

Well, flattening it all into an L^2 vector doesn't really make sense I think: hubbard_n is a bit complicated (something like `Vector{Array{3, Matrix{Complex{T}}}}`) and we really don't want to take inner products with it simply stacked next to the density.

antoine-levitt:
> Well, flattening it all into an L^2 vector doesn't really make sense I think: hubbard_n is a bit complicated (something like Vector{Array{3, Matrix{Complex{T}}}}) and we really don't want to take inner products with it simply stacked next to the density.

Are we not going to do that to mix them?

Technici4n:
We'll probably have to when we get to that at some point, yeah. I don't know what the best mixing/preconditioning is, if it should be included in the Anderson (I would guess yes), etc. However I am quite sure that just stacking the hubbard_n alongside the rho/tau makes no sense: rho/tau are densities, hubbard_n has entries of magnitude < 1. That's why I sidestepped this question in this PR.

antoine-levitt:
That's kind of the point: the preconditioning goes into the pack/unpack routines and make all the quantities of the same magnitude (actually magnitude is not the right word, it's more to do with the hardness of the jacobian), and then everything gets mushed together. I don't know what else to do than that...

Issue #1334, opened 2026-06-30
Separate wavefunction and density grids


Technici4n:
We should introduce two FFT grids: the "psi" grid which is used for operations involving orbitals and always able to fit up to 2*Gmax where Gmax = $2\sqrt{E_\mathrm{cut}}$, and a density or "rho" grid which can be denser. By default, the two grids would be the same, however supersampling should only make the rho grid denser, while keeping the psi grid fixed.

#### Why the psi grid should never be larger than 2*Gmax
The application of a potential in Fourier space is:
```math
(V \psi)(\mathbf G) = \frac{1}{\Omega} \sum_{\mathbf G'} \hat{V}(\mathbf G - \mathbf G') \psi(\mathbf G')
```
Since we only care about coefficients <= Gmax, this means that we only care about $\mathbf G - \mathbf G'$ that fit in a grid of max frequency 2Gmax. So representing a local potential on a denser grid is entirely useless, for the purpose of applying it to wave functions.

#### Why it's useful to have a denser rho grid sometimes
The only term that is actually evaluated in real space is the XC potential. It is sampled on a uniform grid, so the high-frequency components of the XC potential can actually get aliased. This can be a problem for more advanced functionals which are less smooth (e.g. meta-GGA functionals). In that case, increasing the density of the rho grid can fix the problem at a very low cost (since the local potential applications remain in the coarser psi grid).

#### Inside vs outside the XC term
You can actually understand this as an effect internal to the XC term. You use $E_{xc}[\mathrm{upsample}(\rho)]$. The adjoint of upsampling is downsampling (both in Fourier space) so you will naturally do the right thing when constructing the potential. However I would suggest that this should instead live outside the XC term, and that all densities $\rho$ computed by DFTK should be upsampled already, such that we will already have the "dual grid" infrastructure in place for ultrasoft PSPs.

#### My suggestion
So I suggest the following:
- Introducing separate grids for rho and psi. In most cases (supersampling = 2, the default) they will be the same.
- Transferring between grids will be done in Fourier space by padding or truncation.
- I also suggest deprecating supersampling and replacing it by something like `Ecutrho` which would default to `4*Ecut`.

antoine-levitt:
I agree in principle. However, a couple of things
- We might want to be careful to avoid too many copies in the default case
- I think it makes sense to keep the objects (rho, V) on the smallest grid they can/need to be represented on. Can you elaborate on ultrasoft PSPs so we get a better idea of the design space?
[...]

PR #1365, opened 2026-07-30
Optionally use a denser grid for the XC evaluation

Technici4n:
The main idea here is to beat the egg-box effect that typically occurs for meta-GGA functionals. Related work: https://doi.org/10.1088/2516-1075/adc056
Related: #1334.

- We already had the density transfer infrastructure so it was not too bad.
- I added the infrastructure for extra `PlaneWaveBasis` kwargs to be passed to the terms at instantiation time. E.g. `PlaneWaveBasis(model; Ecut, kgrid, supersampling_xc=4)`.

First version written by Claude Opus 5.

### DIFF
--- a/src/PlaneWaveBasis.jl
+++ b/src/PlaneWaveBasis.jl
@@ -400,23 +400,29 @@ end
 
 
 @doc raw"""
+    G_vectors(fft_grid::FFTGrid)
     G_vectors(basis::PlaneWaveBasis)
     G_vectors(basis::PlaneWaveBasis, kpt::Kpoint)
 
-The list of wave vectors ``G`` in reduced (integer) coordinates of a `basis`
-or a ``k``-point `kpt`.
+The list of wave vectors ``G`` in reduced (integer) coordinates of an
+FFT grid, a `basis` or a ``k``-point `kpt`.
 """
-G_vectors(basis::PlaneWaveBasis) = basis.fft_grid.G_vectors
+G_vectors(fft_grid::FFTGrid) = fft_grid.G_vectors
+G_vectors(basis::PlaneWaveBasis) = G_vectors(basis.fft_grid)
 G_vectors(::PlaneWaveBasis, kpt::Kpoint) = kpt.G_vectors
 
 @doc raw"""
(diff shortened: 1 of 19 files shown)


antoine-levitt:
Hm in fact the fft_size logic in planewavebasis is wrong: instead of restricting the symmetries to those that preserve the rgrid, we should adapt fft_size upwards to preserve the symmetries (if possible, this is all very confusing to me I have to say, maybe the AI has better ideas). For the xc we then reuse the same logic with a supersampling parameter

Technici4n:
I think we should move away from the concept of "supersampling" and instead think of a density Ecut; the latter can be linked to physical oscillations of the density. This will be important for ultrasoft as well.

> we should adapt fft_size upwards to preserve the symmetries

Maybe yeah. I wonder if some symmetries can simply never be expressed in an FFT grid? This might also be mainly determined by the prime factors of the fft size? Symmetries never cease to confuse me :D

src/PlaneWaveBasis.jl
```diff
@@ -400,23 +400,29 @@ end
...
     G_vectors(basis::PlaneWaveBasis)
     G_vectors(basis::PlaneWaveBasis, kpt::Kpoint)
 
-The list of wave vectors ``G`` in reduced (integer) coordinates of a `basis`
-or a ``k``-point `kpt`.
+The list of wave vectors ``G`` in reduced (integer) coordinates of an
+FFT grid, a `basis` or a ``k``-point `kpt`.
 """
-G_vectors(basis::PlaneWaveBasis) = basis.fft_grid.G_vectors
+G_vectors(fft_grid::FFTGrid) = fft_grid.G_vectors
```
Technici4n: this method already exists in fft.jl!

src/terms/xc.jl
```diff
@@ -340,8 +422,9 @@ end
 
 
 # stores the input to libxc in a format it likes
-struct LibxcDensities{T}
-    basis::PlaneWaveBasis{T}
+struct LibxcDensities{FFTtype}
```
Technici4n, line 425: we can probably drop the generic entirely

PR #1369, opened 2026-08-03
Delete lowpass_for_symmetry!

### LINKED ISSUE #1366: Remove lowpass_for_symmetry!
### Where it is used now

`lowpass_for_symmetry!` runs inside `symmetrize_ρ` when `do_lowpass=true`. The only production caller with the default is `apply_χ0` (`src/response/chi0.jl:527`), on `δV`. The three density call sites — `compute_density`, `compute_δρ`, `compute_kinetic_energy_density` — all pass `do_lowpass=false` since #671.

### Why it was introduced

#364, fixing #351: with `supersampling=1.5` the SCF died with `Input array not real`. Symmetrization writes into `ρ(G)` for a `G` whose `-G` is outside the box, the array stops being conjugate-symmetric, and the density comes out complex. The filter zeroes those `G`.

This is a property of the box, not of the symmetries: a box is not rotation-invariant, so `S·G` leaves it for a large fraction of `G` even when every symop permutes the r-grid (silicon, all 48 symops preserved: 56% of the grid at supersampling 2).

### It is a no-op for supersampling ≥ 2

`ρ` is a product of two orbitals, so its support is `|G| ≤ 2·Gmax_orbital`. That ball is `S`-invariant and contained in the box, so every `G` the filter touches carries `ρ = 0`. Silicon LDA, `Ecut=10`, `kgrid=[2,2,2]`, norm of the imaginary part that `irfft` discards after symmetrizing:

| supersampling | ‖Im ρ_sym‖ |
|---|---|
| 2.0 | 2.6e-16 |
| 1.2 | 3.1e-4  (‖Re ρ_sym‖ = 2.3) |

### What if supersampling < 2

That regime is already approximate — the density is aliased and χ0 is not self-adjoint (1.4e-4, against the 2e-6 tolerance of the chi0 tests), with or without the filter. And the protection is gone in practice anyway: `compute_density` no longer applies it, and `irfft(x) = real(ifft(x))` silently drops the imaginary part now that `RealFourierArray`'s check is gone. #351's reproducer today returns a density with a ~1e-4 inconsistency instead of erroring.

### Applying it inconsistently, as now, is worse than not applying it

`apply_χ0` symmetrizes `δV` with the filter, `compute_δρ` symmetrizes `δρ` without, so χ0 is no longer sandwiched between the same projector. Self-adjointness defect `|⟨δV₁, χ₀δV₂⟩ - ⟨δV₂, χ₀δV₁⟩|`, silicon, chi0 test tolerance 2e-6:

| | δV only (current) | never | δV and δρ |
|---|---|---|---|
| supersampling 2.0 | 9.5e-10 | 9.6e-10 | 9.6e-10 |
| supersampling 1.2 | **1.4e-4** | 4.0e-11 | 1.2e-10 |

### Proposal

Delete `lowpass_for_symmetry!` and the `do_lowpass` keyword, and warn when `supersampling < 2`, so the invariant the code has silently relied on since #671 is stated rather than half-enforced.

One thing we lose: `symmetrize_ρ` is then not a projector for input with content beyond the density cutoff (0.30 relative defect on random input, machine precision on a density). If we want to keep that property, the alternative is to turn the filter into a mask precomputed with the basis and applied on both sides — `O(n_G)` instead of `O(n_G · n_symm)`.

antoine-levitt:
Fix https://github.com/JuliaMolSim/DFTK.jl/issues/1366. Along the way, fix flaky tests. The test now uses fft_size=nothing instead of the hardcoded one (all sorts of things go bad with supersampling < 2 anyway)

### DIFF
--- a/src/symmetry.jl
+++ b/src/symmetry.jl
@@ -318,39 +318,16 @@ function accumulate_over_symmetries!(ρaccu, ρin, basis::PlaneWaveBasis{T}, sym
     ρaccu
 end
 
-# Low-pass filters ρ (in Fourier) so that symmetry operations acting on it stay in the grid
-# This function is optimized for CPU and GPU.
-function lowpass_for_symmetry!(ρ::AbstractArray, basis; symmetries=basis.symmetries)
-    all(isone, symmetries) && return ρ
-
-    Gs = reshape(G_vectors(basis), size(ρ))
-    fft_size = basis.fft_size
-
-    symm_S = to_device(basis.architecture, [symop.S for symop in symmetries])
-
-    # Loop structure optimized for both CPU and GPU
-    map!(ρ, ρ, Gs) do ρ_i, G
-        acc = ρ_i
-        for S in symm_S
-            idx = index_G_vectors(fft_size, S * G)
-            acc *= isnothing(idx) ? 0 : 1
-        end
-        acc
-    end
-    ρ
-end
-
 """
 Symmetrize a density by applying all the basis (by default) symmetries and forming the average.
 """
 @views @timing function symmetrize_ρ(basis, ρ::AbstractArray{T};
-                                     symmetries=basis.symmetries, do_lowpass=true) where {T}
+                                     symmetries=basis.symmetries) where {T}
     ρin_fourier  = fft(basis, ρ)
     ρout_fourier = zero(ρin_fourier)
     for σ = 1:size(ρ, 4)
         accumulate_over_symmetries!(ρout_fourier[:, :, :, σ],
                                     ρin_fourier[:, :, :, σ], basis, symmetries)
-        do_lowpass && lowpass_for_symmetry!(ρout_fourier[:, :, :, σ], basis; symmetries)
     end
     inv_fft = T <: Real ? irfft : ifft
     inv_fft(basis, ρout_fourier ./ length(symmetries))
@@ -548,5 +525,9 @@ Ensure its real-space equivalent of passed Fourier-space representation is entir
 removing wavevectors `G` that don't have a `-G` counterpart in the basis.
 """
 @timing function enforce_real!(fourier_coeffs, basis::PlaneWaveBasis)
-    lowpass_for_symmetry!(fourier_coeffs, basis; symmetries=[SymOp(-Mat3(I), Vec3(0, 0, 0))])
+    Gs = reshape(G_vectors(basis), size(fourier_coeffs))
+    fft_size = basis.fft_size
+    map!(fourier_coeffs, fourier_coeffs, Gs) do coeff, G
+        isnothing(index_G_vectors(fft_size, -G)) ? zero(coeff) : coeff
+    end
 end
--- a/src/symmetry.jl
+++ b/src/symmetry.jl
@@ -521,10 +521,10 @@ function unfold_kcoords(kcoords, symmetries)
 end
 
 """
-Zero out the Nyquist frequencies of a Fourier-space representation, ie the wavevectors `G`
+Zero out the Nyquist components of a Fourier-space representation, ie the wavevectors `G`
 that don't have a `-G` counterpart in the basis (only present for even `fft_size`).
 """
-@timing function drop_nyquist_frequency!(fourier_coeffs, basis::PlaneWaveBasis)
+@timing function drop_nyquist_components!(fourier_coeffs, basis::PlaneWaveBasis)
     Gs = reshape(G_vectors(basis), size(fourier_coeffs))
     fft_size = basis.fft_size
     map!(fourier_coeffs, fourier_coeffs, Gs) do coeff, G
--- a/src/symmetry.jl
+++ b/src/symmetry.jl
@@ -519,15 +519,3 @@ function unfold_kcoords(kcoords, symmetries)
         normalize_kpoint_coordinate(round.(k; digits) .+ 0.0)
     end
 end
-
-"""
-Zero out the Nyquist components of a Fourier-space representation, ie the wavevectors `G`
-that don't have a `-G` counterpart in the basis (only present for even `fft_size`).
-"""
-@timing function drop_nyquist_components!(fourier_coeffs, basis::PlaneWaveBasis)
-    Gs = reshape(G_vectors(basis), size(fourier_coeffs))
-    fft_size = basis.fft_size
-    map!(fourier_coeffs, fourier_coeffs, Gs) do coeff, G
-        isnothing(index_G_vectors(fft_size, -G)) ? zero(coeff) : coeff
-    end
-end
--- a/src/symmetry.jl
+++ b/src/symmetry.jl
@@ -521,10 +521,10 @@ function unfold_kcoords(kcoords, symmetries)
 end
 
 """
-Ensure its real-space equivalent of passed Fourier-space representation is entirely real by
-removing wavevectors `G` that don't have a `-G` counterpart in the basis.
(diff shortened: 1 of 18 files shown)


src/symmetry.jl
```diff
@@ -548,5 +525,9 @@ Ensure its real-space equivalent of passed Fourier-space representation is entir
 removing wavevectors `G` that don't have a `-G` counterpart in the basis.
 """
 @timing function enforce_real!(fourier_coeffs, basis::PlaneWaveBasis)
```
Technici4n: Can this function be deleted as well? I think a similar argument as for `lowpass_for_symmetry!` can be made (why is it even needed? `irfft` will drop the imaginary part anyway), and it caused an inconsistency in the NLCC force: https://github.com/JuliaMolSim/DFTK.jl/pull/1370

test/chi0.jl
```diff
@@ -31,27 +32,30 @@ function test_chi0(testcase; symmetries=false, temperature=0, spin_polarization=
         basis = PlaneWaveBasis(model; basis_kwargs...)
         ρ0    = guess_density(basis, magnetic_moments)
         ham0  = energy_hamiltonian(basis, nothing, nothing; ρ=ρ0).ham
-        nbandsalg = is_εF_fixed ? FixedBands(; n_bands_converge=6) : AdaptiveBands(model)
+        default_bands = is_εF_fixed ? FixedBands(; n_bands_converge=6) : AdaptiveBands(model)
+        nbandsalg = FixedBands(; n_bands_converge=default_bands.n_bands_compute,
+                               n_bands_compute=default_bands.n_bands_compute + 3,
+                               default_bands.occupation_threshold)
```
Technici4n: what is this change doing?

test/chi0.jl
```diff
@@ -31,27 +32,30 @@ function test_chi0(testcase; symmetries=false, temperature=0, spin_polarization=
...
         res = DFTK.next_density(ham0, nbandsalg; tol, eigensolver)
         scfres = (; ham=ham0, res...)
 
-        # create external small perturbation εδV
+        # create external small perturbation εδV. apply_χ0 (being grounded on a symmetric basis)
+        # only resolves perturbations respecting the symmetries, so we make δV symmetric.
```
Technici4n, line 43: leftover AI generated comment?

test/kernel.jl
```diff
@@ -180,6 +181,15 @@ end
             δΔρ = reshape(δΔρ_real, size(Δρ)...)
             δτ  = mpi_bcast!(randn(size(τ)) / model.unit_cell_volume, basis.comm_kpts)
 
+            # Finite differences walk along the linearised σ + εδσ, while the true σ(ε) is
+            # quadratic where ∇ρ vanishes by symmetry: at those points the stencil leaves
+            # σ ≥ 0, gets clamped by Libxc and straddles a kink. Drop them. (Only the
+            # same-spin channels are constrained; the αβ one is ∇ρα⋅∇ρβ and may be negative.)
```
Technici4n: leftover AI generated comment?

src/symmetry.jl
Technici4n: OK so for derivation we kinda have to keep `enforce_real!`, it's just that the name is misleading. Do also note that `enforce_real!` actually changes the resulting real IFFT a little bit.

Technici4n: Yeah IMO don't delete this but just rename it: https://github.com/JuliaMolSim/DFTK.jl/issues/1371. Either in this PR or another one.

test/chi0.jl
antoine-levitt: Actually converging all the bands - the previous version didn't converge enough the ones with low occupation

test/chi0.jl
antoine-levitt: nope this is mine :-p

test/kernel.jl
antoine-levitt: yeah but i think it's nice to keep

test/chi0.jl
mfherbst: This is more or less:
```suggestion
        nbandsalg = is_εF_fixed ? FixedBands(; n_bands_converge=6) : FixedBands(model)
```

test/kernel.jl
mfherbst: But it's a bit clunky to read. I just had to read it 3 times to get it. Can you rephrase ?

mfherbst:
Fine for me if my two comments are addressed.

test/kernel.jl
antoine-levitt: Yeah, it was a misattribution anyway, the clamp is not the issue, it's a a meta-gga stiffness issue, so I'm just raising the tol for everybody

antoine-levitt: There was actually two issues: the clamp thing, and a meta-gga stiffness issue. Raised the sigma to avoid the clamp, and loosened the tolerance to avoid the stiffness.

test/kernel.jl
```diff
@@ -180,6 +181,17 @@ end
...
+            # Where ∇ρ vanishes by symmetry σ is ~1e-18, far below the reach of the stencil,
+            # which would then straddle the σ ≥ 0 clamp in Libxc at any ε_fd. Lift it clear;
+            # the αβ channel is ∇ρα⋅∇ρβ and may legitimately be negative, so leave it alone.
+            same_spin = size(σ, 1) == 1 ? [1] : [1, 3]
+            σ = copy(σ)
+            σ[same_spin, :] .= max.(σ[same_spin, :], 4ε_fd .* abs.(δσ[same_spin, :]))
+
+            # Loose tolerance: at a few points of small σ the meta-GGAs are stiff enough
+            # that the ε_fd⁴ truncation error still reaches ~2e-6.
+            rtol = 1e-5
```
mfherbst, line 192: Only do this for meta-GGA ?

test/kernel.jl
```diff
@@ -180,6 +181,17 @@ end
             δΔρ = reshape(δΔρ_real, size(Δρ)...)
             δτ  = mpi_bcast!(randn(size(τ)) / model.unit_cell_volume, basis.comm_kpts)
 
+            # Where ∇ρ vanishes by symmetry σ is ~1e-18, far below the reach of the stencil,
+            # which would then straddle the σ ≥ 0 clamp in Libxc at any ε_fd. Lift it clear;
```
mfherbst: Can you define "Lift it clear" please ? Sounds like an LLM artifact.

test/kernel.jl
antoine-levitt: There's not that much point... I'm a bit tired of tolerance chasing, I'd rather have a lenient enough ceiling that we don't have to change it every two weeks. If tests pass up to 1e-5 it means there aren't any blatant mistakes in the code (which is really the only thing that could happen, this is just wrapper code)

test/kernel.jl
antoine-levitt: sounds fine enough to me but sure

mfherbst (approved):
Lgtm

mfherbst:
Test failures in `test/hubbard.jl` are real. @antoine-levitt could you have a look ?

antoine-levitt:
| fft_size | keep Nyquist (PR) | zero Nyquist (master) |
  |---|---|---|
  | 36  | -8.336019 | -8.347792 |
  | 108 | -8.322655 | -8.322655 |

so that's just noise on the nlcc (which would need supersampling or a bigger ecut to resolve properly): both are wrong anyway. The fact that this PR is less wrong here is probably not worth reading too much into (although this PR plausibly discards less information than master and so might be more accurate, but it's debatable). I just adapted the test numbers

PR #1370, opened 2026-08-05
Fix NLCC force missing enforce_real!

Technici4n:
~~This is one possible fix, but upon closer inspection the entire `enforce_real!` function is very suspicious. Let's discuss in #1369.~~ This is the correct fix, it's just that `enforce_real!` is misnamed (should be renamed: https://github.com/JuliaMolSim/DFTK.jl/issues/1371).

### DIFF
--- a/src/terms/xc.jl
+++ b/src/terms/xc.jl
@@ -244,6 +244,9 @@ end
 # Function barrier to work around various type instabilities.
 function _forces_xc(basis::PlaneWaveBasis{T}, Vxc_fourier::AbstractArray{U}, 
                     form_factors, iG2ifnorm, groups) where {T, U}
+    # Adjoint of the enforce_real! call in the construction of the NLCC
+    enforce_real!(Vxc_fourier, basis)
+
     # Pre-allocation of large arrays for GPU Efficiency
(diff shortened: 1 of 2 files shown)


antoine-levitt:
Yeah this kind of makes me want to delete it entirely. Alternatively we might want to drop it in irfft?

antoine-levitt:
Yeah thinking about it, either moving it to irfft or dropping it entirely is the right thing to do. I'll move it to irfft, which should be essentially a non-functional change; then if we can decide if we want to drop it later.

Technici4n:
I don't know about moving it to irfft, and if you do that it will not fix this bug anymore.

antoine-levitt:
Hm, good point,. Fine, let's drop it completely?

Technici4n:
I thought about that, but keeping the Nyquist components will cause trouble with derivatives I think (seems to be a well-known problem in signal processing). So as much as I don't like it I think we should keep zeroing it. See also https://github.com/JuliaMolSim/DFTK.jl/issues/1371#issuecomment-5192579014

mfherbst:
I'll merge this now, rename happens in #1366

PR #1340, opened 2026-07-11
Extract LOBPCG eigensolver into standalone LOBPCGEigensolver.jl package

antoine-levitt:
Extracts the stability-focused **LOBPCG algorithm** out of `src/eigen/` into a new, dependency-light package **[LOBPCGEigensolver.jl](https://github.com/JuliaMolSim/LOBPCGEigensolver.jl)** under JuliaMolSim, which exposes the low-level `lobpcg(A, X, B, precon, tol, maxiter; …)` entry point. DFTK depends on it.

### What moved to the package
- The LOBPCG algorithm itself (formerly `lobpcg_hyper_impl.jl`).
- The GPU-only `compute_λ` helper and the AMDGPU `LinearAlgebra` workarounds (Cholesky / `gesdd!` / 5-arg `mul!`), now in `LOBPCGEigensolver`'s AMDGPU extension — loaded automatically whenever DFTK and AMDGPU are both present.
- The LOBPCG-internal unit tests.

### What stays in DFTK
- The `lobpcg_hyper` convenience wrapper (default tolerance, `converged`/`n_iter` reporting) and the dense `diag_full` reference solver — these aren't LOBPCG per se; they now call `LOBPCGEigensolver.lobpcg`.
- `diagonalize_all_kblocks` / `select_eigenpairs_all_kblocks` (the k-block driver).
- `PreconditionerTPA` / `PreconditionerNone`, which now extend `LOBPCGEigensolver.precondprep!`.
- DFTK's own `columnwise_dots` / `columnwise_norms` (used well beyond the solver).

### Timings are preserved
The solver takes the `TimerOutput` to record into as a `timer` keyword (added in [LOBPCGEigensolver#6](https://github.com/JuliaMolSim/LOBPCGEigensolver.jl/pull/6), released in 0.1.1), and `lobpcg_hyper` hands it `DFTK.timer`. The solver's internal sections therefore show up in the DFTK timer exactly as before the extraction, now nested under `lobpcg_hyper`:

```
self_consistent_field
  lobpcg_hyper
    matvec
      DftHamiltonian multiplication
    ortho! X vs Y
      ortho!
    rayleigh_ritz
    preconditioning
    Update residuals
```

Since `@timing` only compiles away for DFTK's own code, the timer object itself is disabled when the `timer_enabled` preference is off, so the solver stays silent too.

### Incidental fix
[...]

PR #1372, opened 2026-08-05
Move to LOBPCGEigensolver package

abussy:
Move to the newly registered LOBPCGEigensolver package. Tests pass on CPU, CUDA and AMDGPU.


antoine-levitt:
Didn't you see https://github.com/JuliaMolSim/DFTK.jl/pull/1340? It should be about the same, no?

abussy:
I didn't sorry :see_no_evil:. It's the same.

PR #1374, opened 2026-08-12
Update GPU docs (incl. running on cluster)

abussy:
Updates to the GPU documentation, and especially the tricky business of running CUDA-aware MPI calculations without insane start-up time.

### DIFF
--- a/docs/src/tricks/compute_clusters.md
+++ b/docs/src/tricks/compute_clusters.md
@@ -398,7 +398,7 @@ export CUDA_CACHE_PATH="/path_to_existing_folder_on_scratch"
 
 my_project="/absolute_path_to_julia_project"
 
-srun julia --project=$my_project example.jl > example.out
+srun julia --project=$my_project --compiled-modules=strict example.jl > example.out
 ```
 
 ### Special notes on NVIDIA clusters
--- a/docs/src/tricks/compute_clusters.md
+++ b/docs/src/tricks/compute_clusters.md
@@ -313,6 +313,123 @@ srun julia -t 1 --project -e '
 '
 ```
 
+## Running DFTK with MPI and GPUs
+Note that DFTK is implemented for GPU-aware MPI. If your cluster does not support
+such an MPI implementation, keep to CPU execution. This tutorial is illustrated with
+examples from a NVIDIA cluster using CUDA.jl, but adaptation to AMD GPUs is straightforward. We recommend using a [manual download of Julia](https://julialang.org/downloads/manual-downloads/) for finer control.
+
+The easiest way forward is the creation of a Julia environment, with a 
+`LocalPreferences.toml` file that defines where the GPU-aware MPI library
+is found on the system, as well as CUDA details. On AMD clusters, only the
+MPIPreferences are necessary.
+
+```julia
+# This preference can be removed on AMD GPU clusters
+[CUDA_Runtime_jll]
+__clear__ = ["local"]
+version = "13.1" # CUDA version number should match that of the cluster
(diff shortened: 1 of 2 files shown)


docs/src/tricks/gpu.jl
```diff
@@ -72,7 +72,7 @@ scfres = self_consistent_field(basis; tol=1e-6)
 compute_forces(scfres)
 
 # !!! warning "GPU performance"
-#     Our current (December 2025) benchmarks show DFTK to have reasonable performance
+#     Our current (August 2025) benchmarks show DFTK to have reasonable performance
```
mfherbst, line 75: Update year ?

docs/src/tricks/gpu.jl
```diff
@@ -9,10 +9,10 @@
...
 #
 # !!! info "GPU parallelism not supported everywhere"
-#     Not all features of DFTK are ported to the GPU. SCF and forces
-#     with standard Libxc functionals are fully supported. Stresses
-#     and response calculations are work in progress as of December 2025.
-#     In most cases there is no intrinsic limitation and typically it only takes
+#     Not all features of DFTK are ported to the GPU. As of August 2026, 
+#     SCF, forces, stresses, and response calculations are supported with 
+#     all standard Libxc functionals (at LDA, GGA and mGGA level).
+#     In most cases there is no intrinsic limitationm and it typically only takes
```
mfherbst: Typo

mfherbst (approved):
Thanks !

docs/src/tricks/gpu.jl
abussy: I haven't run comprehensive benchmarks recently, so the acceleration numbers are indeed tied to December 2025. I expect that they have improved since, although the numbers are vague enough that they might still be correct. I'll be running benchmarks soon enough though, I can update then.

docs/src/tricks/compute_clusters.md
```diff
@@ -313,6 +313,123 @@ srun julia -t 1 --project -e '
...
+# Add Julia exec to the PATH, and set JULIA_DEPOT_PATH to a place on $SCRATCH
+export PATH="/path_to_julia_installation_dir/julia-1.12.6/bin:$PATH"
+export JULIA_DEPOT_PATH="/path_to_julia_depot_on_scratch/.julia"
+
+# Directory for cached JIT compiled Libxc GPU code (only for NVIDIA)
+export CUDA_CACHE_PATH="/path_to_existing_folder_on_scratch"
+
+my_project="/absolute_path_to_julia_project"
+
+srun julia --project=$my_project example.jl > example.out
```
Technici4n:
```suggestion
srun julia --project=$my_project --compiled-modules=strict example.jl > example.out
```
If you add this it will crash if any module is not precompiled. Better than a deadlock :)

abussy: Learning new stuff everyday

abussy:
As you might appreciate, efficiently running CUDA-aware MPI with DFTK is very tricky, and that is solely due to `Libxc_GPU_jll`. I am planning an update to `Libxc.jl` in order to allow external builds via a LocalPreferences setting. This will allow the following:

1) Running MPI and the rest of Julia with a common CUDA installation (avoid warning proliferation)
2) Avoid insanely long JIT compilation of Libxc, since the external build can be done for the specific GPU architecture. Also avoids the necessity of setting `CUDA_CACHE_PATH`.
3) Enable HIP Libxc on AMD clusters

Human cost will be minimal, since building libxc is very easy (only CMake and CUDA/HIP required).

PR #1379, opened 2026-08-15
fix upf loading for DFTpy pseudos with empty core data and dummy projectors

### LINKED ISSUE #1378: load_psp fails on empty PP_NLCC and 1-point dummy projector
`load_psp` fails on some local-only BLPS UPF files generated by DFTpy: https://dftpy.rutgers.edu/LPP.html 

Example:

```julia
using DFTK
load_psp("al.blps.lda.upf")
``` 

The two cases seem to trigger it:

1. Empty optional PP_NLCC is parsed as Float64[], then broadcast against the full radial grid (see [this line](https://github.com/JuliaMolSim/DFTK.jl/blob/4f9cf1bd11d7605ffcdfc63f0d8a05b90f329b1c/src/pseudo/PspUpf.jl#L157)). 
2. A dummy zero projector with cutoff_radius_index="1" reaches linear_interpolation, which cannot build a linear interpolant from one point ([see here](https://github.com/JuliaMolSim/DFTK.jl/blob/4f9cf1bd11d7605ffcdfc63f0d8a05b90f329b1c/src/pseudo/PspUpf.jl#L162)).

A minimal patch in `src/pseudo/PspUpf.jl` (line 157 and 162) could be: 

```julia
zero_if_missing_or_empty(x) = (isnothing(x) || isempty(x)) ? zeros(length(rgrid)) : x

r2_ρion = pseudo.rhoatom ./ (4π)
r2_ρcore = rgrid .^ 2 .* zero_if_missing_or_empty(pseudo.nlcc)
r2_τcore = rgrid .^ 2 .* zero_if_missing_or_empty(pseudo.taumod)
``` 
and:

```julia
r2_projs_interp = map(r2_projs) do r2_projs_l
    map(r2_projs_l) do proj
        if length(proj) < 2
            linear_interpolation((rgrid[1:2],), zeros(2))
        else
            linear_interpolation((rgrid[1:length(proj)],), proj)
        end
    end
end
```

aouinaayoub:
Fixes JuliaMolSim/DFTK.jl#1378


Technici4n:
Maybe a naive question: why does the file even include the projectors if they are dummy, instead of not listing any projector?

Maybe all we have to do is normalize the files when reading them in PseudoPotentialIO.jl?

aouinaayoub:
I don't know why DFTpy's UPF writer keeps these fields. I think the projector and empty NLCC are artifacts of the generic UPF representation.

Yes, I think it's better to strip out these empty sections when reading them in PseudoPotentialIO.jl

mfherbst:
Fair point @Technici4n. Sorry to keep pushing you around @aouinaayoub, but I agree that fixing that at the PSPio level is the more reasonable approach. Would you mind kick-starting the discussion over there with a PR. Best also include the weird file in a unit test over there, so we can catch bugs early.

Issue #1322, opened 2026-06-12
Refactor coulomb.jl


antoine-levitt:
API discussed with @toschaefer @mfherbst @j-mangold:

```julia
## public API of a kernel
# the fundamental mathematical API: a kernel is a function that depends on p
function eval_kernel_fourier(k, p::AbstractVector3) # needs to be implemented for every kernel, the fallbacks below will take care of the rest
end
# computes a bunch of eval_kernel_fourier at once. Needed for wigner-seitz where we use FFTs to speed up computation
function eval_kernel_fourier(k::InteractionKernel, basis)
    [eval_kernel_fourier(k, G) for G in G_vectors(basis)]
end
# computes the probe charge integral ∫...
function compute_probe_charge_integral(k::InteractionKernel, α)
    error(NotImplementedError(), "This kernel does not support probe charge")
end
# invariants that can be tested for every kernel
@test eval_kernel_fourier(k, basis) == map(G_vectors()) do eval_kernel_fourier(k, G)
@test try compute_probe_charge_integral(k, α) ≈ get_it_by_quadrature() catch e <: NotImplementedError true#= =# end


struct BareCoulomb <: InteractionKernel end

function eval_kernel_fourier(k::BareCoulomb, p)
    1/p^2
end
function compute_probe_charge_integral(k::BareCoulomb, α)
    # ...
end


struct WignerSeitzTruncatedCoulomb <: InteractionKernel end

function eval_kernel_fourier(k::WignerSeitzTruncatedCoulomb, p)
    # get the short-range part analytically, the long-range part by slow fourier transform (explicit sum over R)
end
function eval_kernel_fourier(k::BareCoulomb, basis)
    # get the short-range part analytically, the long-range part by fast fourier transform
end
# compute_probe_charge left unimplemented explicitly, it'll just error

# this should be used for explicitly singular kernels (bare coulomb and long-range coulomb, but not eg spherically truncated or wigner seitz: in these two later cases the implementation of eval_kernel_fourier should be careful to return the right thing at 0)
struct ReplaceSingularity <: InteractionKernel
    inner_kernel <: InteractionKernel
    replacement
end
function eval_kernel_fourier(k::ReplaceSingularity, q)
    q == 0 ? replacement : eval_kernel_fourier(k.inner_kernel, q)
end
# compute_probe_charge left unimplemented explicitly, it'll just error

struct VoxelAverage
    inner_kernel <: InteractionKernel
    voxel
end
[...]

PR #1382, opened 2026-09-04
[WIP] k-point sampling for exact exchange

toschaefer:
A first version of exact exchange for k-meshes beyond $\mathbf \Gamma$-only sampling.

Exchange couples k to k' via the momentum transfer q = k - k'.
Here, we add a q-point logic.

- build_qpoints (PlaneWaveBasis.jl) returns the unique q = k - k' plus a table kprime_mapping[ik, iq] giving the index of k' = k - q. 
- TermExactExchange stores one kernel per q. Energy and operator loop over q.
- Coulomb kernels take a q-point and are evaluated (e.g. $4\pi/(G+q)^2$) on the full FFT grid instead of the smaller spherical grid, since overlap densities require large G. This changed references values in the tests slightly (updated). 
- Coulomb kernels adapted: for instance, spherical cutoff uses the supercell volume for default radius, Wigner-Seitz truncates at the supercell WS cell, voxel averaging takes its mesh from the q-grid. 
- shifted_cholesky for the ACE compression (not sure if there are better solutions to avoid tiny negative eigenvalues...)
- new tests: kernels and energies of a k-grid vs. the equivalent supercell, collinear spin with k-points, AFM chain vs. supercell, explicit/shifted grids.

Missing: 
* MPI parallelization over k-points
* GPU optimization (not tested at all)
* does not exploit any symmetry of BZ (currently requires symmetries=false)
* we can either apply https://github.com/JuliaMolSim/DFTK.jl/issues/1322 here or in a new PR.

Small example (finishes in about a minute):
```julia
using DFTK                                                                              
using AtomsBuilder                                                                      
using PseudoPotentialData                                                               
                                                                                        
pseudopotentials = PseudoFamily("dojo.nc.sr.pbe.v0_5.stringent.upf")                    
system = bulk(:Si)                                                                      
Ecut   = 15                                                                             
kgrid  = [2, 2, 2]                                                                      
                                                                                        
model_pbe  = model_DFT(system; pseudopotentials, functionals=PBE(), symmetries=false)   
basis_pbe  = PlaneWaveBasis(model_pbe; Ecut, kgrid)                                     
scfres_pbe = self_consistent_field(basis_pbe; tol=1e-6)                                 
                                                                                        
model  = model_HF(system; pseudopotentials, exx_kernel=WignerSeitzTruncatedCoulomb())   
basis  = PlaneWaveBasis(model; Ecut, kgrid)                                             
scfres = self_consistent_field(basis; scfres_pbe.ρ, scfres_pbe.ψ,                       
                               scfres_pbe.eigenvalues, scfres_pbe.occupation,           
                               is_converged=ScfConvergenceEnergy(1e-8),                 
                               exxalg=AceExx(), damping=0.4,                            
                               solver=DFTK.scf_damping_solver())                        
```


src/PlaneWaveBasis.jl
```diff
@@ -653,6 +653,63 @@ function scatter_kpts_block(basis::PlaneWaveBasis, data::Union{Nothing,AbstractA
     end
 end
 
+"""
+Computes the unique momentum transfers q = k - k' for a given basis.
+Returns the Kpoint objects for the q-grid and the integer mapping table
+`kprime_mapping[ik, iq]` such that k' = k - q.
+"""
+function build_qpoints(basis::PlaneWaveBasis{T}; tol=1e-12) where {T}
```
Technici4n, line 661: Some relevant helpers that we already have for phonons: https://github.com/JuliaMolSim/DFTK.jl/blob/4a39b3950aa59a7eced1d12e93fab70eb4c82e2a/src/transfer.jl#L180

antoine-levitt:
From the call: use find_equivalent_kpt from transfer.jl to recompute the kpoint corresponding to k+q on the fly, without precomputing it

toschaefer:
Thank you both for the hint. With `find_equivalent_kpt` you only mean to replace my current `kprime_mapping`? Or do you suggest to remove the entire function `build_qpoints` and do the entire q-logic in the fly (e.g. by summing of k' instead)?

After thinking a bit about it, I think it makes sense to build a list of all q points in the beginning, since many (k, k') pairs share the same q. For instance, some Coulomb kernels (like Wigner-Seitz) are relatively expensive to build per q. So having a list of unique q-points is beneficial. 

One further thing: I had a problem with `build_qpoints` and `ExplicitKpoint`. There the stored points are not normalised, e.g. a k-point stored as 0.57 is searched as -0.43 (the same point) and findfirst returns nothing. 
I tried to solve that issue by comparing modulo reciprocal lattice vectors (is_approx_integer(k1 - k2)). The same fix would work in `find_equivalent_kpt`. 
Monkhorst-Pack grids are fine because their coordinates are already normalised.

I used `ExplicitKpoint` to test arbitrarily shifted k-meshes.

antoine-levitt:
Re Cholesky: the issue here is finding a low-rank Atilde such that Atilde X = Y (Y = WX here). Naive option is Atilde = Y X^T but that's not symmetric, so ACE is Atilde = Y M Y^T, with M = (Y^T X)^-1. I'm kind of surprised that the "increasing-shift cholesky" fails here, can you maybe give a reproducer? Otherwise a stable option is to do an eigen decomposition of Y^T X and take out the small eigenvalues : it's small enough that it doesn't matter if we do eigen or cholesky here.

Alternatively, orthonormalize Y = AX first: do Y = QR, write Atilde = Q M Q*, M has to be R (Q* X)^-1. Still have to invert `Q*X`, but that one is possibly better conditioned than `Y*X`?

toschaefer:
Thanks! I already ended up with exactly your second option: eigendecomposition and dropping small eigenvalues.
I think one reason the shifted Cholesky fails is that the kernel can have negative Fourier components (sometimes happens with the ProbeCharge method). I will push it soon and explain in more detail.

Also: I will use find_equivalent_kpt instead of kprime_mapping but I will keep the initial construction of the q-point list.

antoine-levitt:
Re the q points: I think what i'm saying is that you should not have the q be a Kpoint, just a Vec3. If precomputing the coulomb kernel is expensive sure you precompute them

Re explicit kpoints: why don't we just normalize them at basis construction?

antoine-levitt:
ah sorry didn't see your edit, looks like we agree!

toschaefer:
Alright, thanks to your help I think a found a viable solution for both problems (q-points and cholesky).

Regarding q-points: The function build_qpoints is now very short and only returns all q-points but no object kprime_mapping. The latter (k') is build on the fly using find_equivalent_kpt, as suggested. 
Additionally, I slightly modified the function find_equivalent_kpt to make it compatible with `ExplicitKpoints`: essentially an additional call of normalize_kpoint_coordinate was required. Check the diff of transfer.jl. 

Regarding cholesky: I think the origin of the failure for cholesky were negative Fourier component of the Coulomb kernel when using the probe charge method in combination with anisotropic cells. Then `Mk` can become indefinite. In accordance with @antoine-levitt second suggestion, the ACE algorithm now uses a pseudo-inverse based on a diagonalization of the Hermitian matrix Mk:
```julia
# The compressed operator is Wk Mk⁻¹ Wk'. Mk is Hermitian, but depending on the kernel 
# possibly indefinite and nearly singular. We therefore use its pseudo-inverse,        
# dropping eigenvalues below the tolerance.                                            
Mk = Hermitian(ψk' * Wk)                                                               
λ, U = eigen(Mk)                                                                       
cutoff = size(Mk, 1) * eps(real(T)) * maximum(abs, λ)                                  
Dk = Diagonal(map(λi -> abs(λi) > cutoff ? inv(λi) : zero(λi), λ))                     
opk = NonlocalOperator(basis, kpt, Wk * U, Dk)                                         
(; opk, Mk, Dk, Wk)                     
```                                               
I added an explicit test for a kernel with a negative component via `Coulomb(ReplaceSingularity(-100.0))`.
The function shifted_cholesky was removed from src/common/linalg.jl.

toschaefer:
> Re the q points: I think what i'm saying is that you should not have the q be a Kpoint, just a Vec3. If precomputing the coulomb kernel is expensive sure you precompute them

Agreed!

> Re explicit kpoints: why don't we just normalize them at basis construction?

Ok, but could that lead to unexpected behavior if users specify explicit paths in the BZ (for band structures etc)?

antoine-levitt:
> Dk = Diagonal(map(λi -> abs(λi) > cutoff ? inv(λi) : zero(λi), λ))                     

FYI `pinv` works for scalars (and for diagonal matrices). Also don't we want to zero out the positive eigenvalues?

> opk = NonlocalOperator(basis, kpt, Wk * U, Dk)                                         

should that not be Wk, U Dk U* (to do the manipulations on small matrices rather than big ones?) multiplication by unitary matrices can be freely switched around with no accuracy issue.

antoine-levitt:
> Ok, but could that lead to unexpected behavior if users specify explicit paths in the BZ (for band structures etc)?

OK I understood what you were saying: find_equivalent_kpt currently assumes the kpoints are normalized but could be fixed to not assume this, in which case this issue would not be present, right? Then yes I agree it's the better fix

toschaefer:
> FYI `pinv` works for scalars (and for diagonal matrices). Also don't we want to zero out the positive eigenvalues?

hm, no I think zero out positive eigenvalues would really change (not just compress) the operator. Note that negative components of the Fourier kernel are not wrong, but a consequence of the way probe charge works. So the positive eigenvalues are important.

Regarding pinv: I see, do you mean like that?
```
Mk = Hermitian(ψk' * Wk)
opk = NonlocalOperator(basis, kpt, Wk, pinv(Mk))
(; opk, Mk, Wk)
```

antoine-levitt:
> opk = NonlocalOperator(basis, kpt, Wk, pinv(Mk))

I think you need pinv(Hermitian(Mk)) to make sure it will dispatch to an eigenvalue decomposition (otherwise it will use an svd). Also there's kwargs you can use to set the thresold.

> hm, no I think zero out positive eigenvalues would really change (not just compress) the operator. Note that negative components of the Fourier kernel are not wrong, but a consequence of the way probe charge works. So the positive eigenvalues are important.

OK yes I guess that makes sense, there's nothing particularly horrible if that term is negative (I was thinking of something like a metric, where you really want it to be positive)

toschaefer:
> I think you need pinv(Hermitian(Mk)) to make sure it will dispatch to an eigenvalue decomposition (otherwise it will use an svd). Also there's kwargs you can use to set the thresold.

I am still unsure because of GPU support. The eigen part will probably run on GPU but then the pseudo-inverse might read/write individual entries?

toschaefer:
If you are okay, I could implement Issue https://github.com/JuliaMolSim/DFTK.jl/issues/1322 right away (we already have a prototype in PR https://github.com/JuliaMolSim/DFTK.jl/pull/1338). I think that's a straightforward task for Claude Opus 5.5.

I would suggest to tackle the remaining open points (MPI, GPU and symmetry) in separate PRs.

antoine-levitt:
> If you are okay, I could implement Issue https://github.com/JuliaMolSim/DFTK.jl/issues/1322 right away (we already have a prototype in PR https://github.com/JuliaMolSim/DFTK.jl/pull/1338). I think that's a straightforward task for Claude Opus 5.5.

Sure!

> I am still unsure because of GPU support. The eigen part will probably run on GPU but then the pseudo-inverse might read/write individual entries?

No particular reason it should, it's probably a map. If not that's a bug for GPUArrays to tackle. We have CI now to tell us that!

Issue #122, opened 2020-02-09
Thomas-Fermi


antoine-levitt:
So, if we omit any kinetic or nonlocal terms, we get thomas-fermi like models where the density is a functional of the density only. That's quite fun for educational purposes, and also maybe useful for people working on orbital-free DFT. The diagonalization approach could maybe work but is stupid here. We should have it as an example just for the fun of it.

antoine-levitt:
To clarify: the simplest is to redo an SCF by hand

antoine-levitt:
As a first step, the Thomas-Fermi von Weizsacker model looks like a one-band DFT model. The normalization constraint is int of u^2 = N, but one could scale that away to 1 and just do a direct minimization.

PR #1384, opened 2026-09-14
add experimental orbital-free DFT

aouinaayoub:
This adds an experimental orbital-free DFT model and direct density minimization using kinetic-energy functionals from Libxc.

It also contains an implementation for #122.

The results for aluminium using TF+vW agree with those from [DFTpy](https://dftpy.rutgers.edu/).


mfherbst:
Dear @aouinaayoub. Sorry for the slow reply. This would be domething of interest for us. I'll hopefully get round to a closer look and some feedback next week.

PR #1385, opened 2026-09-15
Remove XC GPU logic

abussy:
The latest updates to Libxc.jl and DftFunctionals.jl implement CPU fallbacks when there is no GPU alternative (usual transfer from GPU to CPU, evaluation on the CPU, transfer back). Therefore, there is no need for it in DFTK anymore.

However, I ran into an unforseen issue: `MethodError` start appearing due to ambiguous signature, e.g.:

```
potential_terms(::DFTK.LibxcFunctional{:lda, :x}, ::CUDACore.CuArray{Float64, 2, CUDACore.DeviceMemory})
```

Is ambiguous because of the following 2 definitions:
```
     potential_terms(func::DFTK.LibxcFunctional{:lda}, ρ::AbstractMatrix{Float64})
       @ DFTK /capstor/scratch/cscs/abussy/dftk_santis/DFTK.jl/src/DispatchFunctional.jl:108
       
     potential_terms(func::DftFunctionals.Functional{:lda}, ρ::GPUArraysCore.AbstractGPUArray)
       @ DftFunctionals /capstor/scratch/cscs/abussy/dftk_santis/.julia/packages/DftFunctionals/j8Gg9/src/gpu.jl:26
```

Because either option has either a stronger type for the functional or the array, it's ambiguous. A method for `LibxcFunctional` and `AbstractGPUArray` must be available to resolve that.

I solved this with constructs of the type:
```
DftFunctionals.potential_terms(func::LibxcFunctional{:lda},
                               ρ::Matrix{Float64}) =
    _potential_terms(func, ρ)
DftFunctionals.potential_terms(func::LibxcFunctional{:lda},
                               ρ::AbstractGPUMatrix{Float64}) =
    _potential_terms(func, ρ)
function _potential_terms(func::LibxcFunctional{:lda},
                          ρ::AbstractMatrix{Float64})
    s_ρ, n_p = size(ρ)
    fun = Libxc.Functional(func.identifier; n_spin=s_ρ)
    derivatives = filter(in(Libxc.supported_derivatives(fun)), 0:1)
    terms = Libxc.evaluate(fun; rho=ρ, derivatives)
    e  = libxc_energy_density(terms, ρ)
    Vρ = reshape(terms.vrho, s_ρ, n_p)
    (; e, Vρ)
end
```

I don't particularly like that: it's very verbose, and seems like a step sideways instead of forward. However, I couldn't think of anything better. Suggestions are welcome.


abussy:
cscs-ci run GH200, Mi300

src/DispatchFunctional.jl
```diff
@@ -48,7 +48,11 @@ function libxc_energy_density(terms::NamedTuple, ρ::AbstractMatrix)
     haskey(terms, :zk) ? reshape(terms.zk, 1, size(ρ, 2)) .* sum(ρ; dims=1) : false
 end
 
-function DftFunctionals.energy_density(func::LibxcFunctional, ρ::AbstractMatrix{Float64},
+# Custom input type encompassing both CPU Matrix{T} and GPU AbstractGPUMatrix{T} to avoid
+# ambiguities with methods defined in DftFunctionals with the more generic AbstractMatrix{T} type
+const InputMatrix{T} = Union{Matrix{T}, AbstractGPUMatrix{T}}
```
mfherbst, line 53: DftInputMatrix ? Quite specific for this one use case.

Still weird as this is a concrete and an abstract type as a union ... but certainly better as a solution then what we had before.

